PR #122, opened 2020-04-02
[WIP] Sticky navbar on page scroll

manpenaloza:
pr exists as a result of [this twitter conversation](https://twitter.com/nikolasburk/status/1245712496570351616) :)

As the version 2 docs are not mobile responsive so far, I've wrapped the proper CSS code within a standard media query for large screens (>1024px) to support upcoming mobile/responsive tweaks.

The CSS code uses the `position: sticky` CSS feature, which means that [compatibility is given for the majority of browsers](https://caniuse.com/#search=sticky), except Internet Explorers. 

Done so far:
- [ ] feature code
- [ ] tested manually on Safari @ Mac
- [ ] tested manually on Firefox @ Mac
- [ ] tested manually on Chrome @ Mac
- [ ] run `prettify` node script

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "react-instantsearch-dom": "^6.4.0",
     "react-live": "^2.2.2",
     "react-loadable": "^5.5.0",
+    "react-sticky": "^6.0.3",
     "react-testing-library": "^6.0.0",
     "smooth-scroll": "^16.1.2",
     "styled-components": "^4.1.3",
--- a/src/components/sidebar/index.tsx
+++ b/src/components/sidebar/index.tsx
@@ -1,18 +1,18 @@
 import React from 'react';
+import { StickyContainer, Sticky } from 'react-sticky';
 import Tree from './tree';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import { useAllArticlesQuery } from '../../hooks/useAllArticlesQuery';
 import { AllArticles } from '../../interfaces/AllArticles.interface';
 
-const Sidebar = styled.aside`
+const Sidebar = styled.div`
   margin: 0;
-  max-width: 215px;
-  width: 20%;
-  @media (min-width: 1024px) {
-    align-self: flex-start;
-    position: sticky;
-    top: 0;
-  }
+  overflow-y: auto;
+  ${({ isSticky }) =>
+    isSticky &&
+    css`
+      max-height: 100vh;
+    `};
 `;
 
 const List = styled.ul`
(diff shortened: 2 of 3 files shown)


kuldar:
Hey @manpenaloza, thank you so much for the help!

I vaguely remember going for the `position: sticky` solution at some point earlier as well, but it comes with some problems - namely if you're looking at the docs on a smaller screen, you can't actually scroll the sidebar independendly from the content, which means you can't access the bottom links until you scroll to the very end of the page.

I am not sure how much work-arounds that requires or if `position: sticky` is still the best way to go. But just as an inspiration ,here's an earlier sticky-sidebar experiment made with `react-sticky`. (You can see that at a smaller viewport, the sidebar is still independendly scrollable.)

- https://keen-stonebraker-b8c494.netlify.com/
- https://github.com/kuldar/docs-layout/blob/master/src/App.js

manpenaloza:
yes, much better. I'll update the pr.

manpenaloza:
Hey @kuldar !
PR is updated. It seems there's no natural and clean way to achieve this using 'position: sticky'. So I used the react-sticky package quite similar to how you already used it, which was great. Now the desired behavior works on my machine. But it doesn’t work in the Netlify build/preview. Do you have an idea why? Maybe because of the new dependency? 

**The changes required some adjustments:**
- increased navigation container width from 215 to 235px as the scrollbars (depending on browser) take quite some space
- Increased navigation links tags spacing from 0 to 5% to not being overlayed by scrollbars once you scroll within the navigation. There would be ways to make the scrollbar not visible when scrolling, but I haven’t implemented it due to accessibility issues.

**Undesired side effect:**
The sublinks of navigation links don’t work anymore (i.e., "Setup Prisma" should have sublinks). I haven’t investigated the problem more in-depth and will check this tomorrow.

manpenaloza:
Hey @kuldar !
I've mentioned an undesired side effect in my last comment. The latest changes fixed this, and this pr already got updated. 

But still: the Netlify preview somehow does not show how it should behave (sticky scroll not happening + sublinks not expandable). It seems that styling and the react-sticky dependency include works because markup and styling are appropriately applied. But somehow in the browser, the “sticky-causing” Javascript + sublinks open don't execute. Maybe the client-side Javascript hydrating doesn't work correctly? Can you please check this pr on your machine to come closer to finding a reason for this? Thank you!

nilubava:
@manpenaloza This is a known issue for Netlify deploy previews with path prefix `docs`, and we are working on resolving this. This issue will not appear in `prisma.io/docs` tho'.

manpenaloza:
@nilubava ok, thx for information. so then I think I'm done on that PR (?). If you need anything else on this, just ping me here (or in the new PR #144 ).

nilubava:
@manpenaloza So the sticky sidebar is on production now! Thanks for the support!

manpenaloza:
you're welcome!

PR #134, opened 2020-04-03
Updates sql in postgres example to match the rest of the doc and fixes prop name in insert example

lupomontero:
Hi there,

I'm having my first go at prisma today :rocket:, found a couple of minor things in the docs, ...

Thanks for the great work!


nikolasburk:
Thanks a lot for catching this @lupomontero 🚀

PR #138, opened 2020-04-05
Typescript getting-started fixes

kkuchta:
Fixing a couple typos in the typescript version of "getting started
sql": One command that uses node instead of ts-node, and one that is a
duplicate of the few lines above it.


nikolasburk:
Thanks so much for catching this @kkuchta 🎉

PR #140, opened 2020-04-06
Relations: fix nested write example

2color:


nikolasburk:
Thanks so much for fixing those Daniel 🚀

PR #143, opened 2020-04-06
Fix two typos in relational data docs.

leerob:
Thanks for the great docs! 🙌


2color:
Thanks a lot for cathing this 🙏

PR #147, opened 2020-04-07
Fix typo

lewisvrobinson:
### Problem

There is a typo ("ther" – should be "other") on the [why prisma](https://www.prisma.io/docs/understand-prisma/why-prisma) page.

### Solution

Fix the typo!


nikolasburk:
Thanks so much for catching this @lewisvrobinson 🎉

PR #157, opened 2020-04-07
Update year in footer

arvindell:


nikolasburk:
Thanks a lot for catching this @alexvilchis 🙏  we actually updated the footer ourselves in the meantime to match the one from the website!

PR #158, opened 2020-04-07
Fix missing index missing from many-to-many

teebot:
see https://github.com/prisma/prisma/issues/2123#issuecomment-610477124


nikolasburk:
Thanks a lot for this @teebot 🚀

PR #177, opened 2020-04-10
Fix link #type-modifiers

adrianoresende:
https://github.com/prisma/prisma2-docs/issues/176


nikolasburk:
Thanks a lot for catching this @adrianoresende   🎉

PR #179, opened 2020-04-11
Fix 'Prisma-level' on Prisma schema / relations page

anton62k:


nikolasburk:
Thanks a lot for catching this @anton62k   🎉

PR #182, opened 2020-04-11
Update 04-data-model.mdx

dpilarsk:
Fix Int type instead of Boolean


nikolasburk:
Thanks a lot for catching this @dpilarsk   🎉

PR #196, opened 2020-04-13
Fix typo

jhaemin:
- `apprach` → `approach`
- `Prima` → `Prisma`


nikolasburk:
Thanks so much for catching this @jhaemin 🎉

PR #199, opened 2020-04-13
Fix omission of to be verb

arvindell:


nikolasburk:
Thanks a lot for catching this @alexvilchis 🎉

PR #201, opened 2020-04-13
Improve search

Tom-Bonnike:
I noticed the search was **extremely** slow and made the whole website freeze, which is caused by requests being fired on **every single** keystroke… I started debouncing the search in my first commit but then realized the search wasn’t set up properly, so I tried to improve that in my second commit.

I followed the algolia doc for the implementation. See https://www.algolia.com/doc/guides/building-search-ui/going-further/improve-performance/react/#debouncing.

⚠️ I’d suggest manually testing this PR before merging because there seems to be one more tiny issue: an empty query is fired on page load, which is then cached and flashes before showing the results of the very first search 🤷‍♂️It’s not caused by this PR because I can also see it happening in the network tab of the current website, but this PR kinda makes it more obvious (see the after GIF).

# Before

I start typing at normal speed directly after the GIF starts. The whole UI freezes.

Kapture 2020-04-13 at 17 37 21 

1 query/network request is made per character.

Screenshot 2020-04-13 at 17 41 15  

# After

The UI doesn’t freeze.

Kapture 2020-04-13 at 19 01 36

A single query is made.

Screenshot 2020-04-13 at 19 02 41

You might want to add a loading indicator for people on slow networks, too. 🤷‍♂️

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -39,24 +39,22 @@ const HitsWrapper = styled.div`
 `;
 
 const indexName = config.header.search.indexName;
+const searchClient = algoliasearch(
+  config.header.search.algoliaAppId,
+  config.header.search.algoliaSearchKey
+);
+
+const Results = connectStateResults(({ searchState: state, searchResults: res, children }: any) =>
+  res && res.nbHits > 0 ? (
+    children
+  ) : (
+    <div className="no-results">No results for '{state.query}'</div>
+  )
+);
 
 export default function Search() {
   const [query, setQuery] = useState(``);
-  const searchClient = algoliasearch(
-    config.header.search.algoliaAppId,
-    config.header.search.algoliaSearchKey
-  );
-
   const [showHits, setShowHits] = React.useState(true);
-
-  const Results = connectStateResults(({ searchState: state, searchResults: res, children }: any) =>
-    res && res.nbHits > 0 ? (
-      children
-    ) : (
-      <div className="no-results">No results for '{state.query}'</div>
-    )
-  );
-
   const hideSearch = () => setShowHits(false);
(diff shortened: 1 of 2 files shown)


src/components/search/index.tsx
```diff
@@ -39,24 +39,22 @@ const HitsWrapper = styled.div`
 `;
 
 const indexName = config.header.search.indexName;
+const searchClient = algoliasearch(
+  config.header.search.algoliaAppId,
+  config.header.search.algoliaSearchKey
+);
```
Tom-Bonnike, line 45: This was unnecessarily being called on every render of the Search component.

src/components/search/index.tsx
```diff
@@ -39,24 +39,22 @@ const HitsWrapper = styled.div`
...
+
+const Results = connectStateResults(({ searchState: state, searchResults: res, children }: any) =>
+  res && res.nbHits > 0 ? (
+    children
+  ) : (
+    <div className="no-results">No results for '{state.query}'</div>
+  )
+);
```
Tom-Bonnike, line 53: This React component was also unnecessarily re-created on every render of the Search component.

nilubava (approved):
Tested locally, and everything works good. I have added the 'Searching..' indicator in another PR. About the initial query cache, it seems this happens in default settings, and may have to be removed on page load explicitly. Will create a PR for this too later.

Tom-Bonnike:
Cool. Maybe it will look better once there is the loading indicator already.
It seems you’re right, that’s instantsearch’s default behaviour, I guess it could make sense to show those initial results directly after focusing the input then, instead of waiting for the next query.

janpio:
Thanks @Tom-Bonnike 🚀

PR #226, opened 2020-04-17
Fix typos

jimt:


nikolasburk:
Thanks a lot for catching this @jimt 🎉

PR #237, opened 2020-04-21
Add note about GraphQL specific configuration

### LINKED ISSUE #2121: Document AWS lambda configuration to avoid timeouts
We should document https://github.com/prisma/prisma/issues/1754#issuecomment-601271228
Maybe in https://www.prisma.io/docs/guides/deployment/deploying-to-aws-lambda and in nexus too?

In the issue above the solution was to do this:
```
// Set to false to send the response right away when the callback executes, instead of waiting for the Node.js event loop to be empty.
context.callbackWaitsForEmptyEventLoop = false;
```

2color:
Fixes https://github.com/prisma/prisma/issues/2121


nikolasburk:
Thanks a lot Daniel – note that I adjusted the heading of the new section from `GraphQL` to `A note on deploying GraphQL servers to AWS Lambda`

2color:
Thanks. Much better :)

PR #238, opened 2020-04-21
Remove semicolons for consistency across projects

2color:
Remove semicolons so we're consistent with the [site examples](https://www.prisma.io/) and other repos.


nikolasburk:
🧹 Thanks for sweeping this Daniel 😄 :party-parrot:

PR #240, opened 2020-04-21
[FIX] switch zeit now to Vercel Now for prisma docs

Developerayo:
- Zeit just rebranded to Vercel and this PR changes the deployment guidelines to deployment
- Now still remains Now but ZEIT is no longer zeit but Vercel as of today

### DIFF
--- a/content/04-guides/02-deployment/01-deploying-to-vercel-now.mdx
+++ b/content/04-guides/02-deployment/01-deploying-to-vercel-now.mdx
@@ -1,14 +1,14 @@
 ---
-title: 'Deploying to ZEIT Now'
-metaTitle: 'Deploying to ZEIT Now'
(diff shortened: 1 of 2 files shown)


janpio:
I think Now is actually also just "Vercel" now: https://www.notion.so/Frequently-Asked-Questions-c726facaf7cc41c095ba49aeea90219e

Developerayo:
Yes! that's quite true, thank's for pointing that out

> Changes reflected

janpio:
Additional note: This should probably only be merged after the actual example is also renamed.

Developerayo:
Quite true!
I'll send a separate PR for that on the Prisma-example repo

mhwelander:
Hi! You were way ahead of us with this - I will close the PR only because the changes were included in #374; thank you 💚 so much for taking the time.

PR #242, opened 2020-04-22
Fix typo and remove duplicate table rows

benmccann:
I'm assuming "tun" was supposed to be "run" though the mistake was made twice which made me question if it's an acronym I'm unfamiliar with


janpio:
The "duplicate table rows" are not actually duplicates, but JS variants vs the other TS variants of these examples.

benmccann:
Thanks. I sent a new PR with just the typo fix (that was easiest for me since I'm on my phone)

PR #247, opened 2020-04-22
Fix typo

benmccann:
I'm assuming "tun" was supposed to be "run" though the mistake was made twice which made me question if it's an acronym I'm unfamiliar with


mhwelander:
You are correct - it should be 'run'. 😊 Thank you for taking the time to fix it!

PR #250, opened 2020-04-23
Fix typo

timleslie:


2color:
Thanks so much for correcting this @timleslie 🎉

PR #263, opened 2020-04-25
Fix error in return type description

LaurenceM10:
Fix error in return type description for delete query


mhwelander:
Good spot - thank you! 🙏

PR #292, opened 2020-04-28
Prisma Docs - Introduction | Minor Detail Fixes

robin-macpherson:
## Description

Just a few small improvements that I found:

- a couple of missing commas in a list
- a few small grammar fixes ('use' > 'uses' & 'was' > 'has been')
- this is a small detail but there was a section of bullets that had inconsistent verbs: we had "generate" and then "writing" so I simply made them consistent ✌🏼🤓


robin-macpherson:
cc: @nikolasburk

nikolasburk:
Thanks so much Robin 🚀

PR #313, opened 2020-05-04
Filtering related example

mhwelander:
Added sections on how to use `where` for `select` and `include`


mhwelander:
For this: https://github.com/prisma/prisma/issues/2264

PR #338, opened 2020-05-12
Comparison tables

mhwelander:


content/02-understand-prisma/06-api-comparisons/01-prisma-and-mongoose.mdx
```diff
@@ -0,0 +1,294 @@
...
+    select: {
+      name: true,
+    },
+  });
+```
+
+**Mongoose**
+
+```ts
+  const user = await User.findById("5eb9354624286a04e42401d8").select([
```
nikolasburk: is the extra indentation here intional?

content/02-understand-prisma/06-api-comparisons/01-prisma-and-mongoose.mdx
```diff
@@ -0,0 +1,294 @@
...
+    "name",
+    "email",
+  ]);
+```
+
+## Fetching relations
+
+**Prisma**
+<Code hideCopy={true} languages={["Using include", "Fluent API"]}>
```
nikolasburk: isn't the MDX component called `CodeBlock`? does this render properly?

content/02-understand-prisma/06-api-comparisons/01-prisma-and-mongoose.mdx
```diff
@@ -0,0 +1,294 @@
...
+
+```ts
+  const posts = await prisma.user.findOne({
```
nikolasburk: is the extra indentation here intional?

nikolasburk (changes requested):
This looks like a great start @mhwelander 🚀 

I've found a few issues already:

- lots of code blocks seems to have extra indentation
- I believe the MDX code block component is called `CodeBlock`, does `Code` also work?

For some reason, I can't expand the `API comparisons` item in the preview, so I don't see what the rendered version of this page looks like right now...

content/02-understand-prisma/06-api-comparisons/01-prisma-and-mongoose.mdx
mhwelander: Fixed!

content/02-understand-prisma/06-api-comparisons/01-prisma-and-mongoose.mdx
mhwelander: Fixed!

2color (commented):
This is really useful and will help developers coming from either one of Sequelize, TypeORM, or Mongoose 🎉

- On https://deploy-preview-338--prisma2.netlify.app/understand-prisma/api-comparisons/prisma-and-mongoose/#fetching-relations Switching to Fluent API doesn't work.
- "Mongoose doesn't offer a dedicated API for relation filters. You can get similar functionality by sending a raw SQL query to the database.". There's no raw SQL with Mongoose
- The Batch update section of Sequelize shows batch updating based on a relation. This is a nice feature of Prisma but doesn't compare directly to the Sequelize example. What do you think about simplifying that to a normal batch update or adding both? 


Other ideas:
- What do you think about adding a line about Prisma not being an API and link to why Prisma is not an ORM?

I haven't gone through the TypeORM comparison but it's not a blocker

mhwelander:
* Added a sentence about Prisma being a toolkit & link to 'not an ORM'
* Code block tab issue is a Netlify thing
* Removed 'raw SQL' reference from the Mongo ODM! Changed it to: run query, then filter

PR #348, opened 2020-05-14
Smart module / .prisma folder

### LINKED ISSUE #323: Update new dot-prisma folder documentation & illustrations
The docs currently describe the old @prisma/client generation approach (e.g. https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/generating-prisma-client#the-prismaclient-npm-module). We should update it to reflect the new mechanism using the node_modules/.prisma folder.

mhwelander:


content/01-getting-started/02-setup-prisma/01-add-to-existing-project.mdx
```diff
@@ -451,7 +451,7 @@ To get started with Prisma Client, you need to install it as a node module:
 npm install @prisma/client
 ```
 
-Notice that the [`@prisma/client` node module](../../reference/tools-and-interfaces/prisma-client/configuring-the-prisma-client-api#the-prisma-client-npm-module) is a bit different from "conventional" node modules (e.g. a library like [`lodash`](https://lodash.com/)). When installing the module with `npm install @prisma/client`, npm only downloads an empty package into the `node_modules/@prisma/client` directory. The package is then "filled" with the actual library with the following command of the Prisma CLI:
+Notice that the [`@prisma/client` node module](../../reference/tools-and-interfaces/prisma-client/configuring-the-prisma-client-api#the-prisma-client-npm-module) references a folder named `.prisma\client`. The `.prisma\client` folder contains your generated Prisma client, and is modified each time you change the schema and run the following command:
```
janpio: The target of this link does not really exist on that page. Correct page is `generating-prisma-client` I think.

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-generating-prisma-client.mdx
```diff
@@ -55,28 +55,30 @@ const prisma = new PrismaClient()
 
 ## The `@prisma/client` npm module
```
janpio: Is `npm module` common usage here? I would have expected it to be called "npm package". 

(also applies to all the other content of course.)

But maybe this is actually a fine difference here - then possibly it could be explained.

PR #357, opened 2020-05-15
Index page content 

### LINKED ISSUE #287: Empty `index` pages found in sitemap.xml
I was looking at the sitemap when I found these url ending with `/index`

These pages are empty, example: 
Screenshot_2020-04-28 Getting started

List of urls:
https://www.prisma.io/docs/more/index
https://www.prisma.io/docs/guides/index
https://www.prisma.io/docs/guides/upgrade/index
https://www.prisma.io/docs/guides/deployment/index
https://www.prisma.io/docs/guides/database-workflows/index
https://www.prisma.io/docs/guides/database-workflows/cascading-deletes/index
https://www.prisma.io/docs/guides/database-workflows/foreign-keys/index
https://www.prisma.io/docs/guides/database-workflows/unique-constraints-and-indexes/index
https://www.prisma.io/docs/guides/database-workflows/import-and-export-data/index
https://www.prisma.io/docs/guides/database-workflows/setting-up-a-database/index
https://www.prisma.io/docs/reference/index
https://www.prisma.io/docs/reference/more/index
https://www.prisma.io/docs/reference/database-connectors/index
https://www.prisma.io/docs/reference/tools-and-interfaces/index
https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-cli/index
https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/index
https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/index
https://www.prisma.io/docs/understand-prisma/index
https://www.prisma.io/docs/understand-prisma/prisma-in-your-stack/index
https://www.prisma.io/docs/getting-started/index
https://www.prisma.io/docs/getting-started/setup-prisma/index

From a user and SEO perspective these pages don't bring any value because they are empty.
I would recommend to not generate them at all, that would remove them from the sitemap as a consequence.

### LINKED ISSUE #317: Make use of /index landing pages
## Problem

Our documentation is quite deeply nested in places, but only the lowest level is exposed by navigation. For example, you can't see a 'Data import and export' overview page:

image

This means:

* We're losing out on a lot of tasty SEO-food landing pages where we can write overviews or link to other content - for example, 'Deploying Prisma'.
* Users can only navigate using side nav, which will be difficult if the content gets any more nested or a particular section becomes very big. 
* People who navigate by URL (e.g. by removing a section) will hit a lot of 404s.

## Suggestion

1) Redirect / to index pages. For example:

    * https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/ should always redirect to (and ideally *rewrite* because 'index' is not a nice name) to... 
    * https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/index 

2) If a page isn't a top-level page (e.g. Reference) or staticLink = true (e.g. Tools & Interfaces), take users to the actual page when they click a nav link rather than expanding the menu.  Behaviour should be: click text --> go to page, click arrow --> expand.

3) Create a component that lists the entire sub-tree of the section you are on (fully expanded). Make this an optional component that you add to the page in any location (e.g. `<Subsections Depth="2" />`

image

4) Make **all** breadcrumb entries links - even 'Reference'. The way I can see an overview of what's contained in that entire section without having to use the nav, which is a bit small and cramped.

5) Make it possible to disable auto-generated ToC (e.g. by setting toc: false or sectionpage:true on the page):

image

(remove `<hr />` and language switcher from underneath the title if toc:false/section:true - there's no in-page ToC so it's not required)

Related: https://github.com/prisma/prisma2-docs/issues/131

### LINKED ISSUE #346: Exclude page from navigation
## Problem / Requirement

I would like to be able to create a page anywhere in the site but exclude it from the side nav. 

## Proposed solution

Some frontmatter variable like `excludeFromNav` would be great. Page should be excluded from:

* Side navigation
* In-page section navigation (created by #317 )

nilubava:
PR for #317 
Fixes #287 

1. New flags in frontmatter:
- `hidePage: true` for hiding page from nav 
- `toc: false` for hiding toc from page

2. <Subsections/> component which can be added anywhere
3. `/index` wont appear in urls anymore
4. Breadcrumbs turned to links

TODO:

- [x] Add content for index pages
- [x] <Subsection/> depth


mhwelander:
TOC is not showing up on pages without toc:false (if not specified, it should default to toc:true):

image

mhwelander:
The context page shows up at the bottom of the 'section' list:

image

mhwelander:
The side-nav expand/collapse behaviour is inconsistent - I can't quite put my finger on why, but here are some examples:

I expanded some sections (by clicking the arrow) and then clicked 'Prisma in your stack'. Although the the nav item is highlighted in bold as the context page, the  child pages did not expand:

image

In this example, two nav items are highlighted in bold (should only be one):

image

My suggestions are:

1) Allow user to expand/collapse any number of nav sections with arrows (this already happens)
2) As soon as they click on a page ---> 1) collapse all other nav sections 2) highlight only the context page/tree

mhwelander:
Small one - could we make the spacing match here:

image

Ideally by way of a `sectionList` class so that we don't break the current in-page TOC.

PR #358, opened 2020-05-15
Improve analytics

2color:


src/utils/stats.js
```diff
@@ -0,0 +1,32 @@
...
+        xhr.open('POST', COLLECT_ADDRESS, true)
+        xhr.send(model.get('hitPayload'))
+      })
+    })
+  },
+
+  trackPage(page) {
+    const { hostname } = window.location
+
+    if (hostname.includes('netlify') || hostname.includes('localhost')) {
```
Jolg42:
```suggestion
    const { host } = window.location

    if (host.includes('netlify') || host.includes('localhost')) {
```

Tiny change when I remembered that hostname is not as well supported (not supported by Opera 🤷‍♂️) as host (it's almost the same thing)
Screen Shot 2020-05-18 at 10 45 10
	https://developer.mozilla.org/en-US/docs/Web/API/Location/host

Jolg42 (approved):

src/utils/stats.js
```diff
@@ -0,0 +1,33 @@
...
+const COLLECT_ADDRESS = 'https://stats.prisma.workers.dev'
+
+module.exports = {
+  init() {
+    ReactGA.initialize(GA_TRACKING_ID, {
+      gaAddress: GA_ADDRESS,
+    })
+
+    const ga = ReactGA.ga()
+    ga('set', 'anonymizeIp', true)
```
Jolg42, line 14: ❤️

Jolg42 (approved):

PR #368, opened 2020-05-22
Wrong DATABASE_URL format

garciajg:
This changes `postgresql://` DATABASE_URL to `postgres://`

### DIFF
--- a/content/01-getting-started/02-setup-prisma/03-start-from-scratch-prisma-migrate.mdx
+++ b/content/01-getting-started/02-setup-prisma/03-start-from-scratch-prisma-migrate.mdx
@@ -122,7 +122,7 @@ datasource postgresql {
 In this case, the `url` is [set via an environment variable](../../reference/tools-and-interfaces/prisma-schema/prisma-schema-file#using-environment-variables) which is defined in `prisma/.env`:
 
 ```
-DATABASE_URL="postgresql://johndoe:randompassword@localhost:5432/mydb?schema=public"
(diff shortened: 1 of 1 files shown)


janpio:
Why do you think `postgresql` is wrong?

garciajg:
@janpio I run linux and windows and it uses `postgres://` as the protocol, as well as some people having the same issue in the Slack channel with this as well. But it could be different in whatever system you guys use

janpio:
In general Prisma works with both - we just convert to what works.

But the Postgres docs itself define `postgresql://`: 
https://www.postgresql.org/docs/10/libpq-connect.html#LIBPQ-CONNSTRING
> The URI scheme designator can be either postgresql:// or postgres://.

This is probably the reason why we usually use `postgresql://` in our docs.

garciajg:
@janpio if that works for prisma then that's fine. I'm sure most Linux/macOS users will figure it out then. I sure did. Or maybe just add some sort of comments at the top of the page, something like:
**If `postgresql://` URI gives you issues, try doing `postgres://`**. But it's up you guys at this point.

janpio:
I am sure it will not give you issues - or we have a bug that needs to be fixed :)

But adding a note somewhere in the docs might be worth it, instead of replacing all uses of it.

mhwelander:
Hi folks - I've added a note at the top of this doc #408 that you should be able to use the 2 interchangeably, but that you might need to specify if you have issues. 

Thank you for prompting the discussion! 💚  I will close this PR and merge in the note.

garciajg:
@mhwelander Thanks a lot!

PR #391, opened 2020-05-27
Pagination

### LINKED ISSUE #455: Add pagination section to API docs
## What

In a similar fashion to how we have a [Filtering](https://github.com/prisma/prisma2/blob/283be11cb0776b0c05df96354b0c6f03e0b00176/docs/prisma-client-js/api.md#filtering) section, we should add a section on pagination.

## Why
- The API has changed for `before` and `after` which now accept `*WhereUniqueInput`, e.g. `UserWhereUniqueInput` instead of a `string`. 
- It can be used to enable pagination based on arbitrary uniques.

## Example

Before:
```js
const paginatedPosts3 = await prisma.post.findMany({
  first: 5,
  after: 43834,
})
```

After:
```js
const paginatedPosts3 = await prisma.post.findMany({
  first: 5,
  after: { id: 43834},
})
```

mhwelander:


dpetrick:
This needs an overhaul based on: https://github.com/prisma/specs/pull/472

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
+---
+title: 'Pagination'
+metaTitle: 'Pagination (Reference)'
+metaDescription: 'Prisma Client provides features that are not achievable with relational databases. These features are referred to as "polyfills" and explained on this page.'
+---
+
+## Overview
+
+Prisma Client supports both offset and cursor-based pagination.
```
janpio:
```suggestion
Prisma Client supports both offset- and cursor-based pagination.
```
Not sure if this is correct.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+## Offset pagination
+
+Offset pagination combines `skip` and `take` to skip a certain number of results and select a limited range. The following query skips the first 3 `Post` records and returns records 4 - 7:
```
janpio: nit: "combine" on concept that were not introduced before felt weird to me

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+  take: 4,
+})
+```
+
+![](pagination/offset-skip-take.png)
+
+### ✔ Pros of offset pagination
+
+- You can jump to any page immediately. For example, you can `skip` 200 records and `take` 10, which simulates jumping straight to page 21 of the result set. This is not possible with cursor-based pagination.
+- You can paginate the same result set in any sort order. For example, you can jump to page 21 of a list of `User` records sort by first name. This is not possible with cusor-based pagination, which requires sorting by a unique, sequential column.
```
janpio:
```suggestion
- You can paginate the same result set in any sort order. For example, you can jump to page 21 of a list of `User` records sorted by first name. This is not possible with cursor-based pagination, which requires sorting by a unique, sequential column.
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+```ts line-number
+const results = await prisma.post.findMany({
+  skip: 3,
+  take: 4,
+})
+```
+
+![](pagination/offset-skip-take.png)
+
```
janpio, line 23: Something like this might be handy here:
```suggestion

To implement pages of results, you would just `skip` the number of pages multiplied by the number of results you show per page.

```

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+### ✔ Pros of offset pagination
+
+- You can jump to any page immediately. For example, you can `skip` 200 records and `take` 10, which simulates jumping straight to page 21 of the result set. This is not possible with cursor-based pagination.
+- You can paginate the same result set in any sort order. For example, you can jump to page 21 of a list of `User` records sort by first name. This is not possible with cusor-based pagination, which requires sorting by a unique, sequential column.
+
+### ✘ Cons of offset pagination
+
+- Offset pagination **does not scale** at a database level. For example, if you skip 200,000 records and take the first 10, the database still has to traverse the first 200,000 records before returning the 10 that you asked for - this affects performance.
```
janpio:
```suggestion
- Offset pagination **does not scale** at a database level. For example, if you skip 200,000 records and take the first 10, the database still has to traverse the first 200,000 records before returning the 10 that you asked for - this negatively affects performance.
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+### Use cases for offset pagination
+
+- Shallow pagination of results. For example, a blog interface that allows you to filter `Post` records by author and page through the first few pages.
+
+### Example: Filtering and offset pagination
+
+The following query returns all records where the `email` field contains `"prisma.io"`. The query skips the first 40 records and recods 41 - 51.
```
janpio:
```suggestion
The following query returns all records where the `email` field contains `"prisma.io"`. The query skips the first 40 records and returns records 41 - 50.
```

janpio: quotes around search string intentional inside code block?

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+Without `skip: 1`, the second query returns 4 results after (and _including_) the cursor:
+
+![](pagination/cursor-3.png)
+
+If you `skip: 1`, the cursor is not included:
+
+![](pagination/cursor-2.png)
+
+You can choose to `skip: 1` or not depending on the pagination behavior that you want.
```
janpio, line 149: Comment re implementation, not documentation: This is weird.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+#### Can I guess the value of the cursor?
+
+If you guess the value of the next cursor, you will page to an unknown location in your result set. Although IDs are sequential, you cannot predict the rate of increment (`2`, `20`, `32` is more likely than `1`, `2` `3`, particularly in a filtered result set).
+
+#### Does cursor-based pagination use the concept of a cursor in the underlying database?
+
+No, cursor pagination does not [use cursors in the underlying database](https://www.postgresql.org/docs/9.2/plpgsql-cursors.html).
```
janpio:
```suggestion
No, cursor pagination does not use cursors in the underlying database ([e.g. PostgreSQL](https://www.postgresql.org/docs/9.2/plpgsql-cursors.html)).
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+### ✔ Pros of cursor-based pagination
+
+- Cursor-based pagination **scales**. The underlying SQL does not use `OFFSET`, but instead queries all `Post` records with an ID greater than X.
```
janpio: Possibly also use `OFFSET` above in the description of offset navigation?

janpio: Possibly:
```suggestion
- Cursor-based pagination **scales**. The underlying SQL does not use `OFFSET`, but instead queries all `Post` records with an ID greater than the value of `cursor`.
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+### ✘ Cons of cursor-based pagination
+
+- You must sort by your cursor, which has to be a unique, sequential column.
+- You cannot jump to a specific page number - you must traverse pages 1 - 20 in order to know what the cursor should be for the query that returns page 21.
```
janpio: Couldn't I also do a query with LIMIT 200 (in example with page size 10) and get the ID of the last entry to use as a cursor? Or is that meant by "traverse"?

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+  },
+  orderBy: {
+    id: 'asc',
+  },
+})
+```
+
+### Sorting and cursor-based pagination
+
+Cursor-based pagination requires you to sort by a sequential, unique column such as an ID or a timestamp. This value - known as a cursor - bookmarks your place in the result set and allows you to request the next set.
```
janpio, line 194: This feels repeated from above?

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+### Paging backwards with cusor-based pagination
```
janpio: Is this an example as well?

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+- You can jump to any page immediately. For example, you can `skip` 200 records and `take` 10, which simulates jumping straight to page 21 of the result set. This is not possible with cursor-based pagination.
+- You can paginate the same result set in any sort order. For example, you can jump to page 21 of a list of `User` records sort by first name. This is not possible with cusor-based pagination, which requires sorting by a unique, sequential column.
+
+### ✘ Cons of offset pagination
+
+- Offset pagination **does not scale** at a database level. For example, if you skip 200,000 records and take the first 10, the database still has to traverse the first 200,000 records before returning the 10 that you asked for - this affects performance.
+
+### Use cases for offset pagination
+
+- Shallow pagination of results. For example, a blog interface that allows you to filter `Post` records by author and page through the first few pages.
```
rahul3v:
```suggestion
- Shallow pagination of results. For example, a blog interface that allows you to filter `Post` records by its author and shows the filtered results in N number of pages (with limited results per page as set).
```

rahul3v (changes requested):

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: Added to pros of offset pagination.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: Title updated.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: I did that on purpose so that the structure matches offset pagination - I anticipate that it's something people will trip up on.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: No, not intentional - fixed, thanks!

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: Updated this to say you cannot use *only* a cursor to jump to page X, because you don't know what the cursor is going to be. Removed the word 'traverse'.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: Changed to 'offset pagination and cursor-based pagination'.

content/03-reference/01-tools-and-interfaces/02-prisma-client/08-pagination/index.mdx
mhwelander: Fair - changed to 'uses' in both instances.

PR #396, opened 2020-05-28
Multiple environment files

mhwelander:


content/03-reference/01-tools-and-interfaces/01-prisma-schema/index.mdx
```diff
@@ -159,6 +159,31 @@ dotenv.config() // load the environment variables
...
+    ```
+
+2.  Create an `.env` file in, for example, the root of your project.
+
+3.  Include a reference to `dotenv` at the top of your project's entry file.
+
+    ```ts
+    import { config } from 'dotenv'
+    config()
+    ```
```
janpio: Isn't this an alternative to the global installation of the dotenv CLI?

content/03-reference/01-tools-and-interfaces/01-prisma-schema/index.mdx
```diff
@@ -171,10 +196,10 @@ Here are some different examples:
...
 /// This comment will get attached to the `User` node in the AST
 model User {
-  /// This comment will get attached to the `id` node in the AST
-  id      Int    @default(autoincrement())
-  // This comment is just for you
-  weight  Float /// This comment gets attached to the `weight` node
+/// This comment will get attached to the `id` node in the AST
+id      Int    @default(autoincrement())
+// This comment is just for you
+weight  Float /// This comment gets attached to the `weight` node
```
janpio: this change looks unwanted

content/03-reference/01-tools-and-interfaces/01-prisma-schema/index.mdx
mhwelander: @ryands17 Was there a specific reason why this was suggested to the person on Slack? Want to make sure I don't lose context.

ryands17: @mhwelander Yes it's an alternative if they do not want to use `dotenv-cli` in the command but directly in the file. I had given an example of using that while seeding the database initially.

mhwelander: Ahh, then I misunderstood. To clarify:

1) Either you use dotenv -e .env3 in the command

OR

2) Include this in your file:

```
import { config } from "dotenv";
config({ path: ".env3" });
```

Is that right?

ryands17: Yep that is correct :)

mhwelander: Fixed!

content/03-reference/01-tools-and-interfaces/01-prisma-schema/index.mdx
mhwelander: Fixed!

content/03-reference/01-tools-and-interfaces/01-prisma-schema/index.mdx
```diff
@@ -159,6 +159,33 @@ dotenv.config() // load the environment variables
 console.log(`The connection URL is ${process.env.DATABASE_URL}`)
 ```
 
+### Use multiple .env files or .env files outside the prisma folder
+
+To use multiple `.env` files:
+
+1.  Install [dotenv-cli](https://www.npmjs.com/package/dotenv-cli):
```
janpio, line 166: I think dotenv-cli is not needed at all for the `import { config } from 'dotenv'` part, but you need to install `dotenv` into your project to be able to do that.

PR #400, opened 2020-06-01
[FIX] Start From Scratch - broken TS schema in "Write data into the database" section

SMenigat:
When following along the docs I stumbled over a minor error in the [`Start From Scratch` / `Write data into the database`](https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch-prisma-migrate-typescript-mysql#write-your-first-query-with-prisma-client) section.
When just `copy` & `paste`ing the examples the following TS error gets thrown:

ts-error

A quick look reveals some spelling errors in the create syntax. When fixed as proposed the code execution yields the correct result:

fixed-ts-error


mhwelander:
Good catch and thanks for raising a PR! 💚  I've updated the schema file instead (`posts` and `profile`) as we want to keep the example consistent with other pages.

PR #405, opened 2020-06-04
fix(code): make bg more bright

timsuchanek:
To have higher contrast and ease readability for the code blocks, I suggest using the same background color as the default GitHub syntax highlighting.

## Before

image

## After

image

### DIFF
--- a/src/components/layout.css
+++ b/src/components/layout.css
@@ -16,7 +16,7 @@ body {
   --white-color: #ffffff;
   --link-color: #3182ce;
   --border-color: #e2e8f0;
-  --code-bgd-color: #edf2f7;
+  --code-bgd-color: #f6f8fa;
   --list-bullet-color: #a0aec0;
   --search-highlight-bg-color: #ebf8ff;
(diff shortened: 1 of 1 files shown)


kuldar:
I like the change for big code blocks. 👍 Wondering if we should keep the darker one for inline code snippets though, the lighter background kinda disappears at very small scale.

timsuchanek:
That's a good idea!
And btw the only thing I don't like about the lighter color is that there is less blue in there. We could play around a bit more to find a light color with more blue as well.

mhwelander:
Added an item to the Design board: https://www.notion.so/prismaio/Docs-Review-code-block-color-inline-and-block-9f468709d8794740b513718a248799d8

hrescak:
Following up on this here - I think this is reasonable. As far as the right color to use - it looks like the color of the actual page background is close to the lightness @timsuchanek used and is (obviously) in our palette. Thoughts @kuldar ?

mhwelander:
Taking Design team's comments as approval and merging. :)

PR #413, opened 2020-06-04
Update MySQL Connector Docs

diegocouto:
Hey there 👋 

This update is intended to fix an issue [I've discussed](https://github.com/prisma/prisma/issues/1508#issuecomment-638870425) on prisma/prisma#1508. The `host` argument seems to be the right way to set up a socket connection on PostgreSQL, [but not to MySQL](https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-connection-socket.html).

Thanks a lot!


pantharshit00 (approved):
Thanks

PR #415, opened 2020-06-05
findMany input object type updates

rahul3v:
Well, Now the prisma-client generates "**cursor** and **take**" as of the **findMany**  object type removing "**after**, **before**, **first** and **last**"


mhwelander:
🙏 💚 Thank you!

rahul3v:
You also have to change the same in 'https://www.prisma.io' homepage under **Explore the Prisma Client API** - **Pagination**

I'm not able to find the homepage doc. so that i can refer the changes and few more thing to change in it, either send me the file link i have to change (so that i can request the new commit) or you can directly make the changes

janpio:
The website is a different, for now internal repository. 
I created an issue there with a copy of your message.
And started a discussion if we can also open source that repository.

rahul3v:
Well @janpio,
That will be great, Let me know if anyway I can access, and can make the changes for modification

PR #416, opened 2020-06-05
Proposal for enhancing Prisma Studio docs

### LINKED ISSUE #607: Add feature overview and screenshot to Prisma Studio

robin-macpherson:
## Description

I would like to propose enhancing the documentation for Prisma Studio which is currently very sparse. I absolutely love this tool and think that it truly strikes a fantastic balance of abstraction.

This is just a start, but if we agree that this is a great time to work on this, I'd love to add more detail such as showing a comparison between how tables look in the underlying database and how they are represented in Prisma Studio in a way that is based on the datamodel.

### DIFF
--- a/content/03-reference/01-tools-and-interfaces/06-prisma-studio.mdx
+++ b/content/03-reference/01-tools-and-interfaces/06-prisma-studio.mdx
@@ -7,7 +7,13 @@ experimental: true
 
 ## Overview
 
-Prisma Studio is a visual editor for your database. To run Prisma Studio, run the following command in your terminal:
+Prisma Studio is a very powerful database GUI where you can interact with your data.
+
+## Not Your Typical Database GUI
+
+Prisma Studio is different from a typical database GUI program (such as [TablePlus](https://tableplus.com/)) in that it provides a layer of abstraction which allows you to see your data represented as it is in your Prisma datamodel. This is one of the several ways that Prisma bridges the gap between how you structure and interact with your data in your application and how it is actually structured and represented in the underlying database. One benefit of this is that it helps you to build intuition and understanding of these two linked but separate layers over time.
+
+To run Prisma Studio, run the following command in your terminal:
 
 ```console
 npx prisma studio --experimental


robin-macpherson:
cc: @mhwelander 
bcc: @nikolasburk

content/03-reference/01-tools-and-interfaces/06-prisma-studio.mdx
```diff
@@ -7,7 +7,13 @@ experimental: true
 
 ## Overview
 
-Prisma Studio is a visual editor for your database. To run Prisma Studio, run the following command in your terminal:
+Prisma Studio is a very powerful database GUI where you can interact with your data.
```
nikolasburk, line 10: let's try to be a bit more "neutral" here -> "very powerful database GUI" is a bit too much IMO 🙈

content/03-reference/01-tools-and-interfaces/06-prisma-studio.mdx
```diff
@@ -7,7 +7,13 @@ experimental: true
...
+
+## Not Your Typical Database GUI
```
nikolasburk, line 12: also not too sure about the wording here, sounds a bit too marketing-y for a page in the reference docs to me. if this section is about comparing Prisma Studio to other DB GUIs, maybe something like "How Prisma Studio is different from other database GUIs" might work better?

content/03-reference/01-tools-and-interfaces/06-prisma-studio.mdx
```diff
@@ -7,7 +7,13 @@ experimental: true
...
+
+Prisma Studio is different from a typical database GUI program (such as [TablePlus](https://tableplus.com/)) in that it provides a layer of abstraction which allows you to see your data represented as it is in your Prisma datamodel. This is one of the several ways that Prisma bridges the gap between how you structure and interact with your data in your application and how it is actually structured and represented in the underlying database. One benefit of this is that it helps you to build intuition and understanding of these two linked but separate layers over time.
```
nikolasburk, line 14: > which allows you to see your data represented as it is in your Prisma datamodel

- "datamodel" should be spelled as two words: "data model" (sorry this is confusing, in Prisma 1 we had the convention to spell it as one word but with Prisma 2 it's two words). Ideally it can also link to the "Data model" docs page: https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/data-model

nikolasburk (changes requested):
Thanks so much for this Robin! Overall I think it looks great and we can definitely use a bit more content on this page. I just pointed out a few formulations that sound a bit too matketing-y for a reference docs page IMO, I think the tone here should be as neutral as possible (but would also love to hear @mhwelander's thoughts on this) 🙂 

Also, just FYI, we already have an open (internal) "issue" for the Studio page where we want to add:

- the different "features" of Studio, e.g. browsing data, modifying data, table vs tree view,...
- ideally add a number of screenshots per feature

Just as a side-note in case you want to take a stab at this as well 😄

robin-macpherson:
> Thanks so much for this Robin! Overall I think it looks great and we can definitely use a bit more content on this page. I just pointed out a few formulations that sound a bit too matketing-y for a reference docs page IMO, I think the tone here should be as neutral as possible (but would also love to hear @mhwelander's thoughts on this) 🙂
> 
> Also, just FYI, we already have an open (internal) "issue" for the Studio page where we want to add:
> 
> * the different "features" of Studio, e.g. browsing data, modifying data, table vs tree view,...
> * ideally add a number of screenshots per feature
> 
> Just as a side-note in case you want to take a stab at this as well 😄

Thanks for this feedback, @nikolasburk ! I'll be sure to apply your suggestions and also try to expand things a bit. Also, I'd be happy to take a stab at that internal "issue" you mentioned. Could you link me to the issue or let me know how I should go about taking it on?

I'll work on these changes as soon as I can, hopefully this evening after work!

PR #418, opened 2020-06-06
Fix a minor type

apnerve:
Changed `shsould` to `should`


mhwelander:

PR #419, opened 2020-06-07
Fix user model annotation on 1-1 relations

inakiarroyo:
Example is annotated on the `User` model rather than on the `Profile` one


mhwelander:
Good catch, thank you for contributing! 💚 🚀

Issue #423, opened 2020-06-09
Links do not render correctly on Windows


mhwelander:
## Problem

All links in MDX files render like this on Windows:

```
<a href="C:%5Cgetting-started%5Csetup-prisma%5Cadd-to-existing-project/">Add Prisma to an existing project</a>
```

rahul3v:
I have fixed the issue, how can i push my code for this specific issue

rahul3v:
The issue just resolved, you can close this now.!

PR #463, opened 2020-06-12
Links do not render correctly on Windows #423

rahul3v:
The error has been resolved and tested by running locally ...!


src/layouts/articleLayout.tsx
```diff
@@ -8,6 +8,7 @@ import SEO from '../components/seo'
 import { graphql } from 'gatsby'
 import MDXRenderer from 'gatsby-plugin-mdx/mdx-renderer'
 import { useNavigate } from '@reach/router'
+import { MDXProvider } from "@mdx-js/react"
```
nilubava: @rahul3v This is not the right place to define new MDXProvider components. It should be defined in `src/components/customMdx/index.tsx`
Also, this is a path issue which needs to be resolved on the plugin level, which you can find inside `plugins/` folder

nilubava (changes requested):

src/layouts/articleLayout.tsx
rahul3v: ok @nilubava, Well there are too many problems, even if we re fix the only file root path, the link still breaks , so i have to write code to run specifically for local run, see the next commit

nilubava:
@rahul3v It works on Windows. Thanks!Great job!
Once you resolve the conflicts, we can merge it.

rahul3v:
Well @nilubava, I have reviewed all the links, the updated commit was working fine ...!, the warning links will come just because of using some 'redirects' and avoiding '/index' link format.  So, it's not an issue,
you can merge the last commit

rahul3v:
@nilubava  you can merge it now

PR #469, opened 2020-06-12
temp: remove search

janpio:


sdnts (approved):

ultrox:
Well, I see its flagged with temp. But, I kind of use search all the time. What motivated this action?

janpio:
We had a regression in our search index which caused the site to max out a CPU core for ~30 seconds, often crashing the whole browser tab for people along the way. As I don't directly work on this project and had no idea how to fix this in the moment, removing the search was the most practical way forward until someone could properly take care of it.

And that happened in https://github.com/prisma/docs/commit/4206ea444ff40265fa5f93cb137f172654063367 and search is back and snappier as before :D

ultrox:
@janpio thanks so much for detailed answer, this make sense, I would say even a raise.

PR #480, opened 2020-06-16
Client constructor

### LINKED ISSUE #446: Allow datasource credentials to be passed by client
## Problem
In production, we store secrets in a vault for secrets and we do not expose them as environment variables. The service requests a secret from the vault (database credentials) and receives them if authorized, those secrets are stored in memory only and could be unique for this service (the hashicorp Vault can generate database credentials on the fly). We thus need to tell the connection string at runtime when creating the client, which happens just after retrieving the secrets.

## Solution
 It seems like there is an undocumented way to pass a [datasource](https://github.com/prisma/prisma-client-js/blob/master/packages/photon/src/runtime/getPrismaClient.ts#L45) to the client. It might be a way to override the datasource specified in the schema?

## Additional context
Eventually we will also need the same way of securely passing the secret to the migration, most likely by having a programmatic access to the `migrate up` (if we don't already).

mhwelander:
Client constructor parameters page.


content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
```diff
@@ -0,0 +1,70 @@
...
+2. Instantiate a client and override the connection string of the `my_database` datasource:
+
+   ```ts
+   import { PrismaClient } from '@prisma/client'
+
+   const prisma = new PrismaClient({
+     datasources: {
+   |       my_database: 'file:./dev_qa.db',
+     },
+   })
```
janpio: The moment we document this, this is our public API.

There was an issue about this being inconsistent with the structure of the `datasources` block in the schema, but that was closed: https://github.com/prisma/prisma-client-js/issues/698

Are we sure we are ok with this mismatch and that it is not possible to override the `provider` here?

Ping @prisma/frontend

Jolg42: We can’t add the provider here because the client is generated before and can't change dynamically here (need to generate a new client)

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+See [Logging](logging) for more information.
+
+## `errorFormat`
+
+Use the `errorFormat` constructor parameter to determine the level and formatting of errors returned by Prisma. The following configuration outputs raw error messages without formatting or extended information about the error:
+
+```ts
+const prisma = new PrismaClient({
+  errorFormat: 'minimal', // Defaults to 'pretty'
```
timsuchanek: It defaults to `colorless`
```suggestion
  errorFormat: 'minimal', // Defaults to 'colorless'
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
timsuchanek: As Joël mentioned, we can't have the provider here because it needs to be static in generate time, that's why we made the simplification to directly allow the url instead:


What you would expect if it should 100% match the schema structure:
```ts
{
  datasources: {
    db: {
      provider: 'mysql',
      url: 'file:dev.db'
    }
  }
}
```
What we have instead
```ts
{
  datasources: {
    db: 'file:dev.db'
  }
}
```

We could say, that maybe one day there would be additional things to configure about the data source that can be defined in runtime. Then it would make sense to not only allow the string, but keep it in an object to have space for other params.
However, that is very unlikely, because as soon as we put runtime information outside the connection string, the purpose of the connection string would be defied.

janpio: Ok, `provider` was a bad justification then :p 

When I last looked at the structure it currently has, I could not connect the dots as it was missing the `url` thing. So I misunderstood `db` to be a real key instead of referring to the value from the schema - and it did not work

I think
```
{
  datasources: {
    db: {
      url: 'file:dev.db'
    }
  }
}
```

would be better here and make this connection clearer and more obvious as it is a 1:1 representation of the schema structure.

Jolg42:
Just a random question...

Netlify says here in GitHub `Deploy preview ready` but this page is blank https://deploy-preview-480--prisma2.netlify.app/docs/reference/tools-and-interfaces/prisma-client

So it's not possible to preview the PR I guess?

mhwelander:
@Jolg42 We have a bit of an issue with the /docs subdirectory - if you remove it, you can see the page: https://deploy-preview-480--prisma2.netlify.app/reference/tools-and-interfaces/prisma-client/

Jolg42:
Ok so here it's https://deploy-preview-480--prisma2.netlify.app/reference/tools-and-interfaces/prisma-client/constructor/ and it works thanks <3

Jolg42:
Spotted small things @mhwelander 

Replace 
- `prisma.schema` by `schema.prisma`
- `Define a datasources` by `Define a datasource`
Weird indentation here
Screen Shot 2020-06-17 at 17 41 46

mhwelander:
Fixed - thank you! I'll wait for an update on whether the format of `datasources` will change.

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
```diff
@@ -0,0 +1,70 @@
...
+Use the `datasources` constructor parameter to programatically override the URL of a datasource - for example, as part of an integration test. To override a datasource:
+
+1. Define a datasource in the `schema.prisma` file. In the following example, there is a single datasource named `my_database`:
+
+   ```prisma
+   datasource my_database {
+    provider = "sqlite"
+    url = env("DATABASE_URL")
+   }
+   ```
```
Jolg42:
```suggestion
   ```prisma
   datasource my_database {
     provider = "sqlite"
   }
   ```
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
```diff
@@ -16,10 +16,9 @@ Use the `datasources` constructor parameter to programatically override the URL
 
    ```prisma
    datasource my_database {
-    provider = "sqlite"
-    url = env("DATABASE_URL")
+     provider = "sqlite"
+     url      = env("DATABASE_URL")
    }
```
Jolg42, line 21:
```suggestion
   }
   ```
```

Jolg42:
I broke something and fixed it but in the end the indentation looks better 😄

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
timsuchanek: From my side, nothing speaks against that, it makes sense to me.
Nobody uses this API anyway yet, so we can just change it.
Any objections @Jolg42 ?

timsuchanek: Sorry @mhwelander for turning this into a product meeting :D

Jolg42: I was also +1 to change it when I first encountered it :)

janpio: Ok, let's change it and block @mhwelander until it is done 👍

janpio: https://github.com/prisma/prisma/issues/2801

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
```diff
@@ -0,0 +1,70 @@
...
+   > **Note**: You must re-generate the Prisma Client each time you add or rename a datasource. Datasource names are included in the generated client.
+
+2. Instantiate a client and override the connection string of the `my_database` datasource:
+
+   ```ts
+   import { PrismaClient } from '@prisma/client'
+
+   const prisma = new PrismaClient({
+     datasources: {
+   |   my_database: 'file:./dev_qa.db',
```
rahul3v: I think it has some typo error
```suggestion
       my_database: 'file:./dev_qa.db',
```

rahul3v (changes requested):

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
janpio: That is to highlight this line I think.

janpio (changes requested):
Waiting for https://github.com/prisma/prisma/issues/2801

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
rahul3v: Yup that's correct, you are right about it, I locally run it, My bad 😐😃

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
timsuchanek: I implemented it as @janpio suggested. So we can document it like this now 🙂

Jolg42: @mhwelander Tim just unblocked this PR :)
It now need to be updated to reflect it like this (new url property inside db as an object)
```
{
  datasources: {
    db: {
      url: 'file:dev.db'
    }
  }
}
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
janpio:
```suggestion
   |   my_database: {
   |     url: 'file:./dev_qa.db'
   |   }
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
```diff
@@ -0,0 +1,70 @@
...
+     },
+   })
+   ```
+
+> **Note**: If you define multiple datasources in `schema.prisma`, Prisma defaults to the first entry.
```
janpio, line 46: Why is that? Do we not define which one it should be applied to by using `my_database`?

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
janpio: Suggested the required change: https://github.com/prisma/docs/pull/480/files#r442789126

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
janpio: cc @timsuchanek

Jolg42: There was a enabled property in the schema before that got removed (never saw it, just the remains)

Jolg42:
@janpio I think you can unblock that PR now 😃

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
janpio: What is the connection to this comment? Should it not use the datasource with the name you supply above, `my_database` in this example?

janpio (approved):
One comment left for the Note, rest looks good.

content/03-reference/01-tools-and-interfaces/02-prisma-client/02-constructor.mdx
timsuchanek: This is actually touching on undefined behavior, I think the **Note** is correct, but we need to fix that.
Even though you provide a specific datasource as a constructor arg, it could even be, that in our current implementation, it might not be used, if it's not the first datasource in the schema.
We don't have tests for that and didn't define this behavior clearly yet, but should definitely do that asap.

Jolg42:
Yeah 🎊 😄

PR #490, opened 2020-06-21
Fix typo on docs mysql => sqlite (datasource)

neibla:


mhwelander:
Thanks for taking the time to fix this! 💚 🙏

Issue #485, opened 2020-06-18
Fix 92 broken links (Netlify build logs shows broken links)


Jolg42:
Thread discussed with @mhwelander @nilubava @janpio 
https://prisma-company.slack.com/archives/C5Z9TH6N9/p1592476676060100

Example build
https://app.netlify.com/sites/prisma2/deploys/5eeb43876c6531000770c863
<img width="989" alt="Screen Shot 2020-06-18 at 14 18 03" src="https://user-images.githubusercontent.com/1328733/85019101-8d198400-b16e-11ea-9079-f79f7876475d.png">

Related https://github.com/prisma/docs/issues/483

rahul3v:
Maximum of the warning links will come just **because of using some 'redirects' and avoiding '/index' link format**. So, it's not an issue, I was seeing the same issue while i was fixing #463, but the link works fine
`/prisma-schema-file` is using redirects
`'/index'` we are not using  this format


**But I see some of the warning links here as an issue** because of little incorrectly resolving the paths (see warnings like: 21:364, 136:3, 137:3) in above image,

janpio:
> Maximum of the warning links will come just **because of using some 'redirects' and avoiding '/index' link format**. So, it's not an issue, I was seeing the same issue while i was fixing #463, but the link works fine
> `/prisma-schema-file` is using redirects
> `'/index'` we are not using this format

That is also a problem - and either the link checker needs to be fixed or the links need to be adapted to not warn any more in some way.

rahul3v:
Well, I do have fix it upto **14 warn errors**, Just finding to fix it till 0

rahul3v:
@mhwelander , I want you to confirm , where exactly you wanted to point this links to in that specific files
```
warn 1 broken links found on /getting-started/setup-prisma/add-to-existing-project\
warn 594:94   #nested-writes      (there are two more in other files, but one can help)

warn 1 broken links found on /reference/tools-and-interfaces/prisma-schema/data-model\
warn 251:36   #fields
⠀
warn 1 broken links found on /reference/tools-and-interfaces/prisma-client/filtering\
warn 186:134  #no-and
```

**I have fixed this issue** just need to confirm some confusing **non-linking warning links**

mhwelander:
@rahul3v Fixed the links in your list - thanks for pointing them out, and for all the time and effort you are putting into improving all things link-related on the Prisma doc site!

rahul3v:
[...]

Issue #483, opened 2020-06-18
GitHub Action: Check for broken links


mhwelander:
> (How it would work: Github workflow that just runs that part of the build script, red x when it fails and green checkmark if it is happy)

Full thread: https://prisma-company.slack.com/archives/C5Z9TH6N9/p1592476676060100

rahul3v:
Can I see the Full thread if possible

janpio:
Unfortunately not (as people did not expect this to become public), but I can paraphrase:

- We have a build step that outputs warnings in the Netlify console about broken links:
![image](https://user-images.githubusercontent.com/183673/85197110-3d0dff00-b2df-11ea-88cf-5c35b6ec6b12.png)
- This issue is about the idea to turn that into a Github action that produces a check on all PRs that would fail if there are broken links, and pass if there are not.
- The script or build step can probably be extracted and modified to not only warn, but fail on errors.
- That of course also requires the codebase to be really broken link free when we start.

Does that give you enough context to understand what we meant?

rahul3v:
That will be enough @janpio, No problem at all, Well what I understand if we fix the warning of broken links, I think we can fix this
and I already did that on my local deployment, Just waiting for some response from @mhwelander to confirm about  non-linking of few warning links in #485, then I will make pull request with the fix.

mhwelander:
Removed some of the `<a id="test"></a>` anchor links for now and used a nearby heading (needs must) - feature request #508 here. There are 0 now (ignore Netlify - there's an issue: #506)

Big thanks @rahul3v for all the work you've put into the link checker 🚀

janpio:
But we do not have an action that runs automatically and fails PRs via GH Action yet, right?

nilubava:
Closing this since we do not have a Github Action for broken links but rather a plugin which outputs the broken links as errors during the build process.

PR #502, opened 2020-06-23
Fix : broken links #485, #483

rahul3v:
Fixed the issue with some aditional link changes as needed, enebled the custom-header-Id,  since &lt;a> concept is not working correctly

Fix for  #485,  #483

### DIFF
--- a/content/03-reference/01-tools-and-interfaces/01-prisma-schema/05-models.mdx
+++ b/content/03-reference/01-tools-and-interfaces/01-prisma-schema/05-models.mdx
@@ -8,9 +8,9 @@ metaDescription: 'Learn about defining your application models with Prisma. Mode
 
 This is an extension of the [Data model](data-model) page that discusses _models_ in the data model definition in detail.
 
-Models represent the entities of your application domain. They are defined using `model` blocks in the data model. In the [example data model](data-model#example), `User`, `Profile`, `Post` and `Category` are models.
+Models represent the entities of your application domain. They are defined using `model` blocks in the data model. In the [example data model](data-model#example), `User`, `Profile`, `Post` and `Category` are models. Here's the `User` model from the example for reference:
 
-#### Here's the `User` model from the example for reference {#example}
+<a id="example"/>
 
 ```prisma
 model User {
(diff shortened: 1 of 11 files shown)


content/03-reference/01-tools-and-interfaces/01-prisma-schema/05-models.mdx
```diff
@@ -8,9 +8,9 @@ metaDescription: 'Learn about defining your application models with Prisma. Mode
 
 This is an extension of the [Data model](data-model) page that discusses _models_ in the data model definition in detail.
 
-Models represent the entities of your application domain. They are defined using `model` blocks in the data model. In the [example data model](data-model#example), `User`, `Profile`, `Post` and `Category` are models. Here's the `User` model from the example on the again for reference:
+Models represent the entities of your application domain. They are defined using `model` blocks in the data model. In the [example data model](data-model#example), `User`, `Profile`, `Post` and `Category` are models.
 
-<a id="example" />
+#### Here's the `User` model from the example for reference {#example}
```
mhwelander:
```suggestion

<a id="example">

Here's the `User` model from the example for reference:
```

mhwelander: This shouldn't be a title. `<a id="example">` works as far as I can see: https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/models#example

rahul3v: Yup that's correct, **But** the problem here is we can only able to access **'id' value** of any headers as a link. I really thought about it, and then I created it as H4 header, since H5-H6 formats are not looking well, 

I think I got something now , I have to check it first. then I will let you know. give me little more time

rahul3v: Now it works fine, Now we have both options, 
- Can use short custom links for any headers, Just provide it with custom-id, like `### header {#custom-id}` and use it like `[Mylink](#custom-id)`
- Can use `<a id="link-id"/>` and use it like `[Mylink](page-name#link-id)`

content/03-reference/01-tools-and-interfaces/02-prisma-client/05-filtering.mdx
```diff
@@ -210,7 +210,7 @@ const result = await prisma.post.findMany({
 })
 ```
 
-The following format returns the same results <a id="no-and"></a>:
+#### The following format returns the same results as above **without** `AND` operator {#no-and}
```
mhwelander:
```suggestion
#### The following format returns the same results as above **without** `AND` operator
```

<a id="no-and"></a>

PR #507, opened 2020-06-23
Little update for #502

rahul3v:
@mhwelander, you used `<a>`, so you have to add  it with the page-name, missed in #502


rahul3v:
@mhwelander you have removed the `<a id="no-and"></a>` from this file, and also made a lot of changes in other content files as I have provided the fix in #502 , well may be all of your changes could work now, but you just have merged this change, so it again will break for this page, **I don't know how but**  `<a id="no-and"></a>`  is showing on my commit, but **it's not there in this file changed**

I also checked it to the `prisma:master`, still the same is missing

So, should I re-pull the #502 commits to confirm other changes, if you want it the same, or you can skip it with your changes

Issue #498, opened 2020-06-22
Option to disable line-numbers 


mhwelander:
For console output, for example:

![image](https://user-images.githubusercontent.com/849648/85311725-25d53a00-b4b6-11ea-8952-e2c636c1c4a5.png)

rahul3v:
@mhwelander, where exactly this needed, can you provide the link of the page

PR #512, opened 2020-06-23
Fix : Adding disable line-number option #498

rahul3v:
Use it just by adding `no-lines` at the top of the code where you have to disable line-numbers #498

like 
```
```sql no-lines
-------------
----------------
`` `
```


nilubava (approved):
LGTM :)

PR #519, opened 2020-06-24
docs: minor typo fix

camsloanftc:
change pyublic to public


mhwelander:
Good catch, thank you for contributing! 💚

PR #523, opened 2020-06-25
Remove duplicate example reference of `rest-nextjs`

kingRayhan:
`rest-nextjs` example appears in two times.

image


janpio (changes requested):
One is Javascript, the other Typescript.

kingRayhan:
oh, my mistake

PR #525, opened 2020-06-26
Typo: Filter all `Post` records examples

nickbytes:
I believe there are missing brackets here.

Current:
```js
const filteredPosts = await prisma.post.findMany({
  where: {
    OR: [
      { title: { contains: "hello" },
      { content: { contains: "hello" },
    ],
  },
})
```

Proposed:

```js
const filteredPosts = await prisma.post.findMany({
  where: {
    OR: [
      { title: { contains: "hello" } },
      { content: { contains: "hello" } },
    ],
  },
})
```

Running the example with the Proposed solution works.


mhwelander:
Good spot, thank you for taking the time to submit a pull request! 💚

PR #529, opened 2020-06-27
Fix ahmed tools links

AhmedElywa:
Update `create-nexus-type` and `cascading deletes` links  in upgrade from Prisma 1


mhwelander:
Thank you! 💚

PR #559, opened 2020-07-01
docs(deployment): fix incorrect Serverless link

nampdn:
I think the installation link of the `Serverless Framework CLI` is not correctly documented.


mhwelander:
Thank you! 💚

PR #588, opened 2020-07-03
Update 03-postgresql.mdx

angad777:


mhwelander:
Good spot, thanks for taking the time to raise a PR! :)

PR #598, opened 2020-07-06
 fix(setup-prisma)-2: Show code change as dif

rahul3v:
An addition to #595


mhwelander:
Thank you as always! 💚

PR #604, opened 2020-07-06
Updated transactions section

thebiglabasky:
The transactions page was providing explanations about potentially supported features.
This change removes these to ensure only supported capabilities are documented.


janpio (commented):
👍

PR #622, opened 2020-07-07
Update 06-relations.mdx

ryands17:
- Fixed a line in the doc
- Added the required index for implicit m-n relations
- Added `SERIAL` for better readability


nikolasburk (approved):
Looks great to me but I'll let @mhwelander do the merge in case she has a change request 🙂

PR #625, opened 2020-07-07
Fixing typo in docs for relation table models

kevinlangleyjr:
While trying to create a relation table with Prisma, I found what I believe to be a typo in the docs.

We are referring to the actual models and not the relation model in the example, this PR fixes that.


mhwelander:
Good catch, thanks for raising a PR! 💚

PR #636, opened 2020-07-10
Update Supported Databases to include Yugabyte

vvkgopalan:
Yugabyte YSQL is wire-compatible with Postgres. There is no additional setup required to integrate with Prisma (other than changing the .env file to reflect the appropriate database connection URL - which is necessary for Postgres in the first place). 
For an example app using Yugabyte & Prisma, view https://github.com/vvkgopalan/yugabyte-prisma2-examples/

### DIFF
--- a/content/05-more/03-supported-databases.mdx
+++ b/content/05-more/03-supported-databases.mdx
@@ -21,5 +21,6 @@ Prisma currently supports the following databases:
 | SQLite                | \*      |
 | AWS Aurora            | \*      |
 | AWS Aurora Serverless | \*      |
+| YugabyteDB            | 2.1     |
 
 Note that a fixed version of SQLite is shipped with every Prisma release.


janpio:
We need to discuss internally if we are comfortable with promising full support if we are not running our full test suite on the database. "wire-compatible" is nice in theory, in practice there are a lot of possible gotchas. I will raise this internally.

tomhoule:
We had this discussion, and we do not feel comfortable stating different implementations of the same protocol as supported without running our full test suite on them in CI. @janpio created issues about this for [Yugabyte](https://github.com/prisma/prisma-engines/issues/901) and [Cockroach](https://github.com/prisma/prisma-engines/issues/900).

janpio (changes requested):
See discussion

yoshuawuyts:
We have run the tests for both Yugabyte (https://github.com/prisma/prisma-engines/pull/1139) and Cockroach (https://github.com/prisma/prisma-engines/pull/1024) and unfortunately neither appear to work as-is with our Postgres connector. Even though they share a common network protocol with Postgres, the databases themselves differ in how they store and expose the data.

In order for Prisma to be compatible with Yugabyte and Cockroach we'll need to create dedicated connectors to support them. So unfortunately we can't claim we support either database quite yet. We may in the future though, but that's something that will need to be decided to be part of the roadmap.

PR #678, opened 2020-07-17
Update 08-working-with-json.mdx

2color:
Small typo fix.

Besides that question
In the example:
> The following example returns all users where the `extendedProfile` field does **not** match `json`: 

```
const get = await prisma.user.findMany({
  where: {
    extendedProfile: {
      equals: json,
    },
  },
}
```

Did you intend on using `equals` or `not`?


mhwelander:
Meant to be `not` - fixed as part of your PR. Thank you! 💚

PR #681, opened 2020-07-18
Use the new Nexus url

cktang88:
It's very confusing when it points to the old url!


jasonkuhrt (approved):
LGTM

We could use `nxs.li` to be a bit safer:

- https://nxs.li/website (todo - doesn't exist yet)
- https://nxs.li/plugins/prisma

mhwelander:
Updated to use https://nxs.li/plugins/prisma

Thanks for taking the time to PR! 💚

PR #696, opened 2020-07-23
Add missing `profileId` field in example code (Doc)

TepyThai:
The code example in `One-to-one relations` below: 
```
model User {
  profile   Profile?  @relation(fields: [profileId], references: [id]) // references `id` of `Profile`
}

model Profile {
  id   Int    @id @default(autoincrement())
  user User?
}
```
produces `Error validating` below:

```
Error validating: The argument fields must refer only to existing fields. The following fields do not exist in this model: profileId
```


nikolasburk:
Thanks so much for catching and fixing this @TepyThai 🎉

TepyThai:
My pleasure to be able to help 😊 and thanks for the really informative docs :)

PR #704, opened 2020-07-26
fix(upgrading-prisma-bindings-to-nexus): "Prisma 1"

janpio:
- made it explicit that "Prisma 1 server" is meant


mhwelander:
@janpio Sounds like we should remove the entire note?

janpio:
Possible, not sure. @nikolasburk? @Weakky?

mhwelander:
Closed after discussion with Niko - Prisma server is the name of a component that doesn't exist in Prisma 2.

PR #722, opened 2020-07-30
repetition the same information in doc 

Taykavol:
Deleted unnecessary double times repetition in doc.


mhwelander:
Good catch! Some very peculiar merging happened there 😅 Thanks for the PR!

PR #726, opened 2020-07-31
Add commata to MySQL and PostgreSQL table creation

norraell:
I added "," to the CREATE statements when missing. It may be left, but helps the tutorial reader focus on the statement but not why a comma is missing in a line.


janpio (approved):

mhwelander:

PR #733, opened 2020-08-04
Fixed incorrectly marked example in postgres docs

shaswatsaxena:
The last example is for xml not macaddr


mhwelander:
Thank you!

Issue #750, opened 2020-08-06
Remove inlinecode from anchor links


mhwelander:
Currently using <inlinecode> in headings results in this: `prisma-schema-reference#inlinecodedatasourceinlinecode` - the 'inlinecode' part is included in the anchor link.

nilubava:
@mhwelander Is this issue still there?

PR #755, opened 2020-08-10
fix: removed inlinecode from toc and headings

nilubava:
#750


mhwelander:
@nilubava 👋 Does this need testing from me?

PR #781, opened 2020-08-22
update mysql datasource to use port 3306

manV:
It must be pretty annoying for the people like me who copy and paste this frequently from the docs. Changed it to use default Mysql port.


mhwelander:
Thanks! Probably a PostgreSQL example copy-paste error 😅

PR #785, opened 2020-08-22
docs(error-formatting): add examples for all errorFormats

janpio:
This lets me copy/paste any of the possible values instead of having to copy paste the existing one, and then change it to the one I want.


mhwelander:
👍

PR #792, opened 2020-08-25
Correct Prisma Float type mapping for PostgreSQL

aminmus:
Corrects the type mapping documentation regarding Prisma's Float. 

In reality, Prisma Migrate maps a Prisma **Float** to a PostgreSQL **decimal(65,30)**. This is also what the docs say in the [Prisma Schema API Reference](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/prisma-schema-reference/#float).


mhwelander:
Good catch! 💚 Thanks for raising a PR 👍

PR #795, opened 2020-08-25
Fix missing brackets in docs for Prisma middleware

cimchd:
Missing brackets in first 2 code examples


mhwelander:
💚 Thank you, good spot!

Issue #666, opened 2020-07-15
Clicking on some links is sometimes really slow


Jolg42:
It happens to me sometimes that I click on in the sidebar menu on a link and it takes “a while” to update the page, see example recording:

![22f0bc0a48768854c664a89181c763e3](https://user-images.githubusercontent.com/1328733/87542208-dfcd5980-c6a2-11ea-90ac-52cca0f931fb.gif)

- I click on some links it works fast enough for me until…
- I click on “REST API”, only once (timestamp 8sec)
- And wait for rendering… 7 seconds later :slow-parrot:

This happens kind of randomly but only once per link, looking at the Dev Tools I see it loads:
`/docs/page-data/app-data.json`
`/docs/page-data/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-a-rest-api/page-data.json`

I forgot to open Dev Tools to record what was happening, it could be Netlify slow response or client side rendering taking a long time somehow?

Tested on Firefox Developer Edition 79.0b3 (64-bit) on macOS Catalina

nikolasburk:
Some additional info: https://github.com/prisma/docs/issues/774

nikolasburk:
More info: https://github.com/prisma/docs/issues/810

PR #827, opened 2020-09-08
fix: Techswitcher CPU usage breaking docs

### LINKED ISSUE #511: High memory and CPU usage on pages with techswitchers, going blank
High memory and CPU usage on pages with techswitchers

https://prisma-company.slack.com/archives/C5Z9TH6N9/p1592853060004300

nilubava:
#666


janpio:
What was the reliable reproduction we had found for this? Seems the issue has not been updated to contain it.

mhwelander:
<s>@janpio It's in this issue: https://github.com/prisma/docs/issues/511</s> Never mind, no it isn't! Here: https://prisma-company.slack.com/archives/C5Z9TH6N9/p1598295261120300

janpio:
Ok, so:

> I could just reproduce the "https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch-sql-- will make your browser hang" thing.
> I went to docs.prisma.io and clicked on "Start from scratch (Prisma Migrate) (15 min)" and then tried to use the "Back" button on my mouse (not sure if that step really influenced it at all).
> Afterwards the tab goes bonkers

janpio (approved):
I can still reproduce this on docs.prisma.io, so it was not a fluke.
I can not reproduce this any more in the deploy preview from Netlify in this PR here: https://deploy-preview-827--prisma2-docs.netlify.app

Sometimes the dropdown take a bit to show up, but I think that is something we can fix later if we want to - as long as it does not kill our user's browsers any more this is worth merging.

mhwelander:
My browser still has issues if you go to: https://deploy-preview-827--prisma2-docs.netlify.app/getting-started/setup-prisma/add-to-existing-project

Specifically the 'loading' icon keeps spinning and the full contents of the page don't load (when you scroll, part of the content is cut off)

janpio:
Can confirm @mhwelander  - something is funky there. (loading icon is in the tab of Chrome for me)

mhwelander:
Same - loading icon is in the tab.

nilubava:
@mhwelander this seems to happen only when we are opening the link as https://deploy-preview-827--prisma2-docs.netlify.app/getting-started/setup-prisma/add-to-existing-project and not as https://deploy-preview-827--prisma2-docs.netlify.app/getting-started/setup-prisma/add-to-existing-project-typescript-postgres

Will there be any situations where we have to provide links like this? I thought we always had to link to the language and db in url.

mhwelander:
@nilubava We use the non-technology-specific link (on the site, on Twitter) - I'm also not sure if the link checker would work with the -postgres-typescript part?

janpio:
Even if we would replace all links, it should not really break 🤷

nilubava:
@janpio This was not the assumption. If we need to include the paths without lang and db in the url, we can move this to a separate issue ticket.

nilubava:
Nvm, I will push a fix for this in the same PR

nilubava:
@mhwelander Can you confirm whether the issue has been resolved?

mhwelander:
@nilubava The performance issue is no more!

One small thing - I notice that the default redirect no longer occurs (add-to-existing-project does not become add-to-existing-project-typescript-postgres)

Is that because the URL rewrite was causing the performance issue?

I think we can live with that - however, a small 'bug':

1. Go to add-to-existing-project
2. Change to Javascript and MySQL
3. Try to change back to TypeScript and PostgreSQL

Expected result: URL changes to add-to-existing-project-typescript-postgres

Actual: URL is add-to-existing-project-typescript-mysql (mysql part should update, but doesn't).

janpio:
> I think we can live with that - however, a small 'bug':

That would be problematic as for Google (and other search engines) the same content exists twice on different URLs. If this is really the state to be, it needs to be handled properly with `canonical` etc.

mhwelander:
Alternatively - we do an actual redirect

nilubava:
@mhwelander @janpio That certainly was not an unavoidable bug. I have fixed it now. Except for the slow select dropdown update issue, everything works good now. I have created a new issue for this in https://github.com/prisma/docs/issues/852

PR #834, opened 2020-09-10
add comment about dotenv -- flags

Jolg42:


Jolg42:
Can also be found in https://www.npmjs.com/package/dotenv-cli

content/03-reference/01-tools-and-interfaces/01-prisma-schema/index.mdx
```diff
@@ -189,9 +189,11 @@ You can use `dotenv-cli` if you want to:
 3.  To use the `.env3` file, you can use `dotenv` when you run any Prisma command and specify which `.env` file to use. The following example uses a file named `.env3`:
 
     ```
-    dotenv -e .env3 npx prisma introspect
+    dotenv -e .env3 -- npx prisma introspect
     ```
 
+Note: dotenv doesn't pass the flags to the Prisma command by default, this is why the command includes two dashes `--` before `prisma`, making it possible to use flags like `--force`, `--schema` or `--experimental`.
```
mhwelander:
```suggestion
> **Note:** dotenv doesn't pass the flags to the Prisma command by default, this is why the command includes two dashes `--` before `prisma`, making it possible to use flags like `--force`, `--schema` or `--experimental`.  
```

PR #844, opened 2020-09-12
Fix updateMany return description

akhilome:


mhwelander:
Good eye, thank you for raising a PR! 💚

PR #890, opened 2020-10-05
Misc Fixes

### LINKED ISSUE #432: Specifying a note about MySQL engines
https://github.com/prisma/prisma/discussions/2701
As per the above discussion, we should specify that by default, the MySQL engine is InnoDB and on older versions it's not so we need to explicitly specify this on older versions as otherwise the relations will be lost and not created.

E.g.
```sql
CREATE TABLE Post (
id INT AUTO_INCREMENT PRIMARY KEY,
title VARCHAR(256),
authorId INT,
CONSTRAINT author FOREIGN KEY (authorId) REFERENCES User(id)
) ENGINE = InnoDB;
```

This engine param needs to be specified on older versions and where InnoDB isn't the default.

### LINKED ISSUE #481: Document what happens for deleting a non-existing ID
https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/crud#return-type-5

It would be really nice if the documentation specified the return value for deleting an ID that doesn't exist in the database. I assume it returns null but...I don't know, because it's not documented :)

### LINKED ISSUE #853: Update @@unique with optional name property

### LINKED ISSUE #862: Document `name` attribute for `@index` directive
Example:

```
@@index([field1, field2], name: "myIndexName")
```

Triggered by https://github.com/prisma/migrate/issues/542#issuecomment-691064777

### LINKED ISSUE #863: Clarification around parameterized queries
https://prisma-company.slack.com/archives/C5Z9TH6N9/p1600351377027900

### LINKED ISSUE #885: Document "Optional field is required in @@unique and where clause and doesn't accept null"
Based on this internal comment we should document the bevahiour described in [this issue](https://github.com/prisma/prisma/issues/3772):

image

mhwelander:


content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
```diff
@@ -862,62 +862,108 @@ N/A
...
+* All fields that make up the unique constraint **must** be mandatory fields. The following model is **not** valid because `id` could be `null`:
+
+  ```prisma
+  model User {
+    firstname Int
+    lastname Int
+    id Int?
+    @@unique([firstname,lastname,id])
+  }
+  ```
```
janpio: Semi offtopic:
What is our general stance on "wrong" examples? Do we just put them in and note in the text that they are _not_ what people are supposed to do?

content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
```diff
@@ -862,62 +862,108 @@ N/A
...
 
+
 #### Arguments 
 
 
 | Name       | Required | Type                                   | Description                                                                                                            |
 | ---------- | -------- | -------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
-| `fields` | **Yes**  | `FieldReference[]`  | A list of field names - for example, `["firstname", "lastname"]` |
+| `fields` | **Yes**  | `FieldReference[]`  | A list of field names - for example, `["firstname", "lastname"]`. Fields must be mandatory - see remarks. |
+| `name`   | **No**   | `String`            | The name of the unique combination of fields - defaults to `fieldName1_fieldName2_fieldname3`  |
```
janpio: Probably:
```suggestion
| `name`   | **No**   | `String`            | The name of the unique combination of fields - defaults to `fieldName1_fieldName2_fieldName3`  |
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/030-crud.mdx
```diff
@@ -197,6 +197,34 @@ const result = await prisma.user.findOne({
...
+#### Multi-field unique attribute
+
+Assume your model has a multi-field unique attribute:
+
+```prisma
+model User {
+  firstName String
+  lastName  String
+
+  @@unique(fields: [firstName, lastName], name: "fullname")
```
janpio: nit: highlight this line maybe?

content/03-reference/01-tools-and-interfaces/02-prisma-client/030-crud.mdx
```diff
@@ -779,7 +807,13 @@ export type FirstNameLastNameCompoundUniqueInput = {
 
 ### Return type
 
-`delete` returns a plain old JavaScript object. 
+`delete` returns a plain old JavaScript object or throws a `RecordNotFound` exception if the unique identifier does not exist:
+
+```terminal
+Error for binding '0': RecordNotFound("Record to delete does not exist.")
```
janpio, line 813: Offtopic:
@eemmiillyy, this is an example for a non optimal error message as it includes information that is not relevant for the user `binding '0'` - might be worth creating an issue for a _test_ to test this, and one for improving the error message some time in the future to something more user friendly. I would also be curious, how this compares between databases (MySQL, Postgres, SQLite).

content/03-reference/01-tools-and-interfaces/02-prisma-client/030-crud.mdx
```diff
@@ -779,7 +807,13 @@ export type FirstNameLastNameCompoundUniqueInput = {
...
+```
+
+> **Tip**: If you need to delete multiple records based on some criteria (for example, all `User` records with a `prisma.io` email address, use `deleteMany`)
```
janpio, line 816: Can we link the `deleteMany` to the appropriate docs?

content/03-reference/01-tools-and-interfaces/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -199,10 +199,36 @@ const result = await prisma.queryRaw(
 )
 ```
 
+The following example uses a PostgreSQL query:
+
+```ts
+const userName = 'Sarah'
+const email = 'alice@prisma.io'
+const result = await prisma.queryRaw(
+  'SELECT * FROM User WHERE (name = $1 OR email = $2)', // PostgreSQL variables
```
janpio:
```suggestion
  'SELECT * FROM User WHERE (name = $1 OR email = $2)', // PostgreSQL variables, represented by $1 and $2
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -199,10 +199,36 @@ const result = await prisma.queryRaw(
...
+)
+```
+
 As with tagged templates, Prisma Client escapes all variables.
 
 > **Note**: You cannot pass a table or column name as a variable into a parameterized query. For example, you cannot `SELECT ?` and pass in `*` or `id, name` based on some condition.
 
+#### Parameterized PostgreSQL `ILIKE` query
+
+When you use `ILIKE`, the `%` wildcard character(s) should be included in the variable itself, not the query string:
```
janpio: Hm, is "query string" the correct term here? (I only think of URL query strings here, so may be biased and just not know the correct term)

content/03-reference/02-database-connectors/04-mysql.mdx
```diff
@@ -195,3 +195,7 @@ model Device {
 //  data   xml
 }
 ```
+
+## Engine
+
+If you are using a version of MySQL where InnoDB is not the default engine, you must specify `ENGINE = InnoDB;` when you create a table. If you introspect a database that uses a different engine, relations in the Prisma Schema are not created (or lost, if the relation already existed).
```
janpio: Is that correct that only InnoDB tables are introspected to have relations? What are possible other engines?

content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
mhwelander: There's no general stance - I tend to add a comment next to specific line and say 'this will NOT work' - perhaps I will do that here. But that's interesting - I will chat to Matej about whether we should do something design-related:

image

content/03-reference/01-tools-and-interfaces/02-prisma-client/090-raw-database-access.mdx
mhwelander: Hmm.. you're right, I'll change it to query (`string`) or something like that.

content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
```diff
@@ -862,62 +862,108 @@ N/A
...
-* A model can have any number of unique constraints
+* A `@@unique` block is required if it represents the only unique constraint on a model without an `@id` / `@@id`
+* All fields that make up the unique constraint **must** be mandatory fields. The following model is **not** valid because `id` could be `null`:
+
+  ```prisma
+  model User {
+    firstname Int
+    lastname Int
+    id Int?
+    @@unique([firstname,lastname,id])
```
matthewmueller: nitpick: Spacing could be improved `[firstname, lastname, id]`

matthewmueller (approved):
Looks great!

content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
mhwelander:
```suggestion
    @@unique([firstname, lastname, id])
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/030-crud.mdx
mhwelander:
```suggestion
|  @@unique(fields: [firstName, lastName], name: "fullname")
```

content/03-reference/01-tools-and-interfaces/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -199,10 +199,36 @@ const result = await prisma.queryRaw(
...
```
mhwelander:
```suggestion
When you use `ILIKE`, the `%` wildcard character(s) should be included in the variable itself, not the query (`string`):
```

content/03-reference/02-database-connectors/04-mysql.mdx
mhwelander: @janpio It seems there are many (Blackhole, CSV). MyISAM is specifically the one that doesn't work. I can change it to "If you are using a version of MySQL that uses MyISAM, specify `Engine=InnoDB;` when you create a table". That way we aren't explicitly saying that no other engine works.

content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
mhwelander: I've added this to Design's inbox.

content/03-reference/02-database-connectors/04-mysql.mdx
mhwelander: I've changed it to what I suggested.

content/03-reference/01-tools-and-interfaces/01-prisma-schema/07-prisma-schema-reference.mdx
hrescak: Coming in from Design inbox -> Is generally the entire example 'wrong'? Like is it specifically a counter-example? Or is there usually a line that's worth highlighting as the main culprit? Or both?

hrescak: @mhwelander @nikolasburk

Issue #654, opened 2020-07-13
Escape +, - and | at beginning of a line


nikolasburk:
Sometimes a line in a code block needs to start with a `+`, `-` or `|` but right now that's not possible without triggering the "line highlighting".

Example here: https://www.prisma.io/docs/more/style-guide/mdx-examples#tabbed-blocks

![image](https://user-images.githubusercontent.com/4058327/87296189-abbd3180-c506-11ea-902d-ec4ac718b59e.png)

Potential solutions:

- Escape special character e.g. with `\` (but what if the file needs to start with `\+`? 😅 )
- Additional option to configure line highlights via props of MDX component if necessary

nikolasburk:
Solution: pass selected lines as annotation on a code block:

~~~
```js{4,7-11;edit | 5,13;add}
const typeDefs = `
  type Query {
    info: String!
    feed: [Link!]!
  }

  type Link {
    id: ID!
    description: String!
    url: String!
  }
`
```
~~~

PR #898, opened 2020-10-07
feat: code highlighting redefined

nilubava:
#654 

Usage: 

```
```js file=schema.js highlight=15;normal|4,7-11;edit|5,13;add|1,2;delete
async function main() {
  const allUsers = await prisma.user.findMany();
  console.log(allUsers);
}
......
``````

TODO:

- [ ] Change all the code block highlighting to use the new structure

### CODE REVIEW

**Comment by nikolasburk** [comment]
The highlights work great! But was it intended that we don't render the initial `+`, `-` and `✎` icons at the beginning of the line? I'd imagine those to still be present like with GitHub diff rendering:

```diff
+ adds a line
- removes a line
```

nilubava:
@nikolasburk Updated this in another PR : https://github.com/prisma/docs/pull/1050

PR #906, opened 2020-10-12
Use div for table of contents header and fix page.json

### LINKED ISSUE #747: Change BreadcrumbTitle to a div - currently an H4
Not sure why this is an H4
```js
const BreadcrumbTitle = styled.h4`
  color: var(--code-inner-color) !important;	  color: var(--code-inner-color) !important;
  line-height: 1rem;	  line-height: 1rem;
  font-weight: normal;	  font-weight: normal;
```

2color:


Jolg42:
From what I understand this changes the "CONTENT" from h1 to h2.

From a semantic point of view it doesn't need to be a header I think, there is no added semantic, it's more a nice UI visual title.

I think it's better to reserve h2 for things that have more value.

Screen Shot 2020-10-12 at 18 18 19

Jolg42 (approved):
👍 

Something that can go in this PR or another one is that it's weird to have the breadcrumb as a header h4 too (and even empty here).

We can get rid of the h4 (turn it into span maybe)

Screen Shot 2020-10-12 at 18 33 15

Jolg42 (approved):
Looks great!

src/components/toc.tsx
```diff
@@ -2,15 +2,15 @@ import React from 'react'
 import styled from 'styled-components'
 import { stringify } from '../utils/stringify'
 
-const ChapterTitle = styled.h1`
+const ChapterTitle = styled.div`
```
nilubava, line 5: @2color - Shouldn't this be `h2` ? 🤔

2color: See Joël's [comment](https://github.com/prisma/docs/pull/906#issuecomment-707217532).

mhwelander:
Linked: #747

PR #915, opened 2020-10-15
Corrected some inaccuracies

samirettali:


2color:
Thank you for the contribution @samirettali!

PR #968, opened 2020-11-02
chore: add alt tags to images for accessibility

2color:


Jolg42 (approved):
The diff looks good to me 👍

PR #998, opened 2020-11-07
Fix: Search icons on Firefox aren't sized correctly #992

### LINKED ISSUE #992: Search icons on Firefox aren't sized correctly
Search magnifying glass icon, as well as the `/` shortcut icon should be wider (to match Chromium)

This is Firefox:
Screenshot 2020-11-05 at 15 46 12

And this is Chromium:
Screenshot 2020-11-05 at 15 46 19

Couldn't resist debugging a little, so:
The search icon `svg` has a `16px` width and height, but because its parent is `flex`, you'll need to add a `flex-shrink: 0` to the SVG as well to ensure that the `<input />` sibling does not shrink it. IDK what the flex spec says is "correct", but adding that line to the CSS fixes this issue for me in Firefox :D

rahul3v:
Fix: #992
Now the **slash** and the **icon** are showing correctly

### DIFF
--- a/src/components/search/input.tsx
+++ b/src/components/search/input.tsx
@@ -96,7 +96,7 @@ const SearchBoxDiv = styled.div`
     border: 1px solid ${p => p.theme.colors.gray400};
     border-radius: 4px;
     color: ${p => p.theme.colors.gray400};
-    width: 18px;
+    min-width: 18px;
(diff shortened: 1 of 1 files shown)


hrescak:
Thanks for this @rahul3v!

PR #999, opened 2020-11-08
Rename experimentalFeatures to previewFeatures

zackify:
This fixes the naming that is used in the latest version of prisma


mhwelander:
Good catch, thank you for taking the time to PR! 💚

Issue #991, opened 2020-11-05
Homepage has different search placeholder text compared to other pages


mhwelander:
Search vs Search documentation

(I tried to fix this, but.. I couldn't figure it out)

rahul3v:
@mhwelander
what you suggest, what you need it as always ; `Search Documentation...`  or `Search` or something else

PR #1011, opened 2020-11-11
[draft] Attempting to fix placeholder on home

hrescak:
trying to address #991 - This is a tough because it doesn't reproduce locally, I'm hoping this is not some silly opaque optimization in gatsby or netlfy

### DIFF
--- a/src/components/search/input.tsx
+++ b/src/components/search/input.tsx
@@ -133,7 +133,7 @@ const SearchBox = ({ refine, onFocus, currentRefinement, isOpened, ...rest }: an
   const timeoutId = React.useRef(null)
   const inputEl = React.useRef(null)
   const { width } = useWindowDimensions()
-  const placeholderText = width > 640 ? 'Search Documentation...' : 'Search'
+  const placeholderText = width > 400 ? 'Search Documentation...' : 'Search'
 
   const onChange = (e: any) => {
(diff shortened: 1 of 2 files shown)


src/components/search/input.tsx
```diff
@@ -133,7 +133,7 @@ const SearchBox = ({ refine, onFocus, currentRefinement, isOpened, ...rest }: an
   const timeoutId = React.useRef(null)
   const inputEl = React.useRef(null)
   const { width } = useWindowDimensions()
-  const placeholderText = width > 640 ? 'Search Documentation...' : 'Search'
+  const placeholderText = width > 400 ? 'Search Documentation...' : 'Search'
```
rahul3v, line 136: if you really wanna need only one as always then just put as any one of it, since if you reference #991, i think it only suggest to have one all the time:
```suggestion
  const placeholderText = 'Search Documentation...'
```
or 

```suggestion
  const placeholderText = 'Search'
```

hrescak: THanks for your comment @rahul3v . We'll try to make it work before reverting to one of the two.

hrescak:
OK I'm fairly flabbergasted on this one. Summarized here

https://gist.github.com/hrescak/e67aa920002e1da609a9690cd57d39ce

PR #1016, opened 2020-11-16
Fix wrong titles and descriptions

2color:
This fixes the wrong titles and descriptions on pages such as:
https://www.prisma.io/docs/guides/database-workflows/cascading-deletes/postgresql
https://www.prisma.io/docs/guides/database-workflows/data-validation/postgresql
https://www.prisma.io/docs/guides/database-workflows/unique-constraints-and-indexes/postgresql


nilubava (approved):
LGTM 👍

PR #1031, opened 2020-11-24
Update the Studio usage guidelines

martzoukos:
image

Now that we have a standalone macOS app, we can update the usage instructions to reflect that.

Probably needs some rewording.

Thanks!


martzoukos:
Added a section on the troubleshoot section too:
image

Let me know if that way of linking images is safe or if we prefer to host images on the app instead.

PR #1043, opened 2020-11-29
Update 03-is-prisma-an-orm.mdx

macmillen:
typos


nikolasburk:
thanks so much for catching these @macmillen 🙌

PR #1050, opened 2020-12-01
feat: code highlighting redefined

nilubava:
#654 

Usage: 

```
```js file=schema.js highlight=15;normal|4,7-11;edit|5,13;add|1,2;delete
async function main() {
  const allUsers = await prisma.user.findMany();
  console.log(allUsers);
}
......
``````

You can also use `+`, `-`, `|` etc. as usual too.

TODO:

- [ ] Change all the code block highlighting to use the new structure

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "jquery": "^3.5.1",
     "lodash": "^4.17.15",
     "netlify-lambda": "^2.0.1",
+    "parse-numeric-range": "^1.2.0",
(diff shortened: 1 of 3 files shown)


nikolasburk:
Looks good do me 👌 we probably want to turn off the old behaviour once all code snippets are migrated to the new structure to avoid situations like this:

image

nikolasburk (approved):
Looks good to me 👍

PR #1055, opened 2020-12-03
Update 110-advanced-usage-of-generated-types.mdx

ryands17:


nikolasburk:
Awesome, thanks a lot Ryan! 🚀 

(btw, is the `PromiseReturnType` still exported in the same way or should that by adjusted as well on that page?)

image

ryands17:
@nikolasburk That remains the same. Just the model types are under the `Prisma` namespace.

PR #1059, opened 2020-12-07
Code Example Update

johnthepink:
I think these are missing the closing parenthesis.


mhwelander:
They are - good spot, thank you for submitting a PR for it!

PR #1080, opened 2020-12-10
Expand on how the schema file is resolved

Lanny:
Previously only the `--schema` flag to generate was called out in the docs as a way to specify a schema file other than `./schema.prisma`. Looking at the code there are two other options. This PR documents all four options.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/index.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/index.mdx
@@ -115,7 +115,7 @@ enum Role {
 
 The default name for the schema file is `schema.prisma`. When your schema file is named like this, the Prisma CLI will detect it automatically in the directory where you invoke the CLI command (or any of its subdirectories).
 
-If the file is named differently, you can provide the `--schema` argument to the Prisma CLI with the path to the schema file, e.g.:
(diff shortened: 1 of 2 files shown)


Lanny:
@mhwelander just wanted to bump this and check if there was anything needed from my side. Thanks!

mhwelander:
Hello @Lanny - do pardon the delay (🎄 winter vacation) and thank you for raising this PR!

Schema file resolution docs are already available here: https://www.prisma.io/docs/concepts/components/prisma-schema#prisma-schema-file-location However! You have made corrections, added useful additional info, made it clear that we need to link to that section from each `--schema` table row so that folks can find it. I suggest the following:

[] Add "If yarn is used, steps X and Y are applied for each workspace" to existing page as a note
[] Replace every instance of " `./schema.prisma`, `./prisma/schema.prisma`" in 'Default' column of `--schema` rows with "See: Schema file resolution" + a link to the existing page
[] Change the order of ./schema.prisma and ./prisma/schema.prisma, you are correct that they're in the wrong order

I'm happy to edit the PR if you'd prefer, let me know either way!

molebox:
Hey @Lanny! Have you had a chance to look over @mhwelander comments? If you feel that you dont want to continue with this PR we could take a look instead, no worries 😊

tanberry:
HI @Lanny the docs team will try to pick this up and implement @mhwelander 's suggestions; let us know if you want to do it yourself though.

PR #1087, opened 2020-12-12
Fix: many broken links

rahul3v:
fixed many broken link


mhwelander:
Hi! These broken links occurred due to a bug fix and have been fixed with redirects. Thank you all the same for taking the time to raise this PR!

PR #1090, opened 2020-12-14
Remove (what appears to be a) remainder of cut & paste

Skn0tt:
See title.


mhwelander:
Yikes, good catch - thanks for raising a PR! 💚

PR #1094, opened 2020-12-15
Explicit M-N relations

ryands17:


content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
+---
+title: 'Working with many-to-many relations with Prisma'
+metaTitle: 'Working with many-to-many relations with Prisma'
+metaDescription: 'Learn how you can model and query implicit and explicit many-to-many relations with Prisma'
+---
+
+## Problem
+
+Lots of users have queries about can one model many-to-many relations and query for the data using Prisma. Here are two examples one implicit and one explicit to show how you can handle these with Prisma.
```
nikolasburk:
```suggestion
Modeling and querying many-to-many relations in relational databases can be challenging. This article shows two examples how this can be approached with Prisma. The first example uses an [implicit](../../concepts/components/prisma-schema/relations#implicit-many-to-many-relations) and the second one uses an [explicit](../../concepts/components/prisma-schema/relations#explicit-many-to-many-relations) many-to-many relation.
```

nikolasburk: Side-note: I suggest to not use the term "query" when talking about user requests, just because the terms is already super ambiguous/overloaded 😄 in most cases, I think "question" feels a lot more appropriate.

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
...
+
+## Solution
+
+### Implicit relations
+
+This is a type of many-to-many relation where Prisma handles the join table internally. A basic example for an implicity many-to-many relation would look like this:
```
nikolasburk: Our official term for join table is [relation table](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#relation-tables):

```suggestion
This is a type of many-to-many relation where Prisma handles the [relation table](../../concepts/components/prisma-schema/relations#relation-tables) internally. A basic example for an implicit many-to-many relation would look like this:
```

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
...
+}
+
+model Tag {
+  id    Int        @id @default(autoincrement())
+  name  String     @unique
+  posts PostTags[]
+}
+```
+
+Adding tags to a post would be a create into the join table (`PostTags`) as well as into the tags table (`Tag`):
```
nikolasburk:
```suggestion
Adding tags to a post would be a create into the relation table (`PostTags`) as well as into the tags table (`Tag`):
```

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
...
+          "id": 2,
+          "name": "dev"
+        }
+      }
+    ]
+  }
+]
+```
+
+Sometimes, it's not ideal to show the data for the join table in your UI. In this case, it's best to map the data after fetching it on the server itself and sending that response to the frontend.
```
nikolasburk:
```suggestion
Sometimes, it's not ideal to show the data for the relation table in your UI. In this case, it's best to map the data after fetching it on the server itself and sending that response to the frontend.
```

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
...
+      {
+        "id": 2,
+        "name": "dev"
+      }
+    ]
+  }
+]
+```
+
+So this was how you can implement Implicit and Explicit many-to-many relations and querying for them using Prisma.
```
nikolasburk:
```suggestion
This article showed how you can implement implicit and explicit many-to-many relations and query them using Prisma Client.
```

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
...
+  data: {
+    title: 'Prisma is awesome!',
+    tags: { set: [{ id: 1 }, { id: 2 }], create: { name: 'typescript' } },
+  },
+})
+```
+
+### Explicit relations
+
+Explicit relations mostly need to be created in cases where you would need to store extra fields in the join table or if you're introspecting an existing database that already has many-to-many relations setup. This is the same schema used above but with an explicit join table:
```
nikolasburk:
```suggestion
Explicit relations mostly need to be created in cases where you need to store extra fields in the relation table or if you're [introspecting](https://www.prisma.io/docs/reference/tools-and-interfaces/introspection) an existing database that already has many-to-many relations setup. This is the same schema used above but with an explicit relation table:
```

nikolasburk (changes requested):
Looks great! I made a couple of suggestions that you can just commit, but then it should be good to publish 🚀

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
+---
+title: 'Modeling and querying with many-to-many relations'
```
nikolasburk:
```suggestion
title: 'Modeling and querying many-to-many relations'
```

content/500-support/100-help-articles/200-working-with-many-to-many-relations.mdx
```diff
@@ -0,0 +1,190 @@
...
+metaTitle: 'Modeling and querying with many-to-many relations'
```
nikolasburk:
```suggestion
metaTitle: 'Modeling and querying many-to-many relations'
```

PR #1101, opened 2020-12-17
Fixed: typo

gwauge:
Added missing "to" under "Using .env files" section


mhwelander:

PR #1102, opened 2020-12-18
fix: code with result fix

nilubava:


src/components/customMdx/code.tsx
```diff
@@ -245,7 +245,9 @@ const LineNo = styled.span`
   line-height: ${p => p.theme.space[24]};
   color: ${p => p.theme.colors.gray400};
   display: inline-block;
-  text-align: right; // padding-left: 1em; user-select: none; width: 24px;
+  text-align: right;
+  user-select: none; 
+  width: 24px;
```
rahul3v, line 250:
```suggestion
  width: 30px;
```

As some of code may go beyond **100+** lines codes,  but might not go **1000+** I  believe
So if 3 digit numbers comes in, then, it looks bit odd

layout

So batter go with atleast `30px`

PR #1112, opened 2020-12-23
Proposal for quickstart update

hrescak:
@nikolasburk I took a quick stab at this - I focused on the very beginning to help people jump into it faster. I think the actual flow of the rest of the quickstart is great and at the appropriate depth. Even showing the tables in the sqlite database is IMO more useful than hiding them.


nikolasburk:
I think that looks great, thanks so much Matej! 🙏

Issue #1002, opened 2020-11-09
Align homepage with updated design


mhwelander:
It's currently a regular page with navigation disabled/hidden.

PR #1114, opened 2020-12-23
feat: New homepage for docs 

nilubava:
#1002


hrescak:
Filling in feedback as I come in (some of this is also for @kuldar lol):
- I'd make the main `h` larger - at least the size on individual articles
- the background image of the top section seems very pixelated, I'd consider using an svg maybe
image
- I'd make either the entire cell or the title clickable in these, not just the arrow
image
- the icons up top (overview, components etc... can be pulled from `react-feather`)
- I'd consider making the main call to action buttons large, with icon to the right and not capitalized, to make them consistent everywhere. (the icon can also be found in `react-feather`, I think it's ChevronsRight.

nilubava:
Done -> 

- the background image of the top section seems changed to svg
- Entire cell clickable in link cards, not just the arrow
- the icons up top (overview, components etc...) pulled from `react-feather`
- ChevronsRight pulled from `react-feather`

nilubava:
@hrescak The icon for 'More' seems different from the design. I cannot find the same in `react-feather`. Can you look into this?

hrescak:
I believe it's the `Menu` Icon

thebiglabasky:
One thing my eye somehow caught is the alignment between the round icons and the paragraphs below which felt, although technically aligned, not visually aligned
image

I was tempted to anchor on the icons like so:
image
but this may not be good.
Anyway, I find it good to go otherwise!

thebiglabasky (approved):
LBTWWCH! <- as in Looks Better Than What We Currently Have
So good to merge :D

PR #1116, opened 2020-12-24
The side without a relation scalar must be optional

ahmedosama5200:


nikolasburk:
Thanks a lot for catching this @ahmedosama5200 🚀

PR #1119, opened 2020-12-27
Update index.mdx

heymartinadams:
The schema path can also be called when using `prisma studio`


nikolasburk:
Thanks for the adjustment Martin 🙌

PR #1120, opened 2020-12-27
Fix "existing project" 404ing quickstart link

djm:
Fixes a 404 on the quickstart guide.

See: https://www.prisma.io/docs/getting-started/quickstart-typescript#set-up-prisma-with-your-own-database and click "Add Prisma to Existing Project".

This fix simply points it to the Typescript/Postgres version of the doc by default, and then people can change the defaults if they wish. Probably in an ideal world the old 404ing link would be redirected to the new one long term - but this is a very quick on-github edit.


nikolasburk:
Thanks a lot for catching this @djm 🚀

PR #1121, opened 2020-12-28
Setting the correct path to the start from scratch

RoySegall:
I found out that the start from scratch link is not  set to the correct URL. Hope I did it well :)


nikolasburk:
Thanks a lot for catching this @RoySegall 🚀

PR #1122, opened 2020-12-29
fix: Broken link on docs page

PJijin:
On this page, https://www.prisma.io/docs/concepts/components/prisma-client "Add to existing project" redirects to 404 page. (client-side routing issue)


nikolasburk:
thanks for catching this @PJijin  🚀

PR #1126, opened 2020-12-30
Updated old internal links with new internal links

victorbjorklund:
Updated old internal links with new internal links. Old internal links caused 404-error for client side navigation (but worked because of redirects when hard refresh). Updated to the new correct links to fix issue.


mhwelander:
🙏 You're a gem! 💎

PR #1130, opened 2021-01-01
Add a note explaining something that confused me in db matrix

JakubKoralewski:
Added note to `Misc` in `Database features matrix` that tries to explain that a green ✔️ checkmark in a row for a given database doesn't mean that Prisma supports the feature, but just that the database supports that feature. I know this is said on the overview at the top of the page:

> This page gives an overview of the features which are provided by the databases that Prisma supports.

but what I did and probably what most people do is skim through and look for the feature they want missing the important detail at the top. As a result I [was confused](https://github.com/prisma/prisma/issues/1684#issuecomment-752239611) as to whether a certain feature was available or not.

This is just a suggestion, feel free to put it differently or close the pull request if you feel it is redundant.

### DIFF
--- a/content/400-reference/300-database-reference/01-database-features.mdx
+++ b/content/400-reference/300-database-reference/01-database-features.mdx
@@ -88,7 +88,7 @@ Lock option (MySQL):
 ### Misc
 
 Features are only available if supported by Prisma _and_ the underlying database.
-If the database supports the feature, but the Prisma Client doesn't you should still be able to use [raw database access](../../concepts/components/prisma-schema/raw-database-access) to issue SQL commands.
+If your database supports a feature but Prisma Client does not, consider using [raw database access](../../concepts/components/prisma-schema/raw-database-access) to issue SQL commands.
 
 | Feature                           | PostgreSQL | SQL Server | MySQL | SQLite | Prisma schema                                                                                  | Prisma Client | Prisma Migrate |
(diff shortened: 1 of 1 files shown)


content/400-reference/300-database-reference/01-database-features.mdx
```diff
@@ -87,6 +87,9 @@ Lock option (MySQL):
 
 ### Misc
 
+Note that both the support for the feature from the given database and from Prisma Client is required to use the feature from Prisma.
```
mhwelander:
```suggestion
Features are only available if supported by Prisma _and_ the underlying database.
```

content/400-reference/300-database-reference/01-database-features.mdx
```diff
@@ -87,6 +87,9 @@ Lock option (MySQL):
 
 ### Misc
 
+Note that both the support for the feature from the given database and from Prisma Client is required to use the feature from Prisma. 
+If the database supports the feature, but the Prisma Client doesn't you should still be able to use [raw database access](../../concepts/components/prisma-schema/raw-database-access) to issue SQL commands.
```
mhwelander:
```suggestion
If your database supports a feature but Prisma Client does not, consider using [raw database access](../../concepts/components/prisma-schema/raw-database-access) to issue SQL commands.
```

mhwelander:
Very valid point! There's already a note at the top of the page, but as it's out of context, I think people miss it. I've taken your PR one step further and added a * to each ✔️ with a footnote that reiterates that the feature is only available if natively supported by the db. Thank you so much for bringing this to our attention!

PR #1148, opened 2021-01-07
Prisma CLI best practices guide

ryands17:


Jolg42:
Note the switch from `@prisma/cli` to `prisma` is planned soon I think, so not sure if this page will be still needed or needs to be updated after.

PR #1156, opened 2021-01-08
Updated page title

michielswaanen:
Page title wasn't consistent with the other page titles. They all have an upper-case as first letter.


johannesschobel (approved):
good catch

mhwelander:
Hello! Pardon the delay in responding and thank you for taking the time to raise a PR; Prisma schema is not capitalized because it's not a proper noun - Prisma Client, Prisma Studio, and Prisma Migrate are all 'products', which is why they are capitalized. "Preview features" is also not capitalized.

PR #1158, opened 2021-01-08
Transactions guide

mhwelander:
- [ ] <s>Add feedback re: saga</s>
- [x] Incorporate redirects
- [x] Add meta info
- [x] Revise title of page
- [x] Link from transactions page
- [ ] <s>Link from transactions blog</s>
- [ ] <s>Diagrams and illustrations</s>
- [x] Conclusion


content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+The technique you choose depends on your particular use case.
+
+> **Note**: For the purposes of this guide, _writing_ to a database encompasses creating, updating, and deleting data.
+
+</TopBlock>
+
+## Dependent writes
+
+Writes are considered **dependent** one each other if:
```
thebiglabasky:
```suggestion
Writes are considered **dependent** on each other if:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+  })
+```
+
+If any operation fails, Prisma rolls back the entire transaction. Nested writes are not currently supported by top-level bulk operations like `client.user.deleteMany` and `client.user.updateMany`.
+
+#### When to use nested writes
+
+Consider using nested writes if:
+
+* ✔ You want to write to create two or more records related by ID at the same time (for example, create a blog post and a user)
```
thebiglabasky:
```suggestion
* ✔ You want to write two or more records related by ID at the same time (for example, create a blog post and a user)
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+* ✔ You want to update and create records related by ID at the same time (for example, change a user's name and create a new blog post)
+
+> **Tip:** If you [pre-compute your IDs, you can choose between a nested write or using the `$transaction` API](#scenario-pre-computed-ids-and-the-transaction-api).
+
+#### Scenario: Sign-up flow
+
+Consider the Slack sign-up flow, which:
+
+1. Creates a team
+2. Add one user to that team, which automatically becomes that team's administrator
```
thebiglabasky:
```suggestion
2. Adds one user to that team, which automatically becomes that team's administrator
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+    team: {
+      connect: {
+        id: team.id,
+      },
+    },
+  },
+})
+```
+
+However, this code has an problem - consider the following scenario:
```
thebiglabasky:
```suggestion
However, this code has a problem - consider the following scenario:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+    },
+  },
+})
+
+await prisma.$transaction([createUserOperation, createTeamOperation])
+```
+
+##### Are cascading deletes supported by the Prisma Client nested writes?
+
+No, you cannot use nested writes to perform a cascading delete - for example, delete a team and all of its members. You can, however, use the `$transaction` API to perform a cascading delete. Deleting a team and deleting the team's members are not dependent writes - if you know the ID of the team, you can [perform both operations within a `$transaction`](#scenario-privacy-legislation).
```
thebiglabasky: It could be worth linking to the cascade delete GitHub issue to mention that we are looking at supporting that in the schema or is that not something worth mentioning here?
```suggestion
No, you cannot use Prisma Client's nested writes to perform a cascading delete - for example, delete a team and all of its members. Instead, you can:

* Use the `$transaction` API to perform a cascading delete. Deleting a team and deleting the team's members are not dependent writes - if you know the ID of the team, you can [perform both operations within a `$transaction`](#scenario-privacy-legislation).
* [Configure cascading deletes at the database level](../../general-guides/database-workflows/cascading-deletes).

> **Note**: We are a looking into [cascading deletes support at a schema level](https://github.com/prisma/prisma/issues/2810).
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+Depending on your requirements, Prisma Client has four options for handling independent writes that should succeed or fail together.
+
+### Bulk operations
+
+Bulk writes allow you to write multiple records of the same type in a single transaction - if any operation fails, Prisma rolls back the entire transaction. Prisma currently supports:
+
+* `updateMany`
+* `deleteMany`
+
+The ability to [create and upsert in bulk](https://github.com/prisma/prisma-client-js/issues/332) is planned.
```
thebiglabasky:
```suggestion
The ability to [create and upsert in bulk](https://github.com/prisma/prisma-client-js/issues/332) is being considered: feel free to chime in with your use case and upvote if you're interested.
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+#### When to use bulk operations
+
+Consider bulk operations as a solution if:
+
+* ✔ You want to update a batch of the _same type_ of record, like a batch of emails
```
thebiglabasky, line 310: Maybe not worth using bullet points since there's only one of them?

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+      members: { } // Cannot access members here
+    },
+  })
+```
+
+You an use the [`$transaction` API to perform a cascading delete](#scenario-cascading-delete).
+
+#### Can I use bulk operations with the `$transaction` API?
+
+Yes - for exapmle, you can include multiple `deleteMany` operations inside a `$transaction`.
```
thebiglabasky:
```suggestion
Yes - for example, you can include multiple `deleteMany` operations inside a `$transaction`.
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+    members: {
+      create: {
+        email: 'alice@prisma.io',
+      },
+    },
+  },
+})
+```
+
+Instead of auto-generating IDs, change the `id` fields of `Team` and `User` to a `String` (if you do not provide a value, a UUID is generated automatically). We recommend using UUIDs:
```
thebiglabasky: A minor note: UUIDs are sometimes challenged versus CUIDs etc. Maybe it's fine to leave that as something we recommend out of simplicity, but this may be something people challenge.
```suggestion
Instead of auto-generating IDs, change the `id` fields of `Team` and `User` to a `String` (if you do not provide a value, a UUID is generated automatically). This example uses UUIDs:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+    },
+  },
+})
+```
+
+There's no compelling reason to switch to manually generated IDs and the `$transaction` API if you are already using auto-generated IDs and nested writes.
+
+## Read, modify, write
+
+In some cases you may need to perform custom logic as part of an atomic operation - also known as the [read-modify-write pattern](https://en.wikipedia.org/wiki/Read%E2%80%93modify%E2%80%93write). The following is an example of the read-modify-write pattern:s
```
thebiglabasky:
```suggestion
In some cases you may need to perform custom logic as part of an atomic operation - also known as the [read-modify-write pattern](https://en.wikipedia.org/wiki/Read%E2%80%93modify%E2%80%93write). The following is an example of the read-modify-write pattern:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+* Read a value from the database
+* Run some logic to manipulate that value (for example, contacting an external API)
+* Write the value back to the databases
```
thebiglabasky:
```suggestion
* Write the value back to the database
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+All operations should **succeed or fail together** without making unwanted changes to the databases, but you do not necessarily need to use an actual database transactions. This section of the guide describes two ways to work with Prisma Client and the read-modify-write pattern:
```
thebiglabasky:
```suggestion
All operations should **succeed or fail together** without making unwanted changes to the database, but you do not necessarily need to use an actual database transaction. This section of the guide describes two ways to work with Prisma Client and the read-modify-write pattern:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+* Designing idempotent APIs
+* Optimistic concurrency control
+
+### Idempotent APIs
+
+Idempotency is the ability to run the same logic with the same parameters multiple times with the same result: the **effect on the database** is the same whether you run the logic one or one thousand times. For example:
+
+* **NOT IDEMPOTENT**: Create a user in the database with email address "letoya@prisma.io". The `User` table **does not** enforce unique email addresses. The effect on the database is different if you run the logic once (one user created) or ten times (ten users created).
+* **IDEMPOTENT**: Create a user in the database with email address "letoya@prisma.io". The `User` table **does** unique email addresses. The effect on the database is the same if you run the logic once (one user created) or ten times (subsequent attempts to create a user with the same email address will fail due to `UNIQUE` constraint).
```
thebiglabasky:
```suggestion
* **IDEMPOTENT**: Create a user in the database with the email address "letoya@prisma.io". The `User` table **does** enforce unique email addresses. The effect on the database is the same if you run the logic once (one user created) or ten times (subsequent attempts to create a user with the same email address will fail due to `UNIQUE` constraint).
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+  User               User[]
+|  stripeCustomerId   String?
+}
+```
+
+The upgrade flow looks like this:
+
+1. Count the number of users
+2. Create a subscription in Stripe that includes the number of users
+3. Associate the team with the Stripe custsomer ID to unlock paid features
```
thebiglabasky:
```suggestion
3. Associate the team with the Stripe customer ID to unlock paid features
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+This example has a problem: you can only run the logic _once_. Consider the following scenario:
+
+1. Stripe creates a new customer and subscription, and returns a customer ID
+2. Updating the team **fails** - the team is not marked as a customer in the Slack database
+3. The customer is charged by Stripe, but paid features are not unlocked in Slack because the team lacks a valid `customerId`
+4. Running the same code again either:
+
+   * Results in an error because the team (defined by `externalId`) already exists - Stripe never returns a customer ID
+   * If `externalId` is not subject to a unique constraint, Stripe creates a yet another subscription (**not idempotent**)
```
thebiglabasky:
```suggestion
   * If `externalId` is not subject to a unique constraint, Stripe creates yet another subscription (**not idempotent**)
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+You cannot re-run this code in case of error and you cannot change to another plan without being charged twice.
```
thebiglabasky:
```suggestion
You cannot re-run this code in case of an error and you cannot change to another plan without being charged twice. 
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
+---
+title: Prisma Client transactions guide
+tocDepth: 2
+---
+
+<TopBlock>
+
+A **transaction** is a unit of work that only succeeds if every operation within that transaction succeeds. For example, if you update 200 users within a transaction, each update must succeed - if not, all changes are rolled back and the transaction fails as a whole.
+
+Prisma Client supports six different ways of handling transactions for three different scenarios:
```
mhwelander:
```suggestion
Prisma Client supports five different ways of handling transactions for three different scenarios:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+    id: teamId,
+  }
+})
+```
+
+You can now retry the same logic multiple times with the same input without adverse effect. To further enhance this example, you can introduce a mechanism whereby the subscription is cancelled or temporarily deactivated if the update does not succeed after a set number of attempts.
+
+### Optimistic concurrency control
+
+Optimistic concurrency control (OCC) is a model for handling concurrent operations on a single entity that does not rely on 🔒 locking. Instead, we **optimistically** assume that a record will remain unchanged in between reading and writing, and use some kind of concurrency token (a timestamp or version field) to detect if a record has changed.
```
thebiglabasky:
```suggestion
Optimistic concurrency control (OCC) is a model for handling concurrent operations on a single entity that does not rely on 🔒 locking. Instead, we **optimistically** assume that a record will remain unchanged in between reading and writing, and use a concurrency token (a timestamp or version field) to detect if a record has changed. 
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+})
+```
+
+You can now retry the same logic multiple times with the same input without adverse effect. To further enhance this example, you can introduce a mechanism whereby the subscription is cancelled or temporarily deactivated if the update does not succeed after a set number of attempts.
+
+### Optimistic concurrency control
+
+Optimistic concurrency control (OCC) is a model for handling concurrent operations on a single entity that does not rely on 🔒 locking. Instead, we **optimistically** assume that a record will remain unchanged in between reading and writing, and use some kind of concurrency token (a timestamp or version field) to detect if a record has changed. 
+
+If a conflict ❌ occurs (someone else has changed the record since you read it), cancel the transaction. Depending on your scenario, you can then:
```
thebiglabasky:
```suggestion
If a conflict ❌ occurs (someone else has changed the record since you read it), you cancel the transaction. Depending on your scenario, you can then:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+
+The technique you choose depends on your particular use case.
+
+> **Note**: For the purposes of this guide, _writing_ to a database encompasses creating, updating, and deleting data.
+
+</TopBlock>
+
+## Dependent writes
+
+Writes are considered **dependent** one each other if:
```
mhwelander:
```suggestion
Writes are considered **dependent** on each other if:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+You can now retry the same logic multiple times with the same input without adverse effect. To further enhance this example, you can introduce a mechanism whereby the subscription is cancelled or temporarily deactivated if the update does not succeed after a set number of attempts.
+
+### Optimistic concurrency control
+
+Optimistic concurrency control (OCC) is a model for handling concurrent operations on a single entity that does not rely on 🔒 locking. Instead, we **optimistically** assume that a record will remain unchanged in between reading and writing, and use some kind of concurrency token (a timestamp or version field) to detect if a record has changed. 
+
+If a conflict ❌ occurs (someone else has changed the record since you read it), cancel the transaction. Depending on your scenario, you can then:
+
+* Re-try the transaction (book another cinema seat)
+* Throw an error (alert the user that they will overwrite made by someone else)
```
thebiglabasky: This sentence has an issue and I can't seem to understand exactly what it's supposed to say. Is that "alert the user that they will overwrite a change made by someone else"?
```suggestion
* Throw an error (alert the user that they are about to overwrite changes made by someone else)
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+}
+
+model Movie {
+  id    Int    @id     @default(autoincrement())
+  name  String @unique
+  seats Seat[]
+}
+```
+
+The following sample code finds the first available seat and assigns that seat to a user:s
```
thebiglabasky:
```suggestion
The following sample code finds the first available seat and assigns that seat to a user:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+  })
+```
+
+If any operation fails, Prisma rolls back the entire transaction. Nested writes are not currently supported by top-level bulk operations like `client.user.deleteMany` and `client.user.updateMany`.
+
+#### When to use nested writes
+
+Consider using nested writes if:
+
+* ✔ You want to write to create two or more records related by ID at the same time (for example, create a blog post and a user)
```
mhwelander:
```suggestion
* ✔ You want to create two or more records related by ID at the same time (for example, create a blog post and a user)
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+  data: {
+    claimedBy: userId,
+  },
+  where: {
+    id: availableSeat.id,
+  },
+})
+```
+
+However, this code suppers from the "double booking problem" - it is possible for two people to book the same seats:
```
thebiglabasky:
```suggestion
However, this code suffers from the "double-booking problem" - it is possible for two people to book the same seats:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,773 @@
...
+  throw new Error(`That seat is already booked! Please try again.`)
+}
+```
+
+It is now impossible for two people to book the same seat:
+
+1. Seat 3A returned to Sorcha (`version` is 0)
+2. Seat 3A returned to Ellen (`version` is 0)
+3. Seat 3A claimed by Sorcha (`version` is incremented to 1, booking succeeds)
+4. Seat 3A claimed by Ellen (in-memory `version` (0) does not match database `version` (1) - booking does not succeed)
```
thebiglabasky: Some kind of conclusion would be nice.
I would also add why OCC is useful: the fact that it avoids locking makes the application much more resilient to a high number of concurrent requests and hence is more scalable overall.
Locking may end up impacting the performance of the application, or, in the worst-case scenario, lead to an outage in case the application receives a lot of concurrent requests.

thebiglabasky (approved):
Some nit picks and typos + one main request: adding why OCC is valuable

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+    },
+  },
+})
+
+await prisma.$transaction([createUserOperation, createTeamOperation])
+```
+
+##### Are cascading deletes supported by the Prisma Client nested writes?
+
+No, you cannot use Prisma Client's nested writes to perform a cascading delete - for example, delete a team and all of its members. Instead, you can:
```
matthewmueller: This reads like a "hard no", but this can be worked around pretty easy by doing cascading deletes at the DB level. 
```suggestion
Although you cannot specifically use _nested writes_ to perform a cascading delete (for example, delete a team and all of its members), you can still perform cascading deletes in the following ways:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+You an use the [`$transaction` API to perform a cascading delete](#scenario-cascading-delete).
+
+#### Can I use bulk operations with the `$transaction` API?
+
+Yes - for example, you can include multiple `deleteMany` operations inside a `$transaction`.
+
+### <inlinecode>$transaction</inlinecode> API
+
+The `$transaction` API is generic solution to independent writes that allows you to run multiple operations as a single, atomic operation - if any operation fails, Prisma rolls back the entire transaction. As Prisma Client evolves, use cases for the `$transaction` API will increasingly be replaced by more specialized bulk operations (such as `createMany`) and nested writes.
```
matthewmueller: As the Prisma Client
```suggestion
The `$transaction` API is generic solution to independent writes that allows you to run multiple operations as a single, atomic operation - if any operation fails, Prisma rolls back the entire transaction. As the Prisma Client evolves, use cases for the `$transaction` API will increasingly be replaced by more specialized bulk operations (such as `createMany`) and nested writes.
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+model PrivateMessage {
+  id         Int     @id @default(autoincrement())
+  user       User    @relation(fields: [userId], references: [id])
+  userId     Int  
+  message      String
+}
+```
+
+If a user invokes the right to be forgotten, we must delete three records: the user record, private messages, and posts. It is critical that _all_ delete operations succeed together or not at all, which makes this a use case for a transaction. However, using a single bulk operation like `deleteMany` is not possible in this scenario because we need to delete across three models. Instead, we can use the `$transaction` API to run three operations together - two `deleteMany` and one `delete`:
```
matthewmueller, line 412: Ah, this is an interesting example because the models could be related (user's messages and posts), but we lack the capabilities to do this nested write because we don't support nested deleteMany's. 

Taking a mental note of this.

The way I've thought about $transaction so far is that it'll eventually only be for unrelated records. It's really hard to come up with a use case for operating on unrelated data, so I think this is fine to keep.

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+        team: {
+          id: teamID
+        }
+      },
+    },
+  },
+})
+```
+
+There's no compelling reason to switch to manually generated IDs and the `$transaction` API if you are already using auto-generated IDs and nested writes.
```
matthewmueller, line 525: Glad to see this clarification!

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+* Write the value back to the database
+
+All operations should **succeed or fail together** without making unwanted changes to the database, but you do not necessarily need to use an actual database transaction. This section of the guide describes two ways to work with Prisma Client and the read-modify-write pattern:
+
+* Designing idempotent APIs
+* Optimistic concurrency control
+
+### Idempotent APIs
+
+Idempotency is the ability to run the same logic with the same parameters multiple times with the same result: the **effect on the database** is the same whether you run the logic one or one thousand times. For example:
```
matthewmueller: once or one thousand times
```suggestion
Idempotency is the ability to run the same logic with the same parameters multiple times with the same result: the **effect on the database** is the same whether you run the logic once or one thousand times. For example:
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+* **NOT IDEMPOTENT**: Create a user in the database with email address "letoya@prisma.io". The `User` table **does not** enforce unique email addresses. The effect on the database is different if you run the logic once (one user created) or ten times (ten users created).
+* **IDEMPOTENT**: Create a user in the database with the email address "letoya@prisma.io". The `User` table **does** enforce unique email addresses. The effect on the database is the same if you run the logic once (one user created) or ten times (subsequent attempts to create a user with the same email address will fail due to `UNIQUE` constraint).
```
matthewmueller: Hmmmmmmmmm.... I think we might want a better example here. It's idempotent from the database perspective, but it's not naturally idempotent from the application-layer.

If I run create once it works, if I run create again with the same inputs, it throws. You could make it idempotent with try-catch logic, but it feels like we could simplify.

Upserts are usually the canonical example of idempotency, you run the same script over and over and over and if you pass in the same inputs your application does the same thing and the database is in the same state.

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+Idempotency is something you can actively design, and is not always needed or possible.
```
matthewmueller: I'd maybe add that if you can pull off idempotency you should try. Operationally it will make your life a lot easier. But sometimes you can't and that's ok.

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+      },
+    },
+  },
+})
+
+// Find customer in Stripe
+let customer = await stripe.customers.get({ externalId: teamID })
+
+if (customer) {
+  customer = await stripe.customers.update({ // If team already exists, update
```
matthewmueller: Reduce horizontal scrolling by placing the comments above.
```suggestion
  // If team already exists, update
  customer = await stripe.customers.update({ 
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+    id: teamId,
+  }
+})
+```
+
+You can now retry the same logic multiple times with the same input without adverse effect. To further enhance this example, you can introduce a mechanism whereby the subscription is cancelled or temporarily deactivated if the update does not succeed after a set number of attempts.
+
+### Optimistic concurrency control
+
+Optimistic concurrency control (OCC) is a model for handling concurrent operations on a single entity that does not rely on 🔒 locking. Instead, we **optimistically** assume that a record will remain unchanged in between reading and writing, and use a concurrency token (a timestamp or version field) to detect if a record has changed.
```
matthewmueller: Minor nitpick: I love the use of icons, but it seems like you're switching between `<icon> <keyword>` and `<keyword> <icon>`. See conflict below.

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+Idempotency is the ability to run the same logic with the same parameters multiple times with the same result: the **effect on the database** is the same whether you run the logic one or one thousand times. For example:
+
+* **NOT IDEMPOTENT**: Create a user in the database with email address "letoya@prisma.io". The `User` table **does not** enforce unique email addresses. The effect on the database is different if you run the logic once (one user created) or ten times (ten users created).
+* **IDEMPOTENT**: Create a user in the database with the email address "letoya@prisma.io". The `User` table **does** enforce unique email addresses. The effect on the database is the same if you run the logic once (one user created) or ten times (subsequent attempts to create a user with the same email address will fail due to `UNIQUE` constraint).
+
+Idempotency is something you can actively design, and is not always needed or possible.
+
+#### When to design an idempotent API
+
+* ✔ You need to be able to retry the same logic without creating unwanted side-effects in the databases
```
matthewmueller, line 551: Love the checkmarks, any chance those could **be** the bullet points? 😅

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+Optimistic concurrency control (OCC) is a model for handling concurrent operations on a single entity that does not rely on 🔒 locking. Instead, we **optimistically** assume that a record will remain unchanged in between reading and writing, and use a concurrency token (a timestamp or version field) to detect if a record has changed. 
+
+If a conflict ❌ occurs (someone else has changed the record since you read it), you cancel the transaction. Depending on your scenario, you can then:
+
+* Re-try the transaction (book another cinema seat)
+* Throw an error (alert the user that they are about to overwrite changes made by someone else)
+
+#### When to use optimistic concurrency control
+
+* ✔ You anticipate that conflicts will be _rare_ - in some scenarios, such as editing shared documents, a lock (pessimistic concurrency control) might be more appropriate
```
matthewmueller: We don't really have a good solution for when conflicts are frequent AND you can't use idempotency :-/

The solution for that is the long-running transaction forbidden fruit. 

Might be nice to keep for the sake of transparency, but I could also see us not mentioning that tradeoff. Retries would work, your application would just be stuck spinning for awhile on those front-row seats for that Justin Beiber concert.

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  throw new Error(`That seat is already booked! Please try again.`)
+}
+```
+
+It is now impossible for two people to book the same seat:
+
+1. Seat 3A returned to Sorcha (`version` is 0)
+2. Seat 3A returned to Ellen (`version` is 0)
+3. Seat 3A claimed by Sorcha (`version` is incremented to 1, booking succeeds)
+4. Seat 3A claimed by Ellen (in-memory `version` (0) does not match database `version` (1) - booking does not succeed)
```
matthewmueller, line 803: It ends kind of abruptly. I'm sorry for not writing a conclusion 😅

matthewmueller (approved):
Amazing wordsmithing. I enjoyed seeing this rough rock turn into a polished gem.

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  email String @unique
+  teams Team[] // Many teams
+}
+```
+
+The most straightforward approach is to create a team, then create and attach a user to that team:
+
+```ts
+// Create a team
+await prisma.team.create({
```
nikolasburk:
```suggestion
const team = await prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  data: {
+    name: 'Aurora Adventures',
+  },
+})
+
+// Create a user and assign them to the team
+await client.user.create({
```
nikolasburk:
```suggestion
const user = await prisma.user.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+However, this code has a problem - consider the following scenario:
+
+1. Creating the team succeeds - "Aurora Adventures" is now taken
+2. Creating and connecting the user fails - the team "Aurora Adventures" exists, but has no users
+3. Going through the sign-up flow again and attempting recreate "Aurora Adventures" fails - the team already exists
+
+Creating a team and adding a user should be one atomic operation that **succeeds or fails as a whole**.
+
+To implement atomic writes in a low-level database clients, you must wrap your inserts in `BEGIN`, `COMMIT` and `ROLLBACK` queries. Prisma Client solves the problem with nested writes. The following query creates a team, creates a user, and connects the records in a single transaction:
```
nikolasburk: maybe link nested writes to the docs corresponding docs page here: https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#nested-writes

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+```ts
+await client.team.create({
```
nikolasburk:
```suggestion
const team = await prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+##### Can I perform multiple nested writes - for example, create two new teams and assign users?
+
+Yes, but this is a combination of scenarios and techniques:
+
+* Creating a team and assigning users is a dependent write - use nested writes
+* Creating all teams and users at the same time is an independent write because team/user combination #1 and team/user combination #2 are unrelated writes - use the `$transaction` API
+
+```ts
+// Nested write
+const createOne = client.team.create({
```
nikolasburk:
```suggestion
const createOne = prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+    members: {
+      create: {
+        email: 'alice@prisma.io',
+      },
+    },
+  },
+})
+
+// Nested write
+const createTwo = client.team.create({
```
nikolasburk:
```suggestion
const createTwo = prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  subject String
+  body    String
+  unread  Boolean
+}
+```
+
+Based on this schema, you can use `updateMany` to mark all unread emails as read:
+
+```ts
+await client.email.updateMany({
```
nikolasburk:
```suggestion
await prisma.email.updateMany({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  const deleteMany = await prisma.team.deleteMany({
+    where: {
+      id: {
+        in: [2, 99, 2, 11]
+      }
+    },
+    data: {
+      members: { } // Cannot access members here
+    },
+  })
```
nikolasburk:
```suggestion
await prisma.team.deleteMany({
  where: {
    id: {
      in: [2, 99, 2, 11]
    }
  },
  data: {
    members: { } // Cannot access members here
  },
})
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+    prisma.user.create({
+      data: {
+        id: userID,
+        email: 'alice@prisma.io',
+        team: {
+          id: teamID
+        }
+      }
+    }),
+    client.team.create({
```
nikolasburk:
```suggestion
    prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+```ts
+const id = 9; // User to be deleted
+
+const deletePosts = prisma.post.deleteMany({
+  where: {
+    userId: id
+  }
+})
+
+const deleteMessages = prisma.privatemessage.deleteMany({
```
nikolasburk:
```suggestion
const deleteMessages = prisma.privateMessage.deleteMany({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+await prisma.$transaction([deletePosts, deleteMessages, deleteUser]) // Operations succeed or fail together
+```
+
+#### Scenario: Pre-computed IDs and the `$transaction` API
+
+Dependent writes are not supported by the `$transaction` API - if operation A relies on the ID generated by operation B, use [nested writes](#nested-writes). However, if you _pre-computed_ IDs (for example, by generating GUIDs), your writes become independent. Consider the sign-up flow from the nested writes example:
+
+```ts
+await client.team.create({
```
nikolasburk:
```suggestion
await prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  },
+})
+```
+
+Instead of auto-generating IDs, change the `id` fields of `Team` and `User` to a `String` (if you do not provide a value, a UUID is generated automatically). This example uses UUIDs:
+
+```prisma highlight=2,9;delete|3,10;add
+model Team {
+  id   Int    @id @default(autoincrement())  
+  id   String  @id @default(uuid())
```
nikolasburk:
```suggestion
  id   Int       @id @default(autoincrement())  
  id   String    @id @default(uuid())
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  name String
+  members User[]
+}
+
+model User {
+  id   Int    @id @default(autoincrement())
```
nikolasburk:
```suggestion
  id   Int     @id @default(autoincrement())  
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+Technically you can still use nested writes with pre-computed APIs if you prefer that syntax:
+
+```ts
+import { v4 } from 'uuid'
+
+const teamID = v4();
+const userID = v4();
+
+await client.team.create({
```
nikolasburk:
```suggestion
await prisma.team.create({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
nikolasburk: plus one!  created an issue here: https://github.com/prisma/docs/issues/1180

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+#### Scenario: Upgrading a Slack team
+
+You are creating an upgrade flow for Slack that allows teams to unlock paid features. Teams can choose between different plans and pay per user, per month. You use Stripe as your payment gateway, and extend your `Team` model to store a `stripeCustomerId`. Subscriptions are managed in Stripe.
+
+```prisma
+model Team {
+  id                 Int    @id @default(autoincrement())
+  name               String
+  User               User[]
+|  stripeCustomerId   String?
```
nikolasburk: what's the reason you're not using the new way of line highlight here?

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+All operations should **succeed or fail together** without making unwanted changes to the database, but you do not necessarily need to use an actual database transaction. This section of the guide describes two ways to work with Prisma Client and the read-modify-write pattern:
+
+* Designing idempotent APIs
+* Optimistic concurrency control
+
+### Idempotent APIs
+
+Idempotency is the ability to run the same logic with the same parameters multiple times with the same result: the **effect on the database** is the same whether you run the logic one or one thousand times. For example:
+
+* **NOT IDEMPOTENT**: Create a user in the database with email address "letoya@prisma.io". The `User` table **does not** enforce unique email addresses. The effect on the database is different if you run the logic once (one user created) or ten times (ten users created).
```
nikolasburk: it feels a bit weird that the email gets hyperlinked in the rendered version, maybe add backticks?

```suggestion
* **NOT IDEMPOTENT**: Create a user in the database with email address `"letoya@prisma.io"`. The `User` table **does not** enforce unique email addresses. The effect on the database is different if you run the logic once (one user created) or ten times (ten users created).
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
nikolasburk:
```suggestion
* **IDEMPOTENT**: Create a user in the database with the email address `"letoya@prisma.io"`. The `User` table **does** enforce unique email addresses. The effect on the database is the same if you run the logic once (one user created) or ten times (subsequent attempts to create a user with the same email address will fail due to `UNIQUE` constraint).
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+   * Results in an error because the team (defined by `externalId`) already exists - Stripe never returns a customer ID
+   * If `externalId` is not subject to a unique constraint, Stripe creates yet another subscription (**not idempotent**)
+
+You cannot re-run this code in case of an error and you cannot change to another plan without being charged twice. 
+
+The following refactor (highlighted) introduces a mechanism that checks if a subscription already exists, and either creates the description or updates the existing subscription (which will remain unchanged if the input is identical):
+
+```ts highlight=12-27;normal
+// Calculate the number of users times the cost per user
+const numTeammates = await client.user.count({
```
nikolasburk:
```suggestion
const numTeammates = await prisma.user.count({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  customer = await stripe.customers.create({ // If team does not exist, create customer
+    externalId: teamId,
+    plan: "plan_id",
+    quantity: numTeammates
+  })
+}
+
+// Update the team with the customer id to indicate that they are a customer
+// and support querying this customer in Stripe from our application code.
+await client.team.update({
```
nikolasburk:
```suggestion
await prisma.team.update({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  },
+  where: {
+    id: teamId,
+  }
+})
+```
+
+You can now retry the same logic multiple times with the same input without adverse effect. To further enhance this example, you can introduce a mechanism whereby the subscription is cancelled or temporarily deactivated if the update does not succeed after a set number of attempts.
+
+### Optimistic concurrency control
```
nikolasburk, line 664: we should link to this issue somewhere in the section: https://github.com/prisma/prisma/issues/4988

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+}
+```
+
+The following sample code finds the first available seat and assigns that seat to a user:
+
+```ts
+const movieName = "Hidden Figures";
+
+// Find first available seat
+const availableSeat = await client.seat.findFirst({
```
nikolasburk:
```suggestion
const availableSeat = await prisma.seat.findFirst({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+  },
+})
+
+// Throw an error if no seats are available
+if (!availableSeat) {
+  throw new Error(`Oh no! ${movieName} is all booked.`)
+}
+
+// Claim the seat
+await client.seat.update({
```
nikolasburk:
```suggestion
await prisma.seat.update({
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+Even though Sorcha has successfully booked the seat, the system ultimately stores Ellen's claim. To solve this problem with optimistic concurrency control, add a `version` field to the seat:
+
+```prisma highlight=7;normal
+model Seat {
+  id        Int   @id @default(autoincrement())
+  userId    Int?
+  claimedBy User? @relation(fields: [userId], references: [id])
+  movieId   Int
+  Movie     Movie @relation(fields: [movieId], references: [id])
```
nikolasburk:
```suggestion
  movie     Movie @relation(fields: [movieId], references: [id])
```

nikolasburk (commented):
This looks awesome and I think it'll be an incredibly valuable resources for folks, great job Matt and Martina 👏 

I noticed that often times `client` was used for the variable name of `PrismaClient` and suggested to change them to `prisma`.

Also, a few more suggestions:

- I think we should link to our [transactions blog post](https://www.prisma.io/blog/how-prisma-supports-transactions-x45s1d5l0ww1) somewhere
- For many `PrismaClient` code snippets, I think it could be interesting to consider to add a SQL tab to the same code block so that people see what this code will translate to when it's being executed against the DB
- can we add some diagrams/illustrations to make some concepts in the guide more visual?
- I'd still prefer calling the guide just "Transactions" in the sidenav (or at least "Prisma Client transactions", the "guide" feels a bit superfluous since it's alrteady in the "Guides" section and therefore that's kind of implied)

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
mhwelander: Actually you did, that's my mistake!!

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
mhwelander:
```suggestion
Idempotency is something you can and should actively design into your application wherever possible.
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
```diff
@@ -0,0 +1,778 @@
...
+
+#### When to design an idempotent API
+
+* ✔ You need to be able to retry the same logic without creating unwanted side-effects in the databases
+
+#### Scenario: Upgrading a Slack team
+
+You are creating an upgrade flow for Slack that allows teams to unlock paid features. Teams can choose between different plans and pay per user, per month. You use Stripe as your payment gateway, and extend your `Team` model to store a `stripeCustomerId`. Subscriptions are managed in Stripe.
+
+```prisma
```
mhwelander:
```suggestion
```prisma highlight=5;normal
```

content/300-guides/100-prisma-guides/100-prisma-client-transactions-guide.mdx
mhwelander:
```suggestion
  stripeCustomerId   String?
```

mhwelander:
GH issues for remaining tasks:

* #1185 
* #1184 
* #1183

PR #1159, opened 2021-01-08
fix: make search result block completely clickable

Jolg42:
It happens often to me that I click on the block but the a link doesn't cover the whole block so I it ends up closing the results instead of going to the link.

This will make the search a better to me & others ❤️


hrescak:
we should maybe consider making the entire row clickable?

Jolg42:
@hrescak not sure it makes a big difference 🤔

PR #1164, opened 2021-01-12
add @updateAt remark. 

### LINKED ISSUE #1181: Document @updatedAt attribute behaviour
## Bug description
Not a bug but not a feature request either. :)

The @updatedAt attribute defined in Prisma schema only sets the time the model was updated when the client caller leaves the field as `undefined`. This is useful behaviour and should be documented for future users.

See [here](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat)

matthewmueller:
Closes: https://github.com/prisma/prisma/issues/4941

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -262,7 +262,7 @@ generator client {
 
 Defines a Prisma [model](../../concepts/components/prisma-schema/data-model#defining-models) <span class="concept"></span>.
 
-### Remarks 
+### Remarks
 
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -1263,6 +1263,7 @@ Automatically stores the time when a record was last updated. Prisma Client auto
 
 #### Remarks
 
+* Override `@updatedAt` with your own field value
```
janpio, line 1266: What does that mean? I don't understand.

matthewmueller:
Better PR: https://github.com/prisma/docs/pull/1165

PR #1175, opened 2021-01-14
create another support article

ryands17:


content/500-support/100-help-articles/300-finding-entities-based-on-relation.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+```js
+await prisma.user.findMany({
+  where: { posts: { none: {} } },
+})
+```
+
+This will give all the users that haven't created any posts yet.
+
+In a similar way, you can create a query for _Users who have one or more Posts_ as follows:
```
nikolasburk:
```suggestion
In a similar way, you can query for **users that created at least one post** as follows:
```

content/500-support/100-help-articles/300-finding-entities-based-on-relation.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+```js
+await prisma.user.findMany({
+  where: { posts: { some: {} } },
+})
+```
+
+This will give all the users that have one or mosts made.
```
nikolasburk:
```suggestion
This query returns all the users that created at least one post.
```

content/500-support/100-help-articles/300-finding-entities-based-on-relation.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: 'Finding entities based on their relation existence'
+metaTitle: 'Finding entities based on their relation existence'
```
nikolasburk:
```suggestion
title: 'Querying models based on their relations existence'
metaTitle: 'Querying models based on their relations existence'
```

nikolasburk (changes requested):
This is awesome, thanks a lot Ryan! I've made a few suggestions, mostly with respect to style and formatting but also minor content adjustments:

- I noticed that you provided a solution for a problem that wasn't outlined in the problem section, so I added this to the problem section as well
- In docs, we should never use "I" or "me" (this might be ok in blog posts that have a clear author so people know who the person is that's writing I or me, but in docs we should remain neutral and only talk to the reader directly, see the style guide [here](https://www.prisma.io/docs/about/style-guide#use-you-as-the-pronoun))
- I made sure consistent formatting is used

content/500-support/100-help-articles/300-finding-entities-based-on-relation.mdx
ryands17: @nikolasburk I think **Querying models based on their relations** seems too generic and could be anything but this article is specific should it be something like ****Querying models based on their relations existence**

PR #1179, opened 2021-01-14
Update 001-sql-server-start-from-scratch.mdx - No connection could be made because the target machine actively refused it. (os error 10061)

xstos:
Got stuck in the tutorial with the following error:

Error: Error in connector: Error querying the database: Error querying the database: Error querying the database: An error occured during the attempt of performing I/O: No connection could be made because the target machine actively refused it. (os error 10061)

Figured it will help others get unstuck faster. :)


mhwelander:
Thank you! 💚

PR #1182, opened 2021-01-15
Update 00-deploying-to-heroku.md

abulhuman:
The current push command results in 
```
error: src refspec main does not match any
error: failed to push some refs to 'https://git.heroku.com/your-app-name.git'
```

### DIFF
--- a/content/300-guides/200-deployment/00-deploying-to-heroku.md
+++ b/content/300-guides/200-deployment/00-deploying-to-heroku.md
@@ -234,7 +234,7 @@ The first command will create an SQL migration file based on the Prisma schema i
 Deploy the app by pushing the changes to the Heroku app repository:
 
 ```no-lines
-git push heroku main
+git push heroku master
 ```
 
 This will trigger a build and deploy your application to Heroku.


janpio (changes requested):
I think newer Heroku deployments only use `main`.

abulhuman:
> I think newer Heroku deployments only use `main`.

I followed the guide to that point and only replacing `main` by `master` in the command worked.

abulhuman:
> I think newer Heroku deployments only use `main`.

more like newer version of git is required where the `master` branch is replaced by `main` by default.

And I think it should made clear from the beginning on the prerequisites section of the guide that the command `git init` should result in a `main` branch rather than a `master` branch.

janpio:
Yes, that might be valid - but that should be an issue then and not a PR that just replaces `main` with `master`. If this would be merged, many people could not work with the instructions any more.

abulhuman:
> Yes, that might be valid - but that should be an issue then and not a PR that just replaces `main` with `master`. If this would be merged, many people could not work with the instructions any more.

Understood sir.

2color:
I resolved this by ensuring that in the beginning of the guide, you're instructed to change to the `main` branch: https://github.com/prisma/docs/commit/ea282f1bf57f7e91517e26fe72c7f15b0c4b2d63

PR #1186, opened 2021-01-15
Add output field to generator client definition example in 03-generators.mdx

ultd:
Added output field to generator client definition as some people will wonder how to do this.


johannesschobel:
output is not directly required, right? this is an optional parameter, so we can omit it here

molebox:
Hey @ultd! Specifiying the output path is documented [here](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client#using-a-custom-output-path). Im going to go ahead and close this PR. Thanks for taking the time though 😊

PR #1197, opened 2021-01-21
Fix the Heroku deployment guide

2color:
Change Heroku deployment guide to use `db push`, because currently using Prisma Migrate requires the ability to create the shadow database which isn't permitted in the free PostgreSQL instance provided by Heroku.


nikolasburk:
Awesome, great catch Daniel!

PR #1199, opened 2021-01-21
Fix example to be valid sql statement

deodad:
The example diff resulted in `ALTER TABLE "User" DROP COLUMN "name", RENAME "name" TO "fullname",`.

Changed the diff to result in a valid rename statement.


johannesschobel (approved):

mhwelander:
Thanks for taking the time! 💚

PR #1200, opened 2021-01-21
Document usage with PgBoucner

2color:


content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
```diff
@@ -74,18 +74,45 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
...
 
+### Usage with PgBouncer
+
+Typically PostgreSQL supports only a certain amount of concurrent connections and this limit can be reached quite fast when the service usage goes up – especially in serverless environments.
+
+A connection pooler like PgBouncer prevents your application from exhausing the database's connection limit.
+
+PgBouncer holds a connection pool to the database and proxies incoming client connections by sitting between Prisma Client and the database.
+
+PgBouncer funnels client connections into a pooling application, which reduces the number of processes a database has to handle at any given time. PgBouncer passes on a limited number of connections to the database and queues additional connections for delivery when space becomes available.
```
janpio: First half of the first sentence says the same as the previous paragraph.

content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
```diff
@@ -74,18 +74,45 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
...
+
+To use Prisma Client with PgBouncer from a serverless function, add the `?pgbouncer=true` flag to the PostgreSQL connection string and set the connection limit to `1`:
```
janpio: Setting Connection Limit to 1 is unrelated to usage of PgBouncer. It is a thing you can do in specific circumstances (when you are running on serverless with an application that never uses more than 1 connectino in parallel or you really want to optimize on the number of serverless functions that can run in parallal)

content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
```diff
@@ -74,18 +74,45 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
...
+```
+
+Additionally, for Prisma Client to work reliably, PgBouncer must run in [**Transaction mode**](https://www.pgbouncer.org/features.html).
+
+_Transaction mode_ offers a connection for every transaction. With this mode in addition to the , the Prisma query engine is able to operate normally.
+
+<details>
+  <summary>How `pgbouncer` mode works in Prisma</summary>
+- Prisma will clean up already present prepared statements in the connection by running `DEALLOCATE ALL` before trying to prepare and execute the Prisma Client query.
+- Prisma will triggers a transaction for every query case –even when just reading data– which allows Prisma to use prepared statements.
```
janpio:
```suggestion
- Prisma will trigger a transaction for every query case – even when just reading data – which allows Prisma to use prepared statements.
```

content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
```diff
@@ -74,18 +74,43 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
 mysql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
 ```
 
+### Usage with PgBouncer
+
+Typically PostgreSQL supports only a certain amount of concurrent connections and this limit can be reached quite fast when the service usage goes up – especially in serverless environments.
```
nikolasburk: Typically seems unecessary here?

```suggestion
PostgreSQL supports only a certain amount of concurrent connections and this limit can be reached quite fast when the service usage goes up – especially in serverless environments.
```

content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
```diff
@@ -74,18 +74,43 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
...
+
+To use Prisma Client with PgBouncer from a serverless function, add the `?pgbouncer=true` flag to the PostgreSQL connection string:
+
+```
+postgresql://USER:PASSWORD@HOST:PORT/DATABASE?pgbouncer=true
+```
+
+Additionally, for Prisma Client to work reliably, PgBouncer must run in [**Transaction mode**](https://www.pgbouncer.org/features.html).
+
+_Transaction mode_ offers a connection for every transaction. With this mode in addition to the , the Prisma query engine is able to operate normally.
```
nikolasburk: seems like this sentence is not complete? "in addition to the ..."?

content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
```diff
@@ -74,18 +74,43 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
...
+
+<details>
```
nikolasburk, line 95: Seems like the content in here doesn't render properly as Markdown right now:

image

Maybe try to add new lines before and after the bullet list? If that doesn't work, we should summon Nilu to fix the rendering.

nikolasburk (changes requested):
This looks great Daniel 👏 I've left a few comments but overall I think this will be really helpful content for a lot of folks 🙂

content/200-concepts/100-components/02-prisma-client/190-deployment.mdx
2color: Yep, this was a whitespace problem. Adding one before and after the bullet list fixed it:
Screen Shot 2021-01-26 at 18 53 58

PR #1212, opened 2021-01-25
Update what is Prisma page for Migrate

albertoperdomo:


albertoperdomo:
Images are now inconsistent. Maybe we should ask design to produce one for Migrate that mimics the one for introspection.

One question I have is if we still want the .node_modules detail in the diagram. I think it's an unnecessary detail.

PR #1216, opened 2021-01-26
Fix broken quickstart link

khrj:
Changes
- `https://www.prisma.io/docs/getting-started/quickstart/`
to
- `https://www.prisma.io/docs/getting-started/quickstart-typescript/`


khrj:
Note that if you go to that link directly it redirects, but the Gatsby link on the edited page doesn't

nikolasburk:
Thanks a lot for catching this @KhushrajRathod 🚀

PR #1219, opened 2021-01-26
Grammer: Added some missing words

molebox:
A couple of words were missing, no biggie


nikolasburk:
Thanks so much for catching these @molebox 🙌

PR #1220, opened 2021-01-26
fix(broken-links): fixed quickstart broken links

angeloanan:
There were a few broken links that points to the quickstart page.

This PR attempts to fix those broken links.


johannesschobel (approved):
LGTM

PR #1221, opened 2021-01-26
findOne updated to findUnique on Soft Delete Middleware Example Docs

danwetherald:
Looks like the docs are using an old method findOne that is no longer supported by prisma client.


johannesschobel (approved):
LGTM, although there are a few other `findOne()` mentions in the docs as well.
Maybe we can remove them as well in another PR?

mhwelander:
#1239

PR #1240, opened 2021-01-31
Data model docs minor inconsistency

stefanhar:
In User model's fields from the sample schema post is array and profile is optional in this table this is swapped around


mhwelander:
Markdown tables - a nightmare! Or maybe I should stop using word wrap. 😅 Thanks for fixing this!

PR #1273, opened 2021-02-06
Fix rogue '[' in Extended native types support

khrj:


mhwelander:
Thanks for taking the time to fix!

PR #1274, opened 2021-02-07
fix: DB seed example command

maoosi:
Fixed wrong example for database seeding command.


Jolg42:
Thanks @maoosi for the PR 💚

PR #1276, opened 2021-02-08
Update 06-relations.mdx

meotimdihia:
Fixed wrong key on example SQL.


mhwelander:
Thanks for taking the time to PR!

PR #1314, opened 2021-02-16
fix \$transaction

matthewmueller:
Prettier automatically escapes `$` to `\$`, so I think this will continue to be a problem. Any reason we go with `<inlinecode>$transaction</inlinecode>` and not backticks?

https://www.prisma.io/docs/guides/prisma-guides/prisma-client-transactions-guide#transaction-api

CleanShot 2021-02-16 at 13 22 38@2x


mhwelander:
We need the `inlinecode` part in order to render the code style in navigation; am not sure of the details - I've added `<!-- prettier-ignore -->` to that section in my current branch to stop Prettier from 'fixing' it.

PR #1332, opened 2021-02-18
Fix wording for @@map description

jbroomer:
I think the description in the table for `@@map` should read `The database table name.` instead of `The database column name.` The way it currently stands makes `@map` and `@@map` appear to do the same thing.


mhwelander:
Thanks!

PR #1348, opened 2021-02-24
SQL Server shared default values

pimeys:
In 2.17.0, a shared default object will lead to panic when introspecting. 2.18.0 fixes the panic, but how we handle them should be in our documentation.

### DIFF
--- a/content/200-concepts/100-components/07-preview-features/100-sql-server/index.mdx
+++ b/content/200-concepts/100-components/07-preview-features/100-sql-server/index.mdx
@@ -126,6 +126,50 @@ Certain migrations will cause more changes than you might expect. For example:
 - Adding or removing `autoincrement()`. This cannot be achieved by modifying the column, but requires recreating the table (including all constraints, indices, and foreign keys) and moving all data between the tables.
 - Additionally, it is not possible to delete all the columns from a table (possible with PostgreSQL or MySQL). If a migration needs to recreate all table columns, it will also re-create the table.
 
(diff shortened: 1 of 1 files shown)


mhwelander:
🙏 You're the best!

content/200-concepts/100-components/07-preview-features/100-sql-server/index.mdx
```diff
@@ -126,6 +126,50 @@ Certain migrations will cause more changes than you might expect. For example:
...
+And the first doesn't get the default value introspected:
+
+```prisma file=schema.prisma
+model cats {
+  id   Int     @id @default(autoincrement())
+  name String?
+}
+```
+
+If using Prisma Migrate together with shared defult objects, changes to them must be done manually to the SQL.
```
tomhoule: typo: defult => default

tomhoule (approved):
Do you think we need an issue about this to link in the docs? The docs look good!

pimeys:
> Do you think we need an issue about this to link in the docs? The docs look good!

We do not have one yet. This kind of needs @prisma/product input maybe?

albertoperdomo:
>Do you think we need an issue about this to link in the docs? The docs look good!

You mean to track the potential implementation to support this? I don't think we need a GH issue for every single gap. Instead, I would encourage users to request support via GH if they care about it.

That said, I'm open to suggestions and if we know already users want this we could create it.

PR #1372, opened 2021-03-02
Add extra remark about not disconnecting in serverless environments

Siilwyn:
Clarify keeping connection open in a serverless environment.

The connection management documentation suggest there are nuances that make it sensible
to reuse database connections, linking to this deployment documentation.


Siilwyn:
This is a followup from this Slack thread:
https://app.slack.com/client/T0MQBS8JG/CKQTGR6T0/thread/CKQTGR6T0-1614349810.348400

Additional context: I was accidentally calling `$disconnect` too, since it is used in multiple examples, then I stumbled on this issue:
https://github.com/mikecao/umami/issues/263

PR #1375, opened 2021-03-03
Add bottom padding to TOCList

shreyas44:
- Change the bottom padding of `TOCList` component. 

**Before:**
Screen Shot 2021-03-03 at 10 04 25 AM

**After:**
Screen Shot 2021-03-03 at 10 03 48 AM

Notice the difference in the spacing in the bottom right corner.

It's a really small thing that bothered me for no reason, so just thought of fixing it.

### DIFF
--- a/src/components/toc.tsx
+++ b/src/components/toc.tsx
@@ -15,7 +15,7 @@ const ChapterTitle = styled.div`
 `
 
 const TOCList = styled.ul`
-  padding: 0;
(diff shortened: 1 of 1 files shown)


hrescak (approved):
I defer to @nilubava to check if this interferes with the scroll logic but LGTM from my point

nilubava (approved):
LGTM as well :) Merging

PR #1390, opened 2021-03-05
Update 010-sql-server-connection-string.mdx

Jolg42:
parameter will be `pool_timeout` in 2.19 from what Julius said


janpio:
That will apply to all/most parameters listed here. I think Martina noted that somewhere to do this with 2.19.

Jolg42:
Indeed, closing

PR #1411, opened 2021-03-15
Update 030-crud.mdx

ssengalanto:
Fix typography error under [Cascading Deletes (deleting related records)](https://www.prisma.io/docs/concepts/components/prisma-client/crud#cascading-deletes-deleting-related-records) section, the diff in Post model are both optional.


mhwelander:
Good spot, thank you!

PR #1420, opened 2021-03-15
Fixed logic error in generated types example

dslatkin:
[This page of the docs](https://www.prisma.io/docs/concepts/components/prisma-client/crud) is demonstrating record creation with generated types, but it looks like there was an "else" keyword forgotten for the case when posts shouldn't be included. I believe this should fix it.


mhwelander:
Good spot, thanks for fixing!

PR #1450, opened 2021-03-19
Fix a broken link

enm10k:
I found a broken link on this page https://www.prisma.io/docs/concepts/components/prisma-client .


mhwelander:
Thank you!

PR #1460, opened 2021-03-21
drop a duplicate article

similegian:
noticed a duplicate article while reading the doco, should be removed


mhwelander:
Thank you!

PR #1473, opened 2021-03-23
Added a note about @prisma/client getting installed on first migration

### LINKED ISSUE #1436: Update the start from scratch guide to mention the client gets installed on generate
`generate` is called when the user runs `prisma migrate dev` this installs the client if its not already there. 

In the 'Start from scratch' guide we ask the user to run the migration then install the client. 

It might be helpful to the user to know that they will get the client when they run their first migration. 

We should keep the [install the client](https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch-typescript-postgres#install-and-generate-prisma-client) section, just mention the fact that it gets added on the initial migrate.

molebox:
Fixes: #1436

### DIFF
--- a/content/100-getting-started/02-setup-prisma/02-start-from-scratch.mdx
+++ b/content/100-getting-started/02-setup-prisma/02-start-from-scratch.mdx
@@ -143,10 +143,10 @@ datasource db {
 ```
 
 In this case, the `url` is [set via an environment variable](../../concepts/components/prisma-schema#using-environment-variables) which is defined in `.env`:
+
 ```js file=.env
 DATABASE_URL = 'postgresql://johndoe:randompassword@localhost:5432/mydb?schema=public'
 ```
-
 
 You now need to adjust the connection URL to point to your own database.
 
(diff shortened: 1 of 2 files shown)


content/100-getting-started/02-setup-prisma/02-start-from-scratch.mdx
```diff
@@ -278,6 +278,8 @@ This command does two things:
 1. It creates a new SQL migration file for this migration
 1. It runs the SQL migration file against the database
 
+> Note: `generate` is called when you run a migration, this will check if `@prisma/client` is installed and install it if it's missing.
+
```
Jolg42: Note: This only applies if you have the prisma-client-js generator in your schema.prisma
So for example, doesn’t apply if you have the go client
```suggestion
> Note: `generate` is called when you run a migration, if the `prisma-client-js` generator is defined in your schema this will check if `@prisma/client` is installed and install it if it's missing.

```

content/100-getting-started/02-setup-prisma/02-start-from-scratch.mdx
```diff
@@ -278,6 +278,8 @@ This command does two things:
 1. It creates a new SQL migration file for this migration
 1. It runs the SQL migration file against the database
 
+> Note: `generate` is called when you run a migration, if the `prisma-client-js` generator is defined in your schema this will check if `@prisma/client` is installed and install it if it's missing.
```
Jolg42:
```suggestion
> Note: `generate` is, by default, called after running `prisma migrate dev`, if the `prisma-client-js` generator is defined in your schema this will check if `@prisma/client` is installed and install it if it's missing.
```

Actually found `is called when you run a migration` confusing because generate is also called if no migration is applied.

nikolasburk:
```suggestion
> **Note**: `generate` is called under the hood when you run a migration. If the `prisma-client-js` generator is defined in your schema, this will check if `@prisma/client` is installed and install it if it's missing.
```

PR #1489, opened 2021-03-26
Connection management refactor

### LINKED ISSUE #667: We could give some numbers for connection pool size / concurrency limits for Serverless AWS Lambda or link to GitHub issue
To me my default thinking was that Lambda only have 1 CPU, looking on Google says that sometimes they have 2 but real testing below says it's always 2 so it could be nice to have a reference to that in the docs section for users:

| Memory (MB) | CPU (Cores) | Default Pool Size (`= 2 * CPU + 1`) |
| :---------: | :---------: | --------------------------- |
|     128     |      2      | 5                           |
|     256     |      2      | 5                           |
|     512     |      2      | 5                           |
|    1024     |      2      | 5                           |
|    2048     |      2      | 5                           |
|    3008     |      2      | 5                           |

So now I just changed the memory setting, but the reported cpu value is always the same, meaning we'll give 2 IO threads per lambda and 5 connections.

And to be clear. This is what we see from the OS layer. It might be wrong, it might be less or more. But the system sees 2 and will act based on that.

_Originally posted by @pimeys in https://github.com/prisma/prisma/issues/942#issuecomment-558153490_

Different popular PostgreSQL instances and their limits:

| Instance Size | used_at_start | max_conn | res_for_super | res_for_normal |
| :-----------: | :---------------------: | :---------------------: | :---------------------: | :--------------------: |
|   t2.micro    | 7 |            87          | 3                         | 77                       |
|   t2.small    | 7 |           198        | 3                         | 188 |
| t2.medium | 7 | 413 | 3 | 403 |
| t2.large | 7 | 856 | 3 | 846 |
| t2.xlarge | 7 | 1743 | 3 | 1733 |
| t2.2xlarge | 7 | 3518 | 3 | 3508 |
| m5.large  | 7 | 823 | 3 | 813 |
| m5.xlarge | 7 | 1691 | 3 | 1681 |
| m5.2xlarge | 7 | 3429 | 3 | 3419 |
| m5.4xlarge | 7 | 5000 | 3 | 4990 |

Some of the connections are used already when the instance is running, so I'd base my calculations for the connections reserved for normal users (the last column).

So with the default pool size of five connections:
Lambda Concurrency Limit based on Different popular PostgreSQL instances
| Instance Size | Lambda Concurrency Limit |
| :-----------: | :---------------------: |
| t2.micro | 15 |
| t2.small | 37 |
| t2.medium | 80 |
| t2.large | 169 |
| t2.xlarge | 346 |
| t2.2xlarge | 701 |
| m5.large | 162 |
| m5.xlarge | 336 |
| m5.2xlarge | 683 |
| m5.4xlarge | 998 |

_Originally posted by @pimeys in https://github.com/prisma/prisma/issues/942#issuecomment-558269427_

### LINKED ISSUE #1454: Connection management guide
Extract https://www.prisma.io/docs/guides/deployment/deployment#database-connection-limit to its own guide + expand with related info:

* #1366 
* #667

mhwelander:
Closes: #1454

- [ ] Create PR to redirect connection pool doc error - can't redirect anchors: https://github.com/prisma/prisma-engines/blob/f461292a2242db52d9f4c87995f0237aacd300d2/libs/user-facing-errors/src/query_engine.rs
- [ ] MANY redirects


mhwelander:
Now part of #1537

Issue #1373, opened 2021-03-02
Consider removing $disconnect() from examples as its only required in very specific cases


mhwelander:
## Context

Using `$disconnect()` in examples makes people think they need to disconnect in the context of a long-running app - they should not do that.

## Suggested action

Review all instances of `$disconnect()` and either remove or explain that the example is a script, not part of a long-running app, so we disconnect when the script is done.

mhwelander:
(Or at least clarify)

PR #1493, opened 2021-03-27
Add remark about not disconneting in serverless environments

Siilwyn:
Clarify keeping connection open in a serverless environment.

The connection management documentation suggest there are nuances that make it sensible
to reuse database connections, linking to this deployment documentation.

This is a followup from this Slack thread:
https://app.slack.com/client/T0MQBS8JG/CKQTGR6T0/thread/CKQTGR6T0-1614349810.348400

Additional context: I was accidentally calling `$disconnect` too, since it is used in multiple examples, then I stumbled on this issue:
https://github.com/mikecao/umami/issues/263


mhwelander:
Closing this PR as the connection management guide was in the process of being completely rewritten and is now live: https://www.prisma.io/docs/guides/performance-and-optimization/connection-management#do-not-explicitly-disconnect

Very good point about the examples using `$disconnect()`, have added a new issue here: https://github.com/prisma/docs/issues/1373

PR #1506, opened 2021-03-30
Added emacs note to editor setup

### LINKED ISSUE #1505: Add emacs suuport to editor setup docs
See https://github.com/prisma/prisma/issues/2312

molebox:
Fixes #1505

### DIFF
--- a/content/200-concepts/300-more/100-editor-setup.mdx
+++ b/content/200-concepts/300-more/100-editor-setup.mdx
@@ -2,6 +2,7 @@
 title: 'Editor setup'
 metaTitle: 'Editor and IDE setup (Reference)'
 metaDescription: 'Learn how to configure your editor and IDEs for an optimal developer experience with Prisma.'
+tocDepth: 2
 ---
 
 ## Overview
@@ -14,6 +15,8 @@ If you don't see your editor here, please [open a feature request](https://githu
 
 You can install the [Prisma VS Code extension](https://marketplace.visualstudio.com/items?itemName=Prisma.prisma).
 
-## Emacs
+## Community projects
+
+### Emacs
 
 For Emacs you can use [emacs-prisma-mode](https://github.com/pimeys/emacs-prisma-mode).
--- a/content/200-concepts/300-more/100-editor-setup.mdx
+++ b/content/200-concepts/300-more/100-editor-setup.mdx
@@ -17,6 +17,8 @@ You can install the [Prisma VS Code extension](https://marketplace.visualstudio.
 
 ## Community projects
 
+> **Note**: Community projects are not maintained or officially supported by Prisma and some features may by out of sync. Use at your own discretion.
+
 ### Emacs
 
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
This is great. 

But: This is not an official editor supported by us, so we should probably mention that this is a community thing that might work or not. This also means, the other editors do not have the same features. It should definitely not be listed on the same level as VSCode which we developer and spend a lot of effort on.

molebox:
> This is great.
> 
> But: This is not an official editor supported by us, so we should probably mention that this is a community thing that might work or not. This also means, the other editors do not have the same features. It should definitely not be listed on the same level as VSCode which we developer and spend a lot of effort on.

You know what, i was actually going to ask exactly this question. I can add a community supported section and list this there

content/200-concepts/300-more/100-editor-setup.mdx
```diff
@@ -14,6 +15,8 @@ If you don't see your editor here, please [open a feature request](https://githu
 
 You can install the [Prisma VS Code extension](https://marketplace.visualstudio.com/items?itemName=Prisma.prisma).
 
-## Emacs
+## Community projects
+
```
janpio, line 21: Maybe a disclaimer here that these are not offered by us, might not have the same or complete feature set and so on.

pimeys:
We btw. have language server working on Emacs, thanks to @pantharshit00. So, it seems Emacs is now BETTER than VSCode :)

Meaning, we're in feature-parity, but the underlying editor is better.

nikolasburk:
If we start adding community extensions here, we also might want to include the one for Jetbrains IDEs: https://plugins.jetbrains.com/plugin/14240-prisma

content/200-concepts/300-more/100-editor-setup.mdx
```diff
@@ -2,6 +2,7 @@
 title: 'Editor setup'
 metaTitle: 'Editor and IDE setup (Reference)'
 metaDescription: 'Learn how to configure your editor and IDEs for an optimal developer experience with Prisma.'
+tocDepth: 2
 ---
 
 ## Overview
```
mhwelander: @molebox Since you're working on this - could you replace Overview with `<TopBlock>` around everything before the first h2?

content/200-concepts/300-more/100-editor-setup.mdx
```diff
@@ -13,3 +14,19 @@ If you don't see your editor here, please [open a feature request](https://githu
 ## VS Code
 
 You can install the [Prisma VS Code extension](https://marketplace.visualstudio.com/items?itemName=Prisma.prisma).
+
+## Community projects
+
+> **Note**: Community projects are not maintained or officially supported by Prisma and some features may by out of sync. Use at your own discretion.
```
mhwelander, line 22: Nice 👍

mhwelander:
Looks good to me. :) Left 2 comments, otherwise SHIIIIPP IIIITTT

PR #1510, opened 2021-03-31
Document how to delete all records from all tables

### LINKED ISSUE #451: Document how to delete all records from all tables
Sometimes it's helpful to wipe all data records in the DB but keep the tables. This is currently not possible via the Prisma Client API so we should document a workaround.

One potential approach which currently **doesn't** work because it relies on a _private_ property of `PrismaClient` is this:

```ts

function lowerCase(name: string): string {
  return name.substring(0, 1).toLowerCase() + name.substring(1)
}

const promises = prisma.dmmf.datamodel.models.map(model => {
  return new Promise(async resolve => {
    const modelName = lowerCase(model.name)
    const modelMethods = prisma[modelName]

    if (modelMethods) {
      try {
        resolve(await modelMethods.findMany({}))
      } catch (e) {
        resolve(e)
      }
    }
  }

async function main() {                    
  await Promise.all(promises)
}

main()
```

Alternatively, the following approach (by [@nnennajohn](https://gist.github.com/nnennajohn)) which is based on knex.js can be used:

```ts
import camelCase from 'lodash/camelCase';
import { prisma } from '@cpath/universal/shared/db';
import knexClient from './knex';
// Contents of the knexClient import above
// import Knex from 'knex';
// const knexClient: Knex = Knex({
//   client: 'pg',
//   connection: {
//     user: process.env.POSTGRES_USER,
//     password: process.env.POSTGRES_PASSWORD,
//     host: process.env.POSTGRES_HOST,
//     port: Number(process.env.POSTGRES_PORT),
//     database: process.env.POSTGRES_DB,
//   },
// });
// Contents of the prisma import above
// export const prisma = new PrismaClient();
function deleteRecords(tablesList) {
  const validTables = tablesList.filter((tableName) => tableName[0] !== '_');
  console.log('These are the valid tables', validTables);
  const recordsDeletions = validTables.map((table) => {
    return prisma[camelCase(table)].deleteMany({});
  });
  return Promise.all(recordsDeletions);
}
function listTables() {
  const query =
    'SELECT table_name FROM information_schema.tables WHERE table_schema = current_schema() AND table_catalog = ?';
  const bindings = [knexClient.client.database()];
  return knexClient.raw(query, bindings).then((results) => {
    return results.rows.map((row) => row.table_name);
  });
}
async function main() {
  const tablesList = await listTables();
  console.log('This is the list of knex tables', tablesList);
  if (tablesList && tablesList.length) {
    deleteRecords(tablesList)
      .then((res) => {
        console.log('Successfully deleted table records', res);
      })
      .catch((err) => {
        console.log('Error deleting table records', err);
      });
  }
}
main()
  .catch((e) => console.error(e))
  .finally();
```

We should also document a workaround that allows to do this without an external dependency like knex.js.

molebox:
This PR documents how to delete all records from all tables using `postgres`, `sqlite` and `mysql`. It gives an example for doing this with each database. The info is added to the [CRUD](https://www.prisma.io/docs/concepts/components/prisma-client/crud) page. Though this page might get split up into sections by operation later.

Fixes #451


content/200-concepts/100-components/02-prisma-client/030-crud.mdx
```diff
@@ -777,6 +777,70 @@ To resolve this error, you can:
   const transaction = await prisma.$transaction([deletePosts, deleteUser])
   ```
 
+### Delete all records from all tables
+
+Sometimes you want to remove all data from all tables but keep the actual tables. This can be particularly useful in a development environment.
+
+The following examples show how to remove all data (including relations) from all tables in your database, with `PostgreSQL`, `SQLite` and `MySQL`. The `SQLite` and `MySQL` examples use a `transaction` to batch the queries together so that they are executed on the same connection.
```
nikolasburk: - the DB names are proper nouns and hence shouldn't be written as inline code (ideally, inline code is used only  when referring to things that "appear like this in code", e.g. variable names, class names, model names, function names, strings, ...)
- let's link the mention of `$transaction` to where it's explained in the docs 

```suggestion
The following examples show how to remove all data (including relations) from all tables in your database, with PostgreSQL, SQLite and MySQL. The SQLite and MySQL examples use a [`$transaction`](TODO) to batch the queries together so that they are executed on the same connection.
```

content/200-concepts/100-components/02-prisma-client/030-crud.mdx
```diff
@@ -777,6 +777,70 @@ To resolve this error, you can:
...
+  prisma.$executeRaw`TRUNCATE TABLE players;`,
+  prisma.$executeRaw`SET FOREIGN_KEY_CHECKS = 1;`,
+])
+```
+
+</tab>
+ 
+</TabbedContent>
+
+#### <inlinecode>Prisma Migrate</inlinecode>
```
nikolasburk:
```suggestion
#### Deleting all records with Prisma Migrate
```

content/200-concepts/100-components/02-prisma-client/030-crud.mdx
```diff
@@ -777,6 +777,70 @@ To resolve this error, you can:
...
+
+If you use Prisma Migrate, you can use `migrate reset`, this will:
+
+- Drop the database
+- Create a new database
+- Apply migrations
+- Seed the database with data
```
nikolasburk: iiuc, this is a sequence of steps, so I think it's a good opportunity to use a numbered instead of a bullet list:
```suggestion
1. Drop the database
1. Create a new database
1. Apply migrations
1. Seed the database with data
```

nikolasburk: Here's a quick overview when to use numbered and when to use unordered bullet lists:https://www.prisma.io/docs/about/style-guide#use-tables-bullet-lists-and-numbered-lists

image

content/200-concepts/100-components/02-prisma-client/030-crud.mdx
```diff
@@ -777,6 +777,70 @@ To resolve this error, you can:
   const transaction = await prisma.$transaction([deletePosts, deleteUser])
   ```
 
+### Delete all records from all tables
+
+Sometimes you want to remove all data from all tables but keep the actual tables. This can be particularly useful in a development environment.
+
```
nikolasburk:
```suggestion
Sometimes you want to remove all data from all tables but keep the actual tables. This can be particularly useful in a development environment. In the following two sections, you can learn how to delete all records from all tables with Prisma Client and with Prisma Migrate.

#### Deleting all records with Prisma Client

```

content/200-concepts/100-components/02-prisma-client/030-crud.mdx
molebox: Argh, sorry, yeah. Ive got to remember this

content/200-concepts/100-components/02-prisma-client/030-crud.mdx
```diff
@@ -777,6 +777,70 @@ To resolve this error, you can:
...
+  prisma.$executeRaw`TRUNCATE TABLE players;`,
+  prisma.$executeRaw`SET FOREIGN_KEY_CHECKS = 1;`,
+])
+```
+
+</tab>
+
+<tab>
+
+```ts
```
mhwelander, line 828: Try and see what this outputs:

```ts
const [one, two, three, four] = await prisma.$transaction([
  prisma.$executeRaw`SET FOREIGN_KEY_CHECKS = 0;`,
  prisma.$executeRaw`TRUNCATE TABLE teams;`,
  prisma.$executeRaw`TRUNCATE TABLE players;`,
  prisma.$executeRaw`SET FOREIGN_KEY_CHECKS = 1;`,
])
```

mhwelander:
One other option that is already documented - use `deleteMany` in a particular order, run all queries in a transaction: https://www.prisma.io/docs/guides/prisma-guides/prisma-client-transactions-guide#scenario-privacy-legislation

PR #1512, opened 2021-03-31
Added troubleshooting section to relations page

### LINKED ISSUE #797: Add section to the Prisma docs that explains confusion about "missing relation fields"
I've explained the problem already in this [issue](https://github.com/prisma/prisma/issues/3434) and hope that we can tackle it on a product-level soon. But for the time being, I think we should have a section in the docs (on the Relations page I guess) that explains the issue behind the confusion as well as how to resolve it.

molebox:
I have added a new troubleshooting section to the bottom of the Relations page. Here i have just gone over the two confusions the users had that were linked in this issue. 

Longer term i think it would be a great idea to split this page up. Its very dense with information and those devs that come looking for a quick answer may get discouraged and intimedated when searching the page for an answer.

That being said, i can see what ive added here potentially being moved to a sub page called Troubleshooting or FAQ or whatever in the furture. For now, this should cover people that have the same issues.

Fixes #797


content/200-concepts/100-components/01-prisma-schema/06-relations.mdx
```diff
@@ -1088,6 +1089,108 @@ model Post {
...
+  category      Category
+  @@id([post, category])
+}
+
+model Category {
+  id             Int        @id @default(autoincrement())
+  name           String
+  posts          Post[]
+}
+```
```
nikolasburk: Maybe add some comments to this schema that this is incorrect so that people don't misunderstand it when only glancing over it:

```suggestion
```prisma
// This schema shows how NOT to define an explicit m-n relation

model Post {
  id            Int         @id @default(autoincrement())
  title         String
  categories    Category[] // this should refer to PostCategories
}

model PostCategories {
  postId        User        @relation(fields: [post], references: [id])
  post          Post
  categoryId    Facility    @relation(fields: [category], references: [id])
  category      Category
  @@id([post, category])
}

model Category {
  id             Int        @id @default(autoincrement())
  name           String
  posts          Post[] // this should refer to PostCategories
}
```
```

File: content/200-concepts/100-components/01-prisma-schema/06-relations.mdx [hunk 2]
```diff
@@ -1088,6 +1089,108 @@ model Post {
...
+}
+
+model Category {
+  id             Int      @id @default(autoincrement())
+  name           String
+  posts          Post[]   @relation("Post")
+  posts          Post[]
+}
+```
+
```
nikolasburk, line 1196: You could show a second alternative here that uses the same relation name:

```suggestion
In case you do want to control the name of the underlying relation table, you can do this as follows:

```prisma
model Post {
  id            Int         @id @default(autoincrement())
  title         String
  categories    Category[]  @relation("CategoryPostRelation")
}
model Category {
  id             Int      @id @default(autoincrement())
  name           String
  posts          Post[]   @relation("CategoryPostRelation")
}
```
```

nikolasburk (commented):
This looks great! I've left two minor suggestions (feel free to merge or ignore), but then it should be ready to merge from my end 👍

content/200-concepts/100-components/01-prisma-schema/06-relations.mdx
molebox: Yep! This is what i meant in slack! 😁

PR #1516, opened 2021-04-01
Use consistent path separators

breyed:
Most of the to path separators are slashes, but a few were backslashes. They are now slashes to match the others.

In its former form, the page was a bit confusing to a new reader (me). When I got to the first backslash form, a reference to prisma\.env, since I wasn't expecting a backslash path separator, I thought at first it was some sort of escape. Consistency avoids this kind of confusion.


molebox:
Thanks for the PR @breyed, great catch! 💯

janpio:
This is what happens when part of the team uses Windows @mhwelander :D

PR #1517, opened 2021-04-01
added note about optional fields defaulting to null

### LINKED ISSUE #1490: Bug report: state that the default value of an optional field is null
In this [section](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields) it's not stated that the default value of an optional field is `null`.  This can lead to confusion 

image

molebox:
Fixes #1490


nikolasburk (approved):
Looks great 👌

PR #1521, opened 2021-04-01
Docs: Fix reflink pointing to non existing page.

Torrino:
Currently the **validator** word takes to a non existing page on the docs while it should take to the dedicated page for prisma Validators. Changed the reflink to take to the right location.


mhwelander:

PR #1532, opened 2021-04-06
added typescript note, changed heading

molebox:
Ive made some small amendments to the advanced type safety page after feedback from @2color


2color (approved):
Looks good to me 👌

PR #1537, opened 2021-04-07
Refactor: Deployment / Connection management

### LINKED ISSUE #927: AWS Lambda Deployment question
Im following along in example guide for the AWS lamba guide. But I'm not 100% sure when you need to run `prisma generate`

I have a schema file and a database ready. 

If I understand the guide will the npm packages be included with the serverless deploy command? 
My understanding was that they weren't included

Im just confused how my 'generated client' will be deployed to lambdas

### LINKED ISSUE #1311: Doc request: Update recommended workaround with NextJS & Prisma in development
Our docs have this section describing how to work around the `There are already 10 PrismaClients running` error during development with NextJS: https://www.prisma.io/docs/support/help-articles/nextjs-prisma-client-dev-practices

In practice, there were a few more changes required to get this working correctly to make `tsc` happy:

```
declare global {
  var globalPrisma: PrismaClient
}

let prisma: PrismaClient
if (process.env.NODE_ENV === 'production') {
  prisma = new PrismaClient()
} else {
  if (!global.globalPrisma) {
    global.globalPrisma = new PrismaClient()
  }
  prisma = global.globalPrisma
}
export { prisma }
```

Here are the two changes:

1. We have to add a `declare global` block
2. We need to call this global `prisma` instance something else, otherwise `tsc` thinks we're redeclaring `prisma`. This can also work the other way around, we can rename the real prisma variable `localPrisma` or something and rename it while exporting.

### LINKED ISSUE #1366: Doc request: Improve `connection_limit` / timeout issues
https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-management#connection-pool

Context: (https://prisma-company.slack.com/archives/CBJCQF95M/p1614612869000300)
One issue some users might have in production is that they can get timeout errors for Prisma client.
The pool_timeout is 5s
5s for a query is forever, but the timeout comes from the pool. if you have 4 connections to the database, and you send 1000 queries to prisma without awaiting them (in a loop for example), the queries will go to a channel in the pool if the pool has no ready connections. then checking out a connection from the pool is raced with a timer, because if you don't do this and just push more and more queries too fast to the channel, you'll eat up all the RAM
so now the queries might take 500ms, but if you're pushing lots of queries constantly, the amount of connections is not enough to serve all requests in a timely manner
(queue is FIFO)

Recommended solution:
Increase the `connection_limit` so the engine can open more connections to the database and execute the query queue faster without getting a timeout. (need to check that your database can handle more connections) 

Alternative:
You can set `pool_timeout=0`, but then you will have Out Of Memory issues when you have too many queries.

Note: 
>If the connection_limit argument is omitted, the default number of connections is calculated according to this formula: num_physical_cpus * 2 + 1 where num_physical_cpus represents the number of physical CPUs on your machine.

What we could do?
- Mention the problem / solution
- We set a default for connection_limit and it's safe to customize it to have more connections. (I think it's important to say it's safe).

(@pimeys knows how it works 😉 )

### LINKED ISSUE #1413: Doc request: Add notes about the upload limit for Lambda
Right now we ship multiple query engine binaries: https://prisma-company.slack.com/archives/CRGDU19K6/p1615560739061000?thread_ts=1615496091.057100&cid=CRGDU19K6

This is a problem in Lambda because there's a total upload limit for your function. This limit is [about 50mb](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html). This includes all your application code and the binaries.

There are alternatives like upload to S3 first and link the zip, but the first thing we should do is ask them to exclude the unnecessary binaries in @prisma/engines if they run into this.

https://www.prisma.io/docs/guides/deployment/deploying-to-aws-lambda

### LINKED ISSUE #1414: Deployment recommendations / from local

### LINKED ISSUE #1454: Connection management guide

### LINKED ISSUE #1468: Deployment overview refactor / refine
We're moving connection management into its own guide, which will mean the deployment overview guide needs an overhaul.

Related issues to include: 
#1414
#1413 
#927

### LINKED ISSUE #1471: "Deployment platforms" page
* List deployment platforms that we test (see e2e tests in related thread - no need to specify that we test them)  and all 'combos' - e.g. vercel + redwood, vercel + nextjs https://github.com/prisma/e2e-tests/tree/dev/platforms
* Check with team if there are any other platforms we should list (maybe ones we don't support & why)
* Move AWS beanstalk section in 'Deployment overview' to this page
* Document #841 under AWS Lambda section (AWS Lambda + WebPack combination)
* If we have a deployment guide, link to it

Related context: https://prisma-company.slack.com/archives/CEYCG2MCN/p1616411590042800

molebox:
This PR refactors the the deployment docs.

- All guide under there own section
- Guide section lists where Prisma is deployed to
- Local deployment covered but warned against
- Deploy changes with migrate documented but needs expanding at a later date
- Caveats to deploying to AWS platforms page that covers some use cases raised by uses
- Deploy to a different OS has its own page which indicates how to use a new binary if targeting a different OS. This should be expanded into its own guide at some point too.

Fixes: #927 
Fixes: #1471 
Fixes: #1414 
Fixes: #1468


content/300-guides/200-deployment/120-faq.mdx
```diff
@@ -0,0 +1,40 @@
...
+This page looks to answer common questions raised via GutHub issues, social media, Slack or any other source related to deploying Prisma.
+
+</TopBlock>
+
+## How to Deploy database changes with Prisma Migrate
+
+To apply pending migrations to development, staging, or testing environments, run the `migrate deploy` command as part of your CI/CD pipeline:
+
+```terminal
+prisma migrate deploy
```
janpio: probably `npx prisma ...` or `yarn prisma ...`

mhwelander:
## TODO

- [x] Redirects in homepage-v8

molebox:
This looks great! 💃🏼

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+metaDescription: 'How to approach connection management with Prisma Client in serverless and PaaS applications.'
+tocDepth: 2
+---
+
+<TopBlock>
+
+Databases can handle a limited number of concurrent connections. Each connection requires RAM, which means that simply increasing the connection limit without scaling the database:
+
+* ✔ might allow more processes to connect _but_
+* ✘ significantly affects database **performance**
```
janpio: Are we sure that is the only thing that can happen? Could it not also just crash the database and similar?

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+The way your application **manages connections** also impacts performance. This guide describes how to approach connection management in [serverless environments](#serverless-environments-faas) and [long-running processes](#long-running-processes-paas).
+
+</TopBlock>
+
+## Long-running processes (PaaS)
```
janpio: This is weird to me, as you can have long-running processes on other environments as well: local, normal servers, VPS, insider containers etc.

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+Examples of long-running processes include Node.js applications hosted on a PaaS service like Heroku. Use the following checklist as a guide to connection management in PaaS environments:
+
+- ☐ Tune the [connection limit](#recommended-connection-limit)
```
janpio, line 23:
```suggestion
- ☐ Tune the [connection limit](#recommended-connection-limit) of your database
```

janpio: Is this about the connection limit of your database (max connection) or Prisma connection pool (connection_limit)?

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+- ☐ Make sure you have [**one** global instance of `PrismaClient`](#prismaclient-in-long-running-applications)
+
+### Recommended connection limit
+
+The recommended connection limit for long-running processes is the **default connection limit** (`num_physical_cpus * 2 + 1`) ÷ **number of application instances**.
```
janpio: The default connection limit or default connection limit, where does this comes from as the "recommended limit"? It is quite a random value that we default to in the Engines.

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+```ts file=client.ts
+import { PrismaClient } from "@prisma/client"
+
+let prisma = new PrismaClient() 
+
+export default prisma
+```
+
+The object is [cached](https://nodejs.org/api/modules.html#modules_caching) the first time the module is imported. Subsequent requests return the cached object rather than creating a new `PrismaClient`:
```
janpio, line 59: Is this really required? In an express app the import would only happen once anyway so this makes it to complicated. Mayybe mention that option as well?

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+* You can optionally [tune the pool size](#optimizing-the-connection-pool).
+
+### <inlinecode>PrismaClient</inlinecode> in long-running applications
+
+In **long-running** applications, we recommend that you:
+
+* ✔ Create **one** instance of `PrismaClient` and re-use it across your application
+* ✔ Assign `PrismaClient` to a global variable _in dev environments only_ to [prevent hot reloading from creating new instances](#prevent-hot-reloading-from-creating-new-instances-of-prismaclient)
+
+To re-use a single instance, create a module that exports a `PrismaClient` object:
```
janpio, line 49: This could use a headline in the line before so it is linkable

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+async function main() {
+    const allUsers = await prisma.user.findMany();
+}
+
+main();
+```
+
+## Serverless environments (FaaS)
+
+Examples of serverless environments include Node.js functions hosted on AWS Lambda. Use the following checklist as a guide to connection management in serverless environments:
```
janpio:
```suggestion
Examples of serverless environments include Node.js functions hosted on AWS Lambda, Vercel or Netlify Functions. Use the following checklist as a guide to connection management in serverless environments:
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+- ☐ [Set connection limit](#recommended-connection-limit-1) based on whether you have an external connection pooler
+- ☐ Optionally [tune the pool size](#optimizing-the-connection-pool) if you are using an external connection pooler
+- ☐ [Instantiate `PrismaClient` outside the handler](#instantiate-prismaclient-outside-the-handler) and do not explicitly `$disconnect()`
+- ☐ Configure [function concurrency](#concurrency-limits) and handle [idle connections](#zombie-connections)
+
+### The serverless challenge
+
+In a serverless environment, each function creates **its own instance** of `PrismaClient`, and each client instance has its own connection pool. 
+
+Consider the following example, where a single AWS Lambda function uses `PrismClient` to connect to a database. The `connection_limit` is **3**:
```
janpio:
```suggestion
Consider the following example, where a single AWS Lambda function uses `PrismaClient` to connect to a database. The `connection_limit` is **3**:
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+![An AWS Lambda function connecting to a database.](./serverless-connections.png)
+
+A traffic spike causes AWS Lambda to spawn two additional lambdas to handle the increased load. Each lambda creates an instance of `PrismaClient`, each with a `connection_limit` of **3**, which results in a maximum of **9** connections to the database:
+
+![Three AWS Lambda function connecting to a database.](./serverless-connections-2.png)
```
janpio, line 133: Soooo nice visual!

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+ 
+200 _concurrent functions_ (and therefore 600 possible connections) responding to a traffic spike 📈 can exhaust the database connection limit very quickly.
```
janpio: Maybe mention that functions that are paused, make this worse as they keep teh connection open by default and block them.
```suggestion
200 _concurrent functions_ (and therefore 600 possible connections) responding to a traffic spike 📈 can exhaust the database connection limit very quickly. Furthermore, any functions that are **paused** keep their connections open by default and block them from being used by another function.
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+1. Start by [setting the `connection_limit` is set to `1`](#recommended-connection-limit-1)
```
janpio:
```suggestion
1. Start by [setting the `connection_limit` to `1`](#recommended-connection-limit-1)
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+2. If a lower connection limit is not enough, consider using an [external connection pooler like PgBouncer](#external-connection-poolers)
```
janpio: This listing is surprising and does not match the content below.

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+### Recommended connection limit
+
+The recommended connection limit in serverless environments depends on whether you are using an [external connection pooler](#external-connection-poolers).
+
+#### Without an external connection pooler
+
+We recommend that you set the `connection_limit` to **1** if you are **not** using an external connection pooler. Each incoming request starts a short-lived Node.js process, and many concurrent functions with a high `connection_limit` can quickly **exhaust the _database_ connection limit** during a traffic spike.
```
janpio: I think this is to broad: 
If you function actually sends 2 queries in parallel, in code, then it will have to fall back to doing that serially instead with a connection_limit of 1. That can be worse thatn using more connections, especially if _all_ calls of the functions would do that.

We at least should point that out.

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+The following example demonstrates how to set the `connection_limit` to 1 in your connection string:
+
+<TabbedContent tabs={[<FileWithIcon text="PostgreSQL" icon="database"/>, <FileWithIcon text="MySQL" icon="database"/>]}>
+<tab>
+
+  ```
+  postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
```
janpio: missing the schema
```suggestion
  postgresql://USER:PASSWORD@HOST:PORT/DATABASE?schema=public&connection_limit=1
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+  ```
+  mysql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=1
+  ```
+
+</tab>
+</TabbedContent>
+
+#### With an external connection pooler
+
+If you are using an external connection pooler, use the default connection limit (`num_physical_cpus * 2 + 1`) as a starting point and then [tune the pool size](#optimizing-the-connection-pool). The external connection pooler prevents a traffic spike from overwhelming the database.
```
janpio:
```suggestion
If you are using an external connection pooler, use the default connection limit (`num_physical_cpus * 2 + 1`) as a starting point and then [tune the pool size](#optimizing-the-connection-pool). The external connection pooler should prevent a traffic spike from overwhelming the database.
```

janpio:
```suggestion
If you are using an external connection pooler, use the default connection limit (`num_physical_cpus * 2 + 1`) as a starting point and then [tune the pool size](#optimizing-the-connection-pool). The external connection pooler should help preventing a traffic spike from overwhelming the database.
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+#### Container reuse
+
+There is no guarnatee that subsequent nearby invocations of a function will hit the same container - for example, AWS can choose to create a new container at any time.
+
+Code should assume the container to be stateless and create a connection only if it does not exist - Prisma Client JS already implements this logic.
+
+#### Zombie connections
+
+Containers that are marked to be removed and are not being reused still keep a connection open and can stay in that state for some time (unknown and not documented from AWS). This can lead to sub-optimal utilization of the DB connections.
```
janpio:
```suggestion
Containers that are marked to be removed and are not being reused still keep a connection open and can stay in that state for some time (unknown and not documented from AWS). This can lead to sub-optimal utilization of the database connections.
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+  1. Note that these are recommendations and not best practices. These would vary from system to
+  system.
+</sup>
+<br />
+<sup>
+  2.{' '}
+  <a href="https://github.com/jeremydaly/serverless-mysql">
+    <inlineCode>serverless-mysql</inlineCode>
+  </a>{' '}
+  is a library that implements this idea.
```
janpio: It only implements the second idea.

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+Code should assume the container to be stateless and create a connection only if it does not exist - Prisma Client JS already implements this logic.
+
+#### Zombie connections
+
+Containers that are marked to be removed and are not being reused still keep a connection open and can stay in that state for some time (unknown and not documented from AWS). This can lead to sub-optimal utilization of the DB connections.
+
+Potential solutions<sup>1, 2</sup> include:
+
+* Using a lower idle connection timeout.
```
janpio: This will not work in a serverless environment - the function is paused and you never execute any code that could close the idle connection. Only next time your serverless function wakes up to be used, you would realize "oh, this connection is super mega idle >1hr and should have been closed after 5 min - let's do that now". Now you have no more connection and need to open a new one :/

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+  2.{' '}
+  <a href="https://github.com/jeremydaly/serverless-mysql">
+    <inlineCode>serverless-mysql</inlineCode>
+  </a>{' '}
+  is a library that implements this idea.
+</sup>
+
+#### Concurrency limits
+
+Depending on your serverless concurrency limit (the number of serverless functions running in parallel), you might still exhaust your database's connection limit. This can happen when too many functions are invoked concurrently, each with its own connection pool, which eventually exhausts the database connection limit. To prevent this, you can [set your serverless concurrency limit](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) to a number lower than the connection limit of your database (as you might want to be able to connect from another client for other purposes).
```
janpio:
```suggestion
Depending on your serverless concurrency limit (the number of serverless functions running in parallel), you might still exhaust your database's connection limit. This can happen when too many functions are invoked concurrently, each with its own connection pool, which eventually exhausts the database connection limit. To prevent this, you can [set your serverless concurrency limit](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) to a number lower than the maximum connection limit of your database divided by the connection used by each function invocation (as you might want to be able to connect from another client for other purposes).
```
bad wording here, but not everyone should be using pool size 1

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+### Increasing the connection limit
+
+Increasing the connection limit allows the query engine to process a larger number of queries in parallel. Be aware that your database must be able to support the increased number of concurrent connections, otherwise you will **exhaust the connection limit**.
+
+To increase the connection limit, manually set the `connection_limit` to a higher number:
+
+```prisma
+datasource db {
+  provider = "postgresql"
+  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=40"
```
janpio: postgres should have schema
```suggestion
  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?schema=public&connection_limit=40"
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+```
+
+### Increasing the pool timeout
+
+Increasing the pool timeout gives the query engine more time to process queries in the queue. You might consider this approach in the following scenario:
+
+* You have already increased the `connection_limit`.
+* You are confident that the queue will not grow beyond a certain size, otherwise **you will eventually run out of RAM**.
+
+To incrase the pool timeout, set the `pool_timeout` parameter to a value larger than the default (10 seconds):
```
janpio:
```suggestion
To increase the pool timeout, set the `pool_timeout` parameter to a value larger than the default (10 seconds):
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+```prisma
+datasource db {
+  provider = "postgresql"
+  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5&pool_timeout=20"
+}
+```
+
+### Disabling the pool timeout
+
+Disabling the pool timeout prevents the query engine from throwing an exception and allows the queue to build up. You might consider this approach in the following scenario:
```
janpio:
```suggestion
Disabling the pool timeout prevents the query engine from throwing an exception after x seconds of waiting for a connection and allows the queue to build up. You might consider this approach in the following scenario:
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+```prisma
+datasource db {
+  provider = "postgresql"
+  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5&pool_timeout=0"
+}
+```
+
+## External connection poolers
+
+Connection poolers like [PgBouncer](https://www.pgbouncer.org/) prevents your application from exhausting the database's connection limit.
```
janpio:
```suggestion
Connection poolers like [PgBouncer](https://www.pgbouncer.org/) prevent your application from exhausting the database's connection limit.
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+### PgBouncer
+
+PostgreSQL only supports a certain amount of concurrent connections, and this limit can be reached quite fast when the service usage goes up – especially in [serverless environments](#serverless-environments-faas).
+
+PgBouncer holds a connection pool to the database and proxies incoming client connections by sitting between Prisma Client and the database. This reduces the number of processes a database has to handle at any given time. PgBouncer passes on a limited number of connections to the database and queues additional connections for delivery when space becomes available.
```
janpio:
```suggestion
PgBouncer holds a connection pool to the database and proxies incoming client connections by sitting between Prisma Client and the database. This reduces the number of processes a database has to handle at any given time. PgBouncer passes on a limited number of connections to the database and queues additional connections for delivery when connections becomes available.
```

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The connection pool has a **connection limit** and a **pool timeout**, which are controlled by connection string parameters.
+
+</TopBlock>
+
+## How the connection pool works
+
+The following steps describe how the query engine uses the connection pool:
+
+1. The query engine instantiates a connection pool with a [congfigurable pool size](#setting-the-connection-limit) and [pool timeout](#setting-the-connection-pool-timeout).
```
janpio:
```suggestion
1. The query engine instantiates a connection pool with a [configurable pool size](#setting-the-connection-limit) and [pool timeout](#setting-the-connection-pool-timeout).
```

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -0,0 +1,84 @@
...
+1. The query engine reserves connections from the pool and processes incoming queries.
```
janpio: This might be missing that it also opens additional connection to the database until the pool size is reached and the pool is "full".
```suggestion
1. The query engine opens the number of database connections defined by `connection_limit` and adds them to the connection pool.
1. The query engine reserves connections from the pool and processes incoming queries.
```

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -0,0 +1,84 @@
...
+1. If the query engine cannot reserve a connection from the pool, queries are added to a FIFO (First In First Out) queue in memory. FIFO means that queries are processed in the order they enter the queue.
+1. If the query engine cannot process a query in the queue for **before the [time limit](#default-pool-timeout)**, it throws an exception with error code `P2024`for that query and moves on to the next one in the queue.
```
janpio:
```suggestion
1. If the query engine cannot process a query in the queue for **before the [time limit](#default-pool-timeout)**, it throws an exception with error code `P2024` for that query and moves on to the next one in the queue.
```

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -0,0 +1,84 @@
...
+## Setting the connection limit
+
+You can specify the number of connections by explicitly setting the `connection_limit` parameter in your database connection URL. For example, with the following `datasource` configuration in your [Prisma schema](../../prisma-schema) the connection pool will have exactly five connections:
+
+```prisma
+datasource db {
+  provider = "postgresql"
+  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5"
+}
+```
```
janpio: This adds another headline:

````suggestion
## Connection limit

### Default connection limit

The default number of connections (pool size) is calculated with the following formula:

```bash
num_physical_cpus * 2 + 1
```

`num_physical_cpus` represents the number of physical CPUs on your machine. If your machine has **four** physical CPUs, your connection pool will contain **nine** connections (`4 * 2 + 1 = 9`).

Although the formula represents a good starting point, the [recommended connection limit](../../../../guides/performance-and-optimization/connection-management#recommended-connection-limit) <span class="guide"></span> also depends on your deployment paradigm - particularly if you are using serverless.

### Setting the connection limit

You can specify the number of connections by explicitly setting the `connection_limit` parameter in your database connection URL. For example, with the following `datasource` configuration in your [Prisma schema](../../prisma-schema) the connection pool will have exactly five connections:

```prisma
datasource db {
  provider = "postgresql"
  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5"
}
```
````

File: content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx [hunk 30]
```diff
@@ -0,0 +1,84 @@
...
+You disable the connection pool timeout by setting the `pool_timeout` parameter to `0`:
+
+```prisma
+datasource db {
+  provider = "postgresql"
+  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5&pool_timeout=2"
+}
+```
+
+You can choose to [disable the connection pool timeout if queries **must** remain in the queue](../../../../guides/performance-and-optimization/connection-management#disabling-the-pool-timeout) <span class="guide"></span> - for example, if you are importing a large number of records in parallel and are confident that the queue will not use up all available RAM before the job is complete.
```
janpio: Add another headline:
````suggestion
## Connection pool timeout

### Default pool timeout

The default connection pool timeout is 10 seconds. If the query engine does not process a particular query within that time, it throws an exception and moves on to the next query in the queue.

### Setting the connection pool timeout

You can specify the pool timeout explicitly setting the `pool_timeout` parameter in your database connection URL. In the following example, the pool times out after `2` seconds:

```prisma
datasource db {
  provider = "postgresql"
  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5&pool_timeout=2"
}
```

### Disabling the connection pool timeout

You disable the connection pool timeout by setting the `pool_timeout` parameter to `0`:

```prisma
datasource db {
  provider = "postgresql"
  url      = "postgresql://johndoe:mypassword@localhost:5432/mydb?connection_limit=5&pool_timeout=2"
}
```

You can choose to [disable the connection pool timeout if queries **must** remain in the queue](../../../../guides/performance-and-optimization/connection-management#disabling-the-pool-timeout) <span class="guide"></span> - for example, if you are importing a large number of records in parallel and are confident that the queue will not use up all available RAM before the job is complete.
```
````

File: content/200-concepts/200-database-connectors/03-postgresql.mdx [hunk 31]
```diff
@@ -65,9 +65,9 @@ The following arguments can be used
...
-| `pool_timeout`     | No       | `5`                    | Maximum number of seconds to wait for a new connection from the pool. If all connections are in use, the database will return a `PoolTimeout` error after waiting for the given time. |
+| `pool_timeout`     | No       | `10`                    | Maximum number of seconds to wait for a new connection from the pool. If all connections are in use, the database will return a `PoolTimeout` error after waiting for the given time. |
 | `socket_timeout`   | No       | `5`                    | Maximum number of seconds to wait until a single query terminates                                                                                               |
 | `sslmode`          | No       | `prefer`               | Configures whether to use TLS, possible values: `prefer`, `disable`, `require`                                                                                  |
 | `sslcert`          | No       |                        | Path the the server certificate. Certificate paths are [resolved relative to the `./prisma folder`](../components/prisma-schema/data-sources#ssl-certificates). |
```
janpio:
```suggestion
| `sslcert`          | No       |                        | Path of the server certificate. Certificate paths are [resolved relative to the `./prisma folder`](../components/prisma-schema/data-sources#ssl-certificates). |
```

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/015-instantiate-prisma-client.mdx
```diff
@@ -25,100 +25,30 @@ const prisma = new PrismaClient()
...
-
-main();
-```
-
-### Prevent hot reloading from creating new instances of `PrismaClient`
-
-Frameworks like [Next.js](https://nextjs.org/) support hot reloading of changed files, which enables you to see changes to your application without restarting. However, if the framework refreshes the module responsible for exporting `PrismaClient`, this can result in **additional, unwanted instances of `PrismaClient` in a development environment**.
-
-As a workaround, you can store `PrismaClient` as a global variable in development environments only, as global variables are not reloaded:
+## The the number of `PrismaClient` instances matters
```
janpio:
```suggestion
## The number of `PrismaClient` instances matters
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
mhwelander: The OS will shut down a database with an 'out of memory' exception, I'll add that

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
mhwelander: Changed to 'connection pool size'.

As far as I know, cpus * 2 + 1 is based on `((core_count * 2) + effective_spindle_count)`, which is just a generally accepted formula. I'll change the wording to 'recommended starting point' to indicate that it's not a catch-all

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
mhwelander: @janpio Like so?

content/300-guides/200-deployment/110-deployment-guides/600-deploying-migrations-from-a-local-environment.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: 'Deploying migrations from a local environment'
+metaTitle: 'Deploying migrations from a local environment'
+metaDescription: 'Learn how to deploy Node.js and TypeScript applications that are using Prisma Client locally.'
+tocDepth: 2
+---
+
+<TopBlock>
+
+There are instances whereby you might want to run a deployment migration locally. This page outlines some examples of how you can do that and **why its generally a bad practice**.
```
mhwelander: * Would say 'not recommended practice' instead of bad
* Can we expand on 'might want to run a deployment migration locally' - maybe something like 'There are **two** scenarios where you might consider deploying migrations directly from a local environment to a production environment' and then list the two use cases: you only have a local CI / CD pipeline + you are baselining a production environment

content/300-guides/200-deployment/110-deployment-guides/600-deploying-migrations-from-a-local-environment.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+> **Note**: `prisma migrate deploy` should generally be part of an automated CI/CD pipeline, and we do not recommend running this command locally to deploy changes to a production database. It is not generally considered good practice to store the production database URL locally.
+
+### An alternative approach
+
+An alternative approach would be to let an automated CI/CD pipeline handle your staging and production environments.
+
+If your goal is to prototype your schema and still have 3 environments that mock a real world use case then you might consider using the [`db push`](../../../../concepts/components/prisma-migrate/db-push) command.
+
+## Add migrations to a pre-existing database locally.
```
mhwelander: This section looks like a shorter version of https://deploy-preview-1537--prisma2-docs.netlify.app/docs/guides/database/developing-with-prisma-migrate/add-prisma-migrate-to-a-project

As far as I understand, it's specifically baselining production that one might perform from local --> production. Is that right, @2color? Is there any other way to do it - a one-off script that you run from your CI/CD setup somehow? 🤷 

I would suggest keeping this section part very focused on this: you can do the baselining tasks from local.

* Change title to "Baselining a production database"
* Very short intro - 'When you add Prisma Migrate to an **existing database**, you must baseline the production database (add link). Baselining is performed **once**, and can be done from a local instance: [possible diagram here of laptop -> list of commands run as part of baselining -> production database]

molebox: @mhwelander Are you suggesting removing all the stuff i had written and keeping it short and with a diagram to illustrate?

mhwelander: @molebox That's right, I would really condense it down to only the point about deploying from local

content/300-guides/200-deployment/110-deployment-guides/600-deploying-migrations-from-a-local-environment.mdx
```diff
@@ -0,0 +1,51 @@
...
+tocDepth: 2
+---
+
+<TopBlock>
+
+There are instances whereby you might want to run a deployment migration locally. This page outlines some examples of how you can do that and **why its generally a bad practice**.
+
+</TopBlock>
+
+## Local CI/CD pipeline
```
mhwelander: Similar feedback here - I think we can really shorten this to:

* If you do not have an automated CI / CD process, you can technically deploy new migrations from your local environment to test and production environments in the following way:

    1. Make sure migration history is up to date (run prisma migrate dev to generate migration history from latest changes)
    1. Change connection string to the the environment (show example)
    1. Run migrate deploy

* Strongly discourage this practice because XYZ (I think we should move this out of the note, as the style is a bit soft / muted). You can use an emoji maybe? "We **strong discourage** ⛔"

* These two say similar things (+ repeats the points you make more strongly in another place about dangers of local deploy) - can we combine into a single short paragraph that says "We recommend setting up an automated CI / CD pipeline to handle deployment to staging and production which uses `migrate deploy` in a pipeline step - see deployment guides for examples`

	  > **Note**: `prisma migrate deploy` should generally be part of an automated CI/CD pipeline, and we do not recommend running this command locally to deploy changes to a production database. It is not generally considered good practice to store the production database URL locally.
	  
	  > An alternative approach would be to let an automated CI/CD pipeline handle your staging and production environments.

* You might not need sub-headings (`###`) after shortening the section, but if you keep them, I suggest renaming to something searchable e.g. "Option 1: Deploy from local" and "Option 2: Automated CI / CD"

content/300-guides/200-deployment/150-deploy-database-changes-with-prisma-migrate.mdx
```diff
@@ -0,0 +1,29 @@
...
+```terminal
+prisma migrate deploy
+```
+
+Exactly when to run `prisma migrate deploy` depends on your platform. For example, a simplified [Heroku](./deployment-guides/deploying-to-heroku) workflow includes:
+
+1. Ensuring the `./prisma/migration` folder is in source control
+2. Running `prisma migrate deploy` during the [release phase](https://devcenter.heroku.com/articles/release-phase)
+
+Ideally, `migrate deploy` should be part of an automated CI/CD pipeline, and we do not generally recommend running this command locally to deploy changes to a production database (for example, by temporarily changing the `DATABASE_URL` environment variable). It is not genreally considered good practice to store the production database URL locally.
```
mhwelander: Link to 'deploy from local' doc 🔗

content/300-guides/200-deployment/110-deployment-guides/650-caveats-when-deploying-to-aws-platforms.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<TopBlock>
+
+The following describes some caveats you might face when deploying to different AWS platforms.
+
+</TopBlock>
+
+## AWS RDS Proxy
+
+<!-- NOTE: Move to platform overview, link to connection management -->
```
mhwelander:
```suggestion

```

content/300-guides/200-deployment/110-deployment-guides/650-caveats-when-deploying-to-aws-platforms.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+Prisma is compatible with AWS RDS Proxy. However, there is no benefit in using it for connection pooling with Prisma due to the way RDS Proxy pins connections:
+
+> "Your connections to the proxy can enter a state known as pinning. When a connection is pinned, each later transaction uses the same underlying database connection until the session ends. Other client connections also can't reuse that database connection until the session ends. The session ends when the client connection is dropped." - <a href="https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy.html">AWS RDS Proxy Docs</a>
+
+Prepared statements (of any size) or query statements greater than 16 KB cause RDS Proxy to pin the session. Because Prisma uses prepared statements for all queries, you won't see any benefit when using RDS Proxy with Prisma.
+
+## AWS Elastic Beanstalk
+
+<!-- NOTE: Move to platform overview -->
```
mhwelander:
```suggestion
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,292 @@
...
+* ✔ Assign `PrismaClient` to a global variable _in dev environments only_ to [prevent hot reloading from creating new instances](#prevent-hot-reloading-from-creating-new-instances-of-prismaclient)
+
+#### Re-using a single <inlinecode>PrismaClient</inlinecode> instance
+
+To re-use a single instance, create a module that exports a `PrismaClient` object:
+
+```ts file=client.ts
+import { PrismaClient } from "@prisma/client"
+
+let prisma = new PrismaClient()
```
Jolg42, line 54:
```suggestion
const prisma = new PrismaClient() 
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
mhwelander: Added section about optimizing for parallel requests

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
mhwelander: Made it roughly match the main headings

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
janpio: @mhwelander wrote in Slack:

> Question about the zombie connections in serverless environments - which part is it that will not work:
>
> > This will not work in a serverless environment - the function is paused and you never execute any code that could close the idle connection.
>
> Is it the "cleaning up idle connections in a separate service" or "setting lower timeout"?
> And what does serverless-mysql implement?
>
> (I'm guessing from context cleaning up idle connections is not possible - I can see settings for zombieMaxTimeout)

janpio: This really referred to "Using a lower idle connection timeout." (GH always displays the comment under the line it refers to, or highlights multiple lines if the comment is about multiple ones- in the main PR view this is most obvious, unfortunately not in the code view :/ )

An idle connection can only be closed if some code is running to close it. You can not set a magic "this connection will close itself after x second" when creating a connection. That is usually done by the program that is opening the connection, by running a timer that then closes it (e.g. in the connection pool - Prisma experimentally supports something like this as well since a recent release btw (undocumented as no idea if it actually works for now))

On serverless the code is paused and the environment frozen (with the open connection). So any "idle" setting can only be evaluated when the code is unfrozen again, but then it is already to late at the connection was blocked for long.

---

Now to the question you attributed to the line below:

> Is it the "cleaning up idle connections in a separate service" or "setting lower timeout"?
> And what does serverless-mysql implement?
> (I'm guessing from context cleaning up idle connections is not possible - I can see settings for zombieMaxTimeout)

serverless-mysql cleans up idle connetions after doing the actual work, so basically "cleaning up idle connections in a separate service" but without the "separate service" - at the cost of making all requests slower by the time the cleanup code needs to execute.

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
janpio: Ah. I'll make a not for someone to investigate that a bit, because: https://dba.stackexchange.com/a/228666

content/300-guides/200-deployment/110-deployment-guides/700-deploying-to-a-different-os.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+Prisma Client depends on the [query engine](../../../concepts/components/prisma-client/query-engine) that is running as a binary on the same host as your application.
+
+The query engine is implemented in Rust and is used by Prisma in the form of executable binary files. The binary is downloaded when `prisma generate` is called.
+
+If you have developed your application on a Windows machine for example, and wish to upload to AWS Lambda, which is a Linux environment, you may encounter issues. In fact, this just wont work and you should be presented with a warning.
+
+To solve this, if you know ahead of time that you will be deploying to a different environment, you can use the [binary targets](/concepts/components/prisma-schema/generators#the-native-binary-target) and specify which of the [supported operating systems](/reference/api-reference/prisma-schema-reference#binarytargets-options) binaries should be included.
+
+> **Note**: If your OS isn't supported you can include a [custom binary](/concepts/overview/under-the-hood#using-custom-engine-binaries).
```
janpio: Solid!

content/300-guides/200-deployment/110-deployment-guides/450-caveats-when-deploying-to-aws-platforms.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: 'Caveats when deploying to AWS platforms'
```
janpio: What other platforms than AWS Lambda does this refer to?

janpio: Ah, Beanstalk and RDS Proxy. You do not technically deploy to `RDS Proxy` so you might want to call this "using AWS services" instead.

janpio: Or split "deploying to AWS platforms (Lambda, Beanstalk)" and "using a AWS database"

content/300-guides/200-deployment/110-deployment-guides/450-caveats-when-deploying-to-aws-platforms.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+The following describes some caveats you might face when deploying to different AWS platforms.
+
+</TopBlock>
+
+## AWS Lambda + Serverless Framework + Next.js
+
+The [Serverless Framework](https://www.serverless.com/framework/docs/) is an open source CLI that enables you to manage and deploy serverless applications.
+
+The Serverless Framework can be used to deploy an [AWS Lambda function](https://www.serverless.com/framework/docs/providers/aws/guide/intro/) through Next.Js, you may however encounter issues with the Prisma binary not being copied over into the lambda function.
```
janpio: I think Serverless Framework can deploy _any_ app, including Next.js ones.
```suggestion
The Serverless Framework can be used to deploy Next.js app as an [AWS Lambda function](https://www.serverless.com/framework/docs/providers/aws/guide/intro/), you may however encounter issues with the Prisma binary not being copied over into the lambda function.
```

content/300-guides/200-deployment/110-deployment-guides/450-caveats-when-deploying-to-aws-platforms.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Another known issue is the Prisma query engine losing its executable flag. This is due to the `CopyWebpackPlugin` and the `aws-lambda` component used internally by the `serverless-next.js` package.
+
+The solution is to use `WebpackPermissionsPlugin` and a community created [patched package](https://www.npmjs.com/package/serverless-nextjs-lambda-binary-fix) for the `serverless-next.js` package.
+
+A working example of a next.config.js file and serverless.yml:
```
janpio:
```suggestion
A working example of a `next.config.js` file and `serverless.yml`:
```

content/300-guides/200-deployment/110-deployment-guides/450-caveats-when-deploying-to-aws-platforms.mdx
```diff
@@ -0,0 +1,98 @@
...
+  },
+}
+```
+
+```yml file=serverless.yml
+example:
+  component: serverless-nextjs-lambda-binary-fix@1.11.4
+```
+
+> **Note**: An SQLite database still wont be read. For this type of deployment it is suggested to us another SQL database.
```
janpio: Why is that?

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
mhwelander: Updated:https://607d3769e786c80007ec9d89--prisma2-docs.netlify.app/docs/guides/performance-and-optimization/connection-management#zombie-connections

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
mhwelander: Updated: https://607d3769e786c80007ec9d89--prisma2-docs.netlify.app/docs/guides/performance-and-optimization/connection-management#zombie-connections

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
mhwelander: (Ignore greyed out 'commit suggestion' option - the content in green is what's in the branch now)

content/300-guides/200-deployment/110-deployment-guides/450-caveats-when-deploying-to-aws-platforms.mdx
molebox: @janpio  I deleted this whole section (`AWS Lambda + Serverless Framework + Next.js`) as it was based off of a github issue that was potentially out of date. 

Its a real rabbit hole with about 20 different solutions to problems that are all related to the deployment of prisma through serverless framework to aws lambda. There doesnt appear to be one fix to recommend to users. From reading the issues it looks like the general fix is to generate the the prisam binary after deployment with a postinstall script, or via webpack somehow.

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
janpio: Uhhh, that "suggested change" in my comment really confused me - as it is not correct. Took me a while to figure out that was edited into my comment. I'll open a new comment on these lines with a new suggestion.

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+</TopBlock>
+
+## How the connection pool works
+
+The following steps describe how the query engine uses the connection pool:
+
+1. The query engine instantiates a connection pool with a [configurable pool size](#setting-the-connection-limit) and [pool timeout](#setting-the-connection-pool-timeout).
+1. The query engine opens the number of database connections defined by `connection_limit` and adds them to the connection pool.
+1. The query engine reserves connections from the pool and processes incoming queries.
```
janpio: Follow up to https://github.com/prisma/docs/pull/1537/files#r616067804
```suggestion
1. The query engine creates on connection and adds it to the connection pool.
1. When a query comes in, the query engine reserves a connection from the pool to process query.
1. If there are no idle connections available in the connection pool, the query engine opens additional database connections and adds them to the connection pool until the the number of database connections reaches the limit defined by `connection_limit`.
```

content/200-concepts/100-components/250-preview-features/050-client-preview-features.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Prisma Client preview features
+metaDescription: Prisma Client features that are currently in preview.
+---
+
+<TopBlock>
+
+The following Preview feature flags are available for Prisma Client:
+
+* [`orderByRelations`](../prisma-client/filtering-and-sorting#sort-by-relation-aggregate-preview)
```
janpio, line 10: Heads up: https://github.com/prisma/prisma/issues/3108 indicates there are quite a few more available right now.

content/200-concepts/100-components/250-preview-features/index.mdx
```diff
@@ -0,0 +1,41 @@
...
+To enable a Prisma Client Preview feature:
+
+1. Add the Preview feature flag to the `client` block:
+
+   ```prisma
+    generator client {
+    provider        = "prisma-client-js"
+    previewFeatures = ["orderByRelation"]
+    } 
+    ```
```
janpio:
```suggestion
   ```prisma
   generator client {
      provider        = "prisma-client-js"
      previewFeatures = ["orderByRelation"]
   }
   ```
```
fix formatting

content/200-concepts/100-components/250-preview-features/index.mdx
```diff
@@ -0,0 +1,41 @@
...
+    } 
+    ```  
+
+2. Re-generate the client:
+
+   ```terminal
+   npx prisma generate
+   ```
+
+3. If you are using Visual Studio Code and the Preview feature is not available in your `.ts` file after generating the client, run the **TypeScript: Restart TS server** command.
```
janpio: This should probably link to the new files we have that list CLI and Client preview features individually (also part of this PR)

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,306 @@
...
+main();
+```
+
+## Serverless environments (FaaS)
+
+Examples of serverless environments include Node.js functions hosted on AWS Lambda, Vercel or Netlify Functions. Use the following checklist as a guide to connection management in serverless environments:
+
+- ☐ Familiarize yourself with the [serverless connection management challenge](#the-serverless-challenge)
+- ☐ [Set pool size (`connection_limit`)](#recommended-connection-pool-size-1) based on whether you have an external connection pooler, and optionally [tune the pool size](#optimizing-the-connection-pool)
+- ☐ [Instantiate `PrismaClient` outside the handler](#instantiate-prismaclient-outside-the-handler) and do not explicitly `$disconnect()
```
janpio:
```suggestion
- ☐ [Instantiate `PrismaClient` outside the handler](#instantiate-prismaclient-outside-the-handler) and do not explicitly `$disconnect()`
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,306 @@
...
+
+Code should assume the container to be stateless and create a connection only if it does not exist - Prisma Client JS already implements this logic.
+
+#### Zombie connections
+
+Containers that are marked "to be removed" and are not being reused still **keep a connection open** and can stay in that state for some time (unknown and not documented from AWS). This can lead to sub-optimal utilization of the database connections.
+
+A potential solution is to **clean up idle connections** ([`serverless-mysql`](https://github.com/jeremydaly/serverless-mysql) implements this idea).
+
+> **Note**: This is a suggested solution, not an official recommendation. Refer to the documentation for your chosen serverless platform for official recommendations.
```
janpio: Problem: `serverless-mysql` is not usable with Prisma, but only with the normal `mysql` node package. So the Note does not really make sense here unfortunately - people _can_ not use this suggested solution.
```suggestion
A potential solution is to **clean up idle connections** ([`serverless-mysql`](https://github.com/jeremydaly/serverless-mysql) implements this idea, but cannot be used with Prisma).
```

(We might have an equivalent for this soon-ish, but not yet.)

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
```diff
@@ -0,0 +1,306 @@
...
+
+#### Concurrency limits
+
+Depending on your serverless concurrency limit (the number of serverless functions running in parallel), you might still exhaust your database's connection limit. This can happen when too many functions are invoked concurrently, each with its own connection pool, which eventually exhausts the database connection limit. To prevent this, you can [set your serverless concurrency limit](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) to a number lower than the maximum connection limit of your database divided by the connection used by each function invocation (as you might want to be able to connect from another client for other purposes).
```
janpio:
```suggestion
Depending on your serverless concurrency limit (the number of serverless functions running in parallel), you might still exhaust your database's connection limit. This can happen when too many functions are invoked concurrently, each with its own connection pool, which eventually exhausts the database connection limit. To prevent this, you can [set your serverless concurrency limit](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) to a number lower than the maximum connection limit of your database divided by the number of connections used by each function invocation (as you might want to be able to connect from another client for other purposes).
```

content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
ruohola: @janpio Why is specifying the schema important here? Isn't `schema=public` the default?

PR #1543, opened 2021-04-08
Change SQL examples in concepts, relations

wisnie:
Hey!
While I was reading docs I noticed that in the first SQL example, a group should be `UNIQUE`, instead of the fields itself. In the second SQL example, I think that if that's the 1-N relation, the fields can't be unique either. Correct me if I am wrong :D. Have a nice day!


mhwelander:
👋 Pardon the delay in processing this - you are right, these fields should not be unique. Sort of a strange example now that I think about it, plenty of people with the same first / last name combo! Will review in upcoming refactor of this page. Thank you!

PR #1544, opened 2021-04-09
Update $disconnect remark

matthewmueller:
Reference issue: https://github.com/prisma/prisma/issues/3069

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -2998,11 +2998,7 @@ A nested `disconnect` query breaks the connection between a parent record and a 
 #### Remarks
 
 - `disconnect` is only available if the relation is optional.
-- Prisma Client throws an exception if the provided ID or unique identifier is not connected:
-
-  ```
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -2998,11 +2998,7 @@ A nested `disconnect` query breaks the connection between a parent record and a
 #### Remarks
 
 - `disconnect` is only available if the relation is optional.
-- Prisma Client throws an exception if the provided ID or unique identifier is not connected:
-
-  ```
-  The records for relation `PostToUser` between the `User` and `Post` models are not connected.
-  ```
+- If the relationship was already not connected the operation does nothing.
```
mhwelander:
```suggestion
- If the relationship you are attempting to disconnect does not exist:

	- ([In 2.21.0 and later](https://github.com/prisma/prisma/releases/tag/2.21.0)), the operation does nothing
	- (Before 2.21.0) Prisma Client throws an exception if the provided ID or unique identifier is not connected:
	
	```
    The records for relation `PostToUser` between the `User` and `Post` models are not connected.
    ```
```

PR #1546, opened 2021-04-09
Update 100-prisma-schema-reference.mdx

lucaspere:
Updating the description that the type of the provider value is Array of enums and adding sqlserver as a valid provider.


mhwelander:
Merci! 💚

Issue #1058, opened 2020-12-07
Can't use arrow up/down keys to navigate search results


nikolasburk:
It's awesome that I can open search by hitting `/` but once I typed my search keyword I'm forced to use my mouse again because I can't navigate the search results with my arrow up/down keys.

The docs would be even more awesome if that was possible :)

MartinaTest: Cannot use arrows, screen freezes.
MartinaTest2: Slow, and it maintains the 'select' between searches - e.g. search for 'sql', page down 3x, search for 'data', #3 is selected even though it's a new search

meletj:
Fixed in the new search modal.

PR #1563, opened 2021-04-14
Fixed scroll on search results. Added some types

molebox:
This PR doesnt so much solve as provide a solution to the scroll problem. I also added some typescript types and cleaned up a bit. 
I also changed the React import on most of the files (sorry about that) TypeScript was shouting at us.


hrescak:
@molebox Thank you so much for jumping in to help!! I can't seem to be able to activate the key navigation in the deploy preview - am I missing something?  Thanks for putting in the types as well!

molebox:
> @molebox Thank you so much for jumping in to help!! I can't seem to be able to activate the key navigation in the deploy preview - am I missing something? Thanks for putting in the types as well!

No worries! Hmm, what do you mean by the key navigation? Opening the serachbar via the keyboard? Ive also just noticed the the searchbar isnt max width, thats because ive wrapped the styled div instead of the form, will push a fix

mhwelander:
@molebox Noticed that the ❌ has moved: 
image

I also can't ⬇ in the results - I get this error:

```
Uncaught TypeError: Cannot read property 'indexOf' of undefined
    at p (react-keyboard-event-handler.js:353)
    at react-keyboard-event-handler.js:261
    at Array.find (<anonymous>)
    at t.findMatchedKey (react-keyboard-event-handler.js:260)
    at t.value (react-keyboard-event-handler.js:183)
```

hrescak:
@molebox So the original issue we're trying to solve here is #1058  - the up and down arrow key navigation across search results is super super slow

molebox:
@hrescak This is very strange as i can use the keyboard to go up and down in the preview link. If you go to the docs, click on the search bar, enter some text `include` then tab into the result, you can then press down and up arrow keys. Is this not working for you? Or you @mhwelander ?

This link => [https://6076ea04ae15430008fb87fa--prisma2-docs.netlify.app/docs/](https://6076ea04ae15430008fb87fa--prisma2-docs.netlify.app/docs/)

molebox:
> @molebox Noticed that the ❌ has moved:
> image
> 
> I also can't ⬇ in the results - I get this error:
> 
> ```
> Uncaught TypeError: Cannot read property 'indexOf' of undefined
>     at p (react-keyboard-event-handler.js:353)
>     at react-keyboard-event-handler.js:261
>     at Array.find (<anonymous>)
>     at t.findMatchedKey (react-keyboard-event-handler.js:260)
>     at t.value (react-keyboard-event-handler.js:183)
> ```

Ok cool, just pushed a fix for the cross in the wrong place. You should still be able to use the up and down arrows though. I dont understand why you cant and i can, especially as we are all using the same preview link.... 🤨

https://6076f87c057d610007f8fcde--prisma2-docs.netlify.app/docs/

hrescak:
Yup, just tried again, can not use keys at all - on MacOS and Chrome or Safari. I'm also seeing an extra padding at the bottom of the field here now: 
image

PR #1569, opened 2021-04-14
ci: GH Actions workflow that outputs a list of the changed docs URLs

janpio:
The Netlify "Pages changed" functionality only lists the html file names. This is missing the deployment URL and removing the file extension to make the actual, clickable URL to see the changes.

This PR adds a GitHub Action workflow that uses an action to first get the list of changed files (same list as on "Files changed"), then creates a string that lists the preview deploy URLs, and finally creates (or updates) a comment with this string so you get a clickable list of preview links.

Examples:
- In this PR you can see how this looks when there are not changed files: https://github.com/prisma/docs/pull/1569#issuecomment-1025230295
- And in this other temp PR you can see an actual list of changed pages: https://github.com/prisma/docs/pull/2786#issuecomment-1025235690 (It also shows a challenge with the tech switcher pages which by default create an invalid link)

🚀

### DIFF
--- a/.github/workflows/list-changed-pages.yml
+++ b/.github/workflows/list-changed-pages.yml
@@ -57,3 +57,4 @@ jobs:
         body: |
           Comment with list of links that has been updated on another run
         reactions: hooray
+
--- a/.github/workflows/list-changed-pages.yml
+++ b/.github/workflows/list-changed-pages.yml
@@ -33,6 +33,7 @@ jobs:
             body="$body$link"
           fi   
         done
+        body="$body%0A%0A(Note that links will only be valid after Netlify deploy succeeded)"
         echo "::set-output name=body::$body"
 
     # - uses: LouisBrunner/checks-action@v1.1.1
--- a/.github/workflows/list-changed-pages.yml
+++ b/.github/workflows/list-changed-pages.yml
@@ -31,13 +31,9 @@ jobs:
             body="$body$link"
           fi   
         done
-        echo "ßßßß"
-        echo -e $body
         body="${body//'%'/'%25'}"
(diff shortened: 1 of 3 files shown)


Jolg42:
Looks like it's not happy with the tech switcher (which can be solved later)
https://github.com/prisma/docs/pull/2786#issuecomment-1038865490

Jolg42 (approved):
Nice! Was dreaming about that one 😍

.github/workflows/list-changed-pages.yml
```diff
@@ -0,0 +1,75 @@
...
+    steps:
+    - name: Get list of changed files
+      id: files
+      uses: jitterbit/get-changed-files@v1
+    - name: List Netlify Deploy URLs
+      id: links
+      # Note that this uses `%0A` instead of `\n` because of https://github.com/peter-evans/create-or-update-comment#setting-the-comment-body-from-a-file which I could not get to work here - but this brute force method does.
+      run: |
+        body="This PR changes the following pages (Netlify Preview Deploy links):%0A%0A"
+        no_changed_pages="%0A- This PR does not change any pages the GH Actions workflow could detect."
```
Jolg42:
```suggestion
        no_changed_pages="%0A- This PR does not change any pages the [GH Actions workflow](https://github.com/prisma/docs/blob/main/.github/workflows/list-changed-pages.yml) could detect."
```

keerlu:
> Nice! Was dreaming about that one 😍

I've also been really wanting something like this!

janpio:
> Looks like it's not happy with the tech switcher (which can be solved later)

Yes, that is all special logic that I did not want to bother with. This and more things will become obvious after this has been in production for a bit.

PR #1582, opened 2021-04-20
fix(deployment-guides): Split Serverless and PaaS

janpio:
- also renames some to actually use the platform names, vs. the deployment method
- also removes one that is an implementation detail probably


molebox (approved):

molebox:
Good shout, i thought about this early on. Then re-thought it. This makes sense though

PR #1587, opened 2021-04-20
feat(cli/introduction): add links for npm + yarn

janpio:


molebox (commented):
I kinda wonder if this is necessary? If people are using Prisma we must assume some prio knowledge which includes what package managers are.

janpio:
Not necessarily. I often try things in other languages and appreciate each single link that links to the tools that are "normal" in those ecosystems, especially in Introduction articles. But not strong opinion here, I mostly felt this would be neat here.

PR #1610, opened 2021-04-23
fix(supported-databases): Explicit mention of non support of Data API

janpio:
We got feedback https://twitter.com/mpv/status/1385592683100573696 that this currently was misleading and not clear - so mention this explicitly that not yet supported, with link to issue about this.

image

Could possibly look and sound better - feel free to edit as usual.

https://deploy-preview-1610--prisma2-docs.netlify.app/docs/reference/database-reference/supported-databases

### DIFF
--- a/content/400-reference/300-database-reference/03-supported-databases.mdx
+++ b/content/400-reference/300-database-reference/03-supported-databases.mdx
@@ -8,22 +8,24 @@ metaDescription: 'This page lists all the databases and their versions that are 
 
 Prisma currently supports the following databases:
 
-| Database              | Version |
-| --------------------- | ------- |
-| PostgreSQL            | 9       |
-| PostgreSQL            | 10      |
-| PostgreSQL            | 11      |
-| PostgreSQL            | 12      |
-| PostgreSQL            | 13      |
-| MySQL                 | 5.6     |
-| MySQL                 | 5.7     |
-| MySQL                 | 8       |
-| MariaDB               | 10      |
-| SQLite                | \*      |
-| AWS Aurora            | \*      |
-| AWS Aurora Serverless | \*      |
+| Database                     | Version |
+| ---------------------------- | ------- |
+| PostgreSQL                   | 9       |
+| PostgreSQL                   | 10      |
+| PostgreSQL                   | 11      |
+| PostgreSQL                   | 12      |
+| PostgreSQL                   | 13      |
+| MySQL                        | 5.6     |
+| MySQL                        | 5.7     |
+| MySQL                        | 8       |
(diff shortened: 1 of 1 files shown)


janpio:
PS: The other note could also be changed to a footnote for the `SQLite` row, making it superscript 1 and 2 in the table and below. Might be more readable as well.

mpvosseller:
Great, thank you @janpio!

The one other logical place you might consider adding a note about this is on the AWS caveats page but probably not needed. https://www.prisma.io/docs/guides/deployment/deployment-guides/caveats-when-deploying-to-aws-platforms

Also depending on your optimism (and willingness to discuss possible future enhancements) you might include the word "yet". ;-) 
"This does not **yet** support the Data API for Aurora Serverless"

Because I know you folks will get it done!

janpio:
Oh, that is a good suggestion that I think @mhwelander or someone from that team will happily take up.

I learned that using `yet` is not a good idea if you do not really know if and when something will happen - so I prefer not to. I think the issue link implies enough that we are of course interested in implementing this. But right now our work lies mostly elsewhere.

content/400-reference/300-database-reference/03-supported-databases.mdx
```diff
@@ -8,23 +8,29 @@ metaDescription: 'This page lists all the databases and their versions that are
...
+| AWS Aurora                   | \*      |
+| AWS Aurora Serverless &sup1; | \*      |
 
 Note that a fixed version of SQLite is shipped with every Prisma release.
 
+<FootNote>
+  
+&sup1; This does not include support for [Data API for Aurora Serverless](https://github.com/prisma/prisma/issues/1964).
+
+</FootNote>
```
janpio: This seems to break the link in there @mhwelander

content/400-reference/300-database-reference/03-supported-databases.mdx
```diff
@@ -8,23 +8,29 @@ metaDescription: 'This page lists all the databases and their versions that are
...
+| MariaDB                      | 10      |
+| SQLite                       | \*      |
+| AWS Aurora                   | \*      |
+| AWS Aurora Serverless &sup1; | \*      |
 
 Note that a fixed version of SQLite is shipped with every Prisma release.
 
+<FootNote>
+  
+&sup1; This does not include support for [Data API for Aurora Serverless](https://github.com/prisma/prisma/issues/1964).
```
mhwelander:
```suggestion
&sup1; This does not include support for [Data API for Aurora Serverless](https://github.com/prisma/prisma/issues/1964).
```

PR #1615, opened 2021-04-26
Fix indentation to render a list as a list

Darsstar:
Instead of as the markdown for that list in a code block. Expecially since it contains a link to dotenv.

It is the the code block just above "Connect your database": https://www.prisma.io/docs/concepts/components/preview-features/sql-server/sql-server-start-from-scratch-typescript/


janpio (changes requested):
Although the indentation is indeed wrong there, this is not the right solution unfortunately :/ 
We need to properly intend all the things there, including code blocks and other sentences, so this works properly.

janpio (changes requested):
-

mhwelander:
Seems the page has quite a lot of indentation problems - I'll merge this PR and fix the rest in my current branch. 

Thanks for pointing it out!

PR #1616, opened 2021-04-26
Fix logical error

rahul3v:
The below query is for id="22" and in output it showing '1'

https://www.prisma.io/docs/concepts/components/prisma-client/select-fields

loGicalError


mhwelander:
💚

PR #1718, opened 2021-05-18
Fix redundant code block

41y08h:


mhwelander:
Hi! This code is in quite a few places, so I'd like to keep it consistent. :) The suggested edit is also lacking an `await` and does not allow you to extend error handling. I will close this PR.

PR #1721, opened 2021-05-19
Fix typo in CRUD section of the documentation

wise-introvert:
~an compound id~ => a compound id


mhwelander:
Good catch, thank you!

PR #1731, opened 2021-05-21
JSON filtering preview

mhwelander:


content/200-concepts/100-components/02-prisma-client/051-working-with-fields/working-with-json-fields.mdx
```diff
@@ -46,134 +44,746 @@ export declare type JsonValue = string | number | boolean | null | JsonObject |
...
+  },
+})
+```
+
+</tab>
+</TabbedContent>
+
+- If you are using PostgreSQL, you must pass in an array of objects to match, even if that array only contains one object:
+
+  ```json
```
nikolasburk: `json` doesn't support comments, so probably better to just annotate this as `js` here:

image


```suggestion
  ```js
```

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/working-with-json-fields.mdx
```diff
@@ -46,134 +44,746 @@ export declare type JsonValue = string | number | boolean | null | JsonObject |
...
+  [{ "status": "expired", "insuranceID": 92 }] // PostgreSQL
+  ```
+
+  If you are using MySQL, you must pass in a single object to match:
+
+  ```json
```
nikolasburk:
```suggestion
  ```js
```

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/working-with-json-fields.mdx
```diff
@@ -46,134 +44,746 @@ export declare type JsonValue = string | number | boolean | null | JsonObject |
...
+
+## <inlinecode>Json</inlinecode> FAQs
+
+### Can you select a subset of JSON key/values to return?
+
+No - it is not yet possible to select which JSON elements to return. Prisma Client returns the entire JSON object.
+
+### Can you filter on the presence of a specific key?
+
+No - it is not yet possible to filter on the presence of a specific key.
```
nikolasburk: if we say "not yet", can we link to github issues that request this so that people can subscribe when they're interested in this feature? (if we don't have gh issues, would it be worth creating them ourselves?)

PR #1732, opened 2021-05-21
refactored quickstart - example

molebox:
An example of how we could shorten the Quickstart guide

### DIFF
--- a/content/100-getting-started/01-quickstart.mdx
+++ b/content/100-getting-started/01-quickstart.mdx
@@ -1,50 +1,23 @@
 ---
 title: 'Quickstart'
-# metaTitle: 'Quickstart (5 min)'
-# metaDescription: 'Get started with Prisma in 5 minutes. You will learn how to send queries to a SQLite database in a plain Node.js or TypeScript script using Prisma Client.'
+metaTitle: 'Quickstart (5 min)'
+metaDescription: 'Get started with Prisma in 5 minutes. You will learn how to send queries to a SQLite database in a plain Node.js or TypeScript script using Prisma Client.'
 duration: '5 min'
-langSwitcher: ['typescript', 'node']
-techMetaTitles:
-  [
-    { name: 'node', value: 'Quickstart: Getting started with Node.js & SQLite' },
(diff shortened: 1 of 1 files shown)


nikolasburk:
Just went through the page and I think the majority of the changes are really great! here's quick bullet list with very rough feedback points:

- I'm not in favor of removing the language switcher but open to discuss it
- like the idea of showing the Prisma schema and  mock data in the expandable section
- like that the "add"-notation is used to show the new code
- like that the CLI expandable result section is used
- the note about statically typed result might scare away JS users (IMO we should ensure that JS users are not confronted with any TS concepts when they get started with Prisma)
- why not user the ["edit"-notation](https://www.prisma.io/docs/about/style-guide/mdx-examples#code-block-with-highlighted-code) to adjust the code for the later code snippets?
- idea: instead of showing the data changes in a markdown table, maybe mention prisma studio and show a screenshot with the data?

molebox:
> Just went through the page and I think the majority of the changes are really great! here's quick bullet list with very rough feedback points:
> 
> * I'm not in favor of removing the language switcher but open to discuss it
> * like the idea of showing the Prisma schema and  mock data in the expandable section
> * like that the "add"-notation is used to show the new code
> * like that the CLI expandable result section is used
> * the note about statically typed result might scare away JS users (IMO we should ensure that JS users are not confronted with any TS concepts when they get started with Prisma)
> * why not user the ["edit"-notation](https://www.prisma.io/docs/about/style-guide/mdx-examples#code-block-with-highlighted-code) to adjust the code for the later code snippets?
> * idea: instead of showing the data changes in a markdown table, maybe mention prisma studio and show a screenshot with the data?

Thanks for looking through it Niko! 

- I think having a discussiojn around the language picker would be great, maybe next weeks docs sync?
- You note about the typed result rings true if we dont have the language picker, if its TS then we should show it but just JS then no. Seeing as I droped the picker for this draft i included the types. But i understadn where you are coming from
- Good point about the edit notation! I honestly forgot about it, but that makes total sense
- Love the screenshot idea

PR #1733, opened 2021-05-21
update seed example to async

Jolg42:
Since a seed is usually async making the example async will be easier to copy-paste for users

### DIFF
--- a/content/300-guides/050-database/300-seed-database.mdx
+++ b/content/300-guides/050-database/300-seed-database.mdx
@@ -35,7 +35,7 @@ Prisma's integrated seeding functionality expects the following:
 
       ```ts
       export async function seed() {
-        //  Prisma create query to seed models in database
+        // Prisma create query to seed models in database
       }  
       ```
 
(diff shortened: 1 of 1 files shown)


content/300-guides/050-database/300-seed-database.mdx
```diff
@@ -61,17 +61,17 @@ Prisma's integrated seeding functionality expects the following:
       <tab>
 
       ```ts
-      const seed = () => {
-        // Prisma create query to seed models in database
-      }
+      async function seed() {
+        //  Prisma create query to seed models in database
```
Jolg42:
```suggestion
        // Prisma create query to seed models in database
```

content/300-guides/050-database/300-seed-database.mdx
```diff
@@ -34,7 +34,7 @@ Prisma's integrated seeding functionality expects the following:
       <tab>
 
       ```ts
-      export const seed = () => {
+      export async function seed() {
         //  Prisma create query to seed models in database
```
Jolg42:
```suggestion
        // Prisma create query to seed models in database
```

molebox (approved):
Noice. I wonder where else we have non async code 🤔

PR #1742, opened 2021-05-24
Update 02-schema-incompatibilities.mdx

omar-dulaimi:
This breaking change was undocumented. We would have preferred it better to know about it months ago.  More info here: https://github.com/prisma/prisma/issues/6049#issuecomment-824728007


janpio (changes requested):
Internal note: Before this could be merged, we need to adapt the format to what we already have in this page and also fact check the statement made here. It is clear that Prisma 1 and Prisma 2 are different here, but let's make sure we present in the appropriate way. (This is only a problem for _some_ users)

omar-dulaimi:
> Before this could be merged, we need to adapt the format to what we already have in this page and also fact check the statement made here. It is clear that Prisma 1 and Prisma 2 are different here, but let's make sure we present in the appropriate way. (This is only a problem for _some_ users)

Please make any changes you see fit.

tanberry:
We're going to close this PR for now, @omar-dulaimi , because by now Prisma 2 has been out for a while, and there are dos to address using decimal.js. (https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields#working-with-decimal). Please let us know if you think the docs still need additional info. Thank you!

PR #1756, opened 2021-05-27
Added tip to reader that the whole page is a workaround

### LINKED ISSUE #1708: Document that the view workaround does not work with introspection/migrate
The [SQL views workaround in the docs](https://www.prisma.io/docs/guides/database/advanced-database-tasks/sql-views-mysql) is causing confusion, [as reported by users](https://github.com/prisma/prisma/issues/6015).

I think we should add the following to limitations:

- Models created for views using this workaround will be "deleted" from your schema the next time you introspect, because they do not correspond to tables in the database.
- Migrate will treat these views as tables and try to create them, leading to crashes. If you have view models, you cannot use migrate with them.

molebox:
This PR adds a tip for the user to read that indicates that this page is a workaround due to missing functionality in Prisma. It links the issue and suggets the user follow for updates.

Fixes: #1708

### DIFF
--- a/content/300-guides/050-database/900-advanced-database-tasks/08-sql-views.mdx
+++ b/content/300-guides/050-database/900-advanced-database-tasks/08-sql-views.mdx
@@ -43,6 +43,16 @@ In this guide, you will:
 - Manually update the Prisma schema to include the view as a model.
 - Generate Prisma Client and write a simple Node.js script to read data from the view.
 
+<Tip>
+
+The workaround described in this guide is by no means a best practice, it instead fills in some gaps for missing functionality in Prisma. 
+
+There is an [issue](https://github.com/prisma/prisma/issues/678) currently open which aims to add support for database views in Prisma. 
+
+Subscribe to the issue for updates on progress and timelines.
+
+</Tip>
+
 </TopBlock>
 
 ## Prerequisites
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
Looks good to me! If I have a nitpick, it would be that we could make the important piece of information that it will not work with introspection and migrate more visible by stating it quickly in the `<Tip>` block or above.

content/300-guides/050-database/900-advanced-database-tasks/08-sql-views.mdx
```diff
@@ -73,6 +83,8 @@ Be aware of the following limitations when using views with Prisma:
 - You must _manually_ add each view as a model to the Prisma schema right now. Introspection does not add views to the schema currently.
 - Views must include a unique column - such as an ID.
 - The generated Prisma Client will include queries such `create`, `delete`, and `update`, even though you cannot perform these queries on a view.
+- Models created for views will be **deleted** from your schema the next time you run an introspection. This is because they **do not** correspond to the tables in the database.
+- Prisma Migrate will treat these views as tables and try to create them, this can lead to crashes. If you have view models, you **cannot** use Prisma Migrate with them.
```
tomhoule: Nitpick: it should not lead to crashes, I think we can omit this.

molebox: Cool beans

janpio:
Did the suggested change by @tomhoule happen @molebox?

molebox:
> Did the suggested change by @tomhoule happen @molebox?

Yes

PR #1772, opened 2021-05-30
fix(crud): fix grammar mistake

chaficnajjar:


molebox (approved):
Thanks for the PR, good catch!

PR #1773, opened 2021-05-30
fix(crud): fix grammar mistake

chaficnajjar:
See https://english.stackexchange.com/questions/43820/proper-way-to-handle-plurals-with-whose.


molebox (commented):
Thanks for the PR! This change is grammatically incorrect however. The sentence is already plural, and each user does not have multiple names.

PR #1776, opened 2021-05-31
Add a new Prisma Data Platform section

martzoukos:
Even though we're early in the creation of the Prisma Data Platform, we are starting to have some needs for a permanent place in our docs for some of our current functionality.

I went for a concept page, right below our current Prisma Studio section, but please let me know if I messed the taxonomy up. Also I'd love some wording feedback. 

Thank you!

This is what it looks like:
image


molebox (approved):
This looks great to me! It's a bit sales pitchy but I don't really see an issue with that in the context. I'm going to whack an approved on it. See what @mhwelander thinks

martzoukos:
Yeah, I think that using the mission statement as a description of the Data Platform tends to make it sound too salesy and abstract, indeed. I'll give it another go.

mhwelander:
Made a few small changes, otherwise 👍 look good to me. I just want to check if we should put this in its own section rather than under 'Components'.

martzoukos:
I don't mind moving it, I just don't know what the taxonomy is like. I'll leave the decision to you.

martzoukos:
- I used Niko's nice [one-paragraph description](https://github.com/prisma/prisma/releases/) of the PDP and it is much simpler now :). 
- I though that maybe we can leave it in concepts for now. Once we start having more functionality, we can then decide to move it to a separate thing.


I'm merging this now in order to deploy it as I need to the public URL. 

Thank you for the help!

PR #1827, opened 2021-06-04
New environment variables section

### LINKED ISSUE #1583: Move env var docs to CLI section
https://prisma-company.slack.com/archives/CGMCB82N8/p1618918362224300?thread_ts=1618854273.218900&cid=CGMCB82N8

molebox:
The PR introduces a new Environment variables section under Prisma CLI. 

The sections about using env vars have been moved to their own page and a new page has been created about how to setup and use multiple env files, which is linked to from the integration testing guide (which formed the basis for this PR)

Fixes: #1583 🦊


janpio (changes requested):
env vars are also relevant for Client - see Slack

content/200-concepts/100-components/01-prisma-schema/index.mdx
```diff
@@ -136,6 +136,31 @@ Introspecting based on datasource defined in prisma/schema.prisma …
...
+}
+```
+
+In version 2.7.0 and later, you can expand variables using the format specified by [dotenv-expand](https://github.com/motdotla/dotenv-expand). This is useful when dealing with PaaS or similar products. such as Heroku):
+
+```
+# .env
+DATABASE_URL=postgresql://test:test@localhost:5432/test
+DATABASE_URL_WITH_SCHEMA=${DATABASE_URL}?schema=public
+```
```
janpio: This is not a feature of the schema and `env()`, but the `.env` file (or the library we use to parse that file's content into env vars that are picked up by `env()`).

content/200-concepts/100-components/01-prisma-schema/index.mdx
```diff
@@ -136,6 +136,31 @@ Introspecting based on datasource defined in prisma/schema.prisma …
...
+
+There are a few limitations with `env` at the moment:
```
janpio: This is not going away, so "at the moment" could give a wrong indication. 
Maybe we should better list _where_ it can actually be used (datasource.url, generator.binaryTargets - not sure about others).

content/200-concepts/300-more/050-environment-variables/index.mdx
```diff
@@ -16,29 +16,6 @@ You can use machine-specific variables, `.env` files, or a combination of both.
...
-DATABASE_URL_WITH_SCHEMA=${DATABASE_URL}?schema=public
-```
-
-There are a few limitations with `env` at the moment:
-
-- It is not possible to use environment variables for the `provider` argument in `datasource` and `generator` definitions.
-
 ## Using .env files
 
 For many developer tools, it has become a good practice to define environment variables using [`.env`](https://github.com/motdotla/dotenv) files. The Prisma CLI looks for `.env` files, in order, in the following locations:
```
janpio: We probably need a section about Prisma Client's use of `.env` as well then. Not sure if this is identical, then it should be added here as well.

content/200-concepts/100-components/01-prisma-schema/index.mdx
molebox: That would suggest this is moved to the `Environment files` (nice name suggestion btw)  section instead as its not exclusive to the `env` function

content/200-concepts/100-components/01-prisma-schema/index.mdx
```diff
@@ -136,16 +136,7 @@ Introspecting based on datasource defined in prisma/schema.prisma …
 Run prisma generate to generate Prisma Client.
 ```
 
-## Using environment variables
-
-You can use environment variables to provide configuration options when a CLI command is invoked. Environment variables allow you to:
-
-- Keep secrets out of the schema file
-- Improve portability of the schema file
```
janpio, line 144: I think these 3 lines were actually useful here @molebox - they frame why you would even need to use environment variables in the schema file itself. The `to provide configuration options when a CLI command is invoked` half sentence of course was wrong/to limited, but if you remove that you can include these three lines again.

content/200-concepts/100-components/01-prisma-schema/index.mdx
```diff
@@ -136,16 +136,7 @@ Introspecting based on datasource defined in prisma/schema.prisma …
...
-
-You can use machine-specific variables, `.env` files, or a combination of both.
```
janpio, line 146: This could be replaced by link to how to supply env vars (.env file etc)

content/200-concepts/300-more/050-environment-variables/100-managing-env-file-manually.mdx
```diff
@@ -0,0 +1,48 @@
...
+metaTitle: 'Managing .env files manually'
+metaDescription: 'Learn how to manage .env files manually'
+tocDepth: 2
+---
+
+## Manage `.env` files manually
+
+You can use `dotenv-cli` if you want to:
+
+- Use multiple `.env` files
```
janpio, line 12: Could be a link to `Using multiple .env files` page

content/200-concepts/300-more/050-environment-variables/200-using-multiple-env-files.mdx
```diff
@@ -0,0 +1,68 @@
+---
+title: 'Using multiple .env files'
+metaTitle: 'Using multiple .env files.'
+metaDescription: 'Learn how to setup a dedicated testing environment using multiple .env files.'
+tocDepth: 2
+---
+
+<TopBlock>
+
+There is a risk that your production database could be deleted if you store different connection URLs to each of your environments within a single `.env` file.
```
janpio, line 10: This should probably be clarified as it is suuuuper scary. When and how is this possible?

content/200-concepts/300-more/050-environment-variables/index.mdx
```diff
@@ -0,0 +1,95 @@
+---
+title: 'Environment variables'
+metaTitle: 'Environment variables'
+metaDescription: 'Learn how to use environment variables in your Prisma project'
+tocDepth: 2
+---
+
+<TopBlock>
+
+You can use environment variables to provide configuration options when a CLI command is invoked. Environment variables allow you to:
```
janpio, line 10: Not only CLI, but also Client

PR #1843, opened 2021-06-07
Update Prisma DEBUG

Jolg42:
Context https://prisma-company.slack.com/archives/CEYCG2MCN/p1623056404067700

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
@@ -30,5 +30,5 @@ export DEBUG="prisma:client,prisma:engine"
 If you want to enable all debugging options, you can also set `DEBUG` to `prisma*`:
 
 ```terminal
-export DEBUG="prisma*"
+export DEBUG="*"
 ```
--- a/content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
@@ -11,6 +11,7 @@ You can enable debugging output in Prisma Client via the [`DEBUG`](../../../refe
 - `prisma:engine`: Prints relevant debug messages happening in a Prisma [engine](https://github.com/prisma/prisma-engines/)
 - `prisma:client`: Prints relevant debug messages happening in the Prisma Client runtime
 - `prisma*`: Prints all debug messages from Prisma Client or CLI
+- `*`: Prints all debug messages, including external sources such as `dotenv`
 
 ## Setting the `DEBUG` environment variable
 
@@ -27,7 +28,13 @@ export DEBUG="prisma:client"
 export DEBUG="prisma:client,prisma:engine"
 ```
 
-If you want to enable all debugging options, you can also set `DEBUG` to `*`:
+To enable all `prisma` debugging options, set `DEBUG` to `prisma*`:
+
+```terminal
+export DEBUG="prisma*"
+```
+
+To enable _all_ debugging options, set `DEBUG` to `*`:
 
 ```terminal
 export DEBUG="*"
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
```diff
@@ -17,18 +17,15 @@ You can enable debugging output in Prisma Client via the [`DEBUG`](../../../refe
 Here are examples for setting these debugging options in bash:
 
 ```terminal
-# enable only `prisma:engine`-level debugging output
-export DEBUG="engine"
```
janpio: Why remove this?

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
```diff
@@ -17,18 +17,15 @@ You can enable debugging output in Prisma Client via the [`DEBUG`](../../../refe
...
-
 # enable only `prisma:client`-level debugging output
 export DEBUG="prisma:client"
 
 # enable both `prisma-client`- and `engine`-level debugging output
 export DEBUG="prisma:client,prisma:engine"
 ```
 
-If you want to enable all debugging options, you can also set `DEBUG` to `*`:
+If you want to enable all debugging options, you can also set `DEBUG` to `prisma*`:
```
janpio: Are we sure all Prisma logging is prefaced with `prisma`?

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
```diff
@@ -17,18 +17,15 @@ You can enable debugging output in Prisma Client via the [`DEBUG`](../../../refe
...
 
 ```terminal
-export DEBUG="*"
+export DEBUG="prisma*"
```
janpio, line 34: I do not think replacing this is a good choice, it should probably just be listed above as the most wide catching Prisma level option. We have been telling people to use DEBUG=* for ages in all our issues and will keep doing so as it gives the full picture (also including stuff that might influence Prisma from the outside (like loading env vars in dotenv etc).

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
Jolg42: Yes but in old versions it's possible that some were not prefixed.

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
Jolg42: It doesn't exist

Jolg42: Could be changed to `export DEBUG="prisma:engine"` if we want to keep

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
mhwelander: Is there any way to check so that we can give people a range (if you're using Prisma before version X, do this)?

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
mhwelander: Restored

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
mhwelander:
```suggestion
export DEBUG="*"
```

content/200-concepts/100-components/02-prisma-client/140-debugging.mdx
Jolg42: In theory yes but that means going through a lot of Git history so I would say maybe we stick with `*` for "all" and let people know that "prisma*" is also a thing to isolate prisma logs. (And avoid babel noise for example)

mhwelander: Splendid - that's exactly what I did. Thanks!

PR #1867, opened 2021-06-10
Fix docs

chaficnajjar:


content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -551,7 +551,7 @@ const checkJson = await prisma.user.findMany({
   { "status": "expired", "insuranceID": 92 } // MySQL
   ```
 
-- If your array contains multiple objects, PostgreSQL will only return results if 
+- If your array contains multiple objects, PostgreSQL will only return results if
```
janpio: This is still not a full sentence though, right?

content/200-concepts/100-components/02-prisma-client/055-aggregation-grouping-summarizing.mdx
```diff
@@ -602,9 +602,9 @@ model Play {
   id       Int   @id @default(autoincrement())
   score    Int?  @default(0)
   playerId Int?
-  player   User? @relation(fields: playerId, references: id)
-  gameId Int?
-  game Game? @relation(fields: gameId, references: id)
+  player   User? @relation(fields: [playerId], references: [id])
+  gameId   Int?
+  game     Game? @relation(fields: [gameId], references: [id])
```
janpio, line 607: Do the brackets matter here? Does the formatter auto correct to this format?

content/200-concepts/100-components/04-introspection.mdx
```diff
@@ -204,12 +204,12 @@ Prisma translates this into the following data model:
 
 ```prisma
 model User {
-  id      Int      @default(autoincrement()) @id
+  id      Int      @id @default(autoincrement())
```
janpio, line 207: This is what the formatter does to the schema?

janpio (approved):
<3

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
chaficnajjar: @janpio Right, this sentence seems incomplete. Can you ask someone on your team to fix this? I don't know what the full sentence should look like.

content/200-concepts/100-components/02-prisma-client/055-aggregation-grouping-summarizing.mdx
chaficnajjar: @janpio I didn't use the formatter but it says [here](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#relation) that the type of `fields` and `references` should be `FieldReference[]`, so I assumed that the brackets mattered here. The brackets are also used in the rest of the docs so even if they're not required, I'm guessing that it's better to have them to make the code style across the docs more coherent.

content/200-concepts/100-components/04-introspection.mdx
chaficnajjar: @janpio I didn't use the formatter but it says [here](https://www.prisma.io/docs/concepts/components/introspection#order-of-attributes) that `@id` must come before `@default`. I also used `prisma introspect` in my own project and it did place `@id` before `@default`.

janpio: Then you effectively used the same codebase as the formatter 👍

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
mhwelander:
```suggestion
- If your filter array contains multiple objects, PostgreSQL will only return results if _all_ objects are present - not if at least one object is present.
```

content/200-concepts/100-components/02-prisma-client/055-aggregation-grouping-summarizing.mdx
mhwelander: Fair - both work, but we have generally used square brackets in the docs. I see a few places where this isn't the case but we can clean that up.

mhwelander:
💚 This is great!

PR #1876, opened 2021-06-12
fix: align P2002 message with the implementation

armand1m:


janpio:
The error message documentation is generated, and there can probably be other values of the constraint placeholder.

armand1m:
could you point me to how they get generated?

armand1m:
nvm just found it

PR #1891, opened 2021-06-17
2.25.0 / Drift

mhwelander:


content/200-concepts/100-components/03-prisma-migrate/200-shadow-database.mdx
```diff
@@ -50,6 +50,25 @@ If Prisma Migrate does not detect schema drift, it moves on to [generating new m
 
 > **Note**: The shadow database is not responsible for checking if a migration file has been **edited or deleted**. This is done using a the `checksum` field in the `_prisma_migrations` table.
 
+In [2.25.0](https://github.com/prisma/prisma/releases/tag/2.25.0) and later, Prisma Migrate outputs detailed information about which parts of the database have drifted. In the following example, the `Color` enum in the target database is missing the expected variant `RED` and include the unexpected variant `TRANSPARENT`:
```
mhwelander:
```suggestion
In [2.25.0](https://github.com/prisma/prisma/releases/tag/2.25.0) and later, Prisma Migrate outputs detailed information about which parts of the database have drifted. In the following example output, the schema that would be generated from the target database differs from the expected schema. For example, the `Color` enum in the target database is missing the expected variant `RED` and include the unexpected variant `TRANSPARENT`:
```

content/400-reference/200-api-reference/200-command-reference.mdx
```diff
@@ -552,7 +552,7 @@ npx prisma format
 Environment variables loaded from prisma\.env
 Prisma schema loaded from prisma\schema.prisma
 Error: Schema parsing
-error: The preview feature "orderByRelationnp" is not known. Expected one of: microsoftSqlServer, groupBy, createMany, orderByRelation, nApi
+error: The preview feature "orderByRelation" is not known. Expected one of: microsoftSqlServer, groupBy, createMany, orderByRelation, nApi
```
janpio, line 555: Wasn't showing an error via a wrong preview feature the intent of this example? We probably should replace this with `doesNotExist` or similar.

Although: Why do we document the error case? Nothing is formatted in this case, as the validation fails before formatting starts, right?

janpio:
Many of the changes in this PR look like prettier versions fighting with each other. I see this every few PRs here in the docs repo, that these function calls are changed from one style to another.

PR #1913, opened 2021-06-22
fix(system-requirements): explain Node.js compatib a bit more

janpio:


content/400-reference/100-system-requirements.mdx
```diff
@@ -14,16 +14,19 @@ The [latest version of Prisma](https://www.npmjs.com/package/@prisma/client) req
 | Yarn (optional)       | 1.19.2                   |
 
 Notes:
+- Other Node.js releases [in non-Active-LTS state like _Current_ or _Maintenance-LTS_ , and also odd-numbered versions](https://nodejs.org/en/about/releases/) probably also work, but are not recommended for production use. They are not fully tested by Prisma - the versions listed above are.
```
Jolg42: v12 is "Maintenance LTS" and we support it so could be confusing.
```suggestion
- Other Node.js releases [in non-Active-LTS state like _Current_ or odd-numbered versions](https://nodejs.org/en/about/releases/) probably also work, but are not recommended for production use. They are not fully tested by Prisma - the versions listed above are.
```

janpio: Good point! I think then we can rewrite it this way:
```suggestion
- Prisma supports and tests all _Active TLS_ and _Maintenance LTS_ **Node.js** releases. [Releases that are not in these states like _Current_, and also odd-numbered versions](https://nodejs.org/en/about/releases/) probably also work, but are not recommended for production use.
```
Do you agree @Jolg42?

Jolg42 (approved):

content/400-reference/100-system-requirements.mdx
```diff
@@ -14,16 +14,19 @@ The [latest version of Prisma](https://www.npmjs.com/package/@prisma/client) req
 | Yarn (optional)       | 1.19.2                   |
 
 Notes:
-- TypeScript is only required for TypeScript users.
-- When using Yarn 1, 1.19.2 is the minimum version compatible with Prisma Client.
+- Prisma supports and tests all _Active TLS_ and _Maintenance LTS_ **Node.js** releases. [Releases that are not in these states like _Current_, and also odd-numbered versions](https://nodejs.org/en/about/releases/) probably also work, but are not recommended for production use.
```
Jolg42:
```suggestion
- Prisma supports and tests all _Active LTS_ and _Maintenance LTS_ **Node.js** releases. [Releases that are not in these states like _Current_, and also odd-numbered versions](https://nodejs.org/en/about/releases/) probably also work, but are not recommended for production use.
```

PR #1968, opened 2021-07-02
Reduce the verbosity of the solution in 400-nextjs-prisma-client-dev-practices.mdx

alecmev:


molebox:
Hey @alecmev 👋 Thanks for the PR! We feel that being verbose can, in most cases, improve the readability of the code and help people to understand what is actually happening. For that reason im going to decline this PR.

alecmev:
Hi @molebox! I agree with the sentiment, but I believe an inline union might be a more familiar concept to TypeScript users than creating a new type through the extension of an existing one. And the comment that accompanies it - "add prisma to the NodeJS global type" - might make somebody think that the `global` type has a `prisma` property on it throughout the project, which isn't the case, it has an effect only in this file. I also added `?` to to the property, which reflects the reality better.

molebox:
> Hi @molebox! I agree with the sentiment, but I believe an inline union might be a more familiar concept to TypeScript users than creating a new type through the extension of an existing one. And the comment that accompanies it - "add prisma to the NodeJS global type" - might make somebody think that the `global` type has a `prisma` property on it throughout the project, which isn't the case, it has an effect only in this file. I also added `?` to to the property, which reflects the reality better.

Hey @alecmev , those are some solid reasons. Ive had a chat with our typescript dept and they agree with your reasoning. So im happy to merge this PR 🙌

alecmev:
I appreciate you giving consideration to my "appeal", in most projects it would just get ignored :wink:

PR #1971, opened 2021-07-05
Feat: Smart tab os detection

luanvdw:
@nilubava @janpio 

**Summary:**

With the tabbedContent component being in use with content other than that of just the Platform type (Unix vs Windows), there are a few ways in which to approach overriding the default tab behaviour.

A quick solution might have been to create a separate tabbedComponent that would have this logic contained within this new component itself but then the base of it would have a lot of duplicated functionality so I opted for a way in which we can add additional logic that would override the default behaviour only if certain criteria are met with the flexibility to add more criteria in the future if we needed new defaulted behaviour given a different array of tabs.

This update should not affect areas/files where the tabbedContent component is currently being used, but going forward we can specify if it should.

**Testing:**

I do not have a windows machine that I could use for _actual_ testing so my tests were done through User Agent manipulation (reference provided) It would be great if someone on the team can perhaps just help me test this on an actual windows based machine.

smart-tabs


luanvdw:
As a precursor I have only added the updated OS default override to the tabbed content on the `docs/getting-started/quickstart` page. As far as I could tell, this is the only place where OS based tabbed content is currently in use but once the PR logic has been approved, if someone can maybe help me identify if it is being used anywhere else in the docs then I will add the update there as well 👍

janpio:
When I, on my Windows machine, go to https://deploy-preview-1971--prisma2-docs.netlify.app/docs/getting-started/quickstart/ I see the Windows tab active by default - on my partner's Mac the other one is active, same on iOS 🎆 🥳 

We should probably open two follow up issues:
1. Visually the active tab is almost impossible to see - the contrast is so low I have to squint my eyes to see it.
2. We should add this to all OS switchers as it is awesome! (Searching for "Windows" probably gives all uses)

janpio (approved):
Can confirm this works on Windows, macOS and iOS as expected 👍 

@nilubava for code and solution.

molebox:
Well this looks very slick! 🙌

src/components/customMdx/tabbedContent/index.tsx
```diff
@@ -1,17 +1,42 @@
...
   const [activeIndex, setActiveIndex] = React.useState(0)
   const tabContent =
     children &&
     children.filter((child: any) => child.props && child.props.originalType === 'tab')[activeIndex]
+
+  React.useEffect(() => {
+
+    if (tabs && tabs.length > 1) {
+      // check if any default tab behaviour has been stipulated
+      switch (defaultTabType) {
```
molebox, line 27: Curious why you have used a switch here? 😊

luanvdw: @molebox Great question! Although the switch is not really needed for a single "case", my thinking was that going forward we can use this structure to add different/additional override logic for a different set of tabs.

For example, if we have tabbed content for our _supported languages_ or _supported databases_ and wanted a "funnelled" journey where we set a default based on a users previous selection, we could then use this switch case to check which logic we would like to apply in determining the default value for a set of tabs.

molebox: Awesome, thanks for explaining. 😊

nilubava (approved):
LGTM :)

PR #1980, opened 2021-07-05
fix(data-model): model => data model

janpio:
Fix a few location where it used `model` but `data model` was meant - and especially on this page where we explain how the data model consists of models this seemed important :D


molebox (approved):
seal-of-approval

PR #1985, opened 2021-07-06
style: Use `prettier-plugin-prisma`.

avocadowastaken:
Related to https://github.com/umidbekk/prettier-plugin-prisma/issues/9#issue-938063626

- [x] Setup `lint-staged` (#1989)
- [x] Replace diff tokens with highlight annotation (it breaks default prettier parsers, e.g: JS, GraphQL) (#1986) 
- [x] Format `.json` files with prettier (#1991)
- [x] Format `.js` files with prettier (#1992)
- [x] Format `.md` files with prettier) (#1993)
- [ ] Format `.css` files with prettier) (#2019)
- [ ] Format `.mdx` files with prettier ([embedded-language-formatting](https://prettier.io/docs/en/options.html#embedded-language-formatting): off) (#1995)
- [ ] Format `.mdx` code blocks ([embedded-language-formatting](https://prettier.io/docs/en/options.html#embedded-language-formatting): on)
- [ ] Run prettier on CI (#2020)

___

@janpio I will keep this issue a meta issue for now, because we can see everything that prettier doing wrong here.

PR #1986, opened 2021-07-06
refactor: Replace diff tokens with `highlight` annotation.

avocadowastaken:
Usage of inline diff tokens in code blocks breaks prettier formatting #1985.


nilubava (approved):
LGTM :)

PR #2005, opened 2021-07-08
Document queries thenable

### LINKED ISSUE #1695: Document thenable nature of Prisma Client queries
https://github.com/prisma/prisma/discussions/7033 + https://prisma-company.slack.com/archives/CUXLS0Z6K/p1620760435005700

molebox:
This PR attempts to explain that the instantiated prisma client is thenable

Fixes: #1695


content/200-concepts/100-components/02-prisma-client/index.mdx
```diff
@@ -101,6 +101,12 @@ const newUser = await prisma.user.create({
 const users = await prisma.user.findMany()
 ```
 
+<Admonition type="info">
+
+Note that queries will only run if used with `await` or chained with `then()`. This is due to how [`$transactions`](transactions) work when passing the queries on to the [query engine](../prisma-engines/query-engine). The instantiated client has ["thenable"](https://masteringjs.io/tutorials/fundamentals/thenable) methods that act like promises, and the queries are executed as a batch job.
```
Jolg42:
```suggestion
Note that queries will only run if used with `await` or chained with `.then()`. When using [`$transaction`](transactions) this makes it possible to the client to pass all the queries on to the [query engine](../prisma-engines/query-engine) as a transaction.
The client methods are ["thenable"](https://masteringjs.io/tutorials/fundamentals/thenable) they return a PrismaPromise which is implemented like a Promise except that for the query to be executed they need to be awaited or chained with `.then()`.
```

>This is due to how $transactions work when passing the queries on to the query engine.

This could be confusing if you are not using transaction?

PR #2007, opened 2021-07-08
namedConstraints: Remove that map can be an unnamed argument name for `@id` and `@unique`

janpio:
and explain the unnamed argument names a bit better while I am at it - I hope.

Decision made in https://prisma-company.slack.com/archives/C024SCGH291/p1625677909006700, implementation will be adapted

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -980,7 +980,7 @@ Attributes modify the behavior of a [field](#model-fields) or block (e.g. [model
 - _Field_ attributes are prefixed with `@`
 - _Block_ attributes are prefixed with `@@`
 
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -980,7 +980,7 @@ Attributes modify the behavior of a [field](#model-fields) or block (e.g. [model
 - _Field_ attributes are prefixed with `@`
 - _Block_ attributes are prefixed with `@@`
 
-Some attributes take arguments. Arguments in attributes are always named, but in most cases the argument _name_ can be omitted.
+Some attributes take arguments. Arguments in attributes are always named, but in some cases one of the argument's _name_ can be omitted.
```
do4gr, line 983: I think it can only ever be the first one that can omit the name.

janpio: Is the order important? 
Or what are you suggesting to change here?

PR #2011, opened 2021-07-09
POC Shiki twoslash

molebox:
# DO NOT TRY AND MERGE THIS - ITS JUST A POC

## Whats all this about then?

This is a POC of replacing prism with [shiki twoslash](https://shikijs.github.io/twoslash/). This would enable us to show the power of prismas typescript integration.

## Stuff of note

I added two new models to our schema and ran `db push`, i then mocked the client in each code block so that it would use the schema models. We can extract this mocking and have it at the top level of the mdx page. I dont think we can import it from an single source, like a file at the root or whatever.

There are some files i added just playing around, `client.ts` and `singleton.ts` these can be ignored. 

There is a custom `code.css` file which handles the styling of the code blocks. The theme os the codeblock is set in the `gatsby-config` as an option on the remark plugin.

Ive added it to `http://localhost:8000/reference/api-reference/prisma-client-reference#model-queries`
And the quickstart so far, just testing out places.

Its quite sensitive and if you make a change (like remove the `twoslash` attribute from the codeblock, gatsby will cry and break so you have to build again.

Ill keep on plugging away at it, but feel free to use this PR to have a play around and see what it can do!

## Some GIFS! 

Quickstart example:

prisma-twoslash-2

Model query example:
prisma-twoslash-4


tanberry:
Closing for now, though it does look fun...

PR #1995, opened 2021-07-07
style: Format MDX files with Prettier.

avocadowastaken:
@nilubava thanks for the quick reviews! 

This one is big. Prettier has some bugs with MDX. 

[...]

PR #2013, opened 2021-07-09
feat: Percy visual testing

janpio:
Quick test with Percy to make PRs like https://github.com/prisma/docs/pull/1995 simpler to test and confirm.


janpio:
Unfortunately getting the snapshots on Netlify takes quite a while:
```
9:45:46 PM: [percy] created build #9: https://percy.io/6faef5e0/prisma-docs/builds/11449855
9:45:47 PM: [percy] percy has started.
...
9:51:38 PM: [percy] waiting for 245 snapshots to complete...
9:51:38 PM: [percy] done.
9:51:38 PM: [percy] finalized build #9: https://percy.io/6faef5e0/prisma-docs/builds/11449855
```

6 minutes additional build time is to much.

So we probably would want to extract that into a GH Actions run here that is optimized to do that I would say.

janpio:
And another problem: 
Running Percy on Netlify creates _very_ different results from running it on Gitpod (a dev environment):

image

janpio:
That was quick: 
image
Blocked until someone buys a paid plan for us internally 🤷

avocadowastaken:
> 6 minutes additional build time is to much.
> 
> So we probably would want to extract that into a GH Actions run here that is optimized to do that I would say.

Yeah, looks like Netlify is not optimized for that

> And another problem:
> Running Percy on Netlify creates _very_ different results from running it on Gitpod (a dev environment):

Looks like it uses different fonts, I had same problem with loki js, main reason why I started to use Percy. Strange that it occurs on Percy too. Because the mechanic they use to produce the snapshots should be consistent.

> Blocked until someone buys a paid plan for us internally 🤷

I thought they provide free plans for the OSS  https://www.browserstack.com/open-source

janpio:
> > And another problem:
> > Running Percy on Netlify creates _very_ different results from running it on Gitpod (a dev environment):
> 
> Looks like it uses different fonts, I had same problem with loki js, main reason why I started to use Percy. Strange that it occurs on Percy too. Because the mechanic they use to produce the snapshots should be consistent.

Different fonts and missing code block etc - so possibly some code not executed. Let's hope it is not the case via GH Actions.


> > Blocked until someone buys a paid plan for us internally 🤷
> 
> I thought they provide free plans for the OSS [browserstack.com/open-source](https://www.browserstack.com/open-source)

The money is not the problem, the weekend is 😆 Applied anyway, maybe Browserstack is still working and gets back to us sooner.

janpio:
> > I thought they provide free plans for the OSS [browserstack.com/open-source](https://www.browserstack.com/open-source)
> 
> The money is not the problem, the weekend is 😆 Applied anyway, maybe Browserstack is still working and gets back to us sooner.

They do it turns out (suuuuuper fast!), but Percy is not included :/

janpio:
Update: Confirmed they only include the other services, not Percy. Need to wait for the credit card next week.

avocadowastaken:
@janpio btw did you add github integration? it should add Percy github check status https://docs.percy.io/docs/github#step-3-run-and-review-builds

.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
...
+  gatsby-for-percy:
+    name: Percy screenshots via Gatsby build
+    runs-on: ubuntu-latest
+    env:
+      PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+      - run: yarn
+      - run: yarn gatsby build
```
avocadowastaken: I've copied this file in #2020, so there might be merge conflicts, also should it be
```suggestion
      - uses: actions/setup-node@v2
         with:
           cache: 'npm'
           node-version: '14'
      - run: npm ci
      - run: npm run build
```

janpio:
Yes, but I think it will only pop up after a successful run - which we did not have yet as we are missing payment data. But we are working on that now.

.github/workflows/ci.yml
janpio: Oh lol, why are we not using yarn here... Thanks for the suggestion.

janpio:
Here is the Percy GitHub check and it should also be public. If that works and makes sense, we can merge this to the main branch and you can rebase your PRs and then we should be able to visually see if anything changed.

janpio:
Ugh. Unfortunately it looks like it is also using a different font and not rendering code blocks and so on :/ We'll need to debug this before this can be useful :(

janpio:
Actually, merging this anyway as it gives us _some_ visual feedback.

janpio:
Trying that in https://github.com/prisma/docs/pull/2025

janpio:
Update: Nope, "No diffs" with this change.

PR #2028, opened 2021-07-13
Relations troubleshooting: Implicit any-to-many self-relation

mhwelander:


mhwelander:
Closing because incorrect

PR #2045, opened 2021-07-16
fix(logging): consistent quotes for logging config

janpio:
consistent is better than not


molebox:
Looks like a potential copy paste mistake. Is this PR going to be added to? You've marked it as WIP

janpio:
Ah nope, that was probably just wrong button. You can merge if you agree.

PR #2047, opened 2021-07-16
style: Format mdx code blocks.

avocadowastaken:


content/200-concepts/300-more/400-comparisons/01-prisma-and-typeorm.mdx
```diff
@@ -223,10 +223,7 @@ export class User {
   @Column({ unique: true })
   email: string
 
-  @OneToMany(
-    type => Post,
-    post => post.author
-  )
+  @OneToMany((type) => Post, (post) => post.author)
```
janpio, line 232: That might have been a TypeORM convention in this file.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -1894,37 +1894,35 @@ const whereScalarTitleIs = Prisma.validator<Prisma.PostScalarWhereInput>()({
...
-    id: 1,
-  },
-  data: {
-    content: 'This is some updated content',
-    published: true,
-    title: 'This is a new title',
-  },
-})
+const updatePostByIdWithoutAuthor =
+  Prisma.validator<Prisma.PostUpdateWithWhereUniqueWithoutAuthorInput>()({
```
janpio, line 1901: so much better

src/components/customMdx/prism/prism-prisma.js
```diff
@@ -19,3 +19,5 @@ Prism.languages.insertBefore('prisma', 'function', {
 Prism.languages.insertBefore('prisma', 'punctuation', {
   'type-args': /\b(?:references|fields):/,
 })
+
+Prism.languages.json5 = Prism.languages.js
```
janpio, line 23: What does this do and why?

avocadowastaken: It fits better to show node output like that:

````
```json5
[{ email: 'yewande@prisma.io' }, { email: `raheem@gmail.com` }]
```
````

Because prettier adds semicolon at the begging or in the end depending on the rule, e.g:

````
```js
;[{ email: 'yewande@prisma.io' }, { email: `raheem@gmail.com` }]
```
````

**Comment by janpio** [comment]
Went through the Percy diffs - but fear I forgot to mark some code snippets where the `highlight` needs to be updated as the formatting changed new lines etc.

**Comment by avocadowastaken** [comment]
> Went through the Percy diffs - but fear I forgot to mark some code snippets where the `highlight` needs to be updated as the formatting changed new lines etc.

Yeah, this once are hard to find at first glance. I was thinking about using inline annotations for highlights and had an idea, what if instead of adding symbol that breaks formatting at the beginning, we add it at the end with single line comments, e.g:

```ts line-number

// has parsing error
export const schema = makeSchema({
|types: [Query, User, Post, Profile, Category],
  plugins: [nexusSchemaPrisma()],
  outputs: {
    schema: __dirname + '/../schema.graphql',
    typegen: __dirname + '/generated/nexus.ts',
  },
  typegenAutoConfig: {
    sources: [
      {
        source: '@prisma/client',
        alias: 'prisma',
      },
    ],
  },
})

// works just fine
export const schema = makeSchema({
  types: [Query, User, Post, Profile, Category], // | or highlight=normal
  plugins: [nexusSchemaPrisma()],
  outputs: {
    schema: __dirname + '/../schema.graphql',
    typegen: __dirname + '/generated/nexus.ts',
  },
  typegenAutoConfig: {
    sources: [
      {
        source: '@prisma/client',
        alias: 'prisma',
      },
    ],
  },
})
```

src/components/customMdx/prism/prism-prisma.js
janpio: Ohh interesting - so `json5` is a workaround for stupid prettier rules basically? We should document that somewhere in the guide pages then.

janpio:
Hm that is an interesting idea, but I think you better create a separate issue for it. Right now I think we should be fine with the current system, if we just look out for these a bit and make sure to fix them over time if we mess up. But do create that issue, that would indeed solve a lot of problems to put these into comments.

avocadowastaken:
> Hm that is an interesting idea, but I think you better create a separate issue for it. Right now I think we should be fine with the current system, if we just look out for these a bit and make sure to fix them over time if we mess up. But do create that issue, that would indeed solve a lot of problems to put these into comments.

@janpio yeah, sure, I didn't mean to add them in this PR, I just saw that's how unhandy this top level annotations are when using formatter that can move code between the lines. Even some minor prettier release can mess up highlighting once more.

avocadowastaken:
@janpio I also noticed that a lot of embedded code got "wider", maybe it's worth to change `printWidth` from `100` to `80` for `.mdx` files? But it will affect every line in these files, not only the code blocks.

janpio:
I saw that as well - might be worth trying out in a new PR. If we could limit it to the code blocks that would be better though - as there scrolling is a thing, normal text on the other hand will just wrap.

avocadowastaken:
> I saw that as well - might be worth trying out in a new PR. If we could limit it to the code blocks that would be better though - as there scrolling is a thing, normal text on the other hand will just wrap.

I misread your comment and add this change into this PR 🤦‍♂️

janpio:
I think this looks pretty good now. Do you agree @umidbekk? Then I think we should just merge and move forward.

avocadowastaken:
> I think this looks pretty good now. Do you agree @umidbekk? Then I think we should just merge and move forward.

@janpio sure, I think it's ready to be merged

PR #2058, opened 2021-07-25
Adding missing migration name

nguyenyou:


molebox (approved):
Looks good! Thanks for this PR too, it led us to look at where we document the `--name` flag. As such, we'll be adding a little more info on that. 🙌

PR #2059, opened 2021-07-25
Add missing --schema argument to `migrate dev` command

fcaylus:
Hi !
I noticed that the `--schema` argument was missing from the `migrate dev` command reference, but was present for every other `migrate *` commands.

However, it's a supported option (as documented [here in source code](https://github.com/prisma/prisma/blob/master/src/packages/migrate/src/commands/MigrateDev.ts#L62)).

This PR adds this missing argument, using the same text as the other commands.


Jolg42:
Hi! Thank you for this PR, good catch!

Jolg42 (approved):

PR #2066, opened 2021-07-27
SQLServer connection string add ApplicationName added in 2.28

Jolg42:


janpio (changes requested):
Probably needs mention when it was added?

Jolg42:
@janpio added "Since version 2.28.0."

janpio (approved):

PR #2069, opened 2021-07-30
fix(environment-variables-reference): Elaborate on PRISMA_QUERY_ENGINE_BINARY

janpio:
Explains a bit what influence this env var actually has.

### DIFF
--- a/content/400-reference/200-api-reference/300-environment-variables-reference.mdx
+++ b/content/400-reference/200-api-reference/300-environment-variables-reference.mdx
@@ -125,7 +125,7 @@ PRISMA_QUERY_ENGINE_BINARY=custom/my-query-engine-unix
 ```
 
 For Prisma CLI it allows you to define the query engine file to be used.  
-For Prisma Client on build time (during `prisma generate`) it defines where the query engine file will be copied from into the Prisma Client, on run time (when using the generated Client) it can be used to define the specific  query engine file to be used instead of the included one.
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/300-environment-variables-reference.mdx
```diff
@@ -118,12 +118,15 @@ PRISMA_MIGRATION_ENGINE_BINARY=custom/my-migration-engine-unix
...
 
-`PRISMA_QUERY_ENGINE_BINARY` is used to set a custom location for your own query engine binary.
+`PRISMA_QUERY_ENGINE_BINARY` is used to set a custom location for your own query engine binary. 
 
 ```terminal
 PRISMA_QUERY_ENGINE_BINARY=custom/my-query-engine-unix
 ```
 
+For Prisma CLI it allows you to define the query engine file to be used.  
+For Prisma Client on build time (during `prisma generate`) it defines where the query engine file will be copied from into the Prisma Client, on run time (when using the generated Client) it can be used to define the specific  query engine file to be used instead of the included one.
```
molebox:
```suggestion
For Prisma Client, on build time (during `prisma generate`), it defines where the query engine file will be copied from into the Prisma Client. At run time (when using the generated Client) it can be used to define the specific query engine file to be used instead of the included one.
```

molebox (approved):
Thumbs up ✔

PR #2077, opened 2021-08-04
fix(prisma-client-reference): Specify atomic number operations behavior on null values

olliechick:
Clarify behaviour of null values with atomic number operations

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -4228,6 +4228,7 @@ export type IntFieldUpdateOperationsInput = {
 ### Remarks
 
 - You can only perform **one** atomic update per field, per query.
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4228,6 +4228,7 @@ export type IntFieldUpdateOperationsInput = {
 ### Remarks
 
 - You can only perform **one** atomic update per field, per query.
+- If a field is null, it will not be updated by `increment`, `decrement`, `multiply`, or `divide`.
```
janpio: Probably?
```suggestion
- If a field is `null`, it will not be updated by `increment`, `decrement`, `multiply`, or `divide`.
```

olliechick: Done 👍

thebiglabasky (approved):

janpio:
Thanks @olliechick for the PR 🚀

PR #2104, opened 2021-08-06
Made MongoDB authorId optional

pcbowers:
Fixes a bug in one of the examples where the authorId should be optional and was not.


molebox (approved):
Looks good, thank you!

PR #2106, opened 2021-08-06
Fix typo in docker-compose command

TasinIshmam:
Missing ```-``` in ```docker-compose up -d```


molebox (approved):
While this is true for docker compose v1, docker compose v2 enables you to drop the `-` and is backwards compatible. But seeing as not everyone will be running compose v2 as it is in BETA, we can merge your change.

TasinIshmam:
Thanks for the review! I will merge the changes since you approved @molebox

PR #2107, opened 2021-08-07
Indent admonition to preserve numbering

### LINKED ISSUE #1744: Prisma Generate Docs Possible Typo
I think there might be a typo on https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client.

> Note also that prisma generate is automatically invoked when you're installing the @prisma/client npm package. So, when you're initially setting up Prisma Client, you can typically save the third step from the list above.

It should say the fourth step since that's where the generate command is run.

jboxman:
This just needed indentation to match the rest of the ordered list. Tested and works locally.

Closes #1744


janpio (approved):

molebox (approved):
Looks good 🙌

PR #2112, opened 2021-08-09
Restructure `namedConstraints` preview feature guide page

janpio:
This PR restructures the preview feature guide page a bit.

Preview: https://deploy-preview-2112--prisma2-docs.netlify.app/docs/guides/upgrade-guides/upgrading-to-use-preview-features/enabling-named-constraints


do4gr (approved):

molebox (commented):
The backticks in the mdx files title, which renders as the pages title (Enabling `namedConstraints`) doesn't create an inline codeblock. So we can actually see the backticks on the page, and in the nav menu on the left.

Id remove those. Or we can try and figure out why its not rendering as an inline codeblock 😊

PR #2131, opened 2021-08-10
fix(relation-queries): Fix query description

ultrahumus9000:
lte stands for less or equal to but in the original post it was explained by has more on line 1127

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/037-relation-queries.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/037-relation-queries.mdx
@@ -1111,7 +1111,7 @@ For example:
 For example, the following query returns `User` that meet the following criteria:
 
 - No posts with more than 100 views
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/02-prisma-client/037-relation-queries.mdx
```diff
@@ -1111,7 +1111,7 @@ For example:
 For example, the following query returns `User` that meet the following criteria:
 
 - No posts with more than 100 views
-- All posts have more than 50 likes
+- All posts have less or equal to 50 likes
```
molebox:
```suggestion
- All posts have less than, or equal to 50 likes
```

molebox (changes requested):
Small change then we are good to go!

i-beleive-in-change

molebox (approved):
well-done

PR #2132, opened 2021-08-10
fix(error-reference): Add structure

janpio:
A suggestion for a bit more headline structure on the error reference page

### DIFF
--- a/content/400-reference/200-api-reference/250-error-reference.mdx
+++ b/content/400-reference/200-api-reference/250-error-reference.mdx
@@ -13,7 +13,7 @@ This page describes Prisma exceptions and error codes. For more information abou
 
 ## Prisma Client error types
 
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/250-error-reference.mdx
```diff
@@ -11,7 +11,11 @@ This page describes Prisma exceptions and error codes. For more information abou
 
 </TopBlock>
 
-## <inlinecode>PrismaClientKnownRequestError</inlinecode>
+## Prisma Client error types
+
+Prisma Client throws different kinds of errors, that can have the following exception types with the documented data fields for information:
```
molebox:
```suggestion
Prisma Client throws different kinds of errors. The following lists the exception types, and their documented data fields:
```

molebox (approved):
Im still unsure about this sentence, but we roll with it

PR #2138, opened 2021-08-12
fix(link-handling): Fix Homepage Guides Links

amazingefren:
LinkCard elements in the homepage were assigned href as divs

This prevented the element from sending the client to the guide


molebox (approved):
This is a great fix, thank you!
well-done

PR #2141, opened 2021-08-12
Replace imgur with cloudinary

### LINKED ISSUE #4: Move images from imgur to self-hosted / strategy for hosting images
image

thebiglabasky:
Closes https://github.com/prisma/docs/issues/4


thebiglabasky (commented):
I just double checked the correspondance between source imgur URLs and the corresponding cloudinary ones, we should be good

molebox (approved):
noice


I would make sure we change the image names to reflect what they are so that in the code we have an idea too.

thebiglabasky:
Yes, I thought of that but preferred first to keep them for simplifying the review and ensure it would stay coherent and not introduce mistakes. It would be good as a next step to name things properly and utilize the gatsby plugin

PR #2146, opened 2021-08-13
fix(names-in-underlying-database): Move @@index definitions to correct place

pimeys:


molebox (approved):

janpio (changes requested):
highlight needs to be updated as well @molebox 

visible nicely in Percy btw:
image

PR #2151, opened 2021-08-13
Cyclic ref actions page: Small changes in copy

molebox:
Added some small changes to the copy of the Cyclic ref actions page

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/100-cyclic-referential-actions.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/100-cyclic-referential-actions.mdx
@@ -11,6 +11,24 @@ Circular references can occur between models when each model references another,
 
 With Microsoft SQL Server however, if the [referential actions](./) on the relation field are set to something other than `NoAction`, the data model validator will throw an error about the cyclic relation.
 
+Given the below model:
+
+```prisma
+model Chicken {
+  id        Int      @id @default(autoincrement())
+  cock      Chicken? @relation(name: "a_self_relation")
+  hen       Chicken? @relation(name: "a_self_relation", fields: [chickenId], references: [id])
+  chickenId Int?
+}
+```
+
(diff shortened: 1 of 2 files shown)


janpio (approved):

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/100-cyclic-referential-actions.mdx
```diff
@@ -9,9 +9,9 @@ tocDepth: 2
 
 Circular references can occur between models when each model references another, creating a closed loop. Most SQL databases handle these references out of the box. 
 
-With Microsoft SQL Server, however, if the [referential actions](./) on the relation field is set to something other than `NoAction`, the data model validator will throw an error.
+With Microsoft SQL Server however, if the [referential actions](./) on the relation field are set to something other than `NoAction`, the data model validator will throw an error about the cyclic relation.
```
janpio, line 12: Not a comment on the PR, but the content of the page:

I think here we could also show the SQL Server error message that would pop up when you try to create that schema with the cyclic referential actions. Our data model validation will prevent that from happening of course, but it shows the context _why_ we have this validation in the first place.

janpio (commented):
Technically this page should right now also have a mention that SQL Server is in preview right now and link to that preview feature page probably. Not sure if we bother about that as it will go GA in 2 sprints though.

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/100-cyclic-referential-actions.mdx
molebox: Sure, i can do that. Will find somewhere to add it

molebox:
> Technically this page should right now also have a mention that SQL Server is in preview right now and link to that preview feature page probably. Not sure if we bother about that as it will go GA in 2 sprints though.

I did think about that but reasoned, as you said, that it will be out of preview in a short while.

PR #2165, opened 2021-08-17
Update 400-nextjs-prisma-client-dev-practices.mdx

aavonk:
Hello! I ran into a little error when following the docs for instantiating a Prisma Client in a Next.js app using Typescript. It isn't a bug within Prisma itself, but rather a typing issue of the global object in Node. 

I'm pretty sure ```Global``` type on ```NodeJs.Global``` was removed in @node/types 16.0.0. Getting a "Namespace 'NodeJS' has no exported member 'Global'" error. Replacing it with ```typeof globalThis``` does the trick.

Perhaps this is just due to the version of @types/node that next.js installs automatically, but here are the current versions I'm running:
```
Node - 16.0.0
Next.js -  11.1.0
@prisma/client - 2.29.1
prisma: 2.29.1
```

### DIFF
--- a/content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
+++ b/content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
@@ -23,13 +23,14 @@ The solution in this case is to instantiate a single instance `PrismaClient` and
 ```ts file=./db
 import { PrismaClient } from '@prisma/client'
 
+declare global {
+  var prisma: PrismaClient | undefined
+}
+
 // Prevent multiple instances of Prisma Client in development
-declare const global: typeof globalThis & { prisma?: PrismaClient }
-
-const prisma = global.prisma || new PrismaClient()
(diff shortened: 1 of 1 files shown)


millsp (commented):
Thanks for your contribution! I suggest updating with the following. We can take advantage of TypeScript's module augmentation. And since prisma moved on to `node 12`, we can do the following:

```ts

declare global {
  var prisma: PrismaClient | undefined
}

// Prevent multiple instances of Prisma Client in development
if (process.env.NODE_ENV === 'development' && !globalThis.prisma) {
  globalThis.prisma = new PrismaClient()
}
```

https://v8.dev/features/globalthis

molebox:
Hey @aaronvonkreisler , let us know if you'd like to make the suggested changes to this PR. 😄

aavonk:
Hey @molebox & @millsp, I went ahead and added the suggested changes in a new commit. Let me know if you have any thoughts or suggestions around this new change. Thanks!

molebox (approved):
seal-of-approval

PR #2164, opened 2021-08-17
SQL Server GA

molebox:
This PR address a number of mssql tasks: 

Fixes: #2095
Fixes: #2083 
Fixes: #2094 
Fixes: #2091 
Fixes: #2086 
Fixes: #2082 

The bulk of the work is moving the getting started guide out from the preview section and integrated into our getting started guide.

The rest of the work is moving the stuff out of the preview section into the database connectors section and removing mention of mssql being in preview

PR #2170, opened 2021-08-18
feat: Use netlify dev instead of netlify-lambda

nilubava:
#2160


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
+# . "$(dirname "$0")/_/husky.sh"
```
janpio, line 2: Intentional?

janpio: (Noticed via https://github.com/prisma/docs/pull/2164#discussion_r692795128)

Issue #2157, opened 2021-08-16
Document changes to $queryRaw


molebox:
v3 introduces breaking changes to $queryRaw which need to be documented. See [this PR](https://github.com/prisma/prisma/issues/7142#issuecomment-898812395) for more information

PR #2177, opened 2021-08-19
$queryRaw and $executeRaw GA

molebox:
This PR update the `$queryRaw` and `$executeRaw` docs to reflect the changes made in https://github.com/prisma/prisma/issues/7142

Fixes: #2157


content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -87,13 +87,22 @@ Instead of immediately awaiting the result of each operation when it's performed
 >
 > Refer to the 📖 [transactions guide](../../../guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api) for more examples.
 
-### Interactive Transactions
+### Interactive Transactions (in Preview)
```
janpio: Unrelated change to PR - should probably be merged into the epic branch so it does not show up here.

content/300-guides/100-performance-and-optimization/100-prisma-client-transactions-guide.mdx
```diff
@@ -823,11 +823,20 @@ It is now impossible for two people to book the same seat:
 3. Seat 3A claimed by Sorcha (`version` is incremented to 1, booking succeeds)
 4. Seat 3A claimed by Ellen (in-memory `version` (0) does not match database `version` (1) - booking does not succeed)
 
-### Interactive Transactions
+### Interactive Transactions (in Preview)
```
janpio: also unrelated change here

janpio (commented):
- Line 16 can be adapted to list the new methods as well now
- SQL Injection section can probably be updated now with new methods

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -111,19 +117,48 @@ result.forEach((x) => {
...
-> **Note**: The Prisma Client query engine standardizes the return type for all databases. Using the raw queries does not.
+> **Note**: The Prisma Client query engine standardizes the return type for all databases. **Using the raw queries does not**.
+
+## <inlinecode>$queryRawUnsafe</inlinecode>
+
+The `$queryRawUnsafe` method allows you to pass a raw string (or template string) to the database. 
+
+<Admonition type="warning">
+
+**By using this method you open up the possibility for SQL injection attacks**.
```
thebiglabasky:
```suggestion
**By using this method you open up the possibility for SQL injection attacks.
We therefore strongly advise against using this method and use `queryRaw` instead, unless you have no alternative. If so, be sure to properly escape any variable to mitigate the risk of SQL injections.**
You can [learn more about SQL injections on Wikipedia](https://en.wikipedia.org/wiki/SQL_injection).
```

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -30,15 +30,23 @@ Use cases for raw SQL include:
...
-const result = await prisma.$queryRaw('SELECT * FROM User;')
+const result = await prisma.$queryRaw`SELECT * FROM User`
 ```
 
-The same method is implemented as a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates), which makes it easier to use [variables](#using-variables) by providing **placeholders**. The following example includes an `${email}` placeholder:
+The method is implemented as a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates), which makes it easier to use [variables](#using-variables) by providing **placeholders**. The following example includes an `${email}` placeholder:
 
 ```ts no-lines
-const email = 'edna@prisma.io'
+const email = 'bruce@avengers.io'
```
thebiglabasky: Why switching to Avengers? I'd stick to something as neutral (and diverse) as possible ideally.

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -47,8 +55,8 @@ const result =
 
 ```json5
 [
-  { id: 1, email: 'sarah@prisma.io', name: 'Sarah' },
-  { id: 2, email: 'alice@prisma.io', name: 'Alice' },
+  { id: 1, email: 'bruce@avengers.io', name: 'Bruce' },
+  { id: 2, email: 'tony@avengers.io', name: 'Tony' },
```
thebiglabasky: Cf. above. I'd try something more neutral and diverse.

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -146,7 +181,38 @@ Be aware that:
...
+$executeRaw<T = unknown>(query: TemplateStringsArray | Prisma.Sql, ...values: any[]): PrismaPromise<T>;
+```
+
+## <inlinecode>$executeRawUnsafe</inlinecode>
+
+The `$executeRawUnsafe` method allows you to pass a raw string (or template string) to the database. Like `$executeRaw`, it does **not** return database records, but returns the number of rows affected.
+
+<Admonition type="warning">
+
+**By using this method you open up the possibility for SQL injection attacks**.
```
thebiglabasky:
```suggestion
**By using this method you open up the possibility for SQL injection attacks.
We therefore strongly advise against using this method and use `executeRaw` instead, unless you have no alternative. If so, be sure to properly escape any variable to mitigate the risk of SQL injections.**
You can [learn more about SQL injections on Wikipedia](https://en.wikipedia.org/wiki/SQL_injection).
```

thebiglabasky (changes requested):
Left some comments but generally looks good to me.
The interactive transactions content being there is a bit disturbing, but can work fine

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
molebox: Yeah, you are 100% correct. I had changed one and thought about changing all the others to be consistent. But its not really the place for it to be fair 😅

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -111,19 +119,53 @@ result.forEach((x) => {
...
-> **Note**: The Prisma Client query engine standardizes the return type for all databases. Using the raw queries does not.
+> **Note**: The Prisma Client query engine standardizes the return type for all databases. **Using the raw queries does not**.
+
+## <inlinecode>$queryRawUnsafe</inlinecode>
+
+The `$queryRawUnsafe` method allows you to pass a raw string (or template string) to the database.
+
+<Admonition type="warning">
+
+By using this method you open up the possibility for SQL injection attacks. SQL injection attacks can expose your data, be it confidential or otherwise sensitive, to being deleted, modified, or even destroyed.
```
thebiglabasky:
```suggestion
By using this method you open up the possibility for SQL injection attacks. SQL injection attacks can expose your data, be it confidential or otherwise sensitive, to being modified, or even destroyed.
```
Destroyed vs deleted didn't seem to make sense together?

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -111,19 +119,53 @@ result.forEach((x) => {
...
+
+We strongly advice that you use the `$queryRaw` query instead. For more information on SQL injection attacks, see the [OWASP SQL Injection guide](https://www.owasp.org/index.php/SQL_Injection).
```
thebiglabasky: (Maybe that's my english 😄)
Don't we say "We strongly advise" with an 's'?

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -136,7 +178,7 @@ const result: number =
 Be aware that:
 
 - `$executeRaw` does not support multiple queries in a single string (for example, `ALTER TABLE` and `CREATE TABLE` together).
-- Prisma Client submits prepared statements, and prepared statements only allow a subset of SQL statements (`START TRANSACTION`is not permitted). For example, ["MySQL: SQL Syntax Allowed in Prepared Statements"](https://dev.mysql.com/doc/refman/8.0/en/sql-prepared-statements.html).
+- Prisma Client submits prepared statements, and prepared statements only allow a subset of SQL statements. For example, `START TRANSACTION` is not permitted. You can learn more about [“the syntax that MySQL allows in Prepared Statements here”](https://dev.mysql.com/doc/refman/8.0/en/sql-prepared-statements.html).
```
thebiglabasky:
```suggestion
- Prisma Client submits prepared statements, and prepared statements only allow a subset of SQL statements. For example, `START TRANSACTION` is not permitted. You can learn more about [the syntax that MySQL allows in Prepared Statements here](https://dev.mysql.com/doc/refman/8.0/en/sql-prepared-statements.html).
```
Tiny detail :)

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -146,7 +188,46 @@ Be aware that:
...
+$executeRaw<T = unknown>(query: TemplateStringsArray | Prisma.Sql, ...values: any[]): PrismaPromise<T>;
+```
+
+## <inlinecode>$executeRawUnsafe</inlinecode>
+
+The `$executeRawUnsafe` method allows you to pass a raw string (or template string) to the database. Like `$executeRaw`, it does **not** return database records, but returns the number of rows affected.
+
+<Admonition type="warning">
+
+By using this method you open up the possibility for SQL injection attacks. SQL injection attacks can expose your data, be it confidential or otherwise sensitive, to being deleted, modified, or even destroyed.
```
thebiglabasky:
```suggestion
By using this method you open up the possibility for SQL injection attacks. SQL injection attacks can expose your data, be it confidential or otherwise sensitive, to being modified, or even destroyed.
```

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -235,36 +318,45 @@ When you use `ILIKE`, the `%` wildcard character(s) should be included in the va
...
 
 PostgreSQL [does not support using `ALTER` in a prepared statement](https://www.postgresql.org/docs/current/sql-prepare.html), which means that the following queries will not work:
 
 ```ts
 await prisma.$executeRaw`ALTER USER prisma WITH PASSWORD '${password}'` // Tagged template
-await prisma.$executeRaw(`ALTER USER prisma WITH PASSWORD $1`, newPassword) // Parameterized query
+await prisma.$executeRaw`ALTER USER prisma WITH PASSWORD $1, newPassword` // Parameterized query
 ```
 
 You can use the following query to [work around the issue](https://github.com/prisma/prisma-client-js/issues/940#issuecomment-748014175), but be aware that this is potentially **unsafe** as `${password}` is not escaped:
```
thebiglabasky: I think there might be an issue here since we say $executeRaw is the safe one and $executeRawUnsafe is the unsafe one. How could $executeRaw be unsafe?

thebiglabasky: Shouldn't that then be done using the `Unsafe` signature?

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -235,36 +318,45 @@ When you use `ILIKE`, the `%` wildcard character(s) should be included in the va
...
+  ```ts
+  $queryRaw`` // Tagged template
+  $executeRaw`` // Tagged template
+  $queryRaw`query, variables` // Parameterized query
+  $executeRaw`query, variables` // Parameterized query
+  ```
+
+- Both versions of `$executeRaw` (tagged template and parameterized) can only run **one** query at a time. You cannot append a second query - for example, adding `DROP bobby_tables` to the end of an `ALTER`.
+
+If you cannot use tagged templates or parameterized queries (for example, you need to pass in table or column names as variables), you can instead use [`$queryRawUnsafe`](#query-raw-unsafe) or [`$executeRawUnsafe`](#execute-raw-unsafe) but **be aware that your code may be vulnerable to SQL injection**.
```
thebiglabasky, line 359: Note for later: it could be good to point people to how to properly escape strings etc. to prevent SQL injections whenever using Unsafe signature methods. This can be done in a separate effort, but would be worthwhile as a user, since you otherwise just see you're in this situation but left with the task to figure out how to deal with it on your own.

content/300-guides/100-performance-and-optimization/100-prisma-client-transactions-guide.mdx
```diff
@@ -823,7 +823,7 @@ It is now impossible for two people to book the same seat:
 3. Seat 3A claimed by Sorcha (`version` is incremented to 1, booking succeeds)
 4. Seat 3A claimed by Ellen (in-memory `version` (0) does not match database `version` (1) - booking does not succeed)
 
-### Interactive Transactions (in Preview)
+### Interactive Transactions
```
thebiglabasky: Still unrelated, but why dropping the (Preview) label here? In preparation for the GA of interactive transactions?

thebiglabasky (changes requested):
I've spotted a couple of minor things.
One key question revolves around the ALTER table limitation and suggestion solution where I wonder if there is an error in the doc here.

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
molebox: AHHHHHHHH yes we do! I shall take your advice and change it!

ba-dum-tis

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
molebox: This section i wasnt sure about as it relates to an issue. I asked about this in slack yesterday too.

content/300-guides/100-performance-and-optimization/100-prisma-client-transactions-guide.mdx
molebox: Yes, have i jumped the gun on that? Looking through my notes i can see that its not mentioned for 7th sept release so i think i may have made a mistake here.

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
molebox: I agree, a new section within this page about how to escape strings while using `$executeRawUnsafe` / `$queryRawUnsafe` would be ideal. Ive asked @millsp about that. Would be good to link to it from within the page

millsp: Somewhat related: I'm wondering if we could allow `ALTER`, `CREATE`, `DROP` in `$executeRaw` and instead of doing a prepared statement, we would automatically escape their inputs. I also think that we should look into providing some helper to escape input.

molebox: @millsp is this something you are thinking about for 7th sept release?

thebiglabasky: Those are interesting ideas indeed @millsp and I think we should track them in their individual issues, unless they would come with a breaking change, in such case that would be better to indeed do those before the new major.

thebiglabasky (approved):

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
millsp: That won't be a breaking change, so I'll keep the improvements for later.

PR #2185, opened 2021-08-20
fix(database-connectors/mysql): Remove unsupported `sslmode`

janpio:
`sslmode` is not actually supported for MySQL

Also removes invalid statement what is required to create a SSL connection to your database (both for MySQL and PostgreSQL).


molebox (approved):
seal-of-approval

PR #2199, opened 2021-08-24
Fix SQL Server feature matrix

pimeys:
- No support for `RESTRICT` action
- JSON support is only partial


molebox (approved):
seal-of-approval

PR #2202, opened 2021-08-24
Fix section heading level

TasinIshmam:
The heading level was previously wrong, making it look like certain subsections were empty.

Wrong: 

wrong

Correct: 

correct

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
@@ -225,7 +225,7 @@ This relation expresses the following:
 
 Note that you can also require each user to have a teacher by making the `teacher` field [required](../data-model#optional-and-mandatory-fields).
 
-### One-to-many self-relations in the database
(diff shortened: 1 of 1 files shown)


molebox:
I believe this is currently correct, as the `One-to-many self-relations in the database` section is seen as a child of the `One-to-many self relations` section.

Do you find it misleading when reading it?

TasinIshmam:
Oh alright,  I got it the other way around then. 

> I believe this is currently correct, as the `One-to-many self-relations in the database` section is seen as a child of the `One-to-many self relations` section.

This is not true for ```One-to-one self relations in the database``` and ```One-to-one self relations``` at the moment though. Perhaps this should be fixed?

molebox:
> Oh alright, I got it the other way around then.
> 
> > I believe this is currently correct, as the `One-to-many self-relations in the database` section is seen as a child of the `One-to-many self relations` section.
> 
> This is not true for `One-to-one self relations in the database` and `One-to-one self relations` at the moment though. Perhaps this should be fixed?

Sure thang 😀

PR #2206, opened 2021-08-24
Update full text search examples table

jonahsnider:
- Write example queries as inline code
- Write yes/no instead of true/false (more consistent with existing writing style)


molebox (approved):
I like these changes! Thanks for the PR!

noice

jonahsnider:
Looks like the Markdown parser used in the site is treating the `\|` differently than GitHub's editor preview:
screenshot of website

PR #2212, opened 2021-08-26
Add example of explicit many-to-many relation

### LINKED ISSUE #2201: Lack of Explicit Many-to-Many relation example in Self Relation Docs
In the [Self-Relation](https://github.com/prisma/docs/blob/8f1f1b01657b5263b622fffeef458ee528184c88/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx) guide, there is no example of explicit many-to-many self-relation. 

This makes the proper syntax for explicit many-to-many self relations potentially confusing, especially to new users. An explicit version of the many-to-many schema that's already provided in the article looks like this: 

```prisma
model User {
  id            String  @id @default(autoincrement())
  name          String
  followedBy    Follows[] @relation("follower")
  following     Follows[] @relation("following")
}

model Follows {
  follower    User @relation("followedBy", fields: [followerId], references: [id])
  followerId  String
  following   User @relation("following", fields: [followingId], references: [id])
  followingId String

  @@id([followerId, followingId])
}
```

This example can be added along with a link to the [many-to-many concept guide](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations/#explicit-many-to-many-relations). I believe this would make the article more comprehensive. 

Would love to hear thoughts/feedback on this!

TasinIshmam:
Fixes #2201

Note: I haven't added the explicit many-to-many version for MongoDB.  I'm not sure if this makes sense to demonstrate, considering the way many-to-many is modeled in MongoDB. This is something I'll have to look into a bit further. I hope it's alright that this is not provided at the moment :sweat_smile:

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
@@ -310,7 +310,54 @@ This relation expresses the following:
 - "a user can be followed by zero or more users"
 - "a user can follow zero or more users"
 
-Note that this n-n-relation is [implicit](many-to-many-relations#implicit-many-to-many-relations). This means Prisma maintains a [relation table](../relations/many-to-many-relations#relation-tables) for it in the underlying database:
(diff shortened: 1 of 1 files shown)


molebox (approved):
LGTM, we'll tackle mongo another time. It thinks it's a little bit special 😜

noice

Issue #2217, opened 2021-08-26
Support escaping pipes (eg. \|) inside tables


matthewmueller:
Right now there's no way to add a | inside a markdown table because it will complete the table. It's not possible to escape it either.

This came up in our [full text documentation](https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search#query-format). The syntax below is wrong, it should be cat | fox, but the | closes the table

<img width="765" alt="CleanShot 2021-08-26 at 13 52 53@2x" src="https://user-images.githubusercontent.com/170299/131019711-375959de-6452-486c-a3cf-a0557880be8f.png">

molebox:
A simple solution to this, for now, would be to remove the inline code blocks. So instead of `fox /| cat` just write fox || cat

matthewmueller:
That's what I originally did after I noticed the problem, but it got flipped back later on. To me, this means it's an easy mistake to make.

Another option is to make `<code>fox | cat</code>` equivalent to the backticks. Unfortunately, there's extra padding applied to inline `<code>` blocks, so it looks bad.

molebox:
> That's what I originally did after I noticed the problem, but it got flipped back later on. To me, this means it's an easy mistake to make.
> 
> Another option is to make `<code>fox | cat</code>` equivalent to the backticks. Unfortunately, there's extra padding applied to inline `<code>` blocks, so it looks bad.

We could create a simple component that wraps the code element and removes all padding. Call it Tablecode or something

matthewmueller:
Ah, I just realized you fixed this with `<inlinecode>` in https://github.com/prisma/docs/pull/2218/files

That solves the problem!

PR #2218, opened 2021-08-26

matthewmueller:
Unfortunately pipes can't be escaped in our docs markdown parser right now. Created an issue: https://github.com/prisma/docs/issues/2217


janpio (changes requested):
Current change breaks the table - I fear you have to use html code for the `|` inside the column.
image

molebox (approved):
I fixed the table so that we can use the code blocks 👯‍♀️

PR #2260, opened 2021-09-03
feat: Added tailwind + new lens navbar

luanvdw:
Status

1. Add tailwind to docs project.

The preferred and optimised way to add tailwind to a Gatsby project is with PostCSS, `postcss` +  `gatsby-plugin-postcss`
The latest version for `gatsby-plugin-postcss` is `4.13.0` which is incompatible with `gatsby V2` and required `V3+`

Migrating from Gatsby 2 -> 3 can sometimes be straightforward but should definitely be tackled in a separate PR to isolate any breaking changes that might occur.

Alternatively there are 2 options,
- working through `postcss` +  `gatsby-plugin-postcss` to find compatibility for gatsby `2.32.13` 
- import the default tailwind file, hosted on CDN (as can be seen from the preview link)
  - Slow-ish to load -> `fouc`
  - Conflicting styles causing visual discrepancies

eg,

without tailwind import (correct),

correct

with tailwind import (wrong),

wrong


molebox:
Im not sure how high up on the list this is, but i would create that second PR for upgrading to Gatsby v3 then swing back to this. I dont see the point in going half in on the implementation with v2 when moving forward we are going to have to upgrade to v3 anyway. This seems like a good opportunity to do that 😄

luanvdw:
> Im not sure how high up on the list this is, but i would create that second PR for upgrading to Gatsby v3 then swing back to this. I dont see the point in going half in on the implementation with v2 when moving forward we are going to have to upgrade to v3 anyway. This seems like a good opportunity to do that 😄

I agree, with the Navigation on home being different to that of docs and blog I wanted to see what a plausible interim solution would be. I was able to get the installation to work via `postcss` with `"gatsby-plugin-postcss": "3.7.0"` but tailwinds base, still contains conflicting styles creating quite the number of breaking changes. Busy investigating a possible purge to isolate manually generated styles to only those contained within the Navbar.

PR #2266, opened 2021-09-03
fix: full text search

Ketcap:
Fixed  markdown on full text search.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
@@ -9,27 +9,29 @@ metaDescription: 'This page explains how to search for text within a field.'
 The Prisma Client supports a full-text search API for Postgres databases. With full-text search enabled, you can add search functionality to your application by searching for text within a database column.
 
 <Admonition type="info">
-  **Note**: Full text search can **only** be used in a Postgres database
+
+**Note**: Full text search can **only** be used in a Postgres database
+
 </Admonition>
 
 ## Enable full-text search
 
 The full-text Search API is still in **Preview**. You can enable `fullTextSearch` in your Prisma Schema:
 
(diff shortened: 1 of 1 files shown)


janpio:
I think line 87/8 needs to be changed to triple backtick for this to work.
(You can make the newly added one and the one 2 lines before that also only triple backtick by the way. I see no need why it need to be 4 backticks here)

Ketcap:
It stops there for that section. And i can fix that for that sql line as well 👍

Ketcap:
@janpio I've fixed 2 more places to render properly.
- [x] Prisma code example fixed
- [x] Terminal code example fixed
- [x] Sql code example fixed
- [x] Page render issue fixed by missing " ` "

content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
```diff
@@ -18,24 +18,24 @@ The full-text Search API is still in **Preview**. You can enable `fullTextSearch
...
+```
 
 2: Generate the Prisma Client:
 
-    ```terminal copy
-    npx prisma generate
-    ```
+```terminal copy
+  npx prisma generate
+```
```
janpio: I think a better solution here would be to keep the indentation of the code blocks, and replace `1:` and  `2:` with `1.` and `2.` to make it a numbered list. Then this would look fine.

janpio:
Awesome. I think one of those can be fixed in a better way even - left a comment with the idea.

Then this would be ready to merge I would say.

Ketcap:
@janpio I've updated as you recommended looking good i've also fixed `Admonition` text rendering for bold in the page let me know if it needs more adjustment 👍

janpio (approved):
Great work, thank you! (I'll let @molebox or @matthewmueller merge)

molebox (approved):
looks-good

PR #2267, opened 2021-09-03
fix(prisma-client-transactions-guide): fix referential actions admon + Whats new week 30th sept (#2264)

janpio:


molebox (approved):
Good catch!

PR #2277, opened 2021-09-06
full-text search: fix the sql to create an index. Closes #2216

### LINKED ISSUE #2216: full-text search: GIN index that's documented is wrong
Our current GIN index doesn't work. 

> Current FTS appear to not use defined GIN or GIST index
> 
> Recommended GIN index in doc is also not working
> 
> ERROR:  data type text has no default operator class for access method "gin"
> HINT:  You must specify an operator class for the index or define a default operator class for the data type.
> As far as I know it should be something like this
> 
> CREATE INDEX IF NOT EXISTS "post_body_index" ON "Posts" USING GIN (to_tsvector('english', "body"));
> -- Or/And
> CREATE INDEX IF NOT EXISTS "post_body_index" ON "Posts" USING GIST (to_tsvector('english', "body"));

Reference: https://github.com/prisma/prisma/issues/8950

matthewmueller:
This PR fixes the index query

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
@@ -90,7 +90,7 @@ define your own index in SQL:
 CREATE INDEX post_body_index ON "Post" USING GIN (to_tsvector('english', "body"));
 ```
 
-You can continue using Prisma Migrate as you were before, it will ignore indexes
+And you can continue using Prisma Migrate as you were before, it will ignore indexes
 that it doesn't know about.
 
 ## Limitations
--- a/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
@@ -82,11 +82,12 @@ Here's how the following queries would match that text:
 ## Adding Indexes
 
 To speed up your full-text queries, you should add an index to your database.
-Prisma Migrate currently doesn't support adding search indices, but you can
-easily add one yourself.
+
+While the Prisma Schema doesn't natively support GIN indices yet, you can easily
+define your own index in SQL:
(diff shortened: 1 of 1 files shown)


molebox (approved):

content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
```diff
@@ -82,14 +82,15 @@ Here's how the following queries would match that text:
...
+define your own index in SQL:
 
 ```sql
-CREATE INDEX post_body_index ON posts USING GIN (body);
+CREATE INDEX post_body_index ON "Post" USING GIN (to_tsvector('english', "body"));
 ```
 
-You can continue using Prisma Migrate as you were before, it will ignore indexes
+And you can continue using Prisma Migrate as you were before, it will ignore indexes
 that it doesn't know about.
```
janpio, line 94: What does that sentence mean really? Introspection looks at existing indexes, but ignores the ones it does not understand. Migrate only uses them as the basis for creating migrations, which could lead to indexes not in the Prisma schema to being _removed_  in the database.

tanberry: @Jan, do you mean that with MIgrate, if the search index is ignored, and thus not added to the schema, then it (of course) won't be added to the new database?

I was reading the current sentence as "Don't worry about manually adding a search index, because it will not impact your use of Migrate."

janpio: This is probably kinda a nitpick: `it will ignore indexes that it doesn't know about.` You can only ignore things that exist. The indexes do not exist in the schema, so Prisma migrate can not ignore them.

matthewmueller:
_(Going through old PRs and closing them)_ 

If this is still relevant, feel free to re-open.

PR #2297, opened 2021-09-08
update: lens tailwind integration

luanvdw:
I closed the previous [PR](https://github.com/prisma/docs/pull/2260) associated with the goal of adding the new Navbar in Lens to the project, as it became bloated with a lot hackery and workarounds.

This PR (at the time of creation) contains the base code needed to add the `<Navbar />`;

1. It requires the main provider from /lens, `<LensProvider />` to ensure that the `<Navbar />` component has what it needs to render correctly.
2. The `<Navbar />` itself, replacing the WebsiteHeader

Notes:

The generated tailwind css that is incorporated through the `<LensProvder />` contains these styles associated with tailwind,

```
@tailwind base;
@tailwind components;
@tailwind utilities;
```

This is recommended and needed for a [tailwind installation](https://tailwindcss.com/docs/installation) but this project, the `docs` makes use of default HTML element styles in some cases i.e. 

The `h4` tags in the below section (OVERVIEW, COMPONENTS, DATABASE CONNECTORS, MORE) use default margins;

```
margin-top: 1.33em;
margin-bottom: 1.33em;
```

section

but tailwind contains a [Preflight](https://tailwindcss.com/docs/preflight) Preflight that removes all of the default margins from elements like headings, blockquotes, paragraphs, etc. This is by design for tailwind as they mention [here](https://tailwindcss.com/docs/preflight#:~:text=The%20reason%20for%20this%20is%20two-fold%3A);

We can add our base styles in our tailwind configuration in Lens which will correct this, but this will likely be a timeous proceeder to

1. Make sure we add the desired base styles to match our design system
2. Confirm that all default cases in docs have been covered
3. Confirm and correct the implication this will have on other projects that also use lens such as `cloud` and `homepage`

Matching the preview link up against the live docs as well the visual differences picked up by percy indicates that there are a lot of these occurrences (most of which seems Layout / positioning / spacing orientated)


tanberry:
Hi @luanvdw and @nilubava is this PR still possibly valid, or should it be closed? (I sure don't want to close anything that enhances the beauty of the Docs page, haha.)

luanvdw:
Closing on account of outdated relevance.

PR #2298, opened 2021-09-08
docs: include missing 'as' in text

orimdominic:
Include a missed word `as` in documentation text at [Data modelling on the application level](https://www.prisma.io/docs/concepts/overview/what-is-prisma/data-modeling#data-modeling-on-the-application-level), the penultimate paragraph


molebox (approved):
very-good

PR #2306, opened 2021-09-09
fix: postgresql raw query ILIKE example

simone-sanfratello:
fix: postgresql raw query ILIKE example


molebox:
Thanks for the PR! @millsp can you confirm this is correct please 😄

millsp:
That should already fixed in my other PR :+1:

simone-sanfratello:
thanks for the review! feel free to close if it's not needed :)

tanberry:
Yep, looks like this was already fixed, but thanks @simone-sanfratello!

PR #2313, opened 2021-09-13
fix: added temp Navbar

luanvdw:
The [PR](https://github.com/prisma/docs/pull/2297) to make use of Lens components, that were built with Tailwind, requires a larger discussion and planning for a successful roll out.

This PR serves as a temporary solution to have the new <Navbar /> component within the Docs project.


src/components/navbar/MenuList.tsx
```diff
@@ -0,0 +1,118 @@
...
+        svg {
+          margin-right: 0.5rem;
+          path {
+            fill: ${theme.colors.gray800};
+          }
+        }
+      }
+      &:focus {
+        outline: none;
+      }
```
molebox: Do we want to add some additional styling if we are removing the outline on focus? Keyboard users wont see any difference now

src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,128 @@
...
+const MobileMenuBtn = styled.button`
+  background-color: transparent;
+  display: flex;
+  padding: 0;
+  margin: 0;
+  border: 0;
+  &:focus,
+  &:active {
+    outline: none;
+  }
```
molebox: Do we want to add some additional styling if we are removing the outline on focus? Keyboard users wont see any difference now

molebox (commented):
Looks good! Just left a couple of comments

nilubava (approved):
since this is temporary, looks good to me :)

src/components/navbar/MenuList.tsx
luanvdw: 100%, re-added the default focus ring for this solution.

src/components/navbar/Navbar.tsx
luanvdw: 100%, re-added the default focus ring for this solution.

PR #2314, opened 2021-09-13
Filter json doc

molebox:
This PR adds missing info indicating that the filterJson feature is preview.


janpio (dismissed):
(Using "Request changes" to block accidental merge)
Does not sound right to me. Client team will double check.

tanberry:
Hi @janpio when you have a sec, can you share what you think we should do with this? Perhaps close it.. and if we still need to make any changes about implementing `filterJson` we can add in separate PR? Or... is that feature now GA, by some lucky chance? :-)

janpio:
It is not GA yet unfortunately: https://github.com/prisma/prisma/issues/7135

The current state of the PR looks correct to me - but I might not have the full picture. Best ask in #prisma-client so the team can let you know if this addition is indeed in the right place.

PR #2319, opened 2021-09-13
Add correct import for DeepMockProxy

DalyanParker:
I believe the import here is outdated! Was trying to follow along and found the import to be from /cjs/Mock!


molebox:
Hey @DalyanParker what version of jest mock extended are you on?

PR #2323, opened 2021-09-14
Fix reference to the model name

michalbundyra:
In this context we are selecting posts with authors.


tanberry:
HI @ruheni , can you give this one a quick technical review, please, and then let's merge it (whoever gets to it first)? Thanks @ruheni and also thanks to @michalbundyra for creating this PR!

ruheni (approved):

PR #2324, opened 2021-09-15
Fix seed ts compile options - Add tab section for mac and windows examples

elsangedy:


content/300-guides/050-database/300-seed-database.mdx
```diff
@@ -156,7 +156,7 @@ Here we suggest some specific seed scripts for different situations. You are fre
 
    ```ts
    "prisma": {
-     "seed": "ts-node --compiler-options \"{\\\"module\\\":\\\"commonjs\\\"}\" prisma/seed.ts"
+     "seed": "ts-node --compiler-options {\"module\":\"commonjs\"} prisma/seed.ts"
```
janpio, line 159: I am not confident this would work on Windows out of the box.

elsangedy: The current documentation doesn't work for macos.
Maybe we can put these two options? (for windows, for unix)

janpio: The current documentation was chosen as it was supposed to work on all platforms. If that is not the case any more, we need to find an option that does.

Is there an issue for this not working?

Jolg42:
Closing because duplicate, done with https://github.com/prisma/docs/pull/2338 https://prisma-company.slack.com/archives/CGMCB82N8/p1631898183074400

PR #2346, opened 2021-09-20
Fix query example for deleting all data with raw SQL

cyruskorn:
## Describe this PR

The example is using `$queryRaw` with a parameter of type `string`, even though the function only accepts a parameter of type `TemplateStringsArray | Sql`. This is incorrect and the execution of this code will not work.

## Changes

Replaced `$queryRaw` with `$queryRawUnsafe` that accepts a `string` as a parameter.


Jolg42 (approved):
Thanks! Looks like it has already been fixed though (exactly like you suggested), so I'll close this PR

PR #2347, opened 2021-09-20
Remove false/outdated note

tsongas:
The note states "Note: generate is called under the hood by default, after running prisma migrate dev. If the prisma-client-js generator is defined in your schema, this will check if @prisma/client is installed and install it if it's missing." however when I ran prisma migrate dev, @prisma/client did not get installed, furthermore installing @prisma/client is actually the next step in the tutorial so this note seems outdated?

## Describe this PR

Note is bunk...running prisma migrate dev doesn't install @prisma/client even though it's defined in the schema provided.

## Changes

Deleted note.

## What issue does this fix?

Removes false/outdated information.

Please let me know if I'm wrong here!


janpio:
If `@prisma/client` is not installed, that is a bug. It is absolutely intended behavior that `generate` does install `@prisma/client` in your project if it is not a dependency there yet. Can you please open an issue in https://github.com/prisma/prisma about this? Thanks.

PR #2349, opened 2021-09-20
clarify wording on sql injection risk

jtlindsey:
clarify wording on sql injection risk

## Describe this PR

clarify wording on sql injection risk to make it clear that the risk is present when used with user input.


janpio (approved):
Generally looks good to me. 

Maybe it makes sense to include the function name as well? Although that would make the example more copy-paste-able - which is the opposite of what we want.

Jolg42 (approved):
Looks good, thanks!

tanberry:
Thanks @jtlindsey !!

PR #2354, opened 2021-09-22
fix: changes wrong field name in the increment example

AvantaR:
## Describe this PR

It fixes wrong field name in setting value to posts.

## Changes

It changes `count` to `views` field to be consistent with code example below.

## What issue does this fix?

There is no issue for that.


pantharshit00:
Thanks @AvantaR

PR #2356, opened 2021-09-22
Remove unnecessary info

Warix3:
## Describe this PR

I believe this part is not necessary because it's the same thing said before and it's also missing the command so it's probably a mistake

## Changes

Removed part of docs

## What issue does this fix?

Duplicate info

It's also missing the command it's referencing


Jolg42 (approved):
Thanks!

PR #2358, opened 2021-09-23
Add missing field to documentation

moderndayNeo:
Line 409 references an "assignedBy" field that doesn't exist on the model shown.
This change on line 399 adds the assignedBy field to CategoriesOnPosts, as well as the necessary userId field.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
@@ -396,6 +396,8 @@ model CategoriesOnPosts {
   category   Category @relation(fields: [categoryId], references: [id])
   categoryId Int
   assignedAt DateTime @default(now())
+  assignedBy User     @relation(fields: [userId], references: [id])
+  userId     Int
 
   @@id([postId, categoryId])
 }
@@ -404,7 +406,7 @@ model CategoriesOnPosts {
 In this example, the `assignedAt` field stores additional information about the relation between `Post` and `Category`:
 
 - `assignedAt` stores _when_ the post was added to the category
-- `assignedBy` store _who_ added the post to the category
+- `assignedBy` stores _who_ added the post to the category
 
 Note that the same rules as for [1-n-relations](one-to-many-relations) apply (because `Post`↔ `CategoriesOnPosts` and `Category` ↔ `CategoriesOnPosts` are both in fact 1-n-relations), which means one side of the relation needs to be annotated with the `@relation` attribute.
 


tanberry:
@keerlu will you take a look at this, with your `CategoriesOnPosts` knowledge?

keerlu:
Thanks for spotting this @moderndayNeo, and sorry for the late review. I see that the `assignedBy` field has since been [added as part of another PR](https://github.com/prisma/docs/commit/02ee7e38894aabaf457a79a4663469155f4f0ce2#diff-fb79953d91185a909aee582b8bd3c8810a53237a30d3c57b5a99e41b89fac6da), but with a different type - I expect we should update to yours but would need a technical review first. Maybe @janpio? (let me know if this should go to someone else...)

janpio (changes requested):
That is a difficult question unfortunately: 
- `assignedBy` indeed sounds like a potential relation, which is what @moderndayNeo made it by adding the relation field to `User`. 
- But there _is_ no `User` model in this example (and the "underlying SQL" mentioned below), so what you did @moderndayNeo unfortunately would be broken.

janpio:
@keerlu:

A better solution probably would be to come up with some _other_ additional field that would make more sense on the relation that does not require an additional model by being a relation. A suggestion: Replace the `assignedBy String` with a `visible Boolean`. Then it is clear this is just a 0/1 field to define if this should be visible in the UI or not. 

(At the same the `assignedAt` could maybe be renamed to `createAt` to simplify that part as well)

So conclusion: The current suggestion is more broken than the current state. We should either leave it as it is, or improve the example in general.

tanberry:
@janpio I'd like to suggest that this gets converted to an Issue in a Prisma code repo... how to handle the current or proposed fields is not a decision the docs can make. ;-(

janpio:
Hm, I think you misundersodd slightly: 
This is not about Prisma itself, but about the example that is used here in the docs. Currently the text does not match the code example used. This PR introduces some changes, that are unfortunately not correct and would not work. In my comment I pointed out way to potentially improve the example being used here to both fix the text and make it easier to understand what is what and why.

(Happy to chat about this if this additional information does not help to see what could be done here)

keerlu:
@janpio sorry I overlooked your second comment (even though you tagged me...) 

When I get time I'll take a look and see if I can update the example

content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
```diff
@@ -396,6 +396,8 @@ model CategoriesOnPosts {
   category   Category @relation(fields: [categoryId], references: [id])
   categoryId Int
   assignedAt DateTime @default(now())
+  assignedBy User     @relation(fields: [userId], references: [id])
+  userId     Int
```
keerlu, line 400:
```suggestion
  createdAt DateTime @default(now())
  visible   Boolean
  userId    Int
```

Suggesting this change to avoid references to a `User` model that doesn't exist - see @janpio's comments above

content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
```diff
@@ -404,7 +406,7 @@ model CategoriesOnPosts {
 In this example, the `assignedAt` field stores additional information about the relation between `Post` and `Category`:
 
 - `assignedAt` stores _when_ the post was added to the category
```
keerlu, line 408: @janpio sorry, I've just realised I'd need to update this explanation too - if we switch to `createdAt`, should I remove this?

janpio: Yep I think so. Might be easier to do in an unrelated PR to this one, and then close this when it is merged.

moderndayNeo:
@janpio  @keerlu @tanberry  Thanks for taking a look at this. @janpio as you've pointed out, it does make sense to remove the reference to a `User` relation that doesn't exist.

This is my first open-source contribution, how should I proceed? (Mark as done/close PR etc.)

janpio:
I think @keerlu should suggest a way forward here. Thanks so much @moderndayNeo for initially opening this PR and getting our attention to this problem - this will be much better after it is done.

keerlu:
@janpio I'd be happy to make an updated version of the example in a new PR, in which case I'll close this one.

Thanks very much for this contribution @moderndayNeo and sorry the process has been a bit muddled/slow on my side.

moderndayNeo:
@janpio @keerlu  not a problem! And thanks for working on such a great tool. If either of you have LinkedIn, I'd love to connect. Mine is https://www.linkedin.com/in/adamzdrzalka/

PR #2370, opened 2021-09-29
fix: Remove Admonition code block from search results

### LINKED ISSUE #2369: Exclude the “Admonition” block from search preview

nilubava:
## What issue does this fix?

Fixes #2369

### DIFF
--- a/src/utils/algolia.js
+++ b/src/utils/algolia.js
@@ -53,7 +53,9 @@ const handleRawBody = (node) => {
         !section.includes('cmdResult>') &&
         !section.includes('<Subsections') &&
         !section.includes('cmd>') &&
-        !section.includes('<!-- prettier-ignore -->')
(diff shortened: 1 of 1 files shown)


src/utils/algolia.js
```diff
@@ -53,7 +53,9 @@ const handleRawBody = (node) => {
         !section.includes('cmdResult>') &&
         !section.includes('<Subsections') &&
         !section.includes('cmd>') &&
-        !section.includes('<!-- prettier-ignore -->')
+        !section.includes('<!-- prettier-ignore -->') &&
+        !section.includes('<Admonition type="warning">') &&
+        !section.includes('</Admonition>')
```
janpio, line 58: This only removes these explicit lines, not the content of the whole block, correct @nilubava?

PR #2376, opened 2021-10-01
Typo fix in 07-mongodb.mdx

wh1337:
## Describe this PR

Fixing a type in the mongodb docs. 

re: 

```
In order to use the mongodb provider,
you need to set the mongodb feature flag.
You can define the feature flag like this:

generator client {
    provider = "prisma-client-js"
    previewFeatures = ["mongodb"]
  }
```
which is printed out from the console when using `mongoDb`.

## Changes

- changed `mongoDb` to `mongodb`

## What issue does this fix?

Typo


Jolg42:
@cryptic-ai So the official preview flag is actually `mongoDb` and the message was updated recently to match that here https://github.com/prisma/prisma/pull/9459

So not a typo in the end but thanks for the PR 👍🏼

PR #2394, opened 2021-10-12
fix: integration testing about migrate

Jolg42:
`migrate dev` should not be used in integration testing, `migrate deploy` is the correct command to use

Current docs at https://www.prisma.io/docs/guides/testing/integration-testing#running-the-tests

### DIFF
--- a/content/300-guides/150-testing/150-integration-testing.mdx
+++ b/content/300-guides/150-testing/150-integration-testing.mdx
@@ -462,17 +462,16 @@ You can add some scripts to your projects `package.json` file which will setup t
 
 ```json file=package.json
   "scripts": {
-    "migrate:init": "npx prisma migrate dev --name init",
     "docker:up": "docker-compose up -d",
     "docker:down": "docker-compose down",
-    "test": "yarn docker:up && yarn migrate:init && jest -i && yarn docker:down"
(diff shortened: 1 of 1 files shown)


content/300-guides/150-testing/150-integration-testing.mdx
```diff
@@ -462,17 +462,16 @@ You can add some scripts to your projects `package.json` file which will setup t
 
 ```json file=package.json
   "scripts": {
-    "migrate:init": "npx prisma migrate dev --name init",
     "docker:up": "docker-compose up -d",
     "docker:down": "docker-compose down",
-    "test": "yarn docker:up && yarn migrate:init && jest -i && yarn docker:down"
+    "test": "yarn docker:up && npx prisma migrate deploy && jest -i && yarn docker:down"
```
tomhoule: I'm not familiar enough to say, but does it make sense to use `npx` and `yarn` together?

tomhoule (approved):
Good catch!

content/300-guides/150-testing/150-integration-testing.mdx
Jolg42: Indeed yarn should be enough if the CLI is installed locally

PR #2396, opened 2021-10-13
Link to Firebase/GCP Cloud Functions examples

goleary:
In lieu of proper deployment guides for GCP/Firebase cloud functions, these e2e test readmes are super useful and should be much easier to find.


Jolg42 (commented):
Thanks for the links but I don't think we want to link to e2e-tests repo which might be confusing for users.

Instead it would be better to have clean examples in this repo https://github.com/prisma/prisma-examples/tree/latest/deployment-platforms and then link there

janpio:
As @Jolg42 said, the e2e tests unfortunately are not official and super maintained recommendations - although they are indeed better than having nothing :/ Can you open an issue here so creating these real examples and pages here can be prioritized? Thanks.

goleary:
It's arguably more confusing for users having no example...but I get the point. Full deployment guides would certainly be preferable to linking to these e2e tests.

I opened #2400 to track.

I opened a single issue because although they need to be distinct guides, they are similar enough that they could be complete in one shot.

PR #2399, opened 2021-10-15
Update 03-postgresql.mdx – Add a application_name parameter

jnnkky:
## Describe this PR

Adding `application_name` as a followup to https://github.com/prisma/prisma/issues/7804

## Changes

- Added `application_name` parameter to docs

## What issue does this fix?

Follow up to https://github.com/prisma/prisma/issues/7804

@pimeys

### DIFF
--- a/content/200-concepts/200-database-connectors/03-postgresql.mdx
+++ b/content/200-concepts/200-database-connectors/03-postgresql.mdx
@@ -84,7 +84,7 @@ The following arguments can be used:
 | `pgbouncer`        | No       | `false`                | Configure the Engine to [enable PgBouncer compatibility mode](/guides/performance-and-optimization/connection-management/configure-pg-bouncer/)                                   |
 | `application_name` | No       |                        | Since 3.3.0: Specifies a value for the application_name configuration parameter 
 
-As an example, if you want to connect to a schema called `myschema`, set the connection pool size to `5` and configure a timeout for queries of `3` seconds, you can use the following arguments:
+As an example, if you want to connect to a schema called `myschema`, set the connection pool size to `5` and configure a timeout for queries of `3` seconds. You can use the following arguments:
 
 ```
 postgresql://USER:PASSWORD@HOST:PORT/DATABASE?schema=myschema&connection_limit=5&socket_timeout=3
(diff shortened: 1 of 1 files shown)


Jolg42:
Preview at https://deploy-preview-2399--prisma2-docs.netlify.app/docs/concepts/database-connectors/postgresql#arguments

Looks good, maybe it could mention that it's only available since 3.3.0

To be merged after the release.

content/200-concepts/200-database-connectors/03-postgresql.mdx
```diff
@@ -82,6 +82,7 @@ The following arguments can be used:
 | `host`             | No       |                        | Points to a directory that contains a socket to be used for the connection                                                                                                        |
 | `socket_timeout`   | No       |                        | Maximum number of seconds to wait until a single query terminates                                                                                                                 |
 | `pgbouncer`        | No       | `false`                | Configure the Engine to [enable PgBouncer compatibility mode](/guides/performance-and-optimization/connection-management/configure-pg-bouncer/)                                   |
+| `application_name` | No       |                        | Specifies a value for the application_name configuration parameter
```
Jolg42:
```suggestion
| `application_name` | No       |                        | Since 3.3.0: Specifies a value for the application_name configuration parameter 
```

jnnkky: Fair enough 👍

content/200-concepts/200-database-connectors/03-postgresql.mdx
```diff
@@ -82,6 +82,7 @@ The following arguments can be used:
 | `host`             | No       |                        | Points to a directory that contains a socket to be used for the connection                                                                                                        |
 | `socket_timeout`   | No       |                        | Maximum number of seconds to wait until a single query terminates                                                                                                                 |
 | `pgbouncer`        | No       | `false`                | Configure the Engine to [enable PgBouncer compatibility mode](/guides/performance-and-optimization/connection-management/configure-pg-bouncer/)                                   |
+| `application_name` | No       |                        | Specifies a value for the application_name configuration parameter 
 
 As an example, if you want to connect to a schema called `myschema`, set the connection pool size to `5` and configure a timeout for queries of `3` seconds, you can use the following arguments:
```
tanberry: Let's add a period (.) after "...seconds", and make a new sentence of the last phrase. 

So " ...seconds. You can use the following arguments:"

janpio: I don't think this change is correct. Everything in this sentence belongs together and forms on unit:

> As an example, [...], you can use the following arguments: [...]

(Not sure about the commas though)

tanberry: Ah, so the "if you want to" phrase relates to three things: 1. connect to `myschema`, 2. set the connection pool size, and 3) configure the time out. Got it, then we need to make more clear by something like this:

_As an example, use the following arguments to connect to a schema called `myschema`, set the connection pool size to `5`, and then configure a timeout for queries of `3` seconds:_

The original way was quite hard to parse, imo, because the phrase "set the connection pool size..." was ambiguous, unclear if it was the imperative to do that task, or one of the goals. ;-)

janpio: Yes, the new suggestions sounds correct to me.

PR #2401, opened 2021-10-18
Add missing `name` key in @relation

KoltonG:
## Describe this PR

While reading the docs, I noticed that the `@relation` field was missing the `name` key just like the "PinnedPost" relation at the bottom of this page: https://www.prisma.io/docs/concepts/components/prisma-schema/relations#types-of-relations

CleanShot 2021-10-18 at 11 00 59@2x

## Changes

- Add `name` key to `@relation` field

## What issue does this fix?

No issue, just a typo.

None


KoltonG:
Actually, will close since its mentions here https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#relation that `name` can be omitted.

PR #2427, opened 2021-10-24

sagarsneh:
text update


tanberry:
Thank you so much @sagarsneh, we really appreciate your catching (and fixing!) this.

PR #2430, opened 2021-10-25
Update 03-upgrading-the-prisma-layer.mdx

hoogendjk:
change to remove prisma1 package and not the prisma package

## Describe this PR
changes to upgrade documentation

## Changes
change to remove prisma1 package instead of prisma

## What issue does this fix?
No issue available

### DIFF
--- a/content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/03-upgrading-the-prisma-layer.mdx
+++ b/content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/03-upgrading-the-prisma-layer.mdx
@@ -63,10 +63,10 @@ If you're upgrading your entire project [all at once](./how-to-upgrade#upgrade-s
 
 ```terminal
 # remove global installation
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):
Looks good! Thanks!

janpio:
This unfortunately only applies to users that upgrade to the `prisma1` package, not to the ones still using 1.x versions of `prisma`.

Jolg42:
@janpio the page mentions `Note that the former prisma command for Prisma 1 has been renamed to prisma1. You can learn more about this here.` so it was confusing.

janpio:
I do not remember what we actually did, but wasn't that "just" about the CLI command `prisma` and not the Npm package?

Jolg42:
Indeed command and package of Prisma 1 were renamed.
- package `prisma` -> `prisma1` https://github.com/prisma/prisma1
- command `prisma` -> `prisma1` https://github.com/prisma/prisma1/blob/master/cli/packages/prisma-cli/package.json#L23

PR #2432, opened 2021-10-25
Update 020-use-custom-model-and-field-names.mdx

alexmarqs:
Minor typo - change field name in an example

## Changes

- Fixed example to use correct field name for user <-> profile relation when explaining `@map`


Jolg42:
Thanks for the PR, I came across your Pr while working on that page and did many fixes in https://github.com/prisma/docs/pull/2447
I'll close this one as I think other changes were needed.

PR #2435, opened 2021-10-25
Update 300-seed-database.mdx

omar-dulaimi:
## Describe this PR

Include the `prisma db push` command to the list of automatic seeding.

## Changes

Added the push command with other commands that trigger auto seeding.

PR #2443, opened 2021-10-27
fix: remove mention of seeding done by db push

Jolg42:
Originally found by a user https://github.com/prisma/docs/pull/2435


tanberry:
@Jolg42 these additional tips and clarifications look great to me. I did want to double-check; they do not offer an opinion (one way or the other) on additional ways to auto-trigger seeding, correct?

tanberry (approved):
Looks good, thanks so much for the additional tips.

Jolg42:
The text about `db push` was wrong, so there was confusion about when `db seed` gets triggered automatically

PR #2445, opened 2021-10-28
fix: unique-constraints-and-indexes/01-postgresql

Jolg42:
Live page https://www.prisma.io/docs/guides/general-guides/database-workflows/unique-constraints-and-indexes/postgresql

The "" in the SQL makes the sql valid on TablePlus


Jolg42:
From @do4gr 
> The name -> map change is correct.
For the introspection warning, that is caused by the columns with the unique constraints being nullable so adding additional primary keys is a valid fix. Alternatively we could make the columns Not Null to stay with fewer columns / fields.

PR #2447, opened 2021-10-29
fix guide schemas @@map @map

Jolg42:
Stumbled upon this from https://github.com/prisma/docs/pull/2432 and auto-completion for named constraints work in language-tools


content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/020-use-custom-model-and-field-names.mdx
```diff
@@ -124,109 +128,101 @@ Foreign keys are represented as a combination of a [annotated relation fields](/
...
-
-// Fluent API
-const userByProfile = await prisma.profiles
-  .findUnique({
-    where: { id: 1 },
-  })
-  .user_id();
-```
-
-In both cases, `user_id` is used to refer to an _entire_ "user object", not only to the `user_id` column. -->
```
Jolg42, line 185: Tried and this does not happen, it was commented out, so I removed it.

do4gr (approved):

PR #2452, opened 2021-10-29
mistake: table name missed

vladyslav-n:
UPDATE table SET biography = bio; => UPDATE "Profile" SET biography = bio;


tanberry:
Thanks so much @vladyslav-n ! We will get a technical review and then merge or get back with you.

tomhoule (approved):
Thanks for the fix!

PR #2462, opened 2021-11-01
Update 055-aggregation-grouping-summarizing.mdx

tanberry:
This PR removes the Admonition that Mongo is not supported... it's in prep for release 3.4.0, where we add MongoDB support for using ORDER BY clause with Group aggregation.
Let's merge right after the 3.4.0 release is announced.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/055-aggregation-grouping-summarizing.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/055-aggregation-grouping-summarizing.mdx
@@ -252,7 +252,9 @@ The following constraints apply when you combine `groupBy` and `orderBy`:
 
 #### Order by aggregate group
 
-You can **order by aggregate group**. The following example sorts each `city` group by the number of users in that group (largest group first):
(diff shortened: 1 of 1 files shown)


keerlu:
Netlify preview looks fine to me but something odd is happening when I view the Percy preview. Checking this in the #docs channel before approving

keerlu:
Checked the Percy preview issue, happy to approve this for merge later

content/200-concepts/100-components/02-prisma-client/055-aggregation-grouping-summarizing.mdx
```diff
@@ -252,16 +252,7 @@ The following constraints apply when you combine `groupBy` and `orderBy`:
 
 #### Order by aggregate group
 
-In [2.21.0](https://github.com/prisma/prisma/releases/2.21.0) and later, you can **order by aggregate group**.
```
janpio, line 255: This should probably not be removed, as for users on older versions before 2.21.0 it is relevant to know that this will not work for them. Possibly we could even change this to include that with 3.4.0 this also works for MongoDB.

tanberry: Thanks @Jan, I will add back in the version #s, then going forward let's think on how we want to handle this.

PR #2467, opened 2021-11-03
Add warning about main env file

bennycode:
## Describe this PR

I have had a `.env` and a `.env.test` file and when executing `dotenv -e .env.test -- npx prisma migrate dev` then Prisma was sill prefering the `.env` file over my `.env.test` file. That's why I suggest to highlight that the main `.env` file has to be removed.

## Changes

I added an admonition.

## What issue does this fix?

Spending time on finding it out yourself.

No.

### DIFF
--- a/content/300-guides/125-development-environment/050-environment-variables/200-using-multiple-env-files.mdx
+++ b/content/300-guides/125-development-environment/050-environment-variables/200-using-multiple-env-files.mdx
@@ -20,6 +20,12 @@ One solution is to have multiple `.env` files which each represent different env
 
 </Admonition>
 
+<Admonition type="warning">
+
+When using multiple `.env` files you have to remove your main `.env` file because otherwise it will still be read.
+  
+</Admonition>
+
 Then using a package like [`dotenv-cli`](https://www.npmjs.com/package/dotenv-cli), you can load the correct connection URL for the environment you are working in.
 
 </TopBlock>


janpio:
Hmmm, interesting. I do not think we were aware of this behavior before. Can you open a bug issue in prisma/prisma so we can investigate if we can do anything about it, or at least add a warning to make users aware? Not everyone will see the note in the docs.

tanberry:
Thanks @bennycode for this important Note. Let's move it down to the bottom of this section though, along with the other admonition there. Reason for moving it is because the directions do say to *rename* the default/main `.env` file, which implies that you get rid of the .env file.... so let's add your note after they have read the directions. We could also emphasize the rename part of the directions....

bennycode:
After checking again, I noticed that the root `.env` file can actually stay in the project. It won't be used when running `dotenv -e .env.test` BUT Prisma doesn't tell you that. It actually tells you that it reads from `.env` but it actually reads from `.env.test`. This was causing confusion on my end which is why I created now this ticket: https://github.com/prisma/prisma/issues/10104 (as @janpio suggested). I will close my doc change since the problem sits in the Prisma CLI. Thank you all.

janpio:
Thanks for the update. I will continue the discussion in your new issue - it is unfortunately a bit more tricky :/

PR #2468, opened 2021-11-03
Edit troubleshooting-relations doc

tanberry:
Added a section about the problem of enforced primary key DBs not working with implicit syntax, for m-n relations.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
@@ -268,4 +268,4 @@ Some cloud providers enforce (by default) the existence of primary keys in all t
 
 ### Solution
 
-If your database provider automatically creates a primary key, and you want to use m-n relations, you need to use explicit relation syntax, manually create the join model, and verify that this join model has a primary key. 
+You need to use explicit relation syntax, manually create the join model, and verify that this join model has a primary key. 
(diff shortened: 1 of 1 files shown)


keerlu:
One thing I noticed from the preview is that the documentation uses a mix of 'many-to-many' and 'm-n' terminology, which is a little confusing to me. It looks like 'many-to-many' is being used at the point the concept is introduced in the documentation, e.g. https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations, and for the majority of subheadings, so that may be the best option.

Also unsure of capitalisation for 'Join table' - the linked doc has 'JOIN table' so might want to be consistent.

Otherwise looks good to me

janpio:
In Prisma context the Join table is usually called "relation table" (as it enables relations to exist and be expressed via `@relation`).

content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
```diff
@@ -259,3 +259,13 @@ model Category {
...
+
+## Using m-n relations in databases with enforced primary keys
+
+### Problem
+
+Some cloud providers enforce (by default) the existence of primary keys in all tables. However, any Join tables created by Prisma for m-n relations using implicit syntax in Prisma Schema do not have primary keys. 
+
+### Solution
+
+If your database provider automatically creates a primary key, and you want to use m-n relations, you need to use explicit relation syntax, manually create the join model, and verify that this join model has a primary key.
```
albertoperdomo: Same comment as in the other PR. https://github.com/prisma/docs/pull/2469#pullrequestreview-797898457

```suggestion
You need to use explicit relation syntax, manually create the join model, and verify that this join model has a primary key. 
```

albertoperdomo: I think we should consider adding a link to the explicit relation syntax, here, perhaps? Unless it's really easy to navigate to (scroll up or down to a section that is close).

albertoperdomo (changes requested):

janpio (approved):

Issue #958, opened 2020-10-29
Document limitation of missing primary keys in join tables when using implicit relations and corresponding workaround


albertoperdomo:
Join tables created by Prisma through implicit relation syntax do not have primary keys.

Databases can be configured to enforce the existence of primary keys in all tables. This is true for some cloud providers like DigitalOcean.

We should document this limitation and suggest an available workaround, which is using the explicit relation syntax, creating the join model and making sure this join model has a primary key.

Context: https://github.com/prisma/migrate/issues/596#issuecomment-701487236

janpio:
This limitation came up on Slack again: https://prisma.slack.com/archives/CA491RJH0/p1625592140192700

pavelruzicka:
I have just recently run into this issue again using a managed database provided by DigitalOcean. Obviously the workaround is clear in switching from implicit to explicit relation syntax, however the docs are still yet to mention this limitation and I had to come across these two issue discussion in order to realize what the issue was.

tanberry:
Notes for working this Issue: 
--docs that likely need to be modified include:
1. https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations#rules-for-defining-an-implicit-m-n-relation
2. https://www.prisma.io/docs/concepts/components/prisma-schema/relations/troubleshooting-relations#how-to-use-a-relation-table-with-a-many-to-many-relationship (DONE, PR #2468)
3. https://www.prisma.io/docs/guides/migrate-to-prisma/migrate-from-typeorm#implicit-many-to-many-relations (Done, PR #2469)
4. https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations/#conventions-for-relation-tables-in-implicit-m-n-relations

tanberry:
Hi @pavelruzicka and @albertoperdomo. I created [PR #2468](https://github.com/prisma/docs/pull/2468) to add info about this issue to the troubleshooting doc; feedback/tweaks welcome! I will add the same content to other areas in the docs, so look for more PRs on this.

mhessdev:
Closing because PR from @tanberry

PR #2469, opened 2021-11-03
Update 01-migrate-from-typeorm.mdx

tanberry:
Added warning at the bottom of this topic, in section about implicit syntax is m-n relations, about the enforced primary key issue. See [Issue #958](https://github.com/prisma/docs/issues/958).

### DIFF
--- a/content/300-guides/400-migrate-to-prisma/01-migrate-from-typeorm.mdx
+++ b/content/300-guides/400-migrate-to-prisma/01-migrate-from-typeorm.mdx
@@ -1081,6 +1081,6 @@ This would also result in a more ergonomic and less verbose Prisma Client API to
 
 <Admonition type="warning">
   
-Some providers like DigitalOcean enforce primary keys on all tables for replication. In such a case, the user would have to switch to the explicit syntax, and manually create the join model with a primary key. This is because Join tables created by Prisma for many-to-many relations using implicit syntax in Prisma Schema do not have primary keys. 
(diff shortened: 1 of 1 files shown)


keerlu:
Wondering if this admonition should be earlier in the document (e.g. before 'Here is an example with TypeORM') so that people reading this subsection definitely see it

janpio:
My 2c: `database provider automatically creates a primary key` is pretty uncommon, so an edge case.

content/300-guides/400-migrate-to-prisma/01-migrate-from-typeorm.mdx
```diff
@@ -1078,3 +1078,9 @@ model Post {
 ```
 
 This would also result in a more ergonomic and less verbose Prisma Client API to modify the records in this relation, because you have a direct path from `Post` to `Category` (and the other way around) instead of needing to traverse the `PostToCategories` model first.
+
+<Admonition type="warning">
+  
+If your database provider automatically creates a primary key, and you want to use m-n relations, you need to use explicit relation syntax, manually create the join model, and verify that this join model has a primary key. This is because Join tables created by Prisma for m-n relations using implicit syntax in Prisma Schema do not have primary keys.
```
albertoperdomo: I may be wrong but I think 

>If your database provider automatically creates a primary key,

should be

>If your database provider requires tables to have primary keys,

Some providers like DigitalOcean enforce primary keys on all tables for replication. In such a case, the user would have to switch to the explicit syntax.

albertoperdomo (changes requested):

janpio (approved):

PR #2478, opened 2021-11-05
Make join table names consistent in 300-many-to-many-relations.mdx

keerlu:
## Describe this PR

I noticed there was a mix of using `CategoriesOnPosts`, `CategoryOnPost` and `CategoryToPost` for the example join table. I've updated these to be consistent - though could maybe do with a technical review to check I've picked the right one!

## Changes

Renamed join table to `CategoryToPost`throughout.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
@@ -29,16 +29,16 @@ In this example, the model representing the relation table **defines additional 
 model Post {
   id         Int                 @id @default(autoincrement())
   title      String
-  categories CategoryToPost[]
+  categories CategoriesOnPosts[]
 }
 
 model Category {
   id    Int                 @id @default(autoincrement())
   name  String
-  posts CategoryToPost[]
+  posts CategoriesOnPosts[]
 }
 
-model CategoryToPost {
+model CategoriesOnPosts {
   post       Post     @relation(fields: [postId], references: [id])
   postId     Int // relation scalar field (used in the `@relation` attribute above)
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
```diff
@@ -29,16 +29,16 @@ In this example, the model representing the relation table **defines additional
 model Post {
   id         Int                 @id @default(autoincrement())
   title      String
-  categories CategoriesOnPosts[]
+  categories CategoryToPost[]
```
janpio: Line 24 also includes a mention of the old term.

janpio:
Wow, that is quite a "bug" in the current documentation here. None of this would work of course with these 3 terms that do not match. Great catch! I wonder how many users were already confused by that.

A kinda technical consideration here: 
- `_CategoryToPost` would be the name of an implicit join table created by Prisma for an implicit relation. 
- This documentation is about the explicit m:n relation, so using a pretty similar term like `CategoryToPost` might make it harder to differentiate. `CategoriesOnPosts` on the other hand is less "technical" and more descriptive instead.
- So _maybe_ (and that is very much up for discussion), it might be better to instead unify on the `CategoriesOnPosts` terminology here.

tanberry:
Wow, great catch indeed, @keerlu ! Lucy if you are good with @janpio 's suggestion of standardizing on  'CategoriesOnPosts` do you want to open a separate PR (or Issue if you need to research) for that work?

tanberry (approved):

keerlu:
Thanks, I agree it's good to distinguish the explicit relation table name from the generated implicit one. I've updated the table name to be `CategoriesOnPosts`.

PR #2486, opened 2021-11-07
fix: some ineffective finally clauses

LumaKernel:
## Changes

Code snippets for short living scripts have ineffective finally chain.

```ts
(async()=>{ console.log(1); throw 1 })()
  .catch(() => { console.log(2); process.exit(1) })  // or re-throw.
  .finally(() => { console.log(3) }) // unreachable
```

Finally clause with unsafe catch clause has no effect as _finally_. (It's run when no error, but if it's desired, `.then` is enough.)

I chosen the style below.

```ts
  .then(async () => {
    await prisma.$disconnect()
  })
  .catch(async (e) => {
    console.error(e)
    await prisma.$disconnect()
    process.exit(1)
  })
```

### DIFF
--- a/content/100-getting-started/01-quickstart.mdx
+++ b/content/100-getting-started/01-quickstart.mdx
@@ -76,11 +76,13 @@ async function main() {
 }
 
 main()
-  .catch((e) => {
-    throw e
-  })
-  .finally(async () => {
+  .then(async () => {
     await prisma.$disconnect()
+  })
+  .catch(async (e) => {
+    console.error(e)
+    await prisma.$disconnect()
+    process.exit(1)
   })
 ```
 
--- a/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/250-querying-the-database.mdx
+++ b/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/250-querying-the-database.mdx
@@ -25,11 +25,13 @@ async function main() {
 }
 
 main()
-  .catch((e) => {
-    throw e
-  })
-  .finally(async () => {
+  .then(async () => {
     await prisma.$disconnect()
+  })
+  .catch(async (e) => {
+    console.error(e)
+    await prisma.$disconnect()
+    process.exit(1)
   })
 ```
 
@@ -49,11 +51,13 @@ async function main() {
 }
 
 main()
-  .catch((e) => {
-    throw e
-  })
-  .finally(async () => {
+  .then(async () => {
     await prisma.$disconnect()
+  })
+  .catch(async (e) => {
+    console.error(e)
+    await prisma.$disconnect()
+    process.exit(1)
   })
 ```
 
--- a/content/100-getting-started/02-setup-prisma/100-start-from-scratch/120-mongodb/250-querying-the-database.mdx
+++ b/content/100-getting-started/02-setup-prisma/100-start-from-scratch/120-mongodb/250-querying-the-database.mdx
@@ -27,11 +27,13 @@ async function main() {
 }
 
 main()
-  .catch((e) => {
-    throw e
-  })
-  .finally(async () => {
+  .then(async () => {
     await prisma.$disconnect()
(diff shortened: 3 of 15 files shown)


tanberry:
Thanks @LumaKernel ! We will get a tech review done on this and then hopefully merge soon. We appreciate your contribution!

matthewmueller:
Thanks @LumaKernel, I agree with your style.

I'm tempted to suggest:

```ts
main()
  .catch(err => console.error(err))
  .finally(() => prisma.$disconnect())
```

Since it's simpler. But, there's no easy way to apply a non-zero exit code.

So, your changes LGTM!

janpio:
We could use this opportunity to normalize with https://github.com/prisma/prisma-examples/blob/latest/javascript/script/script.js#L93-L100

LumaKernel:
> We could use this opportunity to normalize with https://github.com/prisma/prisma-examples/blob/latest/javascript/script/script.js#L93-L100

@janpio As I mentioned in the description, its `finally` usage has no effect as `finally` because `catch` will end the process. It's the same as following, and if `main()` throws, `prisma.$disconnect()` wouldn't be run.

```ts
main()
  .then(async () => {
    await prisma.$disconnect()
  })
  .catch((e) => {
    console.error(e)
    process.exit(1)
  })
```

LumaKernel:
@matthewmueller 
I agree with you. No simple default with taking care of exit code. I think such scripts can be reader's important parts like deployments that should be failed with exit code. We can add a note with your suggestion like 'pay attention to unhandled promises. If you are using Node.js v14 or later, you can explicitly use `--unhandled-rejections=throw` for `node`(`ts-node`) or use `process.exit(1)`'

janpio:
> @janpio As I mentioned in the description

Yes, exactly why making sure these two use the same pattern makes sense if the current state doesn't work.

millsp:
Could we move `await prisma.$disconnect()` inside of the `main` function and have a single catch clause. This makes it somewhat easier to read and understand:
```ts
// all of that happens on success
async function main() {
	  // ... prisma stuff
	  
	  await prisma.$disconnect()
}

// all of this happens on error
main().catch(async (e) => {
	  console.error(e)
	  await prisma.$disconnect()
	  process.exit(1)
})
```

janpio:
We would probably want a pattern where we do _not_ expect users to manually $disconnect on success - that should not be necessary at all.

LumaKernel:
I suspect whether it is needed on failure too. I think any short living scripts are obviously not _long running_ and it is against the case described in https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-management#calling-disconnect-explicitly

LumaKernel:
Hi. My last comment is acceptable? If so, I'll add a commit.

tanberry:
Hi @LumaKernel yes, please do make one more commit to wrap up the decisions/convo, and let's get this merged!

tanberry:
@janpio and @LumaKernel Do we want to proceed with that final requested commit, or close this one?

janpio:
This also influences our examples which have a similar code snippet: https://github.com/prisma/prisma-examples/blob/65d21a56d1d8dd928066ecbeb83b611e4a542d9f/javascript/script/script.js#L93-L100 If we want to change something here, we should/need also update our examples over there to match and make sense.

millsp:
If you need help, please stick this PR into the Client team inbox (via "Projects").

janpio:
(I could not figure out how to do that even when using the exact project board name @millsp , `⚡️ Client Team` 🤷 )

matthewmueller:
Fixed in prisma-examples: https://github.com/prisma/prisma-examples/pull/3905

matthewmueller:
Resolved conflicts and merged PR in https://github.com/prisma/docs/pull/3510. Thanks @LumaKernel!

andrew-walford-prisma:
@tanberry @keerlu  This issue is still on our board. It looks like it might be finished. Can we archive it?

PR #2492, opened 2021-11-09
Make clearer that SQL Server arguments are alternatives

janpio:
by making them an unordered list instead of just putting each term in a new line

(Not a great solution to the problem though 🤔 )


tanberry:
Thanks @janpio I think the bullets do help make it clear that each is a separate option.

tanberry (approved):
Do you want to merge it @janpio ?

janpio:
Sure. (Generally I am fine with maintainers of a repo to press the button themselves on PRs that are up for review - so you can go ahead next time if you agree. I try to keep not ready PRs on `Draft`)

PR #2497, opened 2021-11-11
Fixes to Non-Canonicalized Internal Links

mzkrasner:
These are absolute URL fixes to issues we are encountering with non-canonical internal links due to relative link conventions.
We are contracted for SEO work with Prisma - we typically work with Etel, Alex, Justin, and Daniel.

### DIFF
--- a/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/150-using-prisma-migrate.mdx
+++ b/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/150-using-prisma-migrate.mdx
@@ -312,7 +312,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./connect-your-database-node-postgres"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-node-postgres"
   arrowLeft
 >
   Connect your database
@@ -321,7 +321,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./install-prisma-client-node-postgres"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/install-prisma-client-node-postgres"
   arrow
 >
   Install Prisma Client
@@ -338,7 +338,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./connect-your-database-node-mysql"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-node-mysql"
   arrowRight
 >
   Connect your database
@@ -347,7 +347,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./install-prisma-client-node-mysql"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/install-prisma-client-node-mysql"
   arrow
 >
   Install Prisma Client
@@ -364,7 +364,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./connect-your-database-node-sqlserver"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-node-sqlserver"
   arrowRight
 >
   Connect your database
@@ -373,7 +373,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./install-prisma-client-node-sqlserver"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/install-prisma-client-node-sqlserver"
   arrow
 >
   Install Prisma Client
@@ -390,7 +390,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./connect-your-database-typescript-postgres"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-typescript-postgres"
   arrowLeft
 >
   Connect your database
@@ -399,7 +399,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./install-prisma-client-typescript-postgres"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/install-prisma-client-typescript-postgres"
   arrow
 >
   Install Prisma Client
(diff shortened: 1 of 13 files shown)


janpio (changes requested):
Prisma Docs try not to use absolute internal links but let the tooling create the appropriate links. 

What problem is being solved here?

mzkrasner:
Hi @janpio,

Resulting from our routine technical audits & site crawls, these particular internal links are, for some reason, defaulting to the non-canonicalized URL conventions. This is a serious issue for SEO as it confuses search engines as to what the true canonical version of a page is.

Additionally, just as a rule of best practice, absolute links are always recommended for SEO.

janpio:
Thanks @mkrasne2 for the additional information.

Can you please list the actual problems? What does "non-canonicalized URL conventions" mean? We definitely want to fix any problems with links that exist.

But we do not want to inconsistently change random link targets in the source of our documentation as this really makes the writer and development experience worse.

(I see a few fixes with trailing slashes that you modify, that make definite sense to fix - but also many other changes that do not to me)

> Additionally, just as a rule of best practice, absolute links are always recommended for SEO.

I strongly disagree with this statement. Internal links are perfectly fine as relative links.

content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/150-using-prisma-migrate.mdx
```diff
@@ -312,7 +312,7 @@ END CATCH
 <ButtonLink
   color="dark"
   type="primary"
-  href="./connect-your-database-node-postgres"
+  href="https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-node-postgres"
   arrowLeft
 >
   Connect your database
```
Jolg42, line 318: Related to @janpio, I checked the page, here for example https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases-typescript-postgres

Then let's find the button where the link is changed by the PR at the bottom and it's `href` value is
`/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-typescript-postgres/`
Which is indeed relative and could be made absolute by our tooling.


I also see the non-canonicalized problem where the href value on current website has an extra ending `/` at the end and the canonical doesn't.
Canonical value:
`https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases-typescript-postgres`

So 2 different problems where both would probably need some tooling investigation/fix from us?

janpio: The relative link is no problem, and the trailing slash is easy to remove if that is what the problem is about. We just need to update our `ButtonLink` component probably.

janpio: Based on the problem description, I opened issue https://github.com/prisma/docs/issues/2500 so this can properly be fixed. All the changes to the `ButtonLink` component usage would then be not necessary.

content/200-concepts/100-components/01-prisma-schema/index.mdx
```diff
@@ -8,8 +8,8 @@ metaDescription: 'The Prisma schema is the main configuration file when using Pr
 
 The Prisma schema file (short: _schema file_, _Prisma schema_ or _schema_) is the main configuration file for your Prisma setup. It is typically called `schema.prisma` and consists of the following parts:
 
-- [**Data sources**](data-sources): Specify the details of the data sources Prisma should connect to (e.g. a PostgreSQL database)
-- [**Generators**](generators): Specifies what clients should be generated based on the data model (e.g. Prisma Client)
+- [**Data sources**](https://www.prisma.io/docs/concepts/components/prisma-schema/data-sources): Specify the details of the data sources Prisma should connect to (e.g. a PostgreSQL database)
+- [**Generators**](https://www.prisma.io/docs/concepts/components/prisma-schema/generators): Specifies what clients should be generated based on the data model (e.g. Prisma Client)
```
janpio, line 12: Here I can see the difference between the links that are changed, and the ones that are not is that line 13 is using the proper "fully qualified" link that we want to use in the docs vs. lines 11 and 12 do not yet.

If the link in line 13 is fine right now, the fix here (and probably for most other suggested changes) could be:

```suggestion
- [**Data sources**](/concepts/components/prisma-schema/data-sources): Specify the details of the data sources Prisma should connect to (e.g. a PostgreSQL database)
- [**Generators**](/concepts/components/prisma-schema/generators): Specifies what clients should be generated based on the data model (e.g. Prisma Client)
```

janpio:
I had a quick look at the suggested changes:
- As commented the ButtonLink seems to have a bug that I reported in https://github.com/prisma/docs/issues/2500 so it can be fixed
- The removal of trailing slashes in links with `#` look valid and correct to me (although many of these could be turned into internal links
- Most other changes seem to be links that do not follow the pattern of using fully qualified internal links (basically: Absolute link minus domain and `/docs`). That are oversights of a previous maintainers and also just bugs to be fixed one by one.

mzkrasner:
Hi Jan,

Based on the above, it sounds like moving forward when we make these changes and all out that an internal link is using a non canonicalized version, should we be replacing with "fully qualified" internal links? It sounds like this then would fix our issue, and remove the need to use absolute links?

Here's an example: if you navigate to this page: https://www.prisma.io/docs/concepts/components/prisma-client/crud and find the anchor text "Working with relations > Nested writes" you will see that it is linking to a non-canonicalized page (https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries/#nested-writes) whereas the correct version is https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#nested-writes

For that link, I see that this is being used as the rel internal: relation-queries#nested-writes

To fix, are you suggesting that we should simply change to /concepts/components/prisma-client/relation-queries#nested-writes instead of the absolute?

Looping in @imchairmanm as well since I'm assuming this would apply to /dataguide too

janpio:
Yes, fundamentally that is the kind of link we usually want to have and that should be rendered correctly without the trailing slash.

(The example you chose is special in that it also has an anchor added to the link (`#nested-writes`), we _might_ have a bug on that logic that always adds a trailing slash for example. But we would figure that out and then fix this independently)

mzkrasner:
Thank you.

Many of the issues discovered in our crawl contain anchors (we call those deep links) - should we wait until we hear more from you regarding whether there's a global bug related to anchor logic before beginning to correct those (in the event that for those, we might have to use absolute links)?

janpio:
We would prefer to identify the bug and fix it. A simple way to see would be to change one of the absolute links here in your PR to use a fully qualified relative one, and then check the preview deployment if the link presents correctly in code. (I can also take some time tomorrow and do that if you do not have the capacity)

mzkrasner:
Hi Jan,

Again, really appreciate the time you're taking to work with us on this. 

The approach you outlined above sounds good. We can move forward from now on with testing in the preview deployment to see if the fully qualified rel link approach works. In cases where it doesn't, we can flag that in our comments.

janpio:
I split out one class of errors (the links with anchors, deep links as you all them) into a separate PR quickly: https://github.com/prisma/docs/pull/2502 Reviews and confirmations the problems are resolved in the Netlify preview very welcome.

> In cases where it doesn't, we can flag that in our comments.

That sounds great. Feel free to @ mention me as well so I def see it.

janpio:
I think we fixed all of these independently, so can close this one.

PR #2501, opened 2021-11-12
fixing broken command

JacobGrady:


thebiglabasky (commented):
Hi and thanks for the proposal. It is though correct to put the environment variable prior to the command you're calling so the environment variable change is scoped to that command execution only.
We'll thus keep the docs as is.

PR #2502, opened 2021-11-12
fix(seo): Adjust internal links with anchors not to use trailing slashes

janpio:
This PR takes one pattern of wrong links from https://github.com/prisma/docs/pull/2497 that @mkrasne2 created and fixes them: Internal links with an anchor. 

They currently all use a trailing slash, but as the canonical version of the the pages do not use slashes, this requires an unnecessary additional redirect - If we link to the correct version directly instead that is not required.


janpio:
TODO: We should compare the before and after on all these pages, to make sure the links are actually fixed and our internal logic that writes the links does not still break them.

tanberry:
@janpio I'm happy to help test this when you're ready!

janpio:
That would be amazing. 

Unfortunately we have no output of the changed pages via CI, so you will have to go from the filenames. Comparison I was talking about was to look at the existing link target deployed to prisma.io/docs (that would probably include a trailing slash at the end of the URL, before the anchor) and confirm that this is not the case any more with preview deployment in the PR here.

tanberry:
Thanks @janpio  for doing this work. I tested many links(old and new) and the PR looks great.

janpio:
cc @mkrasne2 - this should already fix some of the wrong internal links

mzkrasner:
@janpio - awesome, thanks for looping me in here. I just did another pull of all internal links to non-canonicalized pages, and it looks like we're down about 563 from where we started, which is great to see. Now it looks like we have a little over 8,900 remaining to solve. I'll post an updated crawl in the other chat between us.

PR #2506, opened 2021-11-14
Change wrong filename types.ts to context.ts

### LINKED ISSUE #1342: Bug report: Migration guide to Nexus is missing the src/context.ts file
[The migration guide from prisma-binding to Nexus](https://www.prisma.io/docs/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-prisma-binding-to-nexus) is referencing the `src/context.ts`, but that file is never created during the process. It looks like the context is created in `src/types.ts` along with the type though. But I think the createContext should also receive and return the `request` (?).

```
//index.ts
import { createContext } from './context'
//....
 typegenAutoConfig: {
    contextType: 'Context.Context',
    sources: [
      {
        source: '@prisma/client',
        alias: 'prisma',
      },
      {
        source: require.resolve('./context'),
        alias: 'Context',
      },
    ],
  },
//....
new GraphQLServer({ schema, context: createContext() }).start(() =>
  console.log(`Server ready at: http://localhost:4000`)
)
```

chemicalkosek:
There are references to `types.ts` but the actual file should be called `context.ts`

## What issue does this fix?
Fixes https://github.com/prisma/docs/issues/1342


nikolasburk (approved):
Thanks for catching this 🎉

nikolasburk:
Thank you @chemicalkosek 🙏

PR #2518, opened 2021-11-16
add prisma-tql-types-gen

YassinEldeeb:
## Describe this PR

Added a new community generator

## Changes

- Added prisma-typegraphql-types-generator package to community generators section with a brief description on how it works.

## What issue does this fix?

- Allowing for having a single source of truth when working with prisma and typegraphql where developers can model their data in `prisma.schema` file and typegraphql type classes and enums will be generated and registered with TypegraphQL.
- The generated output is very human readable and doesn't look like generated code what so ever.
- The generated output can be edited so developers can add fields to the generated models that are processed by the field resolvers in the graphql api.
- When developers mess up the types by editing the generated output the generator will correct them.


janpio (approved):

tanberry:
Thank you @YassinEldeeb !

PR #2519, opened 2021-11-16
Cursor-based pagination requires orderBy

ahmedosama7450:


tanberry:
Thanks so much @ahmedosama7450 for your docs contribution! It does look like the example for paging backwards was missing the `orderBy` clause; thanks for adding it. 

I'm now wondering if we should call out explicitly that `orderBy` is required with pagination?

ahmedosama7450:
I think you already mention that within cons [here](https://www.prisma.io/docs/concepts/components/prisma-client/pagination#%E2%9C%98-cons-of-cursor-based-pagination)

janpio (approved):

PR #2522, opened 2021-11-16
Add `tocDepth` to the frontmatter.

tanberry:
Each .md file in our docs has (or should have) a frontmatter section to define metadata and control behaviour of docs once published, etc. The new Platform document was missing the `tocDepth` flag so the right-side ToC was not auto-expanding.


janpio (approved):
If this is what we want on this page:
image

millsp:
Yes, it looks awesome now!

millsp (approved):

PR #2526, opened 2021-11-17
Trailing underscore on "Configuring the name of the relation table in implicit many-to-many relations"

ImNahim:
## Describe this PR

Writing @relation("_MyRelationTable") with 1 underscore results in the creation of a table named "__MyRelationTable" (with 2 underscores), while writing @relation("MyRelationTable") without underscore creates a table named "_MyRelationTable" with 1 underscore.

## Changes

Removed the mention "The only requirement is that it starts with an underscore.", removed the underscore from @relation("_MyRelationTable")

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
@@ -353,17 +353,17 @@ CREATE TABLE "Post" (
 
 #### Configuring the name of the relation table in implicit many-to-many relations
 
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks @ImNahim for your help here, good catch! The text was a bit confusing as it implied that one must manually enter the underscore as part of the name, when in reality it is auto-added.

I do think we might want to leave in the fact that an underscore is required...but add a clarifying sentence to explain that we add it automatically. Thoughts? @tomhoule can you tech check this, pls?

janpio (approved):
I think what @ImNahim made of it is correct, but we should get another technical review before we merge. Might be mistaken.

tomhoule (approved):
yep I think that's accurate

tanberry:
Thank you @ImNahim !!

PR #2528, opened 2021-11-17
Add clarity around minimum field size for cuid()

amcjen:
No real documentation existed outside of old Github issues on the minimum character width needed for `cuid()` fields. Hopefully this prevents others from getting bitten like us.

## Describe this PR

Lack of clarity around the cuid implementation led us to believe that all cuids are 25 characters long. If you use this number in your Prisma schema when creating your tables (i.e. `@id @default(cuid()) @db.Char(25)`), then you'll run into an issue when, sometimes, cuid() output can be 26 or 27 characters long.  You'll then see sporadic errors in the logs about the field not being long enough for the generated ID.

## Changes

One liner in the Remarks section of the docs under cuids!

## What issue does this fix?

None


amcjen:
@all-contributors please add @amcjen for doc

janpio:
@amcjen This could also be something our tooling validates and warns about. Can you maybe open a feature request about this over at prisma/prisma? Thanks.

amcjen:
You got it. I'll do that now.

amcjen:
Posted here: https://github.com/prisma/prisma/issues/10332

tanberry:
Thanks so much @amcjen ! We'll get a quick tech review and then merge; your help is very appreciated.

tomhoule:
Looks good to me :+1: I don't know if 30 is the right safe value, but apparently neither does the author of the cuid library, so that sounds reasonable.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2701,6 +2701,7 @@ Generate a globally unique identifier based on the [`cuid`](https://github.com/e
 
 - Compatible with `String`
 - Implemented by Prisma and therefore not "visible" in the underlying database schema. You can still use `cuid()` when using [introspection](/concepts/components/introspection) by [manually changing your Prisma schema](/concepts/components/prisma-client/working-with-prismaclient/use-custom-model-and-field-names) and [generating Prisma Client](/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client), in that case the values will be generated by Prisma's [query engine](/concepts/components/prisma-engines/query-engine).
+- Since the length of `cuid()` output is [undefined per the cuid creator](https://github.com/ericelliott/cuid/issues/164), a safe minimum field size is 30 characters, otherwise you may see random Prisma errors like `Error: The provided value for the column is too long for the column's type.`
```
floelhoeffel: should this be "a safe **maximum** field" as the error message talks about "too long"?

janpio: stumbled over that as well, but: the field has to be minimum 30 characters to be able to store the values that could potentially be created. What happened is that @amcjen had only 25 defined in the database, and hence got that error message when the generated ID was _too long_. 

Maybe this could be clarified by not mentioning the error message, and just mention what to do and why.

amcjen: Ah yeah, totally fine if you want to reword it. I only included the error message because searching for that is how I found the original old closed Github issues that addressed it in the first place. I leave it up to you all though, as your documentation is so well-written already.

amcjen:
> Looks good to me 👍 I don't know if 30 is the right safe value, but apparently neither does the author of the cuid library, so that sounds reasonable.

Same. I think the issue comes down to implementation differences between platforms--Rust specifically seems to return 25, 26, or 27, based on how many you generate at a given time, but apparently others like Javascript always return 25.  (I came across this issue when running automated Cypress tests, and apparently enough cuids were generated that it triggered a 26 length cuid in at least one case.)

30 seems to be consensus from what I could find in the Prisma Github issue history.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
tanberry: Proposed rewrite; please review and make sure it is still technically accurate:
"Since the length of `cuid()` output is undefined per the cuid creator, a safe minimum field size is 30 characters, in order to allow for enough characters for very large values. If you set the minimum field size as less than 30, and then a large value is entered, you might see Prisma errors such as `Error: The provided value for the column is too long for the column's type.`

amcjen: Looks just about right. I'd reword just this one part myself:

"...If you set the minimum field size as less than 30, and then a larger value is generated by `cuid()`, you might see..."

(as when using `cuid()` in Prisma, we are never entering the values ourselves, they're auto-generated)

tanberry: Thanks so much for that important clarification, @amcjen. Do you want to make/commit that change please, then we can go ahead and merge this PR?

amcjen: Updated!

PR #2546, opened 2021-11-18
fix: Fix all the internal links on "Whats new in Prisma Docs"

janpio:
Make internal links internal.


tanberry:
Thank you @janpio !

PR #2553, opened 2021-11-20
docs: Fix date model name in Prisma Migrate table for MySql conector

procrastagonist:
## Describe this PR
I used `date Datetime` for my model in schema.prisma following to docs, run `npx prisma generate` and got issue - `Type "Datetime" is neither a built-in type, nor refers to another model, custom type, or enum.`

## Changes

correct use is `date DateTime`

@all-contributors please add @mkarkachov for doc


tanberry:
Thanks so much @mkarkachov for catching this! We had it correct (DateTime) in many other places in the same doc, but you caught this wrong one. Your help is much appreciated!

PR #2556, opened 2021-11-22
Add Koyeb as a deployment option

edouardb:
## Describe this PR

Hi there :wave:

Added a deployment guide for Koyeb.

## Changes

Add and reference the Koyeb deployment guide.


Jolg42:
Bonjour Edouard 👋🏼 

Note: it seems that the service is in preview and one can not signup instantly ⬇️ 
Screen Shot 2021-11-22 at 17 17 16

Could be that it's too early to add to our docs because of that and things could also change and add some maintenance of the docs?

edouardb:
Hi @Jolg42 

This should be fine, we’re at the end of the private preview phase.
The platform is actually already used in production and there are not going to be any breaking changes.

content/300-guides/200-deployment/110-deployment-guides/250-deploying-to-koyeb.mdx
```diff
@@ -0,0 +1,202 @@
...
+In this guide, you will set up and deploy a Node.js server that uses Prisma with PostgreSQL to [Koyeb](https://www.koyeb.com). The application exposes a REST API and uses Prisma Client to handle fetching, creating, and deleting records from a database.
+
+Koyeb is a developer-friendly serverless platform to deploy apps globally. The platform lets you seamlessly run Docker containers, web apps, and APIs with git-based deployment, TLS encryption, native autoscaling, a global edge network, and built-in service mesh & discovery.
+
+When using the [Koyeb git-driven deployment](https://www.koyeb.com/docs/apps/build-from-git) method, each time you push code changes to a GitHub repository a new build and deployment of the application are automatically triggered on the Koyeb Serverless Platform.
+This guide uses the latter approach whereby you push your code to the app's repository on GitHub.
+
+The application has the following components:
+
+- **Backend**: Node.js REST API built with Express.js with resource endpoints that use Prisma Client to handle database operations against a PostgreSQL database (e.g., hosted on PlanetScale).
```
Jolg42:
```suggestion
- **Backend**: Node.js REST API built with Express.js with resource endpoints that use Prisma Client to handle database operations against a PostgreSQL database (e.g., hosted on Heroku).
```

Here it kind of implies that PlanetScale is running on PostgreSQL but they are actually MySQL compatible, maybe this change is better here since Heroku is mentioned later for free db

tanberry:
Thanks @edouardb and @Jolg42 for your reviews/edits!

PR #2569, opened 2021-11-23
Update emulated referential actions

pimeys:
The emulation now has more supported actions.


keerlu:
This looks to be covered by more recent work on the referential actions docs. @pimeys please reopen and comment if you disagree!

PR #2578, opened 2021-11-23

PatrikValkovic:
## Describe this PR

Typo in doc. I believe you refer to two one-to-many relations.

## Changes

Fix typo in doc.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
@@ -224,7 +224,7 @@ model Category {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tanberry:
Hi @PatrikValkovic thanks for this... I think the whole section actually needs some work; it is odd to start the section with an example that shows the *incorrect* way to do things. We should start with the correct way to define an implicit many-to-many relationship... and if it helps, then below that add a following section on what might seem logical but is incorrect.

Do you think you have bandwidth to make those bigger changes? If not no worries, we can simply merge your typo correction now and then the Docs team will get to the bigger changes later.

PatrikValkovic:
Hi @tanberry. Unfortunately I don't have the capacity for these changes. I will work with Prisma for a while now, so I may catch some typos here and there, but in a near future I can't participate more than that.

tanberry:
Thank you @PatrikValkovic and apologies we are just now getting back to this!

PR #2579, opened 2021-11-23
Add a missing comma in the example JSON object

aqrln:
This PR adds a missing comma in the example JSON object that is used in the JSON filters documentation. I copied it from the page to the editor and noticed there was a syntax error.


tanberry:
Thank you @aqrln ! Great catch, and thanks for helping keep the docs awesome.

Issue #2538, opened 2021-11-18
Document referential emulation for MongoDB


keerlu:
Related tickets: 
https://github.com/prisma/prisma-engines/issues/2396

Docs to update:
- [Referential actions](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions)
  - update table of databases
  - explain what `NoAction` does
- [Referential integrity](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-integrity) 
  - update intro to explain drawbacks of emulation. 
  -  Make it clearer what's referential integrity in general, and what's emulation. 
  - Explain `foreignKeys` vs `prisma` better

keerlu:
This was done in #2581.

PR #2581, opened 2021-11-24
Update referential integrity and referential actions documentation

keerlu:
## Describe this PR

I've made initial changes to the referential integrity and referential actions docs based on conversations with @tanberry, @pimeys and @dpetrick yesterday. At the moment this is mainly getting up-to-date information in there - we still need to make the organisation of the pages as a whole better.

Comments welcome!

## Changes

- explained difference between emulating in client and handling in the database
- added info on MongoDB support
- some copyediting/formatting changes

## What issue does this fix?

WIP for issue #2538

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -28,7 +28,7 @@ model User {
 }
 ```
 
-In this case, the `onDelete: Cascade` argument specifies the `Cascade` action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
 
 When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -84,6 +84,8 @@ In the case where referential integrity is emulated in the Prisma client using t
 | Cascade            | no         | yes        |
 | NoAction           | no         | yes        |
 
+Currently Prisma only implements the referential actions. Foreign keys also create constraints, which make it impossible to manipulate data in a way that would violate these constraints: instead of executing the query, the database responds with an error. These constraints will not be created if you emulate referential integrity in the client, so if you set the referential action to [`NoAction`](/concepts/components/prisma-schema/relations/referential-actions#noaction) there will be no checks to prevent you from breaking the referential integrity.
+
 Another current limitation when emulating referential integrity in Prisma is that any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
 
 When using the `db pull` command, existing relations are kept in the data model. This has some caveats:
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -35,6 +35,13 @@ When connecting to relational databases that support foreign keys, referential i
 Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons.
 
 <Admonition type="warning">
+  Be aware that currently Prisma only implements the referential actions in this
+  mode, but not the constraints. If you set the referential actions to NoAction
+  there will be no checks to prevent you from breaking the referential
+  integrity.
+</Admonition>
+
+<Admonition type="warning">
   There are significant performance implications to emulating referential
   integrity in the Prisma client. In cases where the underlying database can
   handle referential integrity with foreign keys, it is usually the preferred
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -8,9 +8,9 @@ preview: true
 
 <TopBlock>
 
-To keep data consistent, it is necessary to have rules for what happens when related records are deleted or updated. For example, a `Post` model will need to define what happens if its author is deleted. **Referential integrity** refers to the set of constraints and triggers used to enforce this consistency.
+Referential integrity is a property of a set of data that states that all its references are valid. It requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you will need to define what happens if its author is deleted.
 
-In Prisma, relations between records are defined using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which **referential actions** take place on updating or deleting the related record.
+In Prisma, referential integrity is implemented by defining relations between records using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which referential action take place on updating or deleting the related record.
 
 For example, take the following blog post schema, with a one-to-many relation between `User`s and `Post`s:
 
@@ -32,13 +32,13 @@ In this case, the `onDelete: Cascade` argument specifies the `Cascade` referenti
 
 When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
 
-Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons.
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. It can also be selected for relational databases in cases where foreign keys are not supported, or where it preferred for other reasons (for example, in cases where there are scaling concerns).
 
 <Admonition type="warning">
-  Be aware that currently Prisma only implements the referential actions in this
-  mode, but not the constraints. If you set the referential actions to `NoAction`
-  there will be no checks to prevent you from breaking the referential
-  integrity.
+  Be aware that currently Prisma only implements the referential actions when
+  emulating referential integrity in the client, and not the constraints. If you
+  set the referential actions to `NoAction` there will be no checks to prevent
+  you from breaking the referential integrity.
 </Admonition>
 
 <Admonition type="warning">
@@ -67,21 +67,21 @@ generator js {
 }
 ```
 
-To change the type of referential integrity used, update the `referentialIntegrity` paramater in the [`datasource`](/reference/api-reference/prisma-schema-reference#datasource) block. The available options are:
+To change the type of referential integrity used, update the `referentialIntegrity` parameter in the [`datasource`](/reference/api-reference/prisma-schema-reference#datasource) block. The available options are:
 
 - `foreignKeys`: this handles referential integrity in the database using foreign keys. This is the default option when using all SQL database connectors.
 - `prisma`: this emulates referential integrity in the Prisma client. This is the default (and only available) option when using the MongoDB connector.
 
 <Admonition type="warning">
-  Setting the value to `prisma` will remove all previously created foreign keys
-  in the next migration when using Prisma Migrate. Setting the value to
-  `foreignKeys` will create foreign keys for all relations in the next migration
-  when using Prisma Migrate.
+  When using Prisma Migrate, setting the value to `prisma` will remove all
+  previously created foreign keys in the next migration. Setting the value to
+  `foreignKeys` will create foreign keys for all relations in the next
+  migration.
 </Admonition>
 
 ## Handling the referential integrity in Prisma
 
-If referential integrity is handled by the Prisma Client, there is currently a reduced number of supported [referential actions](/concepts/components/prisma-schema/relations/referential-actions) available in the data model:
+In the case where referential integrity is emulated in the Prisma client using the `prisma` option for the `referentialIntegrity` parameter, there is currently a reduced number of supported [referential actions](/concepts/components/prisma-schema/relations/referential-actions) available in the data model:
 
 | Referential action | `onUpdate` | `onDelete` |
 | :----------------- | :--------- | :--------- |
@@ -91,16 +91,16 @@ If referential integrity is handled by the Prisma Client, there is currently a r
 | Cascade            | no         | yes        |
 | NoAction           | no         | yes        |
 
-In this mode, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
+In this case, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
 
-The relations are needed for the client to be able to join data between models, so when using the `db pull` command to a database with `referentialIntegrity` set to `prisma` in the `datasource`, existing relations are kept in the data model. This has some caveats:
+When using the `db pull` command, existing relations are kept in the data model. This has some caveats:
 
 - When a model in the database is deleted, the relation on the other side gets deleted, if existing.
 - When a scalar field used in a relation field gets deleted or renamed, the relation field is kept as-is, leading to validation errors. The user must fix the relation before continuing to use the data model in this case.
 
 ## Handling the referential integrity with foreign keys
 
-Having the referential integrity handled in the database, all referential actions supported by the database are available:
+In the case where referential integrity is handled in the database using the `foreignKeys` option for the `referentialIntegrity` parameter, all referential actions supported by the database are available:
 
 | action      | `onUpdate` | `onDelete` |
 | :---------- | :--------- | :--------- |
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -95,8 +95,26 @@ In this case, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-
 
 When using the `db pull` command, existing relations are kept in the data model. This has some caveats:
 
-- When a model in the database is deleted, the relation on the other side gets deleted, if existing.
-- When a scalar field used in a relation field gets deleted or renamed, the relation field is kept as-is, leading to validation errors. The user must fix the relation before continuing to use the data model in this case.
+- If you delete a model that is part of a relation, Prisma will delete the relation.
+- If you delete or rename a scalar field used in a relation field, the relation field will still refer to the old version, leading to validation errors. For example, take the following data model:
+
+  ```prisma file=schema.prisma
+  model A {
+    field Int
+    b     B   @relation(fields: [field], references: [id])
+  }
+  ```
+
+  If you then change `field` to `renamedField`, you will end up with the following data model after then next `db pull`:
+
+  ```prisma file=schema.prisma
+  model A {
+    renamedField Int
+    b            B   @relation(fields: [field], references: [id])
+  }
+  ```
+
+  In this case, you need to manually fix the relation to point to `renamedField` instead before continuing to use the data model.
 
 ## Handling the referential integrity with foreign keys
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -28,7 +28,7 @@ model User {
 }
 ```
 
-In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the `Post` if the related `User` is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
 
 When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -85,11 +85,11 @@ If referential integrity is handled by the Prisma Client, there is currently a r
 
 | Referential action | `onUpdate` | `onDelete` |
 | :----------------- | :--------- | :--------- |
-| Restrict           | yes        | yes        |
-| SetNull            | yes        | yes        |
+| Restrict           | no         | yes        |
+| SetNull            | no         | yes        |
 | SetDefault         | no         | no         |
-| Cascade            | yes        | yes        |
-| NoAction           | yes        | yes        |
+| Cascade            | no         | yes        |
+| NoAction           | no         | yes        |
 
 In this mode, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -8,7 +8,7 @@ preview: true
 
 <TopBlock>
 
-Referential integrity is a property of a set of data that states that all its references are valid. It requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you must define what happens if its author is deleted.
+Referential integrity is a property of a data set that states that all its references are valid. Referential integrity requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you must define what happens if its author is deleted.
 
 In Prisma, referential integrity is implemented by defining relations between records using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which referential action take place on updating or deleting the related record.
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -36,7 +36,7 @@ Alternatively, Prisma can emulate referential integrity in the client. This is t
 
 <Admonition type="warning">
   Be aware that currently Prisma only implements the referential actions in this
-  mode, but not the constraints. If you set the referential actions to NoAction
+  mode, but not the constraints. If you set the referential actions to `NoAction`
   there will be no checks to prevent you from breaking the referential
   integrity.
 </Admonition>
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -30,16 +30,9 @@ model User {
 
 In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the `Post` if the related `User` is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
 
-When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
+When using a relational database that supports foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
 
 Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. Emulation can also be selected for relational databases in cases where foreign keys are not supported, or where it is preferred for other reasons (for example, in cases where there are scaling concerns).
-
-<Admonition type="warning">
-  Be aware that currently Prisma only implements the referential actions when
-  emulating referential integrity in the client, and not the constraints. If you
-  set the referential action to `NoAction` there will be no checks to prevent
-  you from breaking the referential integrity.
-</Admonition>
 
 <Admonition type="warning">
   There are significant performance implications to emulating referential
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -8,7 +8,7 @@ preview: true
 
 <TopBlock>
 
-Referential integrity is a property of a set of data that states that all its references are valid. It requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you will need to define what happens if its author is deleted.
+Referential integrity is a property of a set of data that states that all its references are valid. It requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you must define what happens if its author is deleted.
 
 In Prisma, referential integrity is implemented by defining relations between records using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which referential action take place on updating or deleting the related record.
 
@@ -32,12 +32,12 @@ In this case, the `onDelete: Cascade` argument specifies the `Cascade` referenti
 
 When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
 
-Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. It can also be selected for relational databases in cases where foreign keys are not supported, or where it preferred for other reasons (for example, in cases where there are scaling concerns).
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. Emulation can also be selected for relational databases in cases where foreign keys are not supported, or where it is preferred for other reasons (for example, in cases where there are scaling concerns).
 
 <Admonition type="warning">
   Be aware that currently Prisma only implements the referential actions when
   emulating referential integrity in the client, and not the constraints. If you
-  set the referential actions to `NoAction` there will be no checks to prevent
+  set the referential action to `NoAction` there will be no checks to prevent
   you from breaking the referential integrity.
 </Admonition>
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -12,7 +12,7 @@ Referential integrity is a property of a set of data that states that all its re
 
 In Prisma, referential integrity is implemented by defining relations between records using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which referential action take place on updating or deleting the related record.
 
-For example, take the following blog post schema, with a one-to-many relation between `User`s and `Post`s:
+For example, take the following blog post schema, with a one-to-many relation between `Post`s and `User`s:
 
 ```prisma file=schema.prisma highlight=4;normal
 model Post {
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -32,7 +32,7 @@ In this case, the `onDelete: Cascade` argument specifies the `Cascade` referenti
 
 When using a relational database that supports foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
 
-Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. Emulation can also be selected for relational databases in cases where foreign keys are not supported, or where it is preferred for other reasons (for example, in cases where there are scaling concerns).
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. Emulation can also be enabled for relational databases in cases where foreign keys are not supported, or where it is preferred for other reasons (for example, in cases where there are scaling concerns).
 
 <Admonition type="warning">
   There are significant performance implications to emulating referential
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -84,7 +84,7 @@ In the case where referential integrity is emulated in the Prisma client using t
 | Cascade            | no         | yes        |
 | NoAction           | no         | yes        |
 
-In this case, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
+Another current limitation when emulating referential integrity in Prisma is that any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
 
 When using the `db pull` command, existing relations are kept in the data model. This has some caveats:
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
@@ -8,21 +8,46 @@ preview: true
 
 <TopBlock>
 
-Referential integrity means a set of constraints or triggers to handle the consistency of data of a [`@relation`](/reference/api-reference/prisma-schema-reference#relation). It can mean preventing a deletion of a record referred from other records, or it can be a cascading action defined in the field's [`referential actions`](/concepts/components/prisma-schema/relations/referential-actions). 
+To keep data consistent, it is necessary to have rules for what happens when related records are deleted or updated. For example, a blog post data model will need to define what happens to a post if its author is deleted. **Referential integrity** refers to the set of constraints or triggers used to enforce this consistency.
 
-The referential integrity can either be handled in the database using foreign keys, or in the Prisma Client. There are pros and cons on both. In general, if the database supports foreign keys, it is usually the preferred choice and the default for SQL databases when using Prisma.
+In Prisma, relations between records are defined using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which **referential actions** take place on updating or deleting the related record.
 
-In certain cases, either when the database doesn't support foreign keys, or when they are not preferred due to scaling reasons, Prisma can disable them and handle the referential integrity in the client.
+For example, take the following blog post schema, with a one-to-many relation between `User`s and `Post`s:
+
+```prisma file=schema.prisma highlight=4;normal
+model Post {
+  id       Int    @id @default(autoincrement())
+  title    String
+  author   User   @relation(fields: [authorId], references: [id], onDelete: Cascade)
+  authorId Int
+}
+
+model User {
+  id    Int    @id @default(autoincrement())
+  posts Post[]
+}
+```
+
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
+
+When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
+
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons.
+
+<Admonition type="warning">
+  There are significant performance implications to emulating referential
+  integrity in the Prisma client. In cases where the underlying database can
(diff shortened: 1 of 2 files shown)


content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,46 @@ preview: true
 
 <TopBlock>
 
-Referential integrity means a set of constraints or triggers to handle the consistency of data of a [`@relation`](/reference/api-reference/prisma-schema-reference#relation). It can mean preventing a deletion of a record referred from other records, or it can be a cascading action defined in the field's [`referential actions`](/concepts/components/prisma-schema/relations/referential-actions). 
+To keep data consistent, it is necessary to have rules for what happens when related records are deleted or updated. For example, a blog post data model will need to define what happens to a post if its author is deleted. **Referential integrity** refers to the set of constraints or triggers used to enforce this consistency.
```
pimeys:
```suggestion
To keep data consistent, it is necessary to have rules for what happens when related records are deleted or updated. For example, a `Post` model will need to define what happens if its author is deleted. **Referential integrity** refers to the set of constraints and triggers used to enforce this consistency.
```

The integrity defines rules what happens in general if a user is deleted and the user has posts. It can be an error, or it can be a referential action to the child record.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,46 @@ preview: true
...
+  id    Int    @id @default(autoincrement())
+  posts Post[]
+}
+```
+
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
+
+When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
+
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons.
```
pimeys:
```suggestion
Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons. Be aware that currently Prisma only implements the **referential actions** in this mode, but not the **constraint**. If setting the referential actions to `NoAction`, this means nothing will stop you from breaking the referential integrity.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -144,6 +146,8 @@ CREATE TABLE "User" (
 </tab>
 </TabbedContent>
 
+The schema does not explicitly define referential actions on the mandatory `author` field, which means that the default referential actions of `Restrict` for `onDelete` and `Cascade` for `onUpdate` apply. See the SQL tab for details of how this is implemented.
```
pimeys: I think even mentioning SQL here is kind of outside of the scope already. A MongoDB user will be reading this documentation as well, and these examples will just confuse the reader if they're not using an SQL database.

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -166,15 +170,21 @@ Note that [`NoAction`](#noaction) can be used in replace of [`Restrict`](#restri
...
-| SQLServer  | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
+| SQL Server | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
 
 - † On [InnoDB](https://dev.mysql.com/doc/refman/5.6/en/innodb-introduction.html#:~:text=InnoDB%20is%20a%20general%2Dpurpose,clause%20creates%20an%20InnoDB%20table.), `SetDefault` will be ignored and silently use `NoAction`.
 
+From version 3.6.0, referential actions will also be available in MongoDB if the `[referentialIntegrity](/concepts/components/prisma-schema/relations/referential-integrity)` Preview feature is enabled:
+
+| Database | Cascade | Restrict | NoAction | SetNull | SetDefault |
+| :------- | :------ | :------- | :------- | :------ | :--------- |
+| MongoDB  | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
```
pimeys: `SetDefault` is not working yet, and most of the `onUpdate` actions are still unstable, but coming.

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -234,22 +244,23 @@ model User {
...
 - **MySQL**: `NoAction` behaves exactly the same as `Restrict`. See [the MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-key-referential-actions) for more information.
 - **SQLite**: When a related primary key is modified or deleted, no action is taken. See [the SQLite docs](https://www.sqlite.org/foreignkeys.html#fk_actions) for more information.
-- **SQLServer**: When a referenced record is deleted or modified, an error is raised. See [the SQLServer docs](https://docs.microsoft.com/en-us/sql/relational-databases/tables/graph-edge-constraints?view=sql-server-ver15#on-delete-referential-actions-on-edge-constraints) for more information.
+- **SQL Server**: When a referenced record is deleted or modified, an error is raised. See [the SQL Server docs](https://docs.microsoft.com/en-us/sql/relational-databases/tables/graph-edge-constraints?view=sql-server-ver15#on-delete-referential-actions-on-edge-constraints) for more information.
+- **MongoDB** (in preview from version 3.6.0): When a record is modified or deleted, nothing is done to any related records.
```
pimeys, line 241: This is actually a question should we even do it like this. The `NoAction` for mongo means we have no referential integrity at all defined for the relation.

pimeys (dismissed):
Ok, so we need to be extremely clear with what these terms mean. First let's talk about the referential integrity: It's a promise the database gives to the user, that we do not break relationships when modifying our data. The database uses a few tools to implement this, one of them is the foreign key constraint that stops you from modifying things that might break relations, and the other is the referential action that is an addition to the constraint that can trigger additional queries in the database transaction before the actual item gets deleted.

How Prisma's Query Core implements the referential integrity is through a set of referential actions. In our case, using `NoAction` means we break the referential integrity, and the user can modify things that break relations. We should be clear on this and maybe communicate with the client team to see do we pursue with only the actions, or should we also implement the constraint part.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
janpio: This should probably be an admonition or something special, that will also be easy to remove if/when this is fixed.

keerlu: Added this as another admonition.

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: I've removed the SQL tab.

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Crossed out SetDefault, and I've also put the `onUpdate` actions back to 'no' for now on the referential integrity page

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Is there anything you'd suggest instead? A better way of phrasing it, or removing it completely?

pimeys: I was more like thinking is it ok we don't have any integrity here. As a feature, sorry, maybe not the right place for that discussion :D

keerlu: Thanks, got it!

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -34,6 +34,13 @@ When connecting to relational databases that support foreign keys, referential i
 
 Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons.
 
+<Admonition type="warning">
+  Be aware that currently Prisma only implements the referential actions in this
+  mode, but not the constraints. If you set the referential actions to NoAction
```
janpio:
```suggestion
  mode, but not the constraints. If you set the referential actions to `NoAction`
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
 
 <TopBlock>
 
-Referential integrity means a set of constraints or triggers to handle the consistency of data of a [`@relation`](/reference/api-reference/prisma-schema-reference#relation). It can mean preventing a deletion of a record referred from other records, or it can be a cascading action defined in the field's [`referential actions`](/concepts/components/prisma-schema/relations/referential-actions). 
+To keep data consistent, it is necessary to have rules for what happens when related records are deleted or updated. For example, a `Post` model will need to define what happens if its author is deleted. **Referential integrity** refers to the set of constraints and triggers used to enforce this consistency.
```
janpio: While the last sentence is kinda technically correct, I am a bit weirded out that this looks like a definition. And I would define "ref integrity" more like Wikipedia does it: https://en.wikipedia.org/wiki/Referential_integrity (ALso meaning: Contraints and triggers are tools that can help enforce referential integrity - but generally it is an abstract concept instead of a function or tool)

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
 
-The referential integrity can either be handled in the database using foreign keys, or in the Prisma Client. There are pros and cons on both. In general, if the database supports foreign keys, it is usually the preferred choice and the default for SQL databases when using Prisma.
+In Prisma, relations between records are defined using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which **referential actions** take place on updating or deleting the related record.
```
janpio: Why is this relevant here? I think this could use some connection back, that we talk about this here because relations are what ref integrity is about. THis is not explicitly spelled out in the first paragraph before this one.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+  authorId Int
+}
+
+model User {
+  id    Int    @id @default(autoincrement())
+  posts Post[]
+}
+```
+
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
```
janpio:
```suggestion
In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+
+When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
+
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option for MongoDB, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to scaling reasons.
```
janpio:
```suggestion
Alternatively, Prisma can emulate referential integrity in Prisma Client. This is the default option when using MongoDB as the database, which does not support foreign keys, and can also be selected for relational databases in cases where foreign keys are not supported, or not preferred due to any reasons, for example scaling concerns.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+
+<Admonition type="warning">
+  Be aware that currently Prisma only implements the referential actions in this
+  mode, but not the constraints. If you set the referential actions to `NoAction`
```
janpio: What does "this mode" refer to here?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+  there will be no checks to prevent you from breaking the referential
+  integrity.
```
janpio: Does this really belong in this document or in the "ref actions" one where we describe the emulated ref actions?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
 }
 ```
 
-Setting the value to `prisma` will remove all foreign keys previously created in the next migration when using Prisma Migrate and the referential integrity will be handled by the client. The value `foreignKeys` creates foreign keys for all relations in the next migration when using Prisma Migrate, letting the database handle the referential integrity.
+To change the type of referential integrity supported, update the `referentialIntegrity` paramater in the [`datasource`](/reference/api-reference/prisma-schema-reference#datasource) block. The available options are:
```
janpio:
```suggestion
To change the type of referential integrity used, update the `referentialIntegrity` paramater in the [`datasource`](/reference/api-reference/prisma-schema-reference#datasource) block. The available options are:
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
...
+
+- `foreignKeys`: this handles referential integrity in the database using foreign keys. This is the default option when using all SQL database connectors.
+- `prisma`: this emulates referential integrity in the Prisma client. This is the default (and only available) option when using the MongoDB connector.
+
+<Admonition type="warning">
```
janpio, line 68: Do we do headlines in Admonitions? If so, having something like "Impact for users of Prisma Migrate" could be helpful here to make it more obvious who should read this.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
...
-| NoAction   | yes        | yes        |
+| Referential action | `onUpdate` | `onDelete` |
+| :----------------- | :--------- | :--------- |
+| Restrict           | no         | yes        |
+| SetNull            | no         | yes        |
+| SetDefault         | no         | no         |
+| Cascade            | no         | yes        |
+| NoAction           | no         | yes        |
 
 In this mode, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
```
janpio: This feels kind of a limitation of the ref integrity mode `prisma`, so maybe we should call it out as that? Or a "design inhererent limitation" or sth

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
...
 
-The relations are needed for the client to be able to join data between models, so when using the `db pull` command to a database with the `prisma` integrity, existing relations are kept in the data model. This has some caveats:
+The relations are needed for the client to be able to join data between models, so when using the `db pull` command to a database with `referentialIntegrity` set to `prisma` in the `datasource`, existing relations are kept in the data model. This has some caveats:
```
janpio: What does "The relations ..." refer to here? The `@relation` attribute in the schema?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
...
 
 - When a model in the database is deleted, the relation on the other side gets deleted, if existing.
 - When a scalar field used in a relation field gets deleted or renamed, the relation field is kept as-is, leading to validation errors. The user must fix the relation before continuing to use the data model in this case.
```
janpio: I am not sure if I understand what this means.

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -7,11 +7,11 @@ tocDepth: 2
 
 <TopBlock>
 
-In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later, you can define **referential actions** on your relation fields. Referential actions determine what should happen to a record when a related record is deleted or updated.
+Referential actions determine what should happen to a record when a related record is deleted or updated. In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later, you can define referential actions on your relation fields.
```
janpio:
```suggestion
Referential actions determine what should happen to a record when a related record is deleted or updated. In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later, you can define referential actions on your relation fields in your Prisma schema.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -55,9 +55,9 @@ By default a database will reject any operation that violates the referential in
 
 Referential actions are defined in the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute and map to the actions on the **foreign key constraint** in the underlying database. If you do not specify a referential action, [Prisma falls back to a default](#referential-action-defaults).
 
-The following model defines a 1-m relation between `User` and `Post` and a n-m relation between `Post` and `Tag`, with explicitly defined referential actions:
```
janpio, line 58: Why drop the `with explicitly defined referential actions`? Not important to spell out? (Honestly don't know)

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -86,8 +86,10 @@ model Tag {
 }
 ```
 
-- If you delete a `Tag`, the corresponding tag assignment is also deleted `TagOnPost` (`Cascade`)
-- If you delete a `User`, the author is removed from all posts (`SetNull`) - `User` and `userId` must be optional
+This model explicitly defines the following referential actions:
+
+- If you delete a `Tag`, the corresponding tag assignment is also deleted in `TagOnPosts`, using the `Cascade` referential action
+- If you delete a `User`, the author is removed from all posts, using the `SetNull` referential action. To allow this, `User` and `userId` must be optional fields in `Post`.
```
janpio:
```suggestion
- If you delete a `Tag`, the corresponding tag assignment is also deleted in `TagOnPosts`, because of the `Cascade` referential action
- If you delete a `User`, the author is removed from all posts by setting the field value to `Null`, because of the `SetNull` referential action. To allow this, `User` and `userId` must be optional fields in `Post`.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -166,15 +148,21 @@ Note that [`NoAction`](#noaction) can be used in replace of [`Restrict`](#restri
 | PostgreSQL | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
 | MySQL      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
 | SQLite     | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
-| SQLServer  | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
+| SQL Server | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
 
 - † On [InnoDB](https://dev.mysql.com/doc/refman/5.6/en/innodb-introduction.html#:~:text=InnoDB%20is%20a%20general%2Dpurpose,clause%20creates%20an%20InnoDB%20table.), `SetDefault` will be ignored and silently use `NoAction`.
```
janpio: Current sentence is weird somehow.

```suggestion
- † On [InnoDB](https://dev.mysql.com/doc/refman/5.6/en/innodb-introduction.html#:~:text=InnoDB%20is%20a%20general%2Dpurpose,clause%20creates%20an%20InnoDB%20table.), `SetDefault` will be ignored and silently be replaced with `NoAction` instead.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -166,15 +148,21 @@ Note that [`NoAction`](#noaction) can be used in replace of [`Restrict`](#restri
...
 
+From version 3.6.0, referential actions will also be available in MongoDB if the `[referentialIntegrity](/concepts/components/prisma-schema/relations/referential-integrity)` Preview feature is enabled:
```
janpio: Does this render correctly? I think I would have put the backticks _inside_ the link.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: I've made a better attempt at this intro using your comment and @pimeys above... happy for more suggestions...

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Looks to refer to when `referentialIntegrity` is set to `prisma`, updated to be clearer

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Not sure I do either!

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Thanks, think this must have just got lost while rearranging, I put it back in

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Done as part of other changes

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Agreed, hopefully less weird now

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Thanks, fixed

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: @pimeys would this be something you're able to clarify?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
 
 <TopBlock>
 
-Referential integrity means a set of constraints or triggers to handle the consistency of data of a [`@relation`](/reference/api-reference/prisma-schema-reference#relation). It can mean preventing a deletion of a record referred from other records, or it can be a cascading action defined in the field's [`referential actions`](/concepts/components/prisma-schema/relations/referential-actions). 
+Referential integrity is a property of a set of data that states that all its references are valid. It requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you will need to define what happens if its author is deleted.
```
tanberry, line 12: consider: "Referential integrity refers to a property of a data set, which states that all its references are valid. Referential integrity requires that if one record references another, then the referenced record must exist. For example, if a `Post` model defines an author, then the author must also exist. Referential integrity is enforced through constraints that prevent you making from changes that break references, and through defining [referential actions](/concepts/components/prisma-schema/relations/referential-actions) that run when updating or deleting records. In the case of the `Post` model, for example, you must define what action is applied if the referenced `Author` is deleted.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
 
-The referential integrity can either be handled in the database using foreign keys, or in the Prisma Client. There are pros and cons on both. In general, if the database supports foreign keys, it is usually the preferred choice and the default for SQL databases when using Prisma.
+In Prisma, referential integrity is implemented by defining relations between records using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which referential action take place on updating or deleting the related record.
 
-In certain cases, either when the database doesn't support foreign keys, or when they are not preferred due to scaling reasons, Prisma can disable them and handle the referential integrity in the client.
+For example, take the following blog post schema, with a one-to-many relation between `User`s and `Post`s:
+
```
tanberry, line 16: Is there really an apostrophe on User's and Post's?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+  posts Post[]
+}
+```
+
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
+
+When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
+
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. It can also be selected for relational databases in cases where foreign keys are not supported, or where it preferred for other reasons (for example, in cases where there are scaling concerns).
+
```
tanberry: Instead of starting the third sentence with "It" let's say something like: "Emulation can also be explicitly selected for relational databases in cases...."
Missing "is" in "... or where it is preferred for other reasons...."

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
pimeys: Yeah this is badly written. What it means is when we have a situation like this:

```prisma
model A {
  field Int
  b     B   @relation(fields: [field], references: [id])
}
```
You then modify your database, change the `field` name to `renamedField`, which leads to the following data model in the next `db pull`:
```prisma
model A {
  renamedField Int
  b            B   @relation(fields: [field], references: [id])
}
```
As you can see the `fields` parameter points to a non-existing field in the model, and this is something we can't fix in the re-introspection: the user must do it manually.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+<Admonition type="warning">
+  Be aware that currently Prisma only implements the referential actions when
+  emulating referential integrity in the client, and not the constraints. If you
+  set the referential actions to `NoAction` there will be no checks to prevent
+  you from breaking the referential integrity.
+</Admonition>
+
```
tanberry, line 36: I don't see that we talk about or explain "the constraints" earlier in this doc, so this sentence seems alarming cause it's unclear what it means... ? 

Also, need singular on "If you set the referential action to..."

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
 }
 ```
 
-Setting the value to `prisma` will remove all foreign keys previously created in the next migration when using Prisma Migrate and the referential integrity will be handled by the client. The value `foreignKeys` creates foreign keys for all relations in the next migration when using Prisma Migrate, letting the database handle the referential integrity.
+To change the type of referential integrity used, update the `referentialIntegrity` parameter in the [`datasource`](/reference/api-reference/prisma-schema-reference#datasource) block. The available options are:
+
+- `foreignKeys`: this handles referential integrity in the database using foreign keys. This is the default option when using all SQL database connectors.
```
tanberry, line 65: "_This is the default option when using all SQL database connectors._"  and MongoDB?? Though above we imply that having the DB handle RI is only by default for MongoDB, and that other relational (SQL?) DBs you have to explicitly set it? @pimeys ?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Updated with this example

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
 
-The referential integrity can either be handled in the database using foreign keys, or in the Prisma Client. There are pros and cons on both. In general, if the database supports foreign keys, it is usually the preferred choice and the default for SQL databases when using Prisma.
+In Prisma, referential integrity is implemented by defining relations between records using the [`@relation`](/reference/api-reference/prisma-schema-reference#relation) attribute. The `onUpdate` and `onDelete` arguments of this attribute can be used to define which referential action take place on updating or deleting the related record.
```
janpio: This is soooo much better than what we had before 🔥

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
 
-In certain cases, either when the database doesn't support foreign keys, or when they are not preferred due to scaling reasons, Prisma can disable them and handle the referential integrity in the client.
+For example, take the following blog post schema, with a one-to-many relation between `User`s and `Post`s:
```
janpio:
```suggestion
For example, take the following blog post schema, with a one-to-many relation between `Post`s and `User`s:
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+  authorId Int
+}
+
+model User {
+  id    Int    @id @default(autoincrement())
+  posts Post[]
+}
+```
+
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the post if the related user is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
```
janpio:
```suggestion
In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the `Post` if the related `User` is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+
+When connecting to relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
```
janpio:
```suggestion
When using a relational databases that support foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,53 @@ preview: true
...
+
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. It can also be selected for relational databases in cases where foreign keys are not supported, or where it preferred for other reasons (for example, in cases where there are scaling concerns).
+
+<Admonition type="warning">
+  Be aware that currently Prisma only implements the referential actions when
+  emulating referential integrity in the client, and not the constraints. If you
```
janpio: We have no definition of what "the contraints" are before it is mentioned here.

(Should probably be something along the lines of "Foreign keys also create constraints, which make it impossible to manipulate data in a way that these constraints would be violated. Instead of executing the query, the database responds with an error. An example: A foreign key column always has to have an actually existing value from the target field")

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
...
+  When using Prisma Migrate, setting the value to `prisma` will remove all
+  previously created foreign keys in the next migration. Setting the value to
+  `foreignKeys` will create foreign keys for all relations in the next
+  migration.
+</Admonition>
 
 ## Handling the referential integrity in Prisma
 
-If the integrity is handled by the Prisma Client, for now there will be reduced amount of available [referential actions](/concepts/components/prisma-schema/relations/referential-actions) available in the data model:
+In the case where referential integrity is emulated in the Prisma client using the `prisma` option for the `referentialIntegrity` parameter, there is currently a reduced number of supported [referential actions](/concepts/components/prisma-schema/relations/referential-actions) available in the data model:
```
tanberry, line 78: Let's check if "option" and "parameter" are the right words here... I don't think we have called `referentialIntegrity` a pram before have we?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -35,37 +67,43 @@ generator js {
...
+| Referential action | `onUpdate` | `onDelete` |
+| :----------------- | :--------- | :--------- |
+| Restrict           | no         | yes        |
+| SetNull            | no         | yes        |
+| SetDefault         | no         | no         |
+| Cascade            | no         | yes        |
+| NoAction           | no         | yes        |
 
-In this mode, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
+In this case, any [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) database queries will not trigger any of the actions, and they should be handled manually if deleting or updating records using the [`raw`](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) queries.
```
tanberry, line 90: Maybe say something like: "In this case, where emulation in the Prisma Client is used, none of the database queries will trigger any of the actions, and the actions should be handled manually..."

That way we can avoid the possibly confusing negative of "any" + "will not"...

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -7,11 +7,11 @@ tocDepth: 2
 
 <TopBlock>
 
-In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later, you can define **referential actions** on your relation fields. Referential actions determine what should happen to a record when a related record is deleted or updated.
+Referential actions determine what should happen to a record when a related record is deleted or updated. In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later, you can define referential actions on your relation fields in your Prisma schema.
```
tanberry, line 11: It's a nit, but to avoid the future tense, consider rewriting to something like this: "Referential actions determine what happens to a record when a related record is deleted or updated."

(There isn't really a non-awkward way to avoid the passive voice here...)

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -166,15 +148,21 @@ Note that [`NoAction`](#noaction) can be used in replace of [`Restrict`](#restri
 | PostgreSQL | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
 | MySQL      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
 | SQLite     | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
-| SQLServer  | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
+| SQL Server | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
+
+- † On [InnoDB](https://dev.mysql.com/doc/refman/5.6/en/innodb-introduction.html#:~:text=InnoDB%20is%20a%20general%2Dpurpose,clause%20creates%20an%20InnoDB%20table.), `SetDefault` will be ignored and will silently be replaced with `NoAction` instead.
+
+From version 3.6.0, referential actions will also be available in MongoDB if the [`referentialIntegrity`](/concepts/components/prisma-schema/relations/referential-integrity) Preview feature is enabled:
```
tanberry, line 156: Avoid using future tense, so consider rewriting to "From Prisma version 3.6.0 and later, referential actions are also available in MongoDB if..."

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -166,15 +148,21 @@ Note that [`NoAction`](#noaction) can be used in replace of [`Restrict`](#restri
 | PostgreSQL | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
 | MySQL      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
 | SQLite     | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
-| SQLServer  | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
+| SQL Server | **✔️**  | **❌**   | **✔️**   | **✔️**  | **✔️**     |
+
+- † On [InnoDB](https://dev.mysql.com/doc/refman/5.6/en/innodb-introduction.html#:~:text=InnoDB%20is%20a%20general%2Dpurpose,clause%20creates%20an%20InnoDB%20table.), `SetDefault` will be ignored and will silently be replaced with `NoAction` instead.
+
```
tanberry, line 154: ...`SetDefault` is ignored and is silently replaced with... present tense... :-)

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -234,22 +222,23 @@ model User {
 
 ##### Result of using `Restrict`
 
-`User`'s with posts **cannot** be deleted. The `User`'s `id` **cannot** be changed.
+`User`s with posts **cannot** be deleted. The `User`'s `id` **cannot** be changed.
 
 <Admonition type="warning">
 
-The `Restrict` action is **not** available on [Microsoft SQL Server](/concepts/database-connectors/sql-server) and will trigger a schema validation error. Instead, you can use [`NoAction`](#noaction), which produces the same result and is compatible with SQLServer.
+The `Restrict` action is **not** available on [Microsoft SQL Server](/concepts/database-connectors/sql-server) and will trigger a schema validation error. Instead, you can use [`NoAction`](#noaction), which produces the same result and is compatible with SQL Server.
```
tanberry, line 230: "... and triggers a schema validation error..."

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -401,7 +393,7 @@ To make sure you catch these new errors you can adjust your code accordingly.
 
 #### Example of catching errors
 
-The following example uses the below blog schema with a 1-m relationship between `Post` and `User` and sets a [`Restrict`](#restrict) referential actions on the `author` field.
+The following example uses the below blog schema with a one-to-many relationship between `Post` and `User` and sets a [`Restrict`](#restrict) referential actions on the `author` field.
```
tanberry, line 403: The very last section in the doc is the one with the odd title/heading. Please see [this possible rewrite](https://www.notion.so/prismaio/Lucy-Tana-68842761500742819698c21212df965e) of the title and section.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: This is just a backtick in the markup, there's no apostrophe

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: This is correct, `foreignKeys` is the default for SQL only. `prisma` is the default for MongoDB.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Yep, this really needs moving elsewhere, probably to the Referential Actions page as suggested above. Fixed the singular though

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Thanks, done

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Thanks, done

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: I've moved this to the `SetAction` section of the referential actions page and rewritten it a bit - let me know if the new version makes sense, and if it would be better here or somewhere else on the page.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: See comment here: https://github.com/prisma/docs/pull/2581#discussion_r771260070

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Done.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Do you have a suggestion on how to phrase this?

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Done!

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
janpio: Either a headline "limitations" or just preface the paragraph with `Limitation: ` maybe.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
janpio: Link unfortunately does not go anywhere for me :(

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -106,10 +108,7 @@ If you do not specify a referential action, Prisma uses the following defaults:
 | `onDelete` | `SetNull`          | `Restrict`          |
 | `onUpdate` | `Cascade`          | `Cascade`           |
 
-For example - in the following schema, all `Post` records must be connected to an `author`. The schema **does not explicitly define referential actions** on the `author` field, which means that default referential actions apply (see SQL tab):
-
-<TabbedContent tabs={[<FileWithIcon text="schema.prisma" icon="prisma"/>, <FileWithIcon text="SQL" icon="database"/>]}>
-<tab>
+For example, in the following model all `Post` records must be connected to an `author`:
```
janpio:
```suggestion
For example, in the following schema all `Post` records must be connected to a `User` via the `author` relation:
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -125,24 +124,7 @@ model User {
...
-);
-
-CREATE TABLE "User" (
-  "id" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT
-);
-```
-
-</tab>
-</TabbedContent>
+The schema does not explicitly define referential actions on the mandatory `author` field, which means that the default referential actions of `Restrict` for `onDelete` and `Cascade` for `onUpdate` apply.
```
janpio:
```suggestion
The schema does not explicitly define referential actions on the mandatory `author` relation field, which means that the default referential actions of `Restrict` for `onDelete` and `Cascade` for `onUpdate` apply.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -234,22 +222,29 @@ model User {
...
+- **SQL Server**: When a referenced record is deleted or modified, an error is raised. See [the SQL Server docs](https://docs.microsoft.com/en-us/sql/relational-databases/tables/graph-edge-constraints?view=sql-server-ver15#on-delete-referential-actions-on-edge-constraints) for more information.
+- **MongoDB** (in preview from version 3.6.0): When a record is modified or deleted, nothing is done to any related records.
+
+<Admonition type="warning">
+
+If you are [emulating referential integrity](/concepts/components/prisma-schema/relations/referential-integrity#handling-the-referential-integrity-in-prisma) in the Prisma client rather than using foreign keys in the database, you should be aware that currently Prisma only implements the referential actions. Foreign keys also create constraints, which make it impossible to manipulate data in a way that would violate these constraints: instead of executing the query, the database responds with an error. These constraints will not be created if you emulate referential intergrity in the client, so if you set the referential action to `NoAction` there will be no checks to prevent you from breaking the referential integrity.
```
janpio: Need to double check that this information is also present on the referential integrity page.

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
```diff
@@ -8,21 +8,46 @@ preview: true
...
+  id    Int    @id @default(autoincrement())
+  posts Post[]
+}
+```
+
+In this case, the `onDelete: Cascade` argument specifies the `Cascade` referential action, deleting the `Post` if the related `User` is deleted. See [Referential actions](/concepts/components/prisma-schema/relations/referential-actions) for further details of supported actions.
+
+When using a relational database that supports foreign keys, referential integrity can be handled in the underlying database. This is the default option in Prisma when using SQL databases.
+
+Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. Emulation can also be selected for relational databases in cases where foreign keys are not supported, or where it is preferred for other reasons (for example, in cases where there are scaling concerns).
```
janpio:
```suggestion
Alternatively, Prisma can emulate referential integrity in the client. This is the default option when using MongoDB as a database, because it does not support foreign keys. Emulation can also be enabled for relational databases in cases where foreign keys are not supported, or where it is preferred for other reasons (for example, in cases where there are scaling concerns).
```

janpio (approved):

content/200-concepts/100-components/01-prisma-schema/06-relations/420-referential-integrity.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
keerlu: I'd moved it to the referential actions page as you suggested, but I've now added it back to referential integrity as well.

janpio:
Now live at https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-integrity + https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions

Issue #2541, opened 2021-11-18
Document full text index for MySQL


keerlu:
Relevant ticket: https://github.com/prisma/prisma/issues/9412

keerlu:
This is done: https://www.prisma.io/docs/concepts/components/prisma-schema/indexes#full-text-indexes-mysql-and-mongodb

PR #2582, opened 2021-11-25
Update Indexes page with fulltext attribute and type argument

keerlu:
## Describe this PR

WIP for 3.6.0 release, comments welcome. Related docs issue is #2541 

## Changes

- Added documentation of `@@fulltext` attribute
- Added documentation of `type` argument
- Reorganised the page a bit (suggestions welcome on how to do this better...)


content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -14,24 +14,37 @@ From version `3.5.0` onwards, Prisma allows configuration of indexes, unique con
 
 - The `sort` argument is available for all databases on the `@unique`, `@@unique` and `@@index` fields. Additionally, SQL Server also allows it on `@id` and `@@id`.
```
pimeys:
```suggestion
- The `sort` argument is available for all databases on the `@unique`, `@@unique` and `@@index` fields. Additionally, SQL Server also allows it on `@id` and `@@id` and MongoDB on `@@fulltext`.
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +142,58 @@ model Post {
...
 
 In both of the cases above unwanted changes to your database can be prevented by properly specifying these properties in your data model where necessary. **The easiest way to do this is to use `prisma db pull` to retrieve the sort and length arguments where applicable.** Alternatively, you could also add these arguments manually. This should be done before using `prisma db push` or `prisma migrate dev`.
+
+### Upgrading with existing fulltext indexes
+
+Earlier versions of Prisma converted fulltext indexes using the `@@index` attribute rather than the `@@fulltext` attribute. After enabling the `fullTextIndex` preview feature, run `prisma db pull` to convert these indexes to `@fulltext` before migrating.
```
pimeys:
```suggestion
Earlier versions of Prisma converted fulltext indexes using the `@@index` attribute rather than the `@@fulltext` attribute. After enabling the `fullTextIndex` preview feature, run `prisma db pull` to convert these indexes to `@@fulltext` before migrating.
```

pimeys (commented):
I can't write a comment as a suggestion, but we should SCREAM the SQL if we can, and indent accordingly, so these look better:

```sql
CREATE TABLE `Id` (
  `id` VARCHAR(3000) PRIMARY KEY
)
```

```sql
CREATE TABLE `Unique` (
  `unique` INT,
  CONSTRAINT `Unique_unique_key` UNIQUE (`unique` DESC)
)
```

With `extendedIndexes` preview feature, we added a possibility in 3.6.0 to define the algorithm type of an `@@index` attribute on PostgreSQL databases. The allowed values are `BTree` and `Hash`, the default one is always `BTree` and the new syntax we should write about is to change the index type to `Hash`:

```prisma
model A {
  id  Int @id
  val Int
  
  @@index([val], type: Hash)
}
```
This translates to the following SQL:
```sql
CREATE TABLE "A" (
  id INT PRIMARY KEY,
  val INT NOT NULL
);

CREATE INDEX "A_val_idx" ON "A" USING HASH (val);
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Thanks, fixed as part of rearrange

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: I've included this but moved it further down with the other 3.6.0 information

pimeys:
Btw, when we are talking about `BTree` and `Hash` in the context of an index, the correct term is `access method`.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -48,8 +58,8 @@ model Id {
 But this is not not valid in MySQL because it exceeds MySQL's index storage limit and therefore Prisma rejects the data model. The generated SQL would be rejected by the database.
```
pimeys:
```suggestion
But this is not valid in MySQL because it exceeds MySQL's index storage limit and therefore Prisma rejects the data model. The generated SQL would be rejected by the database.
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +139,103 @@ model Post {
...
+
+In each of the cases above unwanted changes to your database can be prevented by properly specifying these properties in your data model where necessary. **The easiest way to do this is to use `prisma db pull` to retrieve any existing constraints.** Alternatively, you could also add these arguments manually. This should be done before using `prisma db push` or `prisma migrate dev`.
+
+## Full text indexes (MySQL and MongoDB)
+
+In version 3.6.0 and later, a new `@@fulltext` attribute is available as a preview feature. This provides support for introspection and migration of full text indexes in MySQL and MongoDB. Existing full text indexes in the database will be added to your Prisma schema after introspecting with `db push`, and new full text indexes added in the schema will be created in the database on migration. This also prevents validation errors in some database schemas that were not working before.
+
+For now we do not enable the full text search commands in the Prisma Client. They will be implemented separately, and the progress can be followed in the MongoDB and MySQL issues.
```
pimeys:
```suggestion
For now we do not enable the full text search commands in the Prisma Client. They will be implemented separately, and the progress can be followed in the [MongoDB](https://github.com/prisma/prisma/issues/9413) and [MySQL](https://github.com/prisma/prisma/issues/10415) issues.
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -1,26 +1,38 @@
 ---
-title: 'Index configuration'
-metaDescription: 'How to configure index functionality to allow sorting and length restriction'
+title: 'Indexes'
```
janpio: Technically this name change should probably also trigger a filename change - but that also requires to setup redirects that the old `/index-configuration` still works and redirects to the new location.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -1,26 +1,38 @@
...
+metaDescription: 'How to configure index functionality and add full text indexes'
 hidePage: false
```
tanberry: Don't we want the metaDescription to still mention `sort` and `length`, as well as the new full text? So maybe something like "_How to configure index functionality to use `sort` and `length` restrictions, and add full text indexes._"

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -78,7 +88,7 @@ model CompoundId {
 
 A similar syntax can be used for the `@@unique` and `@@index` fields.
 
-## New `sort` argument
+### New `sort` argument
```
tanberry: Ditto about using "new"...

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -29,9 +41,7 @@ generator client {
 }
 ```
 
-</TopBlock>
-
-## New `length` argument (MySQL)
+### New `length` argument (MySQL)
```
tanberry: I'd suggest that we always avoid relative terms like "new"... In another half year it won't be so new but we might not remember to edit this. Let's just use a heading of "The `length` argument".

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -1,26 +1,38 @@
...
-In order to enable this as a Preview feature please add it to the `generator` block:
+</TopBlock>
+
+## Index configuration
+
+The `extendedIndexes` preview feature enables new configuration options for indexes, unique constraints and primary key constraints in Prisma.
+
+### Enabling the preview feature
+
+To enable this preview feature, add it to the `generator` block of the `schema.prisma` file:
```
tanberry: Let's avoid the use of vague terms like "this feature" and "it". 

Consider "To enable the Extended Indexes preview feature, add the `extendedIndexes` feature flag to the `generator` block of the `schema.prisma` file:

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -1,26 +1,38 @@
...
+metaDescription: 'How to configure index functionality and add full text indexes'
 hidePage: false
 tocDepth: 2
 preview: true
 ---
 
 <TopBlock>
 
-From version `3.5.0` onwards, Prisma allows configuration of indexes, unique constraints and primary keys constraints using extended syntax options in the Prisma schema language. Two new arguments, `length` and `sort`, can be specified in the `@id`, `@@id`, `@unique`, `@@unique` and `@@index` fields in certain databases:
+Prisma allows configuration of indexes, unique constraints and primary key constraints from version `3.5.0` onwards, using extended syntax options in the Prisma schema language. Two new arguments, `length` and `sort`, can be specified in the `@id`, `@@id`, `@unique`, `@@unique` and `@@index` fields in certain databases:
```
tanberry: A couple of suggestions here:
-- If this intro section is talking about the `extendedIndex` feature (which allows for the use of `length` and `sort`) then let's give the feature a name early on. Maybe something like "Prisma allows configuration of indexes, unique constraints, and primary key constraints from version `3.5.0` onwards, using the Extended Indexes feature to extend the syntax options in the Prisma schema language. ..."

--Let's consider Jan's [proposed structure in Slack](https://prisma-company.slack.com/archives/C02FNFLDUS3/p1637840779468600?thread_ts=1637774947.449500&cid=C02FNFLDUS3), of keeping this as a single topic about Indexes, but separating the extant section about `length` and `sort` from the new section for 3.6, about  using `fullText`. It looks like you have started doing that, but the first sections still try to cover both together.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -1,26 +1,38 @@
...
 
 - The `length` argument is available on MySQL on the `@id`, `@@id`, `@unique`, `@@unique` and `@@index` fields. It allows Prisma to support indexes and constraints on `String` and `Bytes` types.
 
 - The `sort` argument is available for all databases on the `@unique`, `@@unique` and `@@index` fields. Additionally, SQL Server also allows it on `@id` and `@@id`.
```
tanberry: I'd suggest we move the above descriptions of `length` and `sort` to under the section called "Index Configuration". Currently this intro paragraph is more detailed than needed; in here let's be brief and simply state the high-level facts (that we provide index configuration and also setting up indexes to support full text search) then go into the details on each of the two sections. In this intro section, we can say "For more information, see the ____ section below" and link to the anchor.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -1,26 +1,38 @@
...
 
+In version `3.6.0` and later, a new `type` argument is also available on the `@@index` field in PostgreSQL. This argument allows Prisma to support the `Hash` index access method as well as the default `BTree` access method.
+
+Version `3.6.0` also introduces a new `@@fulltext` attribute, available as a preview feature. This provides support for introspection and migration of full text indexes in MySQL and MongoDB. The `sort` argument is also available for `@@fulltext` on MongoDB.
+
 <Admonition type="warning">
 
-This might be a **breaking change** if you have a database that is already making use of these features, so it needs to be explicitly enabled via a Preview feature. See [Upgrading from previous versions](#upgrading-from-previous-versions) for details.
+These might be **breaking changes** if you have a database that is already making use of these features, so they need to be explicitly enabled via preview features. See [Upgrading from previous versions](#upgrading-from-previous-versions) for details.
```
tanberry: Replace "These" with something more specific... perhaps say "The Extended Indexes, the `type` argument, and the Full Text Search attribute are in Preview and thus might be braking changes..."

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Thanks, I'd be happy to do that, just checking how it works - is it a case of updating the redirects [here](https://github.com/prisma/website-v8/blob/main/public/_redirects) and having that go out at the same time as the PR? Is there a particular section of the redirects list it should go in?

janpio: There is also an internal redirects system of the docs, I am not super sure how that works.

I am also not super sure where to optimally put the redirect in the website-v8, but this is a good candidate I think: https://github.com/prisma/website-v8/blob/00d90532be9d372e65f802301ed651f433c9f915/public/_redirects#L423 It is definitely next to other docs redirects that seem to be about renaming and moving pages.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +139,103 @@ model Post {
 }
 ```
 
-## Upgrading from previous versions
+### New `type` argument (PostgreSQL)
+
```
tanberry: If this argument is specifically for defining the index access method, let's fit that into the heading, for reading/scanability and also for SEO.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
janpio: `extendedIndexes` is only a temporary preview feature flag. When it goes away, the feature itself will still be there, which is the ability to configure the index via `length` and `sort` and `type`.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
janpio: Alternatively we could also say something like "Configuring the length of indexes (MySQL only)" or "Configuring the length of indexes with `length` (MySQL)"

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
janpio: Similar to above: "Configuring the access type of indexes (PostgreSQL only)" for example, or "Configuring the access type of indexes with `type` (PostgreSQL only)"

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +139,103 @@ model Post {
...
+In version `3.6.0` and later, a new `type` argument is available in PostgreSQL on the `@@index` attribute.
+This allows you to use `Hash` as the index access method, instead of the default `BTree` access method. The `Hash` access method can be much faster for inserts, but only supports checking for equality, whereas the `BTree` access method supports all comparison operators.
```
janpio: What does this actually do to the index? What is the effect on the queries? Why would someone want to do that?

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +139,103 @@ model Post {
...
+- **Existing hash indexes (PostgreSQL):** earlier versions of Prisma did not support hash indexes in the schema. These need to be added before migrating your database.
+
+In each of the cases above unwanted changes to your database can be prevented by properly specifying these properties in your data model where necessary. **The easiest way to do this is to use `prisma db pull` to retrieve any existing constraints.** Alternatively, you could also add these arguments manually. This should be done before using `prisma db push` or `prisma migrate dev`.
+
+## Full text indexes (MySQL and MongoDB)
+
+In version 3.6.0 and later, a new `@@fulltext` attribute is available as a preview feature. This provides support for introspection and migration of full text indexes in MySQL and MongoDB. Existing full text indexes in the database will be added to your Prisma schema after introspecting with `db push`, and new full text indexes added in the schema will be created in the database on migration. This also prevents validation errors in some database schemas that were not working before.
```
janpio:
```suggestion
In version 3.6.0 and later, a new `@@fulltext` attribute is available as a preview feature. This provides support for introspection and migration of full text indexes in MySQL and MongoDB. Existing full text indexes in the database will be added to your Prisma schema after introspecting with `db pull`, and new full text indexes added in the Prisma schema will be created in the database when using Prisma Migrate. This also prevents validation errors in some database schemas that were not working before.
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +139,103 @@ model Post {
...
+### Upgrading from previous versions
+
+<Admonition type="warning">
+
+These are **breaking changes** when updating from certain existing databases. After updating, always run `prisma db pull` to introspect the existing database before migrating.
+
+</Admonition>
 
-In both of the cases above unwanted changes to your database can be prevented by properly specifying these properties in your data model where necessary. **The easiest way to do this is to use `prisma db pull` to retrieve the sort and length arguments where applicable.** Alternatively, you could also add these arguments manually. This should be done before using `prisma db push` or `prisma migrate dev`.
+Earlier versions of Prisma converted full text indexes using the `@@index` attribute rather than the `@@fulltext` attribute. After enabling the `fullTextIndex` preview feature, run `prisma db pull` to convert these indexes to `@@fulltext` before migrating.
```
janpio:
```suggestion
Earlier versions of Prisma converted full text indexes using the `@@index` attribute rather than the `@@fulltext` attribute. After enabling the `fullTextIndex` preview feature, run `prisma db pull` to convert these indexes to `@@fulltext` before migrating again with Prisma Migrate. If you do not do this, the existing indexes will be dropped instead and normal indexes will be created in their place.
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Updated as part of other changes

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: I've now moved these to Index configuration

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: This is going to get pretty long if we add the `type` argument too. Happy to include if you're keen on it.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: I've moved the details to the index configuration section, the intro now just has a high-level description

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
```diff
@@ -129,17 +141,102 @@ model Post {
...
+model Post {
+  id      Int    @id
+  title   String @db.VarChar(255)
+  content String @db.Text
+
+  @@fulltext([title, content])
+}
+```
+
+On MongoDB, the `fullTextIndex` preview feature can be combined with the `extendedIndexes` preview feature to add ascending or descending columns to your full-text index:
```
pimeys:
```suggestion
On MongoDB, the `fullTextIndex` preview feature can be combined with the `extendedIndexes` preview feature to add fields in ascending or descending order to your full-text index:
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
tanberry: Nah, leave it as is, I agree it could get long.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
pimeys: There are no ascending or descending columns. The fields can be added in ascending or descending order to the index.

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
keerlu: Thanks, I've changed the filename now and [put in a PR](https://github.com/prisma/website-v8/pull/402) on the website-v8 repo

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,242 @@
...
+  previewFeatures = ["fullTextIndex", "extendedIndexes", "mongoDb"]
+}
+
+model Post {
+  id      String @id @map("_id") @db.ObjectId
+  title   String @db.VarChar(255)
+  content String @db.Text
+
+  @@fulltext([title(sort: Desc), content])
+}
```
Jolg42: Seems to me that `@db.VarChar(255)` or `@db.Text` only make sense for MySQL but the example says we are using MongoDB here
```suggestion
generator js {
  provider        = "prisma-client-js"
  previewFeatures = ["fullTextIndex", "extendedIndexes", "mongoDb"]
}

datasource db {
  provider = "mongodb"
  url          = env("DATABASE_URL")
}

model Post {
  id      String @id @map("_id") @db.ObjectId
  title   String
  content String
  
  @@fulltext([title(sort: Desc), content])
}
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,242 @@
...
+model Post {
+  id      Int    @id
+  title   String @db.VarChar(255)
+  content String @db.Text
+
+  @@fulltext([title, content])
+}
+```
+
+On MongoDB, the `fullTextIndex` preview feature can be combined with the `extendedIndexes` preview feature to add ascending or descending columns to your full-text index:
```
Jolg42:
```suggestion
On MongoDB, the preview features `fullTextIndex` and `extendedIndexes`can be combined to add ascending or descending indexed columns to your full-text index:
```

More correct I think?

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,247 @@
...
+  url          = env("DATABASE_URL")
+}
+
+model Post {
+  id      String @id @map("_id") @db.ObjectId
+  title   String
+  content String
+  
+  @@fulltext([title(sort: Desc), content])
+}
```
Jolg42: nitpick, seems I introduced these spaces somehow

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
keerlu: Thanks, this has got changed now as part of @pimeys comments above

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
Jolg42: nitpick, seems I introduced these spaces somehow
```suggestion
datasource db {
  provider = "mongodb"
  url          = env("DATABASE_URL")
}

model Post {
  id      String @id @map("_id") @db.ObjectId
  title   String
  content String

  @@fulltext([title(sort: Desc), content])
}
```

content/200-concepts/100-components/01-prisma-schema/09-index-configuration.mdx
Jolg42: This looks like what you are asking, but I didn't check deeper than that
https://kb.objectrocket.com/postgresql/types-of-indexes-in-postgresql-769
Screen Shot 2021-11-29 at 16 07 15

pimeys: So if you define a `Hash` type as the index access method, it means the index data is stored in a format that is much faster to search and insert; using less disk space as a bonus. It also means only the operations `=` or `<>` (not equals) will use the index to access the data; other operations such as `>`, `<`et.al. will NOT use the index, but do a table scan instead and will be much slower compared to a `BTree` index (the default). Also this kind of index type will not make `ORDER BY` operations any faster. If you know your use of this index is limited to these few operations, it makes sense to use due to much smaller space requirements and faster performance.

keerlu: Thanks, have added more detail

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+title: 'Indexes'
+metaDescription: 'How to configure index functionality and add full text indexes'
+hidePage: false
+tocDepth: 2
+preview: true
+---
+
+<TopBlock>
+
+Prisma allows configuration of database indexes, unique constraints and primary key constraints through the `extendedIndexes` preview feature in version `3.5.0` and later.This provides extended syntax options in the Prisma schema language for configuring the length of indexes (MySQL only), the sort order and the index type (PostgreSQL only).
```
janpio:
```suggestion
Prisma allows configuration of database indexes, unique constraints and primary key constraints through the `extendedIndexes` preview feature in version `3.5.0` and later. This provides extended syntax options in the Prisma schema language for configuring the length of indexes (MySQL only), the sort order and the index type (PostgreSQL only).
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+
+Version `3.6.0` also introduces support for introspection and migration of full text indexes in MySQL and MongoDB through a new `@@fulltext` attribute, available through the `fullTextIndex` preview feature.
+
+<Admonition type="warning">
+
+These changes to index configuration and full text indexes might be **breaking changes** if you have a database that is already making use of these features, so they need to be explicitly enabled via preview features. See [Upgrading from previous versions](#upgrading-from-previous-versions) for details.
```
janpio: The target of this link now exists twice on this page. It will link to the first one probably.

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
janpio:
```suggestion
Prisma allows configuration of database indexes, unique constraints and primary key constraints through the `extendedIndexes` preview feature in version `3.5.0` and later.This provides extended syntax options in the Prisma schema language for configuring the length of indexes (MySQL only), the sort order (all databases) and the index type (PostgreSQL only).
```
Is this suggested edit correct @pimeys?

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+  @@id([title(length: 100, sort: Desc), abstract(length: 10)])
+  @@index([author, created_at(sort: Desc)])
+}
+```
+
+### Configuring the access type of indexes with `type` (PostgreSQL)
+
+The `type` argument is available for configuring the index type in PostgreSQL in version `3.6.0` and later, using the `@@index` attribute. This allows you to use `Hash` as the index access method, instead of the default `BTree` access method.
+
+The `Hash` type will store the index data in a format that is much faster to search and insert, and that will use less disk space. However, only the `=` and `<>` can use the index, so other comparison operators such as `<` and `>` will be much slower than the default `BTree` type.
```
janpio:
```suggestion
The `Hash` type will store the index data in a format that is much faster to search and insert, and that will use less disk space. However, only the `=` and `<>` comparisons can use the index, so other comparison operators such as `<` and `>` will be much slower with `Hash` than when using the default `BTree` type.
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+);
+
+CREATE INDEX "Example_val_idx" ON "Example" USING HASH (val);
+```
+
+### Upgrading from previous versions
+
+<Admonition type="warning">
+
+These index configuration changes can be **breaking changes** when updating from certain existing databases. After updating, always run `prisma db pull` to introspect the existing database before migrating.
```
janpio:
```suggestion
These index configuration changes can be **breaking changes** when activating the functionality for certain, existing Prisma schemas for existing databases. After enabling the preview features required to use them, run `prisma db pull` to introspect the existing database to update your Prisma schema before using Prisma Migrate again.
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+- **Existing hash indexes (PostgreSQL):** earlier versions of Prisma did not support hash indexes in the schema. These need to be added before migrating your database.
+
+In each of the cases above unwanted changes to your database can be prevented by properly specifying these properties in your data model where necessary. **The easiest way to do this is to use `prisma db pull` to retrieve any existing constraints.** Alternatively, you could also add these arguments manually. This should be done before using `prisma db push` or `prisma migrate dev`.
+
+## Full text indexes (MySQL and MongoDB)
+
+The `fullTextIndex` preview feature provides support for introspection and migration of full text indexes in MySQL and MongoDB in version 3.6.0 and later. This can be configured using the `@@fulltext` attribute. Existing full text indexes in the database are added to your Prisma schema after introspecting with `db pull`, and new full text indexes added in the Prisma schema are created in the database when using Prisma Migrate. This also prevents validation errors in some database schemas that were not working before.
```
janpio, line 190:
```suggestion
The `fullTextIndex` preview feature provides support for introspection and migration of full text indexes in MySQL and MongoDB in Prisma version 3.6.0 and later. This can be configured using the `@@fulltext` attribute. Existing full text indexes in the database are added to your Prisma schema after introspecting with `db pull`, and new full text indexes added in the Prisma schema are created in the database when using Prisma Migrate. This also prevents validation errors in some database schemas that were not working before.
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+
+For now we do not enable the full text search commands in the Prisma Client. They will be implemented separately, and the progress can be followed in the [MongoDB](https://github.com/prisma/prisma/issues/9413) and [MySQL](https://github.com/prisma/prisma/issues/10415) issues.
```
janpio, line 194: This should probably be a callout of some sort. (It will also later be removed when we release this support)

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -0,0 +1,249 @@
...
+
+  @@fulltext([title(sort: Desc), content])
+}
+```
+
+### Upgrading from previous versions
+
+<Admonition type="warning">
+
+These are **breaking changes** when updating from certain existing databases. After updating, always run `prisma db pull` to introspect the existing database before migrating.
```
janpio: https://github.com/prisma/docs/pull/2582/files#r758677905

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
keerlu: This would have to be changed all over the place if we want to start doing that, going to leave this one for now

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
keerlu: Thanks for catching this, that's sort of annoying as it's going to look clunky linking to both separately. Or could remove the link completely here. Any preferences on what to do with it?

(It does indeed link to the first one)

keerlu: Removed it for now.

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
keerlu: I've added this in for now as I'm pretty sure it's correct.

PR #2585, opened 2021-11-29
Fixed what looks like an error in the docs

ackinc:


content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -2702,7 +2702,7 @@ A nested `createMany` query adds a new set of records to a parent record. See: [
 
 - `createMany` is available as a nested query when you `create` (`prisma.user.create(...)`) a new parent record or `update` (`prisma.user.update(...)`) an existing parent record.
 - Available in the context of a has-many relation - for example, you can `prisma.user.create(...)` a user and use a nested `createMany` to create multiple posts (posts have one user).
-- **Not** available in the context of a many-to-many relation - for example, you **cannot** you can `prisma.post.create(...)` a post and use a nested `createMany` to create categories (many posts have many categories).
+- **Not** available in the context of a many-to-many relation - for example, you **cannot** `prisma.post.createMany(...)` multiple posts and use a nested `createMany` to create categories (many posts have many categories).
```
janpio: I think the removal of `you can` is correct, but the second part indeed means a `post.createMany` that then has a nested `createMany` for the categories.

janpio (changes requested):
See comment

ackinc:
> See comment

You're right; I've reverted the incorrect language change.

janpio (approved):

PR #2588, opened 2021-11-30
Update 07-prisma-data-platform.mdx

krychaxp:
When i have this changes in `schema.prisma` file , the `npx prisma generate` works

### DIFF
--- a/content/200-concepts/100-components/07-prisma-data-platform.mdx
+++ b/content/200-concepts/100-components/07-prisma-data-platform.mdx
@@ -103,6 +103,7 @@ In the generator block of your `schema.prisma` file, enable the feature flag:
 generator client {
   provider        = "prisma-client-js"
   previewFeatures = ["dataProxy"]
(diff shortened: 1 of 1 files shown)


janpio:
This should not be necessary if you set the env var when running `prisma generate` as documented in line 131 (Documentation is Linux only it seems)

janpio (changes requested):
see comment

tanberry:
HI @krychaxp are you still interested in discussing this PR? If so great, if not we will go ahead and close it in next few days.

I agree with @janpio that best practice would be to define `engineType` as a variable when running the `generate` command (or define it in the .env file), but not define in the schema.

But @janpio what did you mean by "Linux only"?

janpio:
> But @janpio what did you mean by "Linux only"?

The documentation in line 131 shows an example that works on Linux/macOS only, not on Windows (where depending on the terminal you do `set ...` or `$env:...` to set an env var).

PR #2607, opened 2021-12-03
Update 200-working-with-scalar-lists-arrays.mdx

zachzhao1984:
empty array should be '[]', not '{}'

## Describe this PR

empty array should be '[]', not '{}'


tanberry:
Thank you for catching that, @nouh !!

PR #2610, opened 2021-12-04
Update docs for readability

smendoza787:
Ensure data source declaration is in the same order

## Describe this PR
This is just a small change to the order of which we display the modeling code. 

It’s a little easier to soot the main differences between PSQL/MongoDB model code when we’re declaring things in the same order 🙂 

## Changes

Moved one block of example code up a few lines.

## What issue does this fix?

Doesn’t fix an issue, just improves readability.

### DIFF
--- a/content/200-concepts/050-overview/100-what-is-prisma/index.mdx
+++ b/content/200-concepts/050-overview/100-what-is-prisma/index.mdx
@@ -76,14 +76,14 @@ model User {
 <tab>
 
 ```prisma
-generator client {
-  provider        = "prisma-client-js"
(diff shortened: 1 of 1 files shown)


tanberry:
Thank so much, @smendoza787 ! That does help, to have the schema shown reflect the order of the documentation section below: source, generator, model.

janpio:
Note that the new order might not actually be the default order when Prisma schema files are initially created. (Which probably means quite a few examples in the docs are "incorrect" - not a big deal but maybe something to clean up in the future.)

PR #2612, opened 2021-12-06
Updated docs

smitray:
'use' is written 2 times


tanberry:
Thank you @smitray ! We really appreciate your help with keeping the docs clean. :-)

PR #2614, opened 2021-12-07
add mention of the Platform as an option for getting started

tanberry:
I aded two new sections, to introduce the Platform as a way to quickly get started with Prisma.


martzoukos (approved):
This is great.
I like the one paragraph description!

Thank you!

PR #2618, opened 2021-12-11
fix example relation between files

naor88:

### DIFF
--- a/content/300-guides/150-testing/100-unit-testing.mdx
+++ b/content/300-guides/150-testing/100-unit-testing.mdx
@@ -326,7 +326,7 @@ test('should fail if user does not accept terms', async () => {
 
 ```ts
 import { MockContext, Context, createMockContext } from './context'
-import { createUser, updateUsername } from './functions-to-test.ts'
+import { createUser, updateUsername } from './functions-with-context'
 
 let mockCtx: MockContext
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks @naor88 ! We'll get a tech review on this and get it merged; your contributions are appreciated!

tomhoule (commented):
I'm absolutely not familiar with this codebase — maybe @nikolasburk would know how to validate the changes?

nikolasburk:
So, the first change to remove `.ts` from the file name in the import is certainly correct. I'm trying to validate the others as well 👍

nikolasburk:
@TasinIshmam put on his detective hat and followed the types and the `createUser` signature from the article is:

```ts
export async function createUser(user: CreateUser, ctx: Context) {
  if (user.acceptTermsAndConditions) {
    return await ctx.prisma.user.create({
      data: user,
    })
  } else {
    return new Error('User must accept terms!')
  }
}
```

So it does expect only the `context` object and not an instance of `PrismaClient`, so all changes should be correct. Thanks so much for catching this @naor88 🙏

tanberry:
Thanks so much @naor88 !!

naor88:
Happy to help! :)

PR #2619, opened 2021-12-12
Set integration testing docs' script to tear down container

grunklejp:
## Describe this PR
Using `&&` will leave the testing container hanging when `jest -i` fails. 

I think tearing down the container regardless of the results of `jest` is the expected behavior here.

### DIFF
--- a/content/300-guides/150-testing/150-integration-testing.mdx
+++ b/content/300-guides/150-testing/150-integration-testing.mdx
@@ -464,7 +464,7 @@ You can add some scripts to your projects `package.json` file which will setup t
   "scripts": {
     "docker:up": "docker-compose up -d",
     "docker:down": "docker-compose down",
-    "test": "yarn docker:up && yarn prisma migrate deploy && jest -i && yarn docker:down"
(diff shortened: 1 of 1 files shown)


tomhoule:
I think this changes the exit code of failed test runs:

```
prisma-desktop.tom.~/src/prisma-engines λ bash -c 'exit 1'
prisma-desktop.tom.~/src/prisma-engines λ echo $?
1
prisma-desktop.tom.~/src/prisma-engines λ bash -c 'exit 1' && echo 'hi'
prisma-desktop.tom.~/src/prisma-engines λ echo $?                      
1
prisma-desktop.tom.~/src/prisma-engines λ bash -c 'exit 1'; echo 'hi'
hi
prisma-desktop.tom.~/src/prisma-engines λ echo $?                    
0
```

I'm not familiar with the test setup here, but in my projects this would clearly be breaking the expected behaviour. @matthewmueller what do you think?

tomhoule (commented):
.

Jolg42:
Sounds to me that `yarn docker:up` and down should not be part of `test`, from experience it's never included like this. (or rarely)

PR #2621, opened 2021-12-13
Update warning message

millsp:
With https://github.com/prisma/prisma/pull/10637

### DIFF
--- a/content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
+++ b/content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
@@ -9,7 +9,7 @@ metaDescription: 'Best practice for instantiating PrismaClient with Next.js'
 Lots of users have come across this warning while working with [Next.js](https://nextjs.org/) in development:
 
 ```
-warn(prisma-client) Already 10 Prisma Clients are actively running
+warn(prisma-client) There are already 10 Prisma Client engines actively running.
 ```
 
 There's a related [discussion](https://github.com/prisma/prisma/discussions/4399) and [issue](https://github.com/prisma/prisma/issues/5103) for the same.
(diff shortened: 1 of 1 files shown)


content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
```diff
@@ -9,7 +9,7 @@ metaDescription: 'Best practice for instantiating PrismaClient with Next.js'
 Lots of users have come across this warning while working with [Next.js](https://nextjs.org/) in development:
 
 ```
-warn(prisma-client) Already 10 Prisma Clients are actively running
+warn(prisma-client) There are already 10 Prisma Client engines actively running.
 ```
```
Jolg42: Maybe we can try something like this?
```
warn(prisma-client) Already 10 Prisma Clients are actively running
```
Which was renamed since Prisma Client 3.7.0 to:
```
warn(prisma-client) There are already 10 Prisma Client engines actively running.
```

It might help people looking for the old message

tanberry:
@millsp and/or @Jolg42 is 10 the limit or what is the maximum number allowed? I think I can infer from the error message that 10is the upper limit, so you cannot add another, but please confirm. And if Yes, 10 is the limit then we should probably say that.

"There are already 10 Prisma Client engines actively running; this is the maximum number allowed."

janpio:
There is no maximum number, this is just a warning output when we realize something is going wrong - and 10 seemed like a number for that.

tanberry:
Thanks @janpio . So there is not a hard limit... but 10 is the number of engines where we typically see problems start to occur?

I guess I am looking for what action does the error message ask the user to take... do we want to the user to delete some engines?

Ideally error messages have 1) explanation of what happened, and 2) a work-around or fix or resolution steps.

Maybe something like this: "There are already 10 Prisma Client engines actively running; try removing any that are not actively used."

janpio:
This change here just documents the error message that is thrown by Prisma Client. If you want to change and improve the error message, #prisma-client or an issue in prisma/prisma is the way to do it. 

> So there is not a hard limit... but 10 is the number of engines where we typically see problems start to occur?

No, it is a randomly chosen number where we know things probably are not optimal any more (probably = warning, not error)

> I guess I am looking for what action does the error message ask the user to take... do we want to the user to delete some engines?

They find this documentation page and apply the solution suggested.

millsp:
Can we merge this?

janpio (approved):

tanberry:
Thanks @millsp and sorry for my delay!

Issue #2524, opened 2021-11-17
Set the `PRISMA_CLIENT_ENGINE_TYPE` env var to use `dataproxy` 


tanberry:
The Prisma Data Proxy requires users to set the env var  `PRISMA_CLIENT_ENGINE_TYPE` to use the value `dataproxy`.

The value of `dataproxy` refers to the PDP Data Proxy, and *can be used in the schema (if the schema is for a PDP project) to define the engineType, but we do not recommend defining it in the schema; instead use the env var. 

SMEs are Jakub, Daniel, Alberto.

Pages impacted include(they might be others):
-- https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#prisma_client_engine_type
--https://www.prisma.io/docs/concepts/components/prisma-engines/query-engine#defining-the-query-engine-type-for-prisma-client

tanberry:
Just getting back to this. Seems the user will need to add a new line in the .env file, for the `PRISMA_CLIENT_ENGINE_TYPE` environment variable.

janpio:
The user needs to set an environment variable, using the `.env` file is one of multiple supported ways to do that. The current Data Proxy documentation for example sets it manually in the command when running `prisma generate`.

ankur-arch:
Outdated.

PR #2630, opened 2021-12-15
Update 300-environment-variables-reference.mdx

tanberry:
Add a note about using `PRISMA_CLIENT_ENGINE_TYPE` env variable to use the data proxy (if using Platform).


content/400-reference/200-api-reference/300-environment-variables-reference.mdx
```diff
@@ -113,6 +113,11 @@ PRISMA_CLI_QUERY_ENGINE_TYPE=binary
 ```env
 PRISMA_CLIENT_ENGINE_TYPE=binary
 ```
+If you are working in the [Prisma Data Platform](https://cloud.prisma.io) and want to use the [Prisma Data Proxy](/concepts/components/prisma-data-platform#prisma-data-proxy) to manage your database connections, you must edit the `.env` file to configure the `PRISMA_CLIENT_ENGINE_TYPE` environment variable. To do so, add the following line to your [`.env`file](/guides/development-environment/environment-variables#using-env-files):
```
janpio, line 122: empty line after code block and missing space in `env file`:
```suggestion

If you are working in the [Prisma Data Platform](https://cloud.prisma.io) and want to use the [Prisma Data Proxy](/concepts/components/prisma-data-platform#prisma-data-proxy) to manage your database connections, you must edit the `.env` file to configure the `PRISMA_CLIENT_ENGINE_TYPE` environment variable. To do so, add the following line to your [`.env` file](/guides/development-environment/environment-variables#using-env-files):
```

janpio, line 122: Re the `.env` file: https://github.com/prisma/docs/issues/2524#issuecomment-995247669

This applies to _all_ variables on this page. We should probably link to our "env variables" pages in the introduction instead of starting to explain for each one individually.

tanberry:
After learning more... Best Practice when using the Platform and the Data Proxy is exactly the process described in the [Data Proxy docs](https://www.prisma.io/docs/concepts/components/prisma-data-platform#step-5-generate-the-client)  (use the `npx prisma generate` command to set ENGINE _TYPE....).

But we need to add two notes to those docs: 1) don't set the ENGINE_TYPE in the Schema and 2) IF you use the .env file (which most don't) then be aware that you will need to make sure the .env file is loaded, and you can skip Step #5 in Data Proxy docs.

janpio:
Just to make sure: Those are both changed to the Data Proxy docs?

janpio (changes requested):
This PR will become obsolete with dataProxy GA in Prisma Client, and the docs changes that go with it: https://github.com/prisma/docs/pull/3187

janpio:
Not a thing any more.

PR #2632, opened 2021-12-16
Remove Github Issue note from Static IP

martzoukos:
We've been seeing some messages in our now deprecated (for PDP purposes) [Studio repo](https://github.com/prisma/studio/issues) and just realized that we actively suggest them to do that in our Data Platform concept page 😅.

I'm changing that instruction in the Docs with a suggestion to create a new Intercom message.


tanberry:
Oooof good catch!

PR #2633, opened 2021-12-17
Fix plural form in Prisma/TypeORM comparison doc

HurinBerdelon:
## Describe this PR

Just fixed a plural form in a exemple in TypeORM vs Prisma, in Type Safety section.

## Changes

It has a example as follows: 

```ts
const postRepository = getManager().getRepository(Post)
const publishedPosts: Post[] = await postRepository.find({
  where: { published: true },
  select: ['id', 'title'],
})
```

after that, it's called a,

```ts
const post = publishedPost[0]
```

as it was defined as publishedPosts before, I made such a change.

```ts
const post = publishedPosts[0]
```

## What issue does this fix?

This change just correct the name of the variable used in example on the documentation.


tanberry:
Thank you so much, @HurinBerdelon ! We appreciate you catching this.

PR #2634, opened 2021-12-17
Update MongoDB limitations to reflect reality

floelhoeffel:


content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -154,9 +154,10 @@ As MongoDB support is currently in preview there are some known limitations that
 - Currently **no embedded collection support**. See [github.com/prisma/prisma/issues/6708](https://github.com/prisma/prisma/issues/6708).
   - Instead try to [normalize your data models](https://docs.mongodb.com/manual/core/data-model-design/#normalized-data-models). This involves
     describing relationships using references between documents.
-- The [Migrate](/concepts/components/prisma-migrate) and [Introspection](/concepts/components/introspection) workflows are currently **not** supported.
+- Currently no plans to add support for [migrate](/concepts/components/prisma-migrate) as most MongoDB projects do not use schemas
```
Jolg42: Note: I think there are more reasons (like it's not SQL and it would be only data migrations in JavaScript maybe, which Migrate doesn't do (focus on schema migrations))

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -154,9 +154,10 @@ As MongoDB support is currently in preview there are some known limitations that
...
+  - Management of `@unique` indexes is realized through `db push`
```
Jolg42, line 158:
```suggestion
- Management of `@unique` indexes is realized through `db push`
```
Not sure if this space was intended

content/200-concepts/200-database-connectors/07-mongodb.mdx
floelhoeffel: True but I believe the "lack-of-schemas" reasoning should be sufficient

nikolasburk:
```suggestion
- Currently no plans to add support for [Prisma Migrate](/concepts/components/prisma-migrate) as MongoDB projects don't rely on internal schemas where changes need to be managed with an extra tool 
```

nikolasburk (changes requested):
I made a suggestion to adjust the phrasing a little bit but otherwise I think this looks great 👍

nikolasburk (approved):

PR #2637, opened 2021-12-17
fix: Database Guides should list all subpages

janpio:
Check Percy to see what beautiful things this change does.


tanberry (approved):
absolutely beautiful. ;-)

PR #2641, opened 2021-12-20
Fix merge foo in all-contributors file

janpio:
Fix problematic merges https://github.com/prisma/docs/pull/2529 and https://github.com/prisma/docs/pull/2534

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -25,23 +25,30 @@
       ]
     },
     {
-
       "login": "geromegrignon",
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks @Jan, so I needed to add what the exact contribution was (docs)... got it!

janpio:
That usually happens automatically, but I think there were some "resolve conflict" actions in between where the Json file got corrupted/invalid.

PR #2642, opened 2021-12-20
Update 058-transactions.mdx

nikolasburk:
Seems like the example should be nested under the previous section in stead of being an H2?

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
@@ -102,7 +102,7 @@ generator client {
 
 Then you can pass an async function into [`$transaction`](/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api).
 
-## Example
+#### Example
 
 Imagine that you are building an online banking system. One of the actions to perform is to send money from one person to another.
 


tanberry:
Thanks @nikolasburk . I think actually it is OK as an H2... because it is the only example for the entire topic, and is a peer of the other H2s. If there were multiple examples on that page, for diff scenarios, I could see them being nested under a specific section. But I think this stands as an overall example for the Transactions API?

janpio:
This looks like an example for interactive transactions, which only works with the preview feature flag enabled - so the change Niko suggests seems correct. (Might want to double check if the code works without the preview feature flag to confirm)

nikolasburk:
> because it is the only example for the entire topic, and is a peer of the other H2s

Ah, I didn't it read it that way tbh. The example only shows code for the interactive version of the `$transaction` API but doesn't give example for the other kind of transactions Prisma Client offers and that are listed as H2s on the page (i.e. nested writes and batch operations). So I'd personally expect it as a sub-heading in the "The `$transaction` API"-section but I might be missing something 🤔

tanberry:
Thanks @nikolasburk and @Jan. So sounds like we are missing other examples... I read it to be a single topic about Transactions (using the `transactions` API), and this was the one example. 

So in reality, is it that the `transactions` API  is only one of several ways to use transactions? Nested writes and bath/bulk operations are the two other two ways? Do those last two ways not reply on the `transactions` API? (Ah, maybe the "Client API" and the `transactions` API are two diff things?

And... is the "Interactive Transactions" truly a child topic of the `transactions` API?

Thanks for helping me learn! ;-)

nikolasburk:
> Thanks for helping me learn! ;-)

Always!! 💚 

> is it that the transactions API is only one of several ways to use transactions? Nested writes and bath/bulk operations are the two other two ways?

Yes, that's my understanding. 

> Do those last two ways not reply on the transactions AP

What do you mean with "reply on the transactions API" exactly? Or do you mean rely? If it's "rely", then the answer is indeed "no", nested writes and bulk operations have been in the Prisma Client API pretty much from the beginning. The `$transaction` API has only been added later.

> And... is the "Interactive Transactions" truly a child topic of the transactions API?

Good question but I'd say probably yes, mostly because the `$transactions` API can be used in two ways:

- by passing an array of Prisma Client operations that will be executed sequentially
```ts
const [posts, totalPosts] = await prisma.$transaction([
  prisma.post.findMany({ where: { title: { contains: 'prisma' } } }),
  prisma.post.count(),
])
```
- by passing a function in which you can perform a random set of (interactive) operations
```ts
await prisma.$transaction(async (prisma) => {
    // 1. Decrement amount from the sender.
    const sender = await prisma.account.update({
      data: {
        balance: {
          decrement: amount,
        },
      },
      where: {
        email: from,
      },
    })
    // 2. Verify that the sender's balance didn't go below zero.
    if (sender.balance < 0) {
      throw new Error(`${from} doesn't have enough to send ${amount}`)
    }
    // 3. Increment the recipient's balance by amount
    const recipient = prisma.account.update({
      data: {
        balance: {
          increment: amount,
        },
      },
      where: {
        email: to,
      },
    })
    return recipient
  })
```

janpio:
> Good question but I'd say probably yes, mostly because the `$transactions` API can be used in two ways:

Additionally, the second way (interactive transactions) can only be used with the preview feature flag enabled.

> So sounds like we are missing other examples...

They are there - just not with a headline "Example" but directly under `The <inlinecode>$transaction</inlinecode> API`.

When interactive transactions (as a preview feature, and second way to use `$transaction`) were added, the whole page was not rewritten - but an additional subheadline was added at the bottom with an elaborate example (that by the way also includes some of the API documentation, see text `The interactive transaction API has a 2nd parameter with the following configuration options`)

So real solution: Make interactive transactions GA, then restructure the whole section about `$transaction` to make it clearer there are two ways to use this method similar to how Niko did above. (Actually, that might not even need to wait until GA happens for the feature)

tanberry:
thank @nikolasburk and @janpio I'll go ahead and merge this change (that makes the Example into a H4), and also start a working draft on how we want to restructure this page.

PR #2648, opened 2021-12-22
feat(logging): Add parameter logging to example

janpio:


tanberry (approved):

tanberry:
Looks good to me, @janpio, from a words view; do we want to wait for a further tech review or merge?

janpio:
I think it is correct, obviously, but I would not be mad if you found someone from Client team to confirm - especially the example data it logs for these queries - your call.

matthewmueller:
Looks good!

tanberry:
Thanks @janpio !

PR #2652, opened 2021-12-23
fix(homepage): Fix 2 broken links on homepage of docs

janpio:


tanberry:
Yay for vanquishing broken links! Thanks @janpio .

PR #2657, opened 2022-01-03
fix(prisma-migrate): Fix list indentation

janpio:
A few of the list items are wrongly indented.


tanberry:
Thank you @janpio !!

PR #2686, opened 2022-01-08
Fixed for typescript ^4.5.2

meotimdihia:
## Describe this PR

Fixed the file that instance Prisma client with Typescript

## Changes

The current version doesn't work for me in Typescript ^4.5.2 or ^4.3, ...
And even with some old changes like this in the past: 
```typescript
interface CustomNodeJsGlobal extends NodeJS.Global {
  prisma: PrismaClient
}

declare const global: CustomNodeJsGlobal
```

### DIFF
--- a/content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
+++ b/content/500-support/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
@@ -24,8 +24,12 @@ The solution in this case is to instantiate a single instance `PrismaClient` and
 import { PrismaClient } from '@prisma/client'
 
 declare global {
-  var prisma: PrismaClient | undefined
+  interface Global {
+    prisma: PrismaClient | undefined 
+  }
 }
+
+declare const global: Global
 
 export const prisma =
   global.prisma ||


millsp:
What isn't working on your end? This is what I can see on my side on `typescript@4.5`.

Screenshot from 2022-01-13 09-55-36

meotimdihia:
I have got this bug with the code in the documentation:
Code_2022-01-13_17-14-01

millsp:
That is an `eslint` error highlight because it disallows the usage of `var` (which is a very good idea). However, in this context of types it is totally fine, so I suggest that you add this comment `// eslint-disable-next-line no-var` just above your error. And then you are good to go :).

Jolg42:
@millsp since eslint usage is popular we could have that rule in the example? (with a small comment for why I guess)

janpio:
Is there a reason we could not use `let` instead?

meotimdihia:
@janpio we'll get this error when use `let`:
Code_2022-01-13_21-08-17

millsp:
@Jolg42 sounds like a nice idea :+1: 
@janpio that's the way to declare global variables in ts :man_shrugging:

millsp:
PR adding the eslint comment https://github.com/prisma/docs/pull/2719

janpio:
So eslint does not support global variables?

millsp:
It does but discourages any usage of the keyword `var` (that is a common default is eslint configs).

PR #2687, opened 2022-01-08
missing name in Prisma object creation example

moselhy:
Mongoose created the object with the name `Alice` and that was missing in the Prisma example.


tanberry:
Gracious, thank you @moselhy , that looks. much better! :-)

PR #2688, opened 2022-01-08
Fix typo in Migrations docs

andergtk:
## Describe this PR

Fix typo in migrations docs.

## Changes

From:

    even though the schema *end* states match

To:

    even though the schema *and* states match

## What issue does this fix?

None.

None.


andergtk:
@all-contributors please add @andergtk for doc

janpio:
Are you sure this is not actually about the "end state" of the schema?

andergtk:
@janpio oh yep, sorry for the mistake, I'll close these issues

PR #2696, opened 2022-01-11
Update 150-db-push.mdx

tanberry:
Added a Note to tell users of PlanetScale that we recommend `db push` instead of `migrate`. Added a link to the Getting Started docs... but after the PlanetScale-specific docs are published, we can change that link to be more specific. And we can add more details to the upcoming Guide as well, about WHY we recommend `db push`. See [Issue #7282](https://github.com/prisma/prisma/issues/7292#issuecomment-988624732).


janpio (approved):
A similar comment/note might be useful on other `migrate` pages, especially `migrate dev`. (If you are reading the `db push` docs, you are already in the right place)

tanberry:
@janpio agreed about the note might be more useful over on pages that talk about using `prisma migrate`... I'll go ahead and merge this PR, and then look for an eloquent way to do that, because there are a lot of docs about using `migrate`. :-)

PR #2701, opened 2022-01-12
Fix typo in 060-full-text-search.mdx

huzaifahj:
Did you mean `body` instead of `post_body_index`?


keerlu (approved):
Thanks for spotting this, @huzaifahj!

PR #2707, opened 2022-01-12
fix(homepage): Remove trailing slash from URL of Getting Started button

janpio:
No more trailng slash on this button: 
image


tanberry:
Thanks @janpio !

PR #2717, opened 2022-01-13
docs: add missing definition

kamiloox:
All entries of the connection URL were described beside the HOST name.

I've added the missing definition.


tanberry:
Ooops, we missed that one, didn't we? Thanks so much @kamiloox for fixing this!

PR #2731, opened 2022-01-16
fix(upgrading-a-rest-api): Show useful toc

janpio:


content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/07-upgrading-a-rest-api.mdx
```diff
@@ -2,6 +2,7 @@
 title: 'REST API'
 metaTitle: 'Upgrading a REST API from Prisma 1 to Prisma 2'
 metaDescription: 'Learn how to upgrade a REST API from Prisma 1 to Prisma 2.'
+tocDepth: 2
```
janpio, line 5: Hm, this does not seem to be working:
image

Only when scrolling down the headlines appear - but maybe that is even how this is supposed to work @nilubava?

PR #2745, opened 2022-01-19
fix: remove ts file extention

naor88:


janpio (changes requested):
Why this change @naor88?

tanberry:
Hi @naor88 let us us know if you're still interested in discussing this PR, otherwise we will go ahead and close it in a few days. Thanks!

PR #2765, opened 2022-01-25
fix(prisma-migrate): Add details to reset explanation

janpio:
- Adds a footnote about what _exactly_ this means per provider
- Moves the introduction under the headline
- Changes list to be dynamic numbering

### DIFF
--- a/content/200-concepts/100-components/03-prisma-migrate/index.mdx
+++ b/content/200-concepts/100-components/03-prisma-migrate/index.mdx
@@ -281,8 +281,8 @@ This command:
 
 1. Drops the database/schema¹ if possible, or performs a soft reset if the environment does not allow deleting databases/schemas¹
 1. Creates a new database/schema¹ with the same name if the database/schema¹ was dropped
-3. Applies all migrations
-4. Runs seed scripts
+1. Applies all migrations
+1. Runs seed scripts
 
 ¹ For MySQL and MongoDB this refers to the database, for PostgreSQL and SQL Server to the schema, and for SQLite to the database file.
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/03-prisma-migrate/index.mdx
```diff
@@ -263,14 +263,14 @@ The `migrate dev` command will prompt you to reset the database in the following
 - Migration history conflicts caused by [modified or missing migrations](#do-not-edit-or-delete-migrations-that-have-been-applied)
 - The database schema has drifted away from the end-state of the migration history
 
+### Reset the development database
+
 You can also `reset` the database yourself to undo manual changes or `db push` experiments by running:
```
Jolg42, line 268: Unrelated to your changes: I don't understand that part `or db push experiments by running:`

content/200-concepts/100-components/03-prisma-migrate/index.mdx
```diff
@@ -279,10 +279,12 @@ npx prisma migrate reset
...
 This command:
 
-1. Drops the database if possible, or performs a soft reset if the environment does not allow deleting databases
-1. Creates a new database with the same name if the database was dropped
-1. Applies all migrations
-1. Runs seed scripts
+1. Drops the database/schema¹ if possible, or performs a soft reset if the environment does not allow deleting databases/schemas¹
+1. Creates a new database/schema¹ with the same name if the database/schema¹ was dropped
+3. Applies all migrations
+4. Runs seed scripts
```
Jolg42:
```suggestion
1. Applies all migrations
1. Runs seed scripts
```
The rendering will do the rest

Jolg42 (approved):

content/200-concepts/100-components/03-prisma-migrate/index.mdx
janpio: I also stumbled, but above I think it explains somewhere that you can use `db push` to just quickly experiment with different schemas. At least I assumed that 🤷

Jolg42: Maybe it would make more sense with `db push --force-reset`, not sure if it helps though

janpio: For me generally teh docs do not work that are like a story, where you have to have read everything before. Seems like this would still be the case with that change.

PR #2708, opened 2022-01-12
Guide to using Prisma with PlanetScale

tanberry:
Created a stub file for PS Guide so that we can collaboratively get it written asap.

https://deploy-preview-2708--prisma2-docs.netlify.app/docs/guides/database/using-prisma-with-planetscale

PR #2766, opened 2022-01-25
make index bullet a sub item of ref actions/integrity

janpio:
See https://deploy-preview-2766--prisma2-docs.netlify.app/docs/guides/database/using-prisma-with-planetscale

This was what I meant by "make it a sub bullet point" (or however I expressed it). The Index requirement follows from the ref actions/integrity change.

image

Suggestion re https://github.com/prisma/docs/pull/2708


keerlu:
Sorry, I completely missed this PR. I considered a sub-bullet but just one looks very odd to me visually, more like a mistake than a deliberate subsection. I think this having the bullet immediately follow the 'referential actions' bullet is clear enough. Any thoughts @tanberry?

tanberry:
Yeah, I agree that having the single, indented sub-bullet looks like a formatting mistake. I think a solution is to leave it as a regular bullet, right beneath Referntial Actions/Integrity bullet, but add a sentence at the very beginning saying something like:

 "Related to the use of referential actions is the use of foreign keys."
 
 Also, @keerlu instead of using the word "As..." use the word "Because..." (That's another translation thing... the word "as" is just too squirelly a word.)

janpio:
To me the "visual mistake" actually represents the visual hierarchy of the content.. But you take it from here.

janpio:
(I restored and changed the target branch to make it visible again what this PR really changes)

PR #1900, opened 2021-06-18
MongoDB

mhwelander:
TODO After merge: https://github.com/prisma/docs/pull/1900/#issuecomment-875331682

PR #2770, opened 2022-01-26
Remove duplicate copy

pawel-cebula:
## Describe this PR

Removes duplicate copy from the docs

## Changes

Same text already appears on lines 212-216

## What issue does this fix?

Docs

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/index.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/index.mdx
@@ -326,12 +326,6 @@ The following list of `Post` documents each have a `userId` field which referenc
 ]
 ```
 
-## Relation fields
-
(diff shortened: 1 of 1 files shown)


janpio:
This seems to have been added with the MongoDB PR:
https://github.com/prisma/docs/pull/1900/files#diff-65ea14636cdcbfb2c5214e01a71556401fc7f171ee4cd22112ed087038ec8f7dR216
https://github.com/prisma/docs/pull/1900/files#diff-65ea14636cdcbfb2c5214e01a71556401fc7f171ee4cd22112ed087038ec8f7dR334
🤔

keerlu:
Thanks for spotting this, @pawel-cebula! It does look like something odd happened when this page was updated for MongoDB. 

I think that the example below starting with 'Consider these two models...' also needs to be moved to the 'Annotated relation fields' below, as it refers to annotated fields... @pawel-cebula / @janpio would you agree with that?

janpio:
(I am lost on that page, so will not comment on that)

tanberry:
@keerlu can you please take another look at this page; I do still see the duplicated content there .

keerlu:
Sorry that this got so delayed. This particular duplicate got fixed separately in #3598 , but the page turned out to have several other bits of duplicate/out-of-place content. Fixed now in #4193.

PR #2772, opened 2022-01-27
Prisma + PlanetScale: Word referential integrity emulation more positively

janpio:
Suggestion to word that paragraph more positively and highlight hte additional functionality Prisma gives you.


content/300-guides/050-database/850-using-prisma-with-planetscale.mdx
```diff
@@ -40,7 +40,8 @@ PlanetScale's branching model and design for scalability means that there are al
...
+- **Referential actions and integrity.** To support scaling across multiple database servers, PlanetScale [_does not allow the use of foreign key constraints_](https://docs.planetscale.com/learn/operating-without-foreign-key-constraints), which are normally used in relational databases to enforce relationships between data in different tables, and asks users to handle this manually in their applications.
```
keerlu:
```suggestion
- **Referential actions and integrity.** To support scaling across multiple database servers, PlanetScale [does not allow the use of foreign key constraints](https://docs.planetscale.com/learn/operating-without-foreign-key-constraints), which are normally used in relational databases to enforce relationships between data in different tables, and asks users to handle this manually in their applications.   
```

Happy with the rewording but the italics in links look very odd to me, don't think we are using that elsewhere...

keerlu (approved):

content/300-guides/050-database/850-using-prisma-with-planetscale.mdx
janpio: I wanted to highlight the fact what is actually the difference. I agree that link plus italics is not the most common, but 🤷 

You decide if you want to undo that change - happy either way.

PR #2777, opened 2022-01-27
fix: fix expect call

m1pl:


Jolg42 (changes requested):
Since the function is defined like 
```
export async function createUser(user: CreateUser, ctx: Context) {
  if (user.acceptTermsAndConditions) {
    return await ctx.prisma.user.create({
      data: user,
    })
  } else {
    return new Error('User must accept terms!')
  }
}
```

So the function returns the error (it doesn't throw it)
It looks like the example is correct to me and this PR not needed.

Though I agree I usually throw errors and test them with `.rejects.toThrowError...`

Jolg42:
Thanks for the PR ❤️

m1pl:
Oh sorry, didn't pay attention. I copy/pasted only the test, but my function throws, hence the misunderstanding.

PR #2781, opened 2022-01-28
MongoDB: Introspection works, and migration is possible via `db push`

janpio:
Possibly we want to keep some information that Migrate works differently or something - but the current admonition is just wrong.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
@@ -253,13 +253,6 @@ Your data model reflects _your_ application domain. For example:
 
 ## Introspection and migration
 
-<Admonition type="warning">
-
-**Not supported by the MongoDB connector**<br />
-You cannot currently use Prisma Migrate or Introspection with the MongoDB connector. Instead, you must manually define a schema that matches the structure of your database.
-
-</Admonition>
-
 There are two ways to define a data model:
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
```diff
@@ -253,13 +253,6 @@ Your data model reflects _your_ application domain. For example:
 
 ## Introspection and migration
```
tanberry, line 255:
```<Admonition type="warning">

** For MongoDB:**<br />
For migration with MongoDB, use the `db push` command, not `migrate`. For introspection, you can use the command `db pull` but be aware that with MongoDB you can only introspect once. We don't support re-introspection at this time. Introspection uses heuristics to guess the data types. If data is inconsistent, you will be prompted and you might need to tweak your data to make it consistent. Additionally, with MongoDB, we do not infer relations, so after you introspect, you will need to add and maintain the relations yourself.

</Admonition>
```

tanberry:
@Jan, thanks for catching this; the admonition was very negative and gave no other options. Thanks @matthewmueller for the overview of how instrospection can work with MongoDB... can I get a review/tweak/approval of my suggested change to the admonition?

janpio:
I suggest we move the creation of new wording into a new PR so we can get the misleading "Introspection is not supported" removed here.

PR #2782, opened 2022-01-29

valerii15298:
Text "The quick brown fox jumps over the lazy dog" contains word 'dog' but does not contain word 'cat' so expression `+dog -cat` in MySQL full text search must be true.


tanberry:
HI @valerii15298 your logic seems solid; it is my logic I doubt. ;-) @keerlu can you apply your logic and confirm, please?

keerlu:
I also agree that the current logic is wrong - @valerii15298 thanks very much for noticing this!

PR #2785, opened 2022-01-30
Fix tiny typo "rest-nuxjs" to "rest-nuxtjs"

stanislav-suslov:


tanberry:
Whoops. Thank you so much @stanislav-suslov for catching this!

PR #2787, opened 2022-01-31
Add `--transpile-only` flag to seed script

tsongas:
Failing to use the `--transpile-only` flag in this context is a catastrophic mistake that consumes massive amounts of RAM when typechecking the seed script(s) unnecessarily, in addition to transpiling them, thereby completely crashing Node in any container with limited RAM i.e. Docker, Heroku, etc. when trying to seed.

### DIFF
--- a/content/300-guides/050-database/300-seed-database.mdx
+++ b/content/300-guides/050-database/300-seed-database.mdx
@@ -46,6 +46,7 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key 
 
 Example:
 `"seed": "ts-node --transpile-only prisma/seed.ts"`
+
 This can be useful to reduce memory usage (RAM) and increase execution speed of the seed script.
 
 </Admonition>  
--- a/content/300-guides/050-database/300-seed-database.mdx
+++ b/content/300-guides/050-database/300-seed-database.mdx
@@ -25,6 +25,17 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key 
   "seed": "ts-node prisma/seed.ts"
 },
 ```
+  
+<Admonition type="info">
+
+`ts-node` does transpiling and typechecking by default; typechecking can be disabled with the following flag `--transpile-only`.
+
+Example:
+`"seed": "ts-node --transpile-only prisma/seed.ts"`
+
+This can be useful to reduce memory usage (RAM) and increase execution speed of the seed script.
+
+</Admonition>   
 
 </tab>
 
@@ -38,18 +49,7 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key 
 
 </tab>
 
-</TabbedContent>
-
-<Admonition type="info">
-
-`ts-node` does transpiling and typechecking by default, typechecking can be disabled with the following flag `--transpile-only`.
-
-Example:
-`"seed": "ts-node --transpile-only prisma/seed.ts"`
-
-This can be useful to reduce memory usage (RAM) and increase execution speed of the seed script.
-
-</Admonition>  
+</TabbedContent> 
 
 ## Integrated seeding with Prisma Migrate
 
--- a/content/300-guides/050-database/300-seed-database.mdx
+++ b/content/300-guides/050-database/300-seed-database.mdx
@@ -43,6 +43,9 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key 
 <Admonition type="info">
 
 `ts-node` does transpiling and typechecking by default, typechecking can be disabled with the following flag `--transpile-only`.
+
+Example:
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks @tsongas !! We will get a tech review and then get this merged; we appreciate your contribution here.

Jolg42:
@tsongas Hi! 
I understand that not doing the type-checks will be faster and use less resources, though for a small seed script it should be reasonable or something some users might want to find type errors (which they won't see after).

I guess some people mainly use seed locally.
You seem to use seed remotely, maybe on some staging setup? (I'm curious to know more about your workflow)

Did you experience the mentioned problems because the seed script was large/imported many other files, maybe?

Thanks!

tsongas:
@Jolg42 thanks for following up! Since I already figured this out, just wanted to save others a huge amount of frustration.

Seed data isn't just for local development it's also needed for automated tests that run in containers in CI/CD pipelines i.e. GitHub Actions, CircleCI, etc., plus in staging and production environments.

I'm only seeding ~100 records, a trivial amount, so I don't think the size of the seed script is an issue, it's that just starting the TS transpiler with even a trivially small script crashes an upgraded Heroku container with 1GB RAM.

Regarding type errors, anyone not writing code with Notepad is going to see them in their editor, which is exactly where I would want to see them...I don't think doing type checks while running the seed files adds any value.

tsongas:
P.S. not to mention people should have a pre-push git hook that runs `tsc` before pushing anything to GitHub.

tanberry:
Hi @tsongas, I'm proposing that I edit your PR to simply add an admonition about using the `--transpile-only` flag can prevent performance issues, instead of changing the default syntax. So basically adding that Note right below the tabbed examples in "[How to seed....](https://www.prisma.io/docs/guides/database/seed-database#how-to-seed-your-database-in-prisma)".  Sound ok? I'll make the edits, actually, and then you can see what I mean.

Also, does this `--transpile-only` flag apply to both Typescript and JS?

Jolg42:
@tanberry this is a TypeScript only change

Jolg42:
Currently looks like this
https://deploy-preview-2787--prisma2-docs.netlify.app/docs/guides/database/seed-database 
Screen Shot 2022-02-10 at 08 53 57

I think we should make it clear it only applies to the TypeScript tab somehow

content/300-guides/050-database/300-seed-database.mdx
```diff
@@ -40,6 +40,12 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key
 
 </TabbedContent>
 
+<Admonition type="info">
+
+When using containers with limited RAM, consider using the `--transpile-only` flag to avoid potential performance issues. The process of typechecking the seed script(s), in addition to transpiling them, can consume large amounts of RAM.
```
Jolg42:
```suggestion
`ts-node` does transpiling and typechecking by default, typechecking can be disabled with the following flag `--transpile-only`.
This can be useful to reduce memory usage (RAM) and increase execution speed of the seed script.
```
Maybe?

tsongas: I think an admonition addresses the issue nicely, my only suggestion is to show the actual code to use i.e. `"seed": "ts-node --transpile-only prisma/seed.ts"` somewhere rather than just showing the flag and leaving it to the user to figure out where/how to use it, sometimes flags can be tricky and not work if added in the wrong order.

Jolg42: Sounds good 👍🏼

content/300-guides/050-database/300-seed-database.mdx
```diff
@@ -40,6 +40,13 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key
 
 </TabbedContent>
 
+<Admonition type="info">
+
+`ts-node` does transpiling and typechecking by default, typechecking can be disabled with the following flag `--transpile-only`.
```
tanberry:
```suggestion
`ts-node` does transpiling and typechecking by default, typechecking can be disabled with the following flag `--transpile-only`.

Example:
`"seed": "ts-node --transpile-only prisma/seed.ts"`
```

content/300-guides/050-database/300-seed-database.mdx
```diff
@@ -40,6 +40,16 @@ Prisma's integrated seeding functionality expects a command in the `"seed"` key
...
+
+Example:
+`"seed": "ts-node --transpile-only prisma/seed.ts"`
```
tanberry:
```suggestion
`"seed": "ts-node --transpile-only prisma/seed.ts"`

```

tanberry:
Thanks so much @tsongas for this important info, and @Jolg42 for your tech reviews! Merging now, but let me know if further changes are needed.

PR #2790, opened 2022-01-31
improve: pin jest-mock-extended version

ruheni:
Specify jest-mock-extended version required for installation


tanberry:
Thank you @ruheni !!

content/300-guides/150-testing/100-unit-testing.mdx
```diff
@@ -64,7 +64,7 @@ To ensure your unit tests are isolated from external factors you can mock the Pr
 This guide will cover two approaches to mocking the client, a singleton instance and dependency injection. Both have their merits depending on your use cases. To help with mocking the client the [`jest-mock-extended`](https://github.com/marchaos/jest-mock-extended) package will be used.
 
 ```terminal
-npm install jest-mock-extended --save-dev
+npm install jest-mock-extended@2.0.4 --save-dev
```
janpio, line 67: Why do we want to pin a specific version of this library permanently for this guide?

ruheni: This seems to be my bad I pinned a specific version but ideally we would like to pin only major versions of other libraries we use in our guides to reduce the maintenance overhead in the long run.

PR #2819, opened 2022-02-04
Update 058-transactions.mdx

KATT:
## Describe this PR

Proposal to name the callback's argument `tx` (for transaction) not to confuse it with the global `prisma` client


tanberry:
Hullo @KATT apologies for us losing this PR, and such a delated reply. Your idea makes sense to me, let me get a tech review on it and see what others think.

millsp:
Hey @KATT 👋

We initially decided on this convention so that one could not mistakenly (or by habit) use the non-transactional client in a transaction, effectively shadowing the non-transactional `prisma` that lives in the higher scope. WDYT?

PR #2821, opened 2022-02-05
More comparable queries for Mongoose vs Prisma documentation

HommeSauvage:
## Describe this PR

Some queries comparisons were different between mongoose and prisma, it's better to keep the sam queries and data to highlight the differences accurately.

## Changes

Queries and some data to for a more accurate comparison


janpio (approved):

tanberry:
Thanks so much, @HommeSauvage !

PR #2825, opened 2022-02-07
Update 05-static-ips.mdx

tanberry:
removed link to AWS for now, until we can figure out how to preserve the capitalization in the URL (which is required in order for link to resolve)


janpio:
> until we can figure out how to preserve the capitalization in the URL (which is required in order for link to resolve)

You need to use a HTML link (instead of a Markdown one), so in this case:

```
<a href="https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.RDSSecurityGroups.html">security groups</a>
```

There is an issue about this for ages.

PR #2778, opened 2022-01-27
feat(client): mongo raw queries

millsp:
https://deploy-preview-2778--prisma2-docs.netlify.app/docs/concepts/components/prisma-client/raw-database-access

PR #2826, opened 2022-02-07
Fix intra-page navigation for $executeRaw

aganoza:
## Describe this PR

Fix intra-page navigation for $executeRaw

Without this change $executeRaw doesn't appear in the sidebar nav and is not clickable either.

### Before:
https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#executeraw

imagen

### After:
https://deploy-preview-2826--prisma2-docs.netlify.app/docs/concepts/components/prisma-client/raw-database-access#executeraw

imagen

@all-contributors please add @aganoza for doc

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
@@ -200,9 +200,7 @@ const result = await prisma.$queryRawUnsafe(
 
 > **Note**: Using `%$2` as an argument would not work
 
-<h2 id="executeraw">
-  <inlinecode>$executeRaw</inlinecode>
-</h2>
+### <inlinecode>$executeRaw</inlinecode>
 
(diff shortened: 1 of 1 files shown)


janpio:
Anyone an idea why this was done via HTML before?

aganoza:
I see $executeRaw was changed to the h2 form here https://github.com/prisma/docs/pull/2778/files#diff-e519588d56e575428e4e74aa7a73e86a43d66f2ab3fddf4ce79bc75dbe873b5eR203-R205 

In the same pull request $queryRawUnsafe wasn't https://github.com/prisma/docs/pull/2778/files#diff-e519588d56e575428e4e74aa7a73e86a43d66f2ab3fddf4ce79bc75dbe873b5eR117

keerlu:
Thanks @aganoza! @janpio we introduced some HTML when we were trying to make the tech switcher component play nicely with headings, and then ended up removing the tech switcher when we ran into other problems with it. I thought I'd changed them all back but looks like I missed one.

PR #2827, opened 2022-02-07
Update 05-static-ips.mdx

tanberry:
Edited to use html to support capitalization in link. Thanks @janpio !


janpio:
Page does not exist any more this way.

PR #2828, opened 2022-02-08
chore: add section on pinning package versions

ruheni:
Added a section in the style guide to pin major versions

### DIFF
--- a/content/600-about/200-prisma-docs/20-style-guide/index.mdx
+++ b/content/600-about/200-prisma-docs/20-style-guide/index.mdx
@@ -366,6 +366,14 @@ npm install prisma
 
 Always use `npm` commands instead of `yarn`.
 
+### Package versioning
+
+Include the major version of a package in the installation step. This ensures compatibility and reduces the maintenance overhead when a package introduces a breaking change:
+
+```terminal
+npm install --save-dev typescript@4
+```
+
 ## Spelling
 
 ### Hyphens


content/600-about/200-prisma-docs/20-style-guide/index.mdx
```diff
@@ -366,6 +366,14 @@ npm install prisma
...
 Always use `npm` commands instead of `yarn`.
 
+### Package versioning
+
+Include the major version of a package in the installation step. This ensures compatibility and reduces the maintenance overhead when a package introduces a breaking change:
+
+```terminal
+npm install --save-dev typescript@4
+```
+
```
nikolasburk, line 376: Maybe include another paragraph about what the pinned version looks like inside of `package.json`.

content/600-about/200-prisma-docs/20-style-guide/index.mdx
```diff
@@ -366,6 +366,14 @@ npm install prisma
 
 Always use `npm` commands instead of `yarn`.
 
+### Package versioning
+
+Include the major version of a package in the installation step. This ensures compatibility and reduces the maintenance overhead when a package introduces a breaking change:
```
janpio, line 371: What about the other side of the additional maintenance overhead by having pinned old major versions although the market moved on?

ruheni: That makes sense. 

This is where the prisma-examples or the e2e tests would come in handy :) – or wherever we choose to store the examples used in the docs. It could be slightly more manageable because I think we can only bump up the major versions in the docs once the tests for the examples pass. 

We can create an initiative to set this up. WDYT? Is it worth the effort? This would need an owner too 🤔

nikolasburk: > What about the other side of the additional maintenance overhead by having pinned old major versions although the market moved on?

That's a great question and reflects the tradeoff we're targeting here! I've formed a fairly strong opinion about this and believe that it's _much more important for an educational resource to be functional than necessarily up to date_. 

If the educational resource is broken because of a breaking change in a package that it uses, there's pretty much no value for the reader any more. They read instructions that are leading to error messages and that are frustrating. On the other hand, if something is "only" outdated, it will work and the reader will get the value they were seeking.  

Maintenance is an important question though and maybe it would make sense to track instances of pinned dependencies that can be revisited on a regular basis or some similar kind of process...

tanberry: What a great conversation! Agree about it's more important that resources are functional than up-to-date (ideally both of course). On the question of where to store samples, what do you all think of having a /gist repo, a single place that we store code snippets and samples? Maybe the current /samples repo could be converted to a /gist repo?

My thinking is this, a two-fold approach: 

1. Yes, we need to state explicitly the exact versions that we support in any install/configure instructions, but that info will be covered in the "Considerations" section, which contains admonitions, pre-reqs, etc. This "Considerations" section is part of a [new template](https://www.notion.so/prismaio/Templates-c414899d290d458b892faaac01526835) that I want to introduce. I don't think exact versions shuold be in the actual steps of the procedure... to fine-grained for easy maintenance. ;-)

2. For the exact code samples used in the steps, what do you all think of the /gist repo idea? That way we mantain them in one place, with the current version-specific syntax, etc.

janpio: What is a `/gist repo`? 

The benefit of having the full examples in a repository would be that they can be tested, run and automation like Renovate can run to keep dependencies up to date (if the updates are compatible with the current code)

nikolasburk:
Let's close this one for now and discuss later how we want to approach package versioning (and then also make sure to go through the docs to adhere to one consistent approach).

https://github.com/prisma/docs/issues/5493

PR #2843, opened 2022-02-10
fix(links): Add missing link targets (or just remove some empty links instead)

janpio:
more broken link fixes


tanberry (commented):
Thanks so much @janpio ! Question:does the change from yarn to npmjs, in the package-lock.js file, impact any sample code snippets or anything elsewhere in the docs?

janpio:
No, that is purely internal to the docs build process and actually a bit weird that a Npm lock file uses the Yarn proxy (which is serving the same exact data as the Npm source CDN).

PR #2849, opened 2022-02-12
Replace `integer` with `bigint` for PostgreSQL in default BigInt type mappings (#2690)

### LINKED ISSUE #2690: BigInt Prisma type maps to `integer` postgres type
The documentation specifies `BigInt` Prisma types have a default mapping to the `integer` type for postgres, but it looks like `BigInt`s in Prisma do map to `bigint`s in postgres. unless I'm misunderstanding 😅 

Docs: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#bigint
Example:
```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}

model User {
  id BigInt @id
}
```
```
postgres=# \d "User"
               Table "public.User"
 Column |  Type  | Collation | Nullable | Default 
--------+--------+-----------+----------+---------
 id     | bigint |           | not null | 
Indexes:
    "User_pkey" PRIMARY KEY, btree (id)
```

ertgl:
Fixes #2690

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -550,7 +550,7 @@ True or false value.
 
 | Connector     | Default mapping |
 | ------------- | --------------- |
-| PostgreSQL    | `integer`       |
+| PostgreSQL    | `bigint`        |
 | Microsoft SQL | `int`           |
 | MySQL         | `INT`           |
 | MongoDB       | `Long`          |


Pwuts:
Can this be reopened? It's a valid fix since Prisma does actually map BigInt to the postgres type BIGINT by default.

ertgl:
> Can this be reopened? It's a valid fix since Prisma does actually map BigInt to the postgres type BIGINT by default.

I agree. This may have been closed automatically because I deleted my fork.

janpio:
No "reopen" button on our side unfortunately.

PR #2859, opened 2022-02-15
remove mention of dble-ticks in raw-database-access.mdx

tanberry:
Removed admonition about using double-ticks


content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -284,15 +284,13 @@ In 2.10.0 and later, you can use `.$executeRaw()` and `.$queryRaw()` inside a [t
 
 `$executeRaw` and `$queryRaw` are implemented as [**tagged templates**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). Tagged templates are the recommended way to use variables with raw SQL in the Prisma Client.
 
-The following example includes a placeholder named `${userId}`. Note the double backticks (` `` `) instead of `()`:
+The following example includes a placeholder named `${userId}`.:
```
janpio:
```suggestion
The following example includes a placeholder named `${userId}`:
```

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -284,15 +284,13 @@ In 2.10.0 and later, you can use `.$executeRaw()` and `.$queryRaw()` inside a [t
...
 
 ```ts
 const userId = 42
 const result = await prisma.$queryRaw`SELECT * FROM User WHERE id = ${userId};`
 ```
 
-> **Important**: You _must_ use double backticks - the same query using `$queryRaw()` is not secure because `${userId}` is not escaped.
```
janpio, line 294: This would actually make sense:

> **Important**: You _must_ use backticks - the same query using `$queryRaw()` is not secure because `${userId}` is not escaped.

I am not sure if that API and difference still exists though or if we moved that into the `unsafe` method completely.

millsp: This API is now gone, we now throw if you don't pass a template. One can use the raw function with the backticks or just passing the result of `Prisma.Sql` helper, but not strings.

https://github.com/prisma/prisma/blob/main/packages/client/src/runtime/getPrismaClient.ts/#L888

janpio (approved):

PR #2865, opened 2022-02-16
MongoDB: Explain when to use findRaw or aggregateRaw instead of $runCommandRaw

### LINKED ISSUE #2845: Update `$runCommandRaw` example to not use aggregate command
Update to an example that doesn't use `aggregate`, and make it clear that you need to use the separate raw query methods for `find` and `aggregate`.

See Slack thread [here](https://app.slack.com/client/T0MQGDA3Y/C5Z9TH6N9/thread/C5Z9TH6N9-1644012842.536569):

> Hey, I don't know where to post this but there's this tweet showing an aggregate command with runCommandRaw which won't really work https://twitter.com/prisma/status/1489546239985307650?s=21 The query will run but if there's more than a hundred documents, the response will only show the first page and there's no way for users to fetch the additional pages. Might not be a big deal but I don't know if our docs encapsulates this The same applies for find, which is the reason we have these separate raw queries

keerlu:
## Describe this PR

Explain when and why to use `findRaw` or `aggregateRaw` instead of `$runCommandRaw`, and change example that wrongly uses `aggregate`

## What issue does this fix?

Fixes #2845

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
@@ -397,6 +397,49 @@ generator client {
 }
 ```
 
+### <inlinecode>$runCommandRaw</inlinecode>
+
+`$runCommandRaw` runs a raw MongoDB command against the database. For example, this query inserts two records with the same `_id`, bypassing normal document validation:
+
+```ts no-lines
+prisma.$runCommandRaw({
+  insert: 'Pets',
+  bypassDocumentValidation: true,
+  documents: [
+    {
+      _id: 1,
+      name: 'Felinecitas',
+      type: 'Cat',
+      breed: 'Russian Blue',
+      age: 12,
+    },
+    {
+      _id: 1,
+      name: 'Nao Nao',
+      type: 'Dog',
+      breed: 'Chow Chow',
+      age: 2,
+    },
+  ],
+})
+```
+
+<Admonition type="warning">
+
+Note that the `$runCommandRaw` command should not be used for queries which contain the `"find"` or `"aggregate"` commands, as you may be unable to fetch all data. This is because MongoDB returns a [cursor](https://docs.mongodb.com/manual/tutorial/iterate-a-cursor/) that is attached to your MongoDB session, and you may not hit the same MongoDB session every time. For these queries, you should use the specialised [`findRaw`](#findraw) and [`aggregateRaw`](#aggregateraw) methods instead.
+
+</Admonition>
+
+#### Return type
+
+`$runCommandRaw` returns a `JSON` object whose shape depends on the inputs.
+
+#### Signature
+
+```ts no-lines
+$runCommandRaw(command: InputJsonObject): PrismaPromise<JsonObject>;
+```
+
 ### <inlinecode>findRaw</inlinecode>
 
 `<model>.findRaw` returns actual database records. It will find zero or more documents that match the filter on the `User` collection:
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -397,6 +397,49 @@ generator client {
...
 ```
 
+### <inlinecode>$runCommandRaw</inlinecode>
+
+`$runCommandRaw` runs a raw MongoDB command against the database. For example, this query inserts two records with the same `_id`, bypassing normal document validation:
+
+```ts no-lines
+prisma.$runCommandRaw({
+  insert: 'Pets',
+  bypassDocumentValidation: true,
```
matthewmueller, line 407: I don't think we want to promote this option `bypassDocumentValidation`. If MongoDB users are using JSON schema validation, that's good for us since it keeps the data consistent.

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
```diff
@@ -397,6 +397,49 @@ generator client {
...
+      breed: 'Chow Chow',
+      age: 2,
+    },
+  ],
+})
+```
+
+<Admonition type="warning">
+
+Note that the `$runCommandRaw` command should not be used for queries which contain the `"find"` or `"aggregate"` commands, as you may be unable to fetch all data. This is because MongoDB returns a [cursor](https://docs.mongodb.com/manual/tutorial/iterate-a-cursor/) that is attached to your MongoDB session, and you may not hit the same MongoDB session every time. For these queries, you should use the specialised [`findRaw`](#findraw) and [`aggregateRaw`](#aggregateraw) methods instead.
```
matthewmueller, line 429: Nice warning!

content/200-concepts/100-components/02-prisma-client/090-raw-database-access.mdx
matthewmueller: I'd suggest just removing it and not talking about it in the paragraph above.

keerlu: In that case we'll need a different example, right? I think Dom's idea was to provide a query that isn't something the Prisma client would allow you to do, as a use case for raw queries. So `bypassDocumentValidation` is included to allow two entries with the same id.

Happy to put in something else if you have suggestions? Otherwise I can explain more clearly that this isn't something you'd normally do, etc.

matthewmueller: Ah, I didn't realize this property was necessary for the example to work.

Still, I think we can come up with a better example. Untested, but I'd suggest we document the [findAndModify](https://docs.mongodb.com/manual/reference/method/db.collection.findAndModify/#mongodb-method-db.collection.findAndModify) command. Prisma's upsert is not nearly as powerful.


Untested, but something like:
```ts
prisma.$runCommandRaw({
  findAndModify: 'Pets',
  query: { age: { $gt: 10 } },
  update: { type: "Old Cat" }
})
```

Let me know if you want me to test that the code above works as expected. I'll be a bit held up until after the release, but can check it after. In the meantime, feel free to merge this as-is and see about improving the example after.

keerlu: Thanks! I'll leave the current one for the release, and we can improve this later.

PR #2874, opened 2022-02-16
fix(data-model): Optional and mandatory fields with MongoDB

janpio:
The current wording makes it sound like a problem of the connector, but it is just something that does not exist on the MongoDB database level so can not be supported.

Discussion at: https://prisma-company.slack.com/archives/C013RMLGE72/p1645025663925709


content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
```diff
@@ -526,14 +526,9 @@ model Tag {
...
-- **Relational databases**: Required fields are represented via `NOT NULL` constraints in the underlying database.
-
-<Admonition type="info">
-
-The MongoDB connector does not enforce constraints at database level.
-
-</Admonition>
-
+- **Databases**
+  - **Relational databases**: Required fields are represented via NOT NULL constraints in the underlying database.
```
janpio:
```suggestion
  - **Relational databases**: Required fields are represented via `NOT NULL` constraints in the underlying database.
```

PR #2878, opened 2022-02-16
fix(referential-actions): Combine "types" tables into one

janpio:
No need to have MongoDB separate from other databases.


tanberry:
Thank you @Jan; having it all in one table, and still saying when Mongo was introduced, is great.

PR #2885, opened 2022-02-16
Update 09-indexes.mdx

tanberry:
Removed MySQL from an admonition about not supporting full text search via Client.


janpio (commented):
While the change is correct, I think this actually requires a link to the search API for MySQL where it is supported to make sense. Probably before the Admonition.

tanberry:
I'm going to merge this one just to have the correct info about MySQL supporting full text search, and will open a separate issue to address your point @Jan. Though, I am not sure where you mean we should link to...? "...requires a link to the search API for MySQL where it is supported..."

PR #2887, opened 2022-02-16
feat: Scalar Types Reference

janpio:
Proof of concept of an idea to split out the scalar types details into its own page that is just linked from the schema reference instead.

Relevant previews:

- [/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types](https://deploy-preview-2887--prisma2-docs.netlify.app/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types)
- [/docs/reference/api-reference/prisma-scalar-types-reference](https://deploy-preview-2887--prisma2-docs.netlify.app/docs/reference/api-reference/prisma-scalar-types-reference)


janpio:
I still think this is a good idea and we should do that. Seems has to be redone though, as the content that is split out here probably changed and that is hard to fix via conflict UI in GitHub.

nikolasburk:
Maybe turn this into an issue instead if we need to redo the PR anyways?

nikolasburk:
Tracking this in an issue: https://github.com/prisma/docs/issues/5496

PR #2908, opened 2022-02-19
Update 03-postgresql.mdx

danasmith242:
Fixed typo, removed extra character.


tanberry:
Thanks @danasmith242 for catching that!

PR #2914, opened 2022-02-21
Update 055-aggregation-grouping-summarizing.mdx

krychaxp:
replace `avg` to `_avg`


tanberry:
Thank you @krychaxp we will run a quick tech review on this, and them get it merged. Those leading underscores are important, thanks for helping us out.

PR #2935, opened 2022-03-03
Fix Typo

hariombalhara:
## Describe this PR

Fixes a Typo `managed` => `managees`


janpio:
Is that really a typo (as it is used different somewhere else) or just another choice of words? Both work in this situation I would say.

hariombalhara:
@janpio Yes in the context it is a typo. I got confused by it. You can see that here it is trying to explain how the error can be fixed by updating the `@relation` attribute for field `manager` and suddenly `managees` also got changed to `managed`.

See the screenshot for reference.

Screenshot 2022-03-03 at 2 52 28 PM

janpio:
Thanks, great catch! (I think we recently clarified the wording here a bit, and seems we forgot a few bits)

Someone from our docs team will see if this affects other examples or text here as well to make sure it is now all uniform.

tanberry:
Good catch, @hariombalhara ! Yes in a sentence above the fix you made, we use the term "managees", so let's go with that. 

"*The following model describes a self-relation where an Employee can have a manager and managees, referencing entries of the same model.*"

And for sure @janpio I'll take a look and make sure we are consistent elsewhere.

PR #2954, opened 2022-03-05
feat: redirect anchor links using browser apis

### LINKED ISSUE #2757: Redirects for link anchors
Example why this would be useful: https://github.com/prisma/prisma/pull/11338

nilubava:
Fixes #2757

### DIFF
--- a/config.js
+++ b/config.js
@@ -26,6 +26,10 @@ const config = {
       from: '/concepts/components/prisma-data-platform#prisma-data-proxy',
       to: '/concepts/data-platform/data-proxy',
     },
+    {
+      from: '/concepts/overview/what-is-prisma/data-modeling#data-modeling-without-sometext',
+      to: '/concepts/overview/what-is-prisma/data-modeling#data-modeling-without-prisma',
+    },
   ],
   header: {
     secondLevelHeaderMenuItems: [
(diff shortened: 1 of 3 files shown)


janpio:
https://deploy-preview-2954--prisma2-docs.netlify.app/docs/concepts/components/prisma-data-platform#prisma-data-proxy returns a 404, then redirects client side to the new page 👍 

Would that also work _inside_ the same page? Does it also work if the original page still exists?

nilubava:
@janpio I completely missed your queries - It does show a 404 flash and I will dig into this deeper. It also works if the original page still exists. 
> Would that also work inside the same page?
 
I didnt exactly understand what you meant here ☝️

janpio:
> It does show a 404 flash and I will dig into this deeper.

I don't think you can avoid this, as the page is literally not there any more and it needs to load the 404 page to be able to execute the javascript that does the redirection.



> > Would that also work inside the same page?
> 
> I didnt exactly understand what you meant here ☝️

Can I redirect from one anchor/headline in a document (that might not exist any more as we renamed the headline) to another one?

tanberry:
@nilubava do we know yet if this work will indeed allow us to redirect from one heading to another heading within the same document?

Is this ready to be merged?

nilubava:
@tanberry Sorry for the delay in replying. It looks like "redirect from one heading to another heading within the same document" is not supported in this. I will work on this

nilubava:
@tanberry It should redirect from one heading to another heading within the same document now. Please test and let me know

PR #2962, opened 2022-03-10
Add query logging documentation for MongoDB

keerlu:
## Describe this PR

Add example of query logging for MongoDB.

May want to change the example


content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/130-logging.mdx
```diff
@@ -112,4 +116,59 @@ Duration: 2ms
...
+</cmd>
+<cmdResult>
+
+```terminal
+Query: db.User.deleteMany({ _id: ( $in: [ “62261e0b18139c6099ba7097”, ], }, })
+Params: [0]
+Duration: 3ms
+Query: db.User.deleteMany({ _id: ( $in: [ “6226277a96069500743edcf9”, ], }, })
+Params: [2, 7, 18, 29]
+Duration: 3ms
```
janpio: The two queries currently seem to include the params, that is probably not correct (or MongoDB logging works different than I would expect)

keerlu: Oops, thanks, that was a copy paste error (I haven't been able to test this yet). Removed the params for now

cprieto: both properties (duration and params) are passed but they will be `undefined`, see 
image

cprieto (approved):

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/130-logging.mdx
keerlu: Ah, ok. I've removed both parameters from the example and added a note to the reference page that these are undefined in MongoDB.

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/130-logging.mdx
```diff
@@ -65,6 +65,10 @@ To use event-based logging:
 
 The following example subscribes to all `query` events and write the `duration` and `query` to console:
 
+<TabbedContent tabs={[<FileWithIcon text="Relational databases" icon="database"/>, <FileWithIcon text="MongoDB" icon="database"/>]}>
+
+<tab>
+
 <CodeWithResult expanded="{true}">
 <cmd>
```
janpio, line 74: In the line below this one, the end of the second highlight should be 26 and not 25.
image

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/130-logging.mdx
```diff
@@ -65,6 +65,10 @@ To use event-based logging:
 
 The following example subscribes to all `query` events and write the `duration` and `query` to console:
 
+<TabbedContent tabs={[<FileWithIcon text="Relational databases" icon="database"/>, <FileWithIcon text="MongoDB" icon="database"/>]}>
```
janpio, line 68: Distance below the tab to the code is surprisingly big:
image

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -156,6 +156,8 @@ export type QueryEvent = {
 }
 ```
 
+Note that for MongoDB, the `params` and `duration` fields will be undefined.
```
janpio, line 159: I expect that type above is not correct any more then, correct @millsp?

janpio:
Unrelated comment to the edited section: The "Hide CLI results" is wrong for the output example. I think we can override that with a custom string in the component.

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/130-logging.mdx
keerlu: Hm, maybe the TabbedContent and CodeWithResult components don't play together well? (I wasn't sure whether it would work at all...)

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
millsp: `NaN` is a `number` :trollface: but `undefined` is not a `string`, so that one is incorrect yes.

keerlu:
Tested query logging this morning and changed the example logged queries to be a bit more interesting/similar to the SQL ones

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/130-logging.mdx
keerlu: Done

keerlu:
> Unrelated comment to the edited section: The "Hide CLI results" is wrong for the output example. I think we can override that with a custom string in the component.

@janpio : wrong in what way? Shouldn't be able to hide at all, or the text should be different?

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: What should I update it to?

millsp: No worries, you can leave it as is because it is technically correct. We will need to fix the types after the release https://github.com/prisma/prisma/issues/12319

janpio:
The text should be different. This not CLI output, but output of an imaginary script running if at all.

keerlu:
> The text should be different. This not CLI output, but output of an imaginary script running if at all.

Hm, I don't understand. I was testing this by running it in the terminal, is that not CLI output?

image

janpio:
No. This is output in the terminal. The CLI usually refers to "Prisma CLI" in our context, which means you run `npx prisma`. This here runs `npx ts-node index.ts`, so he CLI of `npx` to run the package `ts-node` with the file `index.ts`. But that is not what we usually refer to as "CLI".

PR #2963, opened 2022-03-10
Wrong variable name with TRUNCATE example in MySQL

Exilz:
## Describe this PR

The example regarding how to truncate tables is wrong in MySQL. The deconstructed variable doesn't match the name of the column that's being selected in the SQL query.

## Changes

- Fixed the deconstruction in mySQL truncate example

@all-contributors please add @Exilz for doc


tanberry:
Hey thank you very much @Exilz !!

PR #2973, opened 2022-03-14
Fix the type of `_count` argument of `groupBy`

aqrln:
## Describe this PR

Fix the type of `_count` argument of `groupBy`.

## Changes

The correct type was copied from the "Reference" section below the table.

## What issue does this fix?

n/a

A user didn't see the "Reference" section and the example below the table and opened an issue in the main repo (https://github.com/prisma/prisma/issues/10915) because they thought there was a bug and the type was incorrect in the generated client.


aqrln:
Now the problem is the type contains the `|` character, which breaks the table as is, but if I try to escape it with a backslash, it displays as a literal `\|` combination.

aqrln:
I split the type into two code spans with `|` in regular font between them as a workaround

Jolg42 (approved):

keerlu (approved):

keerlu:
Thanks for fixing the type @aqrln - that's annoying about the formatting but at least we have a workaround

PR #2983, opened 2022-03-15
Fix numbering on `SQL views`

frec-mike:
## Describe this PR

Current numbering jumps from 6 to 9. This just fixes that.


keerlu:
Thanks for fixing this, @frec-mike!

PR #2999, opened 2022-03-22
Add composite types to client reference

### LINKED ISSUE #2939: Add embedded types content to Reference section (Epic 2)
Add to Client API Reference

keerlu:
Draft of reference material for composite types

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -4316,7 +4316,7 @@ Use `unset` to unset the value of a composite type. Unlike `set: null`, this rem
 
 #### Examples
 
-##### Remove the <inlinecode>billingAddress</inlinecode> from an order
+##### Remove the <inlinecode>billingAddress</inlinecode> from an <inlinecode>order</inlinecode>
 
 ```ts
 const order = await prisma.order.update({
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -4463,7 +4463,7 @@ const orders = await prisma.order.findMany({
 })
 ```
 
-##### Find products with photos that match the given <inlinecode>url</inlinecode>s
+##### Find products with photos that match all of a list of <inlinecode>url</inlinecode>s
 
 ```ts
 const product = prisma.product.findMany({
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -4257,8 +4257,6 @@ Available for MongoDB only in Prisma `3.10.0` and later.
 
 Composite type methods allow you to create, update and delete [composite types](/concepts/components/prisma-client/composite-types).
 
-### Reference
-
 ### <inlinecode>set</inlinecode>
 
 Use `set` to overwrite the value of a composite type.
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -4276,7 +4276,7 @@ photos: [
 
 #### Examples
 
-##### Set the <inlinecode>shippingAddress</inlinecode> composite type within a new order
+##### Set the <inlinecode>shippingAddress</inlinecode> composite type within a new <inlinecode>order</inlinecode>
 
 ```ts
 const order = await prisma.order.create({
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -4249,41 +4249,177 @@ const posts = await prisma.post.findMany({
 
 ## Composite type methods
 
+<Admonition type="warning">
+
+Available for MongoDB only in Prisma 3.12.0 and later.
+
+</Admonition>
+
+Composite type methods allow you to create, update and delete [composite types](/concepts/components/prisma-client/composite-types).
+
 ### Reference
 
 ### <inlinecode>set</inlinecode>
 
+Use `set` to overwrite the value of a composite type.
+
 #### Remarks
 
-- `set` is optional - you can set the value directly:
+- The `set` keyword is optional - you can set the value directly:
+
+```ts
+photos: [
+  { height: 100, width: 200, url: '1.jpg' },
+  { height: 100, width: 200, url: '2.jpg' },
+]
+```
 
 #### Examples
+
+##### Set the <inlinecode>shippingAddress</inlinecode> composite type within a new order
+
+```ts
+const order = await prisma.order.create({
+  data: {
+    // Normal relation
+    product: { connect: { id: 'some-object-id' } },
+    color: 'Red',
+    size: 'Large',
+    // Composite type
+    shippingAddress: {
+      set: {
+        street: '1084 Candycane Lane',
+        city: 'Silverlake',
+        zip: '84323',
+      },
+    },
+  },
+})
+```
+
+##### Set an optional composite type to <inlinecode>null</inlinecode>
+
+```ts
+const order = await prisma.order.create({
+  data: {
+    // Embedded optional type, set to null
+    billingAddress: {
+      set: null,
+    },
+  },
+})
+```
 
 ### <inlinecode>unset</inlinecode>
 
-#### Remarks
+Use `unset` to unset the value of a composite type. Unlike `set: null`, this removes the field entirely.
 
 #### Examples
+
+##### Remove the <inlinecode>billingAddress</inlinecode> from an order
+
+```ts
+const order = await prisma.order.update({
+  where: {
+    id: 'some-object-id',
+  },
+  data: {
+    billingAddress: {
+      // Unset the billing address
+      // Removes "billingAddress" field from order
+      unset: true,
+    },
+  },
+})
+```
 
 ### <inlinecode>update</inlinecode>
 
-#### Remarks
+Use `update` to update fields within a composite type.
 
 #### Examples
+
+##### Update the zip field of a <inlinecode>shippingAddress</inlinecode> composite type
+
+```ts
+const order = await prisma.order.update({
+  where: {
+    id: 'some-object-id',
+  },
+  data: {
+    shippingAddress: {
+      // Update just the zip field
+      update: {
+        zip: '41232',
+      },
+    },
+  },
+})
+```
 
 ### <inlinecode>upsert</inlinecode>
 
-#### Remarks
+Use `upsert` to update an existing composite type if it exists, and otherwise set the composite type.
 
 #### Examples
+
+##### Create a new <inlinecode>billingAddress</inlinecode> if it doesn't exist, and otherwise update it
+
+```ts
+const order = await prisma.order.update({
+  where: {
+    id: 'some-object-id',
+  },
+  data: {
+    billingAddress: {
+      // Create the address if it doesn't exist,
+      // otherwise update it
+      upsert: {
+        set: {
+          street: '1084 Candycane Lane',
+          city: 'Silverlake',
+          zip: '84323',
+        },
+        update: {
+          zip: '84323',
+        },
+      },
+    },
+  },
+})
+```
 
 ### <inlinecode>push</inlinecode>
 
-#### Remarks
+Use `push` to push values to the end of a list of composite types.
 
 #### Examples
 
+##### Add a new photo to the <inlinecode>photos</inlinecode> list
+
+```ts
+const product = prisma.product.update({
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+
+#### Remarks
+
+- The `set` keyword is optional - you can set the value directly:
+
+```ts
+photos: [
+  { height: 100, width: 200, url: '1.jpg' },
+  { height: 100, width: 200, url: '2.jpg' },
+]
```
janpio, line 4272: The first example beign a special case is somewhat weird.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+})
+```
+
+### <inlinecode>unset</inlinecode>
+
+Use `unset` to unset the value of a composite type. Unlike `set: null`, this removes the field entirely.
+
+#### Examples
+
+##### Remove the <inlinecode>billingAddress</inlinecode> from an order
```
janpio:
```suggestion
##### Remove the <inlinecode>billingAddress</inlinecode> from an <inlinecode>order</inlinecode>
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+```ts
+photos: [
+  { height: 100, width: 200, url: '1.jpg' },
+  { height: 100, width: 200, url: '2.jpg' },
+]
+```
+
+#### Examples
+
+##### Set the <inlinecode>shippingAddress</inlinecode> composite type within a new order
```
janpio:
```suggestion
##### Set the <inlinecode>shippingAddress</inlinecode> composite type within a new <inlinecode>order</inlinecode>
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+<Admonition type="warning">
+
+Available for MongoDB only in Prisma 3.12.0 and later.
+
+</Admonition>
+
+Composite type methods allow you to create, update and delete [composite types](/concepts/components/prisma-client/composite-types).
+
+### Reference
+
```
keerlu: I'm missing a type signature here

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: I was following the format of the remarks for scalar list methods. Not sure I like it that much either

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+Available for MongoDB only in Prisma 3.12.0 and later.
+
+</Admonition>
+
+Composite type filters allow you to filter the contents of [composite types](/concepts/components/prisma-client/composite-types).
+
+### Reference
+
+```ts file=index.d.ts
```
keerlu, line 4432: Need to check these types are correct

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
 })
 ```
 
+## Composite type methods
+
+<Admonition type="warning">
+
+Available for MongoDB only in Prisma 3.12.0 and later.
```
dpetrick: This is not exactly right, we included a lot of of those in 3.10 already.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+    billingAddress: {
+      set: null,
+    },
+  },
+})
+```
+
+### <inlinecode>unset</inlinecode>
+
+Use `unset` to unset the value of a composite type. Unlike `set: null`, this removes the field entirely.
```
dpetrick: I'd be precise here and say "this removes the field entirely from the [matching?] Mongo document."

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+        zip: '41232',
+      },
+    },
+  },
+})
+```
+
+### <inlinecode>upsert</inlinecode>
+
+Use `upsert` to update an existing composite type if it exists, and otherwise set the composite type.
```
dpetrick: A few things:
- `upsert` can only be used on optional composites (`com_field Composite?`).
- The above `update` **can't** be used to update values in an optional composite (this info is also missing above).

There will be questions why people are not able to simply `update` an optional composite and must use upsert. so here it is: The problem is with `null` or `undefined` (`undefined` == doesn't exist on the document!) is the question of what the `update` operation is supposed to do when it hits one of those values. The options were:
- Error out
- Create partial data
- Do not offer `update`

Erroring out is a terrible idea and is an obvious design problem with the API. Partial data doesn't work either because there's no guarantee that you can create a valid composite out of the partial update data. A clear example of that is:
```prisma
model Test {
    ...
    composite Composite?
}

type Composite {
    req String
    opt String?
}
```

If you update the `opt` field in your operation, but hit a `null`/`undefined`, we can't construct a valid `Composite` because `req` is _required_ to be present for it to be valid.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+    },
+  },
+})
+```
+
+## Composite type filters
+
+<Admonition type="warning">
+
+Available for MongoDB only in Prisma 3.12.0 and later.
```
dpetrick: 3.11

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
dpetrick: Is this a common thing to have in our docs, these type definitions? Seems odd to me, as it is outdated quickly and doesn't add much over your editor showing you the types.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+  every?: PhotoWhereInput
+  some?: PhotoWhereInput
+  none?: PhotoWhereInput
+  isEmpty?: boolean
+}
+```
+
+### <inlinecode>equals</inlinecode>
+
+Use `equals` to filter results by matching a composite type or a list of composite types. Requires all fields of the composite type to match.
```
dpetrick: Omits a few details that should be included:
- All required fields of the composite are required for `equals`.
- If you omit an optional field, it will match with `undefined` (missing) fields of the document, but NOT null. You need to be precise in what you're querying for.
- Same goes for `null`: If you say `equals: { ... field: null ... }` then it will only match documents with an explicit null value for that field, not `undefined`.
- Order of fields and order of objects is important. This directly mirrors how Mongo handles comparisons. If you compare `{ "a": "1", "b": "2" }` to `{ "b": "2", "a": "1" }` it will fail, same with lists (which is probably more obvious): `[ { "a": 1 }, { "a": 2 } ] != [ { "a": 2 }, { "a": 1 } ]`

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+        street: '555 Candy Cane Lane',
+        city: 'Wonderland',
+        zip: '52337',
+      },
+    },
+  },
+})
+```
+
+##### Find products with photos that match the given <inlinecode>url</inlinecode>s
```
dpetrick: Headline doesn't really fit the example, does it?

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,353 @@ const posts = await prisma.post.findMany({
...
+        { height: 200 },
+      }
+    }
+  },
+})
+```
+
+### <inlinecode>some</inlinecode>
+
+Use `some` to filter for lists of composite types where some items in the list match the condition.
```
dpetrick: "one or more items" or "at least one item"

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: It's used throughout this Client reference page, not sure about elsewhere.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Done

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Fixed

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Done

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Done

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: I've now put 3.10.0 for the composite type methods, and 3.11.0 for the filters

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Done

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Not sure where to find this, can someone point me to it?

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: I've added optional/required restrictions to upsert/update. Will go into the reasoning if I get time.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Via @millsp 'There aren't any composite type methods, since they are always embedded in a model, so they are always created, updated, deleted via their parent model.' So I'll remove this section.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,371 @@ const posts = await prisma.post.findMany({
 })
 ```
 
+## Composite type methods
```
millsp, line 4250: Should we call this a method? I'm lending more towards calling it "operations". wdyt?

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -4247,6 +4247,371 @@ const posts = await prisma.post.findMany({
...
+
+<Admonition type="warning">
+
+Available for MongoDB only in Prisma `3.10.0` and later.
+
+</Admonition>
+
+Composite type methods allow you to create, update and delete [composite types](/concepts/components/prisma-client/composite-types).
```
millsp, line 4258: Should we remove this too?

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Ah, I was going for consistency with scalar field methods and filters elsewhere in the reference. Will leave for now as may have to be changed in a number of places, but happy to update later.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Sorry, which bit needs removing?

millsp: Same remark as above, we (client team) call it operations so I found it confusing.

PR #3001, opened 2022-03-22
Clarifying what is 'mydb'

Gui-Alucard:
Is just an improvement to help those are new with technology

### DIFF
--- a/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
+++ b/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
@@ -80,7 +80,7 @@ datasource db {
 In this case, the `url` is [set via an environment variable](/concepts/components/prisma-schema#accessing-environment-variables-from-the-schema) which is defined in `.env`:
 
 ```bash file=.env
-DATABASE_URL="mysql://johndoe:randompassword@localhost:3306/mydb"
+DATABASE_URL="mysql://johndoe:randompassword@localhost:3306/mydbschema"
 ```
 
 You now need to adjust the connection URL to point to your own database.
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
`mydb` is the "database" in MySQL context, not a "schema". Schemas as object do not really exist there.

Gui-Alucard:
> 
@janpio 
If you are using MySQL Workbench, then where is 'mydb' you need to type the name of the schema to connect. I just tried that and decide to bring this up.
Captura de Tela 2022-03-22 às 17 15 03

janpio:
That is unfortunate and confusing wording from MySQL, which unfortunately is all over the place. In their docs they also call the `mysql` database a "schema", or say things like "INFORMATION_SCHEMA is a database" further muddling the two terms.

From Prisma perspective MySQL only has the concept of a "database" that is defined after the hostname in the connection string.

Gui-Alucard:
> That is unfortunate and confusing wording from MySQL, which unfortunately is all over the place. In their docs they also call the `mysql` database a "schema", or say things like "INFORMATION_SCHEMA is a database" further muddling the two terms.
> 
> From Prisma perspective MySQL only has the concept of a "database" that is defined after the hostname in the connection string.

Thank you to teach me that =)
I wrote a note about that instead those changes.

content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
```diff
@@ -85,6 +85,8 @@ DATABASE_URL="mysql://johndoe:randompassword@localhost:3306/mydb"
 
 You now need to adjust the connection URL to point to your own database.
 
+  > **_NOTE:_**  In MySQL docs the mysql database ('mydb' in this case) is called as "schema" too.
```
tanberry:
```suggestion
  > **_NOTE:_**  In MySQL documentation the MySQL database ('mydb' in this case) is labeled as "schema" too.
```

tanberry:
Thanks @Gui-Alucard for your help in clarifying this! I did tweak your wording a little; if you are ok with that change please add it and then we can merge this PR. Sound ok to you @janpio?

janpio:
No, this is not correct unfortunately. They only call the `mysql` database (which is the system database which has a specific function) a schema. Both the current suggestion from @Gui-Alucard and your edit suggestion make it sound like if the "database" was a "schema" - and that is not the case. Sorry.

(Open for MySQL documentation links that teach and show me otherwise of course.)

Gui-Alucard:
> No, this is not correct unfortunately. They only call the `mysql` database (which is the system database which has a specific function) a schema...

@janpio I followed this suggestion and did a new commit. I am concerned that I've been naughty and causing problems, I just wanted to clear things up for new developers.

janpio:
You are not naughty or causing problems at all, I just do not agree with the way you phrase this. For me (and according to MySQL documentation) in generall what comes after the hostname is a "database". (It might work a bit like PostgreSQL schemas, but it is still called "database" in context of MySQL usually. Unfortunately there is some inconsistency in tools like MySQL Workbench or on some special databases in their documentation - but nothing that implies that you generally call the things which has the value "mydb" here a "schema")

Gui-Alucard:
I agree with you and understand your point. So, seeing that (MySQL inconsistency) like a limitation I will close this PR though.

janpio:
Thanks for understanding. When we ever understand the terminology properly, we might write a page in our [DataGuide](https://www.prisma.io/dataguide/) and link to it from everywhere. But right now we are still pretty much confused ourselves about the inconsistencies.

PR #3004, opened 2022-03-23
Renamed db push concept page

floelhoeffel:
Under concepts > migrate there was a page titled `db push`. As this doesnt mean much to the user, I renamed it "Schema prototyping".


floelhoeffel:
I reverted the change so it only affects the page title and none of the URLs. I think this could be merged as is.

janpio (approved):
Looks good as an instant fix

tanberry:
Thank you so much @floelhoeffel ! A very sensible change to a much better and more descriptive title.

PR #3013, opened 2022-03-28
Add type mappings to Mongo connector docs

### LINKED ISSUE #2738: Document MongoDB type mapping (Epic 2)
We have a table mapping how types for most databases map to the Prisma Schema types. Here's the [Postgres to Prisma Schema table](https://www.prisma.io/docs/concepts/database-connectors/postgresql#type-mapping-between-postgresql-to-prisma-schema)

MongoDB doesn't have one: https://www.prisma.io/docs/concepts/database-connectors/mongodb

In code, we have the following mappings: https://github.com/prisma/prisma-engines/blob/ccf3dc944acdabb431947150e12b984b34c538cd/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/field_type.rs#L29-L58

Let me know if you have trouble parsing this.

keerlu:
## What issue does this fix?

Fixes #2738

### DIFF
--- a/content/200-concepts/200-database-connectors/07-mongodb.mdx
+++ b/content/200-concepts/200-database-connectors/07-mongodb.mdx
@@ -229,7 +229,7 @@ When introspecting a MongoDB database, Prisma uses the relevant [scalar types](/
 
 | MongoDB (Type \| Aliases) | Prisma   | Supported | Native database type attribute | Notes |
 | ------------------------- | -------- | :-------: | :----------------------------- | :---- |
-| `ObjectId`                | `String` |    ✔️     | `@db.ObjectId`                 |       |
+| `objectId`                | `String` |    ✔️     | `@db.ObjectId`                 |       |
 
 [Introspection](/concepts/components/introspection) adds native database types that are **not yet supported** as [`Unsupported`](/reference/api-reference/prisma-schema-reference#unsupported) <span class="api"></span> fields:
 
--- a/content/200-concepts/200-database-connectors/07-mongodb.mdx
+++ b/content/200-concepts/200-database-connectors/07-mongodb.mdx
@@ -209,66 +209,6 @@ The MongoDB connector maps the [scalar types](/concepts/components/prisma-schema
 
 When introspecting a MongoDB database, the database types are mapped to Prisma according to the following table:
 
-| MongoDB (Type \| Aliases)                   | Prisma        | Supported | Native database type attribute           | Notes                                                                                                                                                                   |
-| ------------------------------------------- | ------------- | :-------: | :--------------------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `bigint` \| `int8`                          | `BigInt`      |    ✔️     | `@db.BigInt`\*                           | \*Default mapping for `BigInt` - no type attribute added to schema.                                                                                                     |
-| `boolean` \| `bool`                         | `Bool`        |    ✔️     | `@db.Boolean`\*                          | \*Default mapping for `Bool` - no type attribute added to schema.                                                                                                       |
-| `timestamp with time zone` \| `timestamptz` | `DateTime`    |    ✔️     | `@db.Timestamptz(x)`                     |
-| `time without time zone` \| `time`          | `DateTime`    |    ✔️     | `@db.Time(x)`                            |
-| `time with time zone` \| `timetz`           | `DateTime`    |    ✔️     | `@db.Timetz(x)`                          |
-| `numeric(p,s)` \| `decimal(p,s)`            | `Decimal`     |    ✔️     | `@db.Decimal(x, y)`                      |
-| `real` \| `float`, `float4`                 | `Float`       |    ✔️     | `@db.Real`                               |
-| `double precision` \| `float8`              | `Float`       |    ✔️     | `@db.DoublePrecision`\*                  | \*Default mapping for `Float` - no type attribute added to schema.                                                                                                      |
-| `smallint` \| `int2`                        | `Int`         |    ✔️     | `@db.SmallInt`                           |                                                                                                                                                                         |
-| `integer` \| `int`, `int4`                  | `Int`         |    ✔️     | `@db.Int`\*                              | \*Default mapping for `Int` - no type attribute added to schema.                                                                                                        |
-| `smallserial` \| `serial2`                  | `Int`         |    ✔️     | `@db.SmallInt @default(autoincrement())` |
-| `serial` \| `serial4`                       | `Int`         |    ✔️     | `@db.Int @default(autoincrement())`      |
-| `bigserial` \| `serial8`                    | `Int`         |    ✔️     | `@db.BigInt @default(autoincrement()`    |
-| `character(n)` \| `char(n)`                 | `String`      |    ✔️     | `@db.Char(x)`                            |
-| `character varying(n)` \| `varchar(n)`      | `String`      |    ✔️     | `@db.VarChar(x)`                         |
-| `money`                                     | `Decimal`     |    ✔️     | `@db.Money`                              |
-| `text`                                      | `String`      |    ✔️     | `@db.Text`\*                             | \*Default mapping for `String` - no type attribute added to schema.                                                                                                     |
-| `timestamp`                                 | `DateTime`    |    ✔️     | `@db.TimeStamp`\*                        | \*Default mapping for `DateTime` - no type attribute added to schema.                                                                                                   |
-| `date`                                      | `DateTime`    |    ✔️     | `@db.Date`                               |
-| `enum`                                      | `Enum`        |    ✔️     | N/A                                      |
-| `inet`                                      | `String`      |    ✔️     | `@db.Inet`                               |
-| `bit(n)`                                    | `String`      |    ✔️     | `@Bit(x)`                                |
-| `bit varying(n)`                            | `String`      |    ✔️     | `@VarBit`                                |
-| `oid`                                       | `Int`         |    ✔️     | `@db.Oid`                                |
-| `uuid`                                      | `String`      |    ✔️     | `@db.Uuid`                               |
-| `json`                                      | `Json`        |    ✔️     | `@db.Json`                               |
-| `jsonb`                                     | `Json`        |    ✔️     | `@db.JsonB`\*                            | \*Default mapping for `Json` - no type attribute added to schema.                                                                                                       |
-| `bytea`                                     | `Bytes`       |    ✔️     | `@db.ByteA`\*                            | \*Default mapping for `Bytes` - no type attribute added to schema.                                                                                                      |
-| `xml`                                       | `String`      |    ✔️     | `@db.Xml`                                |
-| Array types                                 | `[]`          |    ✔️     |
-| `citext`                                    | `String`      |   ✔️\*    | `@db.Citext`                             | \* Only available if [Citext extension is enabled](/concepts/components/prisma-schema/features-without-psl-equivalent#enable-extensions-for-native-database-functions). |
-| `interval`                                  | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `cidr`                                      | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `macaddr`                                   | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `tsvector`                                  | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `tsquery`                                   | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `int4range`                                 | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `int8range`                                 | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `numrange`                                  | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `tsrange`                                   | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `tstzrange`                                 | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `daterange`                                 | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `point`                                     | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `line`                                      | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `lseg`                                      | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `box`                                       | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `path`                                      | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `polygon`                                   | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| `circle`                                    | `Unsupported` |  Not yet  |                                          |                                                                                                                                                                         |
-| Composite types                             | n/a           |  Not yet  |                                          |                                                                                                                                                                         |
-| Domain types                                | n/a           |  Not yet  |                                          |                                                                                                                                                                         |
-
-[Introspection](/concepts/components/introspection) adds native database types that are **not yet supported** as [`Unsupported`](/reference/api-reference/prisma-schema-reference#unsupported) <span class="api"></span> fields:
-
-```prisma file=schema.prisma
-model Device {
-  id   Int                   @id @default(autoincrement())
-  name String
-  data Unsupported("circle")
-}
-```
+| MongoDB (Type \| Aliases) | Prisma     | Supported | Native database type attribute | Notes |
+| ------------------------- | ---------- | :-------: | :----------------------------- | :---- |
+| `ObjectId`                | `ObjectId` |    ✔️     | `@db.ObjectId`                 |       |
--- a/content/200-concepts/200-database-connectors/07-mongodb.mdx
+++ b/content/200-concepts/200-database-connectors/07-mongodb.mdx
@@ -197,17 +197,18 @@ There's also an option to run the replica set locally with this guide: https://d
 
 The MongoDB connector maps the [scalar types](/concepts/components/prisma-schema/data-model#scalar-fields) from the Prisma [data model](/concepts/components/prisma-schema/data-model) to MongoDB's native column types as follows:
 
-| Prisma     | MongoDB               |
-| ---------- | --------------------- |
-| `String`   | `String`              |
-| `Boolean`  | `Bool`                |
-| `Int`      | `Int`                 |
-| `BigInt`   | `Long`                |
-| `Float`    | `Double`              |
+| Prisma     | MongoDB                                                                |
+| ---------- | ---------------------------------------------------------------------- |
+| `String`   | `string`                                                               |
+| `Boolean`  | `bool`                                                                 |
+| `Int`      | `int`                                                                  |
+| `BigInt`   | `long`                                                                 |
+| `Float`    | `double`                                                               |
 | `Decimal`  | [Currently unsupported](https://github.com/prisma/prisma/issues/12637) |
-| `Datetime` | `DateTime`            |
-| `Json`     |                       |
-| `Bytes`    | `BinData`             |
+| `DateTime` | `timestamp`                                                            |
+| `Date`     | `date`                                                                 |
+| `Bytes`    | `binData`                                                              |
+| `Json`     |                                                                        |
 
 MongoDB types that are currently unsupported:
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,36 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+| Prisma    | MongoDB      |
+| --------- | ------------ |
+| `String`  | `String`     |
+| `Bool`    | `Boolean`    |
+| `Double`  | `Double`     |
+| `Int32`   | `Int32`      |
+| `Int64`   | `Int64`      |
+| `Decimal` | `Decimal128` |
+| `Date`    | `DateTime`   |
+| `BinData` | `Binary`     |
```
keerlu: Need to add Json to this list - how do I interpret the mapping [here](https://github.com/prisma/prisma-engines/blob/ccf3dc944acdabb431947150e12b984b34c538cd/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/field_type.rs#L37)?

Also, is the [mapping in the schema reference](http://localhost:8000/reference/api-reference/prisma-schema-reference#mongodb-preview-7) still correct?

dpetrick: - You don't interpret this mapping at all in my opinion, this is internal conversion code and uses a non-Prisma schema type typing. Prisma schema doesn't have `Int32` or `Int64`, for example.

- Second link goes to localhost.

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,36 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+The MongoDB connector maps the [scalar types](/concepts/components/prisma-schema/data-model#scalar-fields) from the Prisma [data model](/concepts/components/prisma-schema/data-model) to MongoDB's native column types as follows:
+
+| Prisma    | MongoDB      |
+| --------- | ------------ |
+| `String`  | `String`     |
+| `Bool`    | `Boolean`    |
+| `Double`  | `Double`     |
+| `Int32`   | `Int32`      |
+| `Int64`   | `Int64`      |
+| `Decimal` | `Decimal128` |
```
janpio: I think this is to be removed.

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,36 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
 
 There's also an option to run the replica set locally with this guide: https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set
+
+## Type mapping between MongoDB and the Prisma schema
+
+> Alternatively, see the [Prisma schema reference](/reference/api-reference/prisma-schema-reference#model-field-scalar-types) for type mappings organized by Prisma type.
+
+The MongoDB connector maps the [scalar types](/concepts/components/prisma-schema/data-model#scalar-fields) from the Prisma [data model](/concepts/components/prisma-schema/data-model) to MongoDB's native column types as follows:
+
+| Prisma    | MongoDB      |
```
janpio: The Prisma list here is inaccurate - it can only contain scalar types that Prisma supports, which would be the same terms as on https://www.prisma.io/docs/concepts/database-connectors/postgresql#prisma-migrate

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,36 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+| ------------------------- | ---------- | :-------: | :----------------------------- | :---- |
+| `ObjectId`                | `ObjectId` |    ✔️     | `@db.ObjectId`                 |       |
+
+[Introspection](/concepts/components/introspection) adds native database types that are **not yet supported** as [`Unsupported`](/reference/api-reference/prisma-schema-reference#unsupported) <span class="api"></span> fields:
+
+```prisma file=schema.prisma
+model Device {
+  id   Int                   @id @default(autoincrement())
+  name String
+  data Unsupported("circle")
```
janpio:
```suggestion
  data Unsupported("RegularExpression")
```
Not sure `Device` is a good example then any more though.

content/200-concepts/200-database-connectors/07-mongodb.mdx
dpetrick: Going by [this reference](https://www.mongodb.com/docs/v5.0/reference/bson-types/) as a base.

In my opinion, the mapping should be in reverse and be the following (the table denotes _valid_ mappings):
| MongoDB    | Prisma      |
| --------- | ------------ |
| `String`  | `String`     |
| `ObjectId`  | `String` or `Binary`     |
| `Bool`    | `Boolean`    |
| `Double`  | `Double`     |
| `Int`   | `Int`      |
| `Long`   | `Int` or `BigInt`      |
| `Date`    | `DateTime`   |
| `Timestamp`    | `DateTime`   |
| `BinData` | `Binary`     |

Unsupported MongoDB types (e.g. they do not have a valid Prisma mapping as of now):
- `Decimal128`
- `Undefined`
- `DBPointer`
- `Null`
- `Symbol`
- `MinKey`
- `MaxKey`
- `Object`
- `Javascript`
- `JavascriptWithScope`
- `Regex`

Note:
- There's also `Array` on the Mongo side, which get's translated as Prisma list type of the whatever the inner type is (e.g. `String[]`).
- Prismas `Json` type can be used to emulate an `Any` type, but will not always work. Some types and values can not be converted into JSON, see the [Extended JSON reference](https://www.mongodb.com/docs/v5.0/reference/mongodb-extended-json/).

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,36 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+| `Int64`   | `Int64`      |
+| `Decimal` | `Decimal128` |
+| `Date`    | `DateTime`   |
+| `BinData` | `Binary`     |
+
+When introspecting a MongoDB database, the database types are mapped to Prisma according to the following table:
+
+| MongoDB (Type \| Aliases) | Prisma     | Supported | Native database type attribute | Notes |
+| ------------------------- | ---------- | :-------: | :----------------------------- | :---- |
+| `ObjectId`                | `ObjectId` |    ✔️     | `@db.ObjectId`                 |       |
```
dpetrick: I don't know how accurate this is, please get a review from a schema team member on this.

dpetrick (dismissed):

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: Thanks, this ordering of the columns makes more sense to me - I was trying to match what's currently in the other connectors. The issue I'm following had the link to the prisma-engines mapping but it indeed didn't seem to map to Prisma types well.

janpio: Other connectors currently have the table under a "Migrations" headline: https://www.prisma.io/docs/concepts/database-connectors/mysql#type-mapping-between-mysql-to-prisma-schema So there it means "If I have this in my Prisma schema, what will it be translated to in my database when I migrate?"

keerlu: Ah, ok, thanks. I found the 'Migrations' headline very confusing and had no idea what it referred to (as it was just that single word). I'd taken it out, but will either replace it with a different heading or include some words of explanation of what the table means.

The heading to the second table on the other connectors, 'Native type mappings', is also confusing - native to Prisma or native to the underlying database?

keerlu: An annoying thing about having the mapping order as Prisma -> MongoDB is that there's no easy way to include MongoDB types which are unsupported in Prisma

Would these go in a separate list, if we kept the ordering the same as for the other connectors?

keerlu: I've made some updates:

- added table headings back in but made them more descriptive
- kept the mapping the current way round, and copied the list of Prisma types from the Postgres connector
- added a list of Mongo types unsupported in Prisma

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: Done as part of other edits

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: Done as part of other edits

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: I've updated it, though I don't have a good idea for an example model atm

content/200-concepts/200-database-connectors/07-mongodb.mdx
tomhoule: I think the `Prisma` column should say `String`, in the sense that the field would look like this:

```prisma
model Test {
  id String @db.ObjectId @map("_id") @default(auto()) @id
}
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+
+[Introspection](/concepts/components/introspection) adds native database types that are **not yet supported** as [`Unsupported`](/reference/api-reference/prisma-schema-reference#unsupported) <span class="api"></span> fields:
+
+```prisma file=schema.prisma
+model Example {
+  id    Int                              @id @default(autoincrement())
```
tomhoule: This schema wouldn't pass validation because it is missing the `@map("_id")`, I think.

content/200-concepts/200-database-connectors/07-mongodb.mdx
janpio: I think those tables (and all other places that describe scalar types or native types are definitely up for a complete rework, see my PR where I suggested extracting some of it into its own page: https://github.com/prisma/docs/pull/2887 I think it would be a _great_ idea to do a project soon to fix that some way and would be happy to help)

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+- Undefined
+- DBPointer
+- Null
+- Symbol
+- MinKey
+- MaxKey
+- Object
+- Javascript
+- JavascriptWithScope
+- Regex
```
janpio:
```suggestion
- `Decimal128`
- `Undefined`
- `DBPointer`
- `Null`
- `Symbol`
- `MinKey`
- `MaxKey`
- `Object`
- `Javascript`
- `JavascriptWithScope`
- `Regex`
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+The MongoDB connector maps the [scalar types](/concepts/components/prisma-schema/data-model#scalar-fields) from the Prisma [data model](/concepts/components/prisma-schema/data-model) to MongoDB's native column types as follows:
+
+| Prisma     | MongoDB               |
+| ---------- | --------------------- |
+| `String`   | `String`              |
+| `Boolean`  | `Bool`                |
+| `Int`      | `Int`                 |
+| `BigInt`   | `Long`                |
+| `Float`    | `Double`              |
+| `Decimal`  | Currently unsupported |
```
janpio: Do we have an issue for adding support again when the driver supports it @dpetrick? That would be good for users to complain and also link. (I think we can be pretty open here that MongoDB just does not support that right now and there is not much we can do unfortunately.)

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
 One simple way for this is to use [MongoDB Atlas](https://www.mongodb.com/cloud/atlas) to launch a free instance that has replica set support out of the box.
 
 There's also an option to run the replica set locally with this guide: https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set
+
+## Type mapping between MongoDB and the Prisma schema
+
+> Alternatively, see the [Prisma schema reference](/reference/api-reference/prisma-schema-reference#model-field-scalar-types) for type mappings organized by Prisma type.
+
+### Mapping from Prisma to MongoDB types on migration
```
janpio, line 196:
```suggestion
### Mapping from Prisma to MongoDB types on Migrations
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+- Null
+- Symbol
+- MinKey
+- MaxKey
+- Object
+- Javascript
+- JavascriptWithScope
+- Regex
+
+### Mapping from MongoDB to Prisma types on introspection
```
janpio:
```suggestion
### Mapping from MongoDB to Prisma types on Introspection
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+
+When introspecting a MongoDB database, the database types are mapped to Prisma according to the following table:
```
janpio: I think that introduction should mention that usually the scalar types are used, and only some special ones get additional native type annotations.

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+
+| MongoDB (Type \| Aliases) | Prisma     | Supported | Native database type attribute | Notes |
+| ------------------------- | ---------- | :-------: | :----------------------------- | :---- |
+| `ObjectId`                | `ObjectId` |    ✔️     | `@db.ObjectId`                 |       |
```
janpio, line 232: Should this maybe also include that it would be a `String` scalar type?
Do we have a similar table in other connector pages?

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+- Object
+- Javascript
+- JavascriptWithScope
+- Regex
+
+### Mapping from MongoDB to Prisma types on introspection
+
+When introspecting a MongoDB database, the database types are mapped to Prisma according to the following table:
+
+| MongoDB (Type \| Aliases) | Prisma     | Supported | Native database type attribute | Notes |
```
janpio: I know this is just copied from PostgreSQL, but is "Native database type attribute" the correct Prisma term here?

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+| `Boolean`  | `Bool`                |
+| `Int`      | `Int`                 |
+| `BigInt`   | `Long`                |
+| `Float`    | `Double`              |
+| `Decimal`  | Currently unsupported |
+| `Datetime` | `DateTime`            |
+| `Json`     |                       |
+| `Bytes`    | `BinData`             |
+
+MongoDB types that are currently unsupported:
```
janpio, line 213: This list seems to be missing `RegularExpression` which was there before.

content/200-concepts/200-database-connectors/07-mongodb.mdx
janpio:
```suggestion
| `ObjectId`                | `String` |    ✔️     | `@db.ObjectId`                 |       |
```

janpio: Is `ObjectId` the correct capitalization from the MongoDB side @dpetrick?

content/200-concepts/200-database-connectors/07-mongodb.mdx
janpio: `autoincrement` is also not supported here. 

I think this should work:
```suggestion
  id    String                              @id @db.ObjectId @default(auto()) @map("_id")
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: That makes it make more sense to me... done

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: It's now `Regex` https://github.com/prisma/docs/pull/3013#discussion_r836248360

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: Hm, this sounds odd to me. It's Prisma Migrate but we normally talk about 'migrations', lower case, elsewhere in docs

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+| Prisma     | MongoDB               |
+| ---------- | --------------------- |
+| `String`   | `String`              |
+| `Boolean`  | `Bool`                |
+| `Int`      | `Int`                 |
+| `BigInt`   | `Long`                |
+| `Float`    | `Double`              |
+| `Decimal`  | Currently unsupported |
+| `Datetime` | `DateTime`            |
+| `Json`     |                       |
```
janpio: The MongoDB value for "Json" seems to be missing.

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+- `JavascriptWithScope`
+- `Regex`
+
+### Mapping from MongoDB to Prisma types on Introspection
+
+When introspecting a MongoDB database, Prisma uses the relevant [scalar types](/concepts/components/prisma-schema/data-model#scalar-fields). Some special types also get additional native type annotations:
+
+| MongoDB (Type \| Aliases) | Prisma   | Supported | Native database type attribute | Notes |
+| ------------------------- | -------- | :-------: | :----------------------------- | :---- |
+| `ObjectId`                | `String` |    ✔️     | `@db.ObjectId`                 |       |
```
janpio: Fix the spelling of objectId:
```suggestion
| `objectId`                | `String` |    ✔️     | `@db.ObjectId`                 |       |
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
```diff
@@ -188,3 +188,55 @@ To resolve this, we suggest you change your deployment to one with a replica set
...
+| ---------- | --------------------- |
+| `String`   | `String`              |
+| `Boolean`  | `Bool`                |
+| `Int`      | `Int`                 |
+| `BigInt`   | `Long`                |
+| `Float`    | `Double`              |
+| `Decimal`  | Currently unsupported |
+| `Datetime` | `DateTime`            |
+| `Json`     |                       |
+| `Bytes`    | `BinData`             |
```
janpio:
```suggestion
| Prisma     | MongoDB               |
| ---------- | --------------------- |
| `String`   | `string`              |
| `Boolean`  | `bool`                |
| `Int`      | `int`                 |
| `BigInt`   | `long`                |
| `Float`    | `double`              |
| `Decimal`  | Currently unsupported |
| `DateTime` | `timestamp`           |
| `Date`     | `date`                |
| `Bytes`    | `binData`             |
| `Json`     |                       |
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
janpio: Matt created one:
```suggestion
| `Decimal`  | [Currently unsupported](https://github.com/prisma/prisma/issues/12637) |
```

content/200-concepts/200-database-connectors/07-mongodb.mdx
janpio: Hmm there probably is none :/

content/200-concepts/200-database-connectors/07-mongodb.mdx
janpio: This can also be Prisma scalar type `Bytes`.

content/200-concepts/200-database-connectors/07-mongodb.mdx
keerlu: Updated.

janpio (approved):

PR #3015, opened 2022-03-28
BigInt is mapped to BIGINT for Postgres

### LINKED ISSUE #2690: BigInt Prisma type maps to `integer` postgres type

Pwuts:
Change default mapping of Prisma's `BigInt` for Postgres from `integer` to `BIGINT`. #2849 is closed, so here is another PR to fix this. The library itself already maps it like this.

Fixes #2690.

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -550,7 +550,7 @@ True or false value.
 
 | Connector     | Default mapping |
 | ------------- | --------------- |
-| PostgreSQL    | `integer`       |
+| PostgreSQL    | `BIGINT`        |
 | Microsoft SQL | `int`           |
 | MySQL         | `INT`           |
 | MongoDB       | `Long`          |
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -550,7 +550,7 @@ True or false value.
 
 | Connector     | Default mapping |
 | ------------- | --------------- |
-| PostgreSQL    | `BIGINT`        |
+| PostgreSQL    | `bigint`        |
 | Microsoft SQL | `int`           |
 | MySQL         | `INT`           |
 | MongoDB       | `Long`          |


tomhoule (approved):
looks good to me — is the uppercase name consistent with the rest of the docs?

Pwuts:
In that table column alone, 3 different capitalization styles are used. `BigInt` in `schema.prisma` generates `BIGINT` in the migration, so it makes sense to me like this.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -550,7 +550,7 @@ True or false value.
 
 | Connector     | Default mapping |
 | ------------- | --------------- |
-| PostgreSQL    | `integer`       |
+| PostgreSQL    | `BIGINT`        |
```
janpio: The capitalization used in this table is the "typical" for each of these databases.

Here an example of the next type down:
```
| Connector     | Default mapping    |
| ------------- | ------------------ |
| PostgreSQL    | `double precision` |
| Microsoft SQL | `float(53)`        |
| MySQL         | `DOUBLE`           |
| MongoDB       | `Double`           |
| SQLite        | `REAL`             |
```

So I would suggest:
```suggestion
| PostgreSQL    | `bigint`        |
```

Pwuts: What is "the typical"? Prisma generates Postgres queries in all upper case.

janpio: While that might be the case (really?), the documentation obviously uses something else here and it would be nice if changes to the docs could adhere to that.

janpio: Thanks.

janpio (approved):

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
tanberry: Thanks @Pwuts and @janpio ! This is kinda a messy one. 

From what I see, the [official Postgres data type is called `bigint`](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT), while at Prisma [we map that to our PSL data type called `BigInt`](https://www.prisma.io/docs/concepts/database-connectors/postgresql#native-type-mappings). (And we do use `BigInt` consistently in our docs, when we are talking about Prisma types.) Soit seems that as least as far as Postgres, there is no place for `BIGINT` in our docs (that I see).

If above is accurate, these are the changes we need to make:

1. The edit that is in this PR now, which will replace `integer` with `bigint` in the table called "[Default type mappings"](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#default-type-mappings-3).

2. Do a much better job in the many "mapping" tables that we have in our docs! I think we are overloading the term "native" in the column headers, and sometimes it is unclear if we are talking about a type being native to the DB, or native to Prisma. We need an intro sentence to each table explaining what the columns really are.

janpio: The Prisma construct that exists (internally) are "native types" or "native database types" (vs. "scalar types"). They express in PSL what PostgreSQL/MySQL/... type should be used in the database. And yes, this is super messy right now and deserves a cleanup.

Pwuts: I agree with all of the above. In addition, if consistency is really desired, I suggest looking into how types are mapped from e.g. `BigInt` to `BIGINT` by Prisma internally, since the (PSQL) migration queries Prisma 3.10 generates contain all uppercase types, which is what caused me to apply this to my PR in the first place. The example queries in the Postgres docs have these types in lowercase, in accordance with the documented type names (in lowercase).

janpio: Can you maybe open an issue in prisma/prisma about the generated, PostgreSQL untypical SQL in the migration files? I think we might want to take a look at that. This SQL is an "API" after all, so it should feel natural and not like 1999 :)

Issue #3005, opened 2022-03-23
MongoDB: Document "undefined" behavior


matthewmueller:
We're making some changes in 3.11.1 around how records with undefined are queried

This [internal document](https://www.notion.so/prismaio/MongoDB-GA-effcea80826544ab9948f6425777e953#b778622fc3af40df983e781827ddbf11) has the details.

I asked Dom for a bit of clarity:

<img width="845" alt="CleanShot 2022-03-23 at 14 46 19@2x" src="https://user-images.githubusercontent.com/170299/159783275-a4e1789e-234e-417e-8a0a-8bf100f734f6.png">

Suggestion:
- Release notes or small docs page on breaking changes in the upgrade (undefineds are now ignores)
- Document `isSet` as a new query operation
- Document that undefined values in MongoDB are not matched by default, you use isSet to match them.

tanberry:
Docs impacted by 3.11.1 patch release:
ASAP:
--https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined ([PR #3009](https://github.com/prisma/docs/pull/3009))
-- https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting
--https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#filter-conditions-and-operators

Later polishing:
-- https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#event-types ("Note that for MongoDB, the params and duration fields will be undefined.")
-- a lot of the troubleshooting topics

matthewmueller:
Closed by: https://github.com/prisma/docs/pull/3017

Issue #2795, opened 2022-01-31
Getting Started Documentation with MongoDB on Prisma should talk about Replica Set requirement (Epic 3)


matthewmueller:
~Currently our getting started guide assumes you know how to get a MongoDB connection URL _from somewhere_:~
- Update: we'll do the above in the Prisma Data Guide because getting a connection URL is not only relevant to Prisma Users.

https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/mongodb/connect-your-database-typescript-mongodb

This assumes that the MongoDB url is configured as a replica set. The default local MongoDB does not work like this and you'll end up with an error like this:

```
Transactions are not supported by this deployment
```

There's a bunch of good suggestions in [this issue](https://github.com/prisma/prisma/issues/8266) on how to configure your local MongoDB to support the replicaset.

Extra: consider also updating Getting Started Guide example schema to use embedded types.

matthewmueller:
Update, we'll probably move this to the Prisma Data Guide. See [this discussion](https://prisma-company.slack.com/archives/CFM8D5VHC/p1643660083691719)

Jolg42:
I think we should document the current MongoDB docker compose setup Pierre created, it might be the easiest one command setup (if you have docker compose).

https://github.com/prisma/prisma/blob/main/docker/docker-compose.yml#L91-L99
So 
```docker-compose.yml
version: '3.7'

services:
  mongodb:
    build: ./mongodb-replica
    environment:
      MONGO_INITDB_ROOT_USERNAME: root
      MONGO_INITDB_ROOT_PASSWORD: prisma
      MONGO_REPLICA_HOST: localhost
      MONGO_REPLICA_PORT: 27017
    ports:
      - '27017:27017'
```
Note I changed the port from current setup to use the default port (so we'll need to do a quick check if that still works after, it should, but let's verify)

https://github.com/prisma/prisma/blob/main/docker/mongodb_replica/Dockerfile
```Dockerfile
FROM mongo:5

# we take over the default & start mongo in replica set mode in a background task
ENTRYPOINT mongod --port $MONGO_REPLICA_PORT --replSet rs0 --bind_ip 0.0.0.0 & MONGOD_PID=$!; \
# we prepare the replica set with a single node and prepare the root user config
[...]

Issue #2794, opened 2022-01-31
Document how to migrate and deal with inconsistent data when your schema changes in MongoDB (Epic 3)


matthewmueller:
How you migrate old data to new schemas is more complicated than SQL. In SQL, the integrity is maintained throughout all your past data, (e.g. if your zip code is an int column, all stored zip codes are ints). 

In MongoDB, this is not the case. You could start sending zip codes as strings and MongoDB will be happy as a clam. There's a number of strategies for dealing with this "feature":  https://github.com/prisma/prisma/issues/7305

We should document a few common scenarios that people will face and guide them through it.

dgradwell-ams:
MongoDB has built-in json schema support. Perhaps this could be used to check consistency.

tanberry:
Covered this topic in https://github.com/prisma/docs/pull/3017

matthewmueller:
Closed by: https://github.com/prisma/docs/pull/3017

Issue #2780, opened 2022-01-28
Document how to add relations to MongoDB after introspection (Epic 2)


matthewmueller:
MongoDB doesn't support automatically adding relations after introspection, so you need to do it by hand. We should document how to do this.

matthewmueller:
Notes with Docs:

- Available in getting started: https://www.prisma.io/docs/getting-started/setup-prisma/add-to-existing-project/mongodb/introspection-typescript-mongodb#tweaking-the-schema
- Don't seem to have a regular place, but there's a slack thread: https://prisma-company.slack.com/archives/C013RMLGE72/p1647951893323199

matthewmueller:
Closed by: https://github.com/prisma/docs/pull/3017

PR #3017, opened 2022-03-28
Using Prisma with MongoDB guide

keerlu:
Guide to MongoDB with Prisma, which will cover

- mapping the `_id` field
- Null vs undefined (#3005 )
- Replica set requirement (#2795 )
- Manually adding relations after introspection (#2780 )
- Migrating old data (#2794 )


content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,242 @@
...
+
+<TopBlock>
+
+This guide discusses the concepts behind using Prisma and MongoDB, explains the commonalities and differences between MongoDB and other database providers, and leads you through the process for configuring your application to integrate with MongoDB.
+
+</TopBlock>
+
+## What is MongoDB?
+
+MongoDB is a NoSQL database that stores data in a JSON-like document format.
```
dpetrick: That JSON-like format is called BSON.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,242 @@
...
+
+**Filtering for `null` and undefined fields.** MongoDB makes a distinction between setting a field to `null` and not setting it at all, which is not present in relational databases. Prisma currently does not express this distinction, which means that you need to be careful when filtering for `null` and undefined fields. For more information, see [How to filter for `null` and undefined fields](#how-to-filter-for-null-and-undefined-fields).
+
+**Replica sets.**
+
+## How to migrate old data to new schemas
+
+Types:
+
+- On-demand: when you add a new field to your schema, the past data isn't touched. But when you go and read the collection, the collection is adjusted, so when you write the data back, your data now matches your schema.
```
dpetrick: What is this exactly referring to?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,242 @@
...
+Prisma still returns `name: null`, but if you look in the database directly you will see that the record has no `name` field defined at all:
+
+```terminal
+_id: 6242c4af032bc76da250b208
+email: "user2@prisma.io"
+```
+
+Prisma returns the same result in both cases, because we currently don't have a way to specify this difference in MongoDB between fields that are `null` in the underlying database, and fields that are not defined at all — see [this Github issue](https://github.com/prisma/prisma/issues/12555) for more information.
+
+This means that you currently have to be careful when searching for `null` and undefined fields. Searching for records with `name: null` will return both records:
```
dpetrick: This is unfortunately the wrong way around - Prisma will return only one of the two, the one with explicit `null`. To include the other record as well you'll need to use `OR .. isSet: false` to get it too.

keerlu: Oops, thanks, that's a pretty major mistake. I searched for both at once when testing this and then muddled the order. Fixed the examples but let me know if it needs better explanation.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: This is copypasted draft text and will be rewritten

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Yep, will most likely include that when this placeholder gets rewritten

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,244 @@
...
+- connect to your database, using the [`mongodb` database connector](/concepts/database-connectors/mongodb)
+- use [Introspection](/concepts/components/introspection) for existing projects if you already have a MongoDB database
+- use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to the database
+- use [Prisma Client](/concepts/components/prisma-client) in your application to query your database
+
+## Differences to consider
+
+MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences you need to be aware of:
+
+**Defining IDs.** MongoDB documents have an `_id` field that contain an [ObjectID](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid). This needs to be mapped to a Prisma field. For more information, see [Defining IDs in MongoDB](/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).
```
keerlu: It makes sense to have this requirement in a MongoDB guide, but we also already have it written up [here](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).

Options are:
- link out to it, like I'm doing currently
- duplicate the content
- remove it from the other place

Any ideas?

tanberry: My vote is to link to it, as you have. Eventually we might decide to add it to the guide, and to link to it form the current location, but I think for now what you have is great.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+- model your database with the [Prisma Schema Language](/concepts/components/prisma-schema)
+- connect to your database, using the [`mongodb` database connector](/concepts/database-connectors/mongodb)
+- use [Introspection](/concepts/components/introspection) for existing projects if you already have a MongoDB database
+- use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to the database
+- use [Prisma Client](/concepts/components/prisma-client) in your application to query your database
+
+## Differences to consider
+
+MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences you need to be aware of:
```
tanberry:
```suggestion
MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences that you need to be aware of:
```

tanberry:
```suggestion
MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences that you need to be aware of:
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
tanberry: About formatting this list of differences: let's use bullets and colons, and keep it bolded like you have. So for example:

- **Defining IDs**: MongoDB documents have an.....

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+**Filtering for `null` and undefined fields.** MongoDB makes a distinction between setting a field to `null` and not setting it at all, which is not present in relational databases. Prisma currently does not express this distinction, which means that you need to be careful when filtering for `null` and undefined fields. For more information, see [How to filter for `null` and undefined fields](#how-to-filter-for-null-and-undefined-fields).
+
+**Using a replica set.** Prisma uses [MongoDB transactions](https://www.mongodb.com/docs/manual/core/transactions/) internally to avoid partial writes on nested queries. When using transactions, MongoDB requires you to have a [replica set](https://www.mongodb.com/docs/manual/replication/) configured — this is a group of MongoDB processes that maintain the same data set, providing redundancy and high availability. If you use MongoDB's [Atlas](https://www.mongodb.com/atlas/database) hosting service, the replica set is configured for you, but if you are running MongoDB locally you will need to set this up yourself. For more information, see [How to configure a replica set](#how-to-configure-a-replica-set)
+
```
tanberry, line 122: I think we need a transition to the How To topics... a H2 titled something like "How to integrate MongoDB with Prisma". (Which means the current H2 headings would need to all change to H3s, or ###.)

Then an intro sentence like "The following sections explain how to implement certain functionality and accomplish specific tasks when using MongoDB with Prisma."

tanberry:
@keerlu I added content to the How To section for handling old data, and also added an intro to the How To section. I meant to do it as a PR on your branch, but looks like it merged my commits into your PR branch. Hmmmm... I think it all worked out, somehow. ;-)
@matthewmueller and @janpio and @keerlu please carefully review the content that I added; anything correct in there I learned from Matt, anything wrong in there I made up or misunderstood.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Thanks for adding this!

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,307 @@
...
+</TopBlock>
+
+## What is MongoDB?
+
+MongoDB is a NoSQL database that stores data in [BSON](https://bsonspec.org/) format, a JSON-like document format designed for storing data in key-value pairs.
+
+MongoDB is commonly used in JavaScript application development because the document model maps easily to objects in application code, and there is built in support for high availability and horizontal scaling.
+
+MongoDB stores data in collections that do not need a schema to be defined in advance, as you would need to do with tables in a relational database. The structure of each collection can also be changed over time. This flexibility can allow rapid iteration of your data model, but it does mean that there are a number of differences when using Prisma to work with your MongoDB database.
+
```
keerlu, line 20: I realise this is 3 paragraphs starting with 'MongoDB' in a row, which doesn't look great. Will sort this out once the content's been settled.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+- connect to your database, using the [`mongodb` database connector](/concepts/database-connectors/mongodb)
+- use [Introspection](/concepts/components/introspection) for existing projects if you already have a MongoDB database
+- use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to the database
+- use [Prisma Client](/concepts/components/prisma-client) in your application to query your database
+
+## Differences to consider
+
+MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences that you need to be aware of:
+
+- **Defining IDs**: MongoDB documents have an `_id` field that contain an [ObjectID](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid). This needs to be mapped to a Prisma field. For more information, see [Defining IDs in MongoDB](/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).
```
dpetrick: This sounds like `_id` fields are always `ObjectID`s, which they're not.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **Migrating old data to new schemas**: In relational databases, all your data must match your schema. If you change the type of a particular field in your schema when you migrate, all the data must also be updated to match. In contrast, MongoDB does not enforce any particular schema, so you need to take care when migrating. For more information, see [How to migrate old data to new schemas](#how-to-migrate-old-data-to-new-schemas).
+
+- **Introspection**: When you introspect an existing database, you will get a schema with no relations and will need to add the missing relations in manually. For more information, see [How to add in missing relations after Introspection](#how-to-add-in-missing-relations-after-introspection).
```
dpetrick: I'd lean towards naming this `Introspection and Prisma relations`.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **Using a replica set**: Prisma uses [MongoDB transactions](https://www.mongodb.com/docs/manual/core/transactions/) internally to avoid partial writes on nested queries. When using transactions, MongoDB requires you to have a [replica set](https://www.mongodb.com/docs/manual/replication/) configured — this is a group of MongoDB processes that maintain the same data set, providing redundancy and high availability. If you use MongoDB's [Atlas](https://www.mongodb.com/atlas/database) hosting service, the replica set is configured for you, but if you are running MongoDB locally you will need to set this up yourself. For more information, see MongoDB's [guide to deploying a replica set](https://www.mongodb.com/docs/manual/tutorial/deploy-replica-set/).
```
dpetrick: This sounds scary, until you realize that you can have a replica set of size 1, not changing much. I'd somehow get that info in to make it less scary.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+## How to implement Prisma with MongoDB
+
+This section provides instructions for how to carry out tasks that require steps specific to MongoDB.
+
+### How to migrate old data to new schemas
+
+Migrating your database over time is an important part of the development cycle. During development, you will need to update your schema file (for example, toads new fields), then update the data in your your development environment’s database, and eventually push both the updated schema and the new data to the production database.
```
dpetrick: I assume frogs have nothing to do with MongoDB, even if they're generally green. `toads` -> `to add`.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **Using a replica set**: Prisma uses [MongoDB transactions](https://www.mongodb.com/docs/manual/core/transactions/) internally to avoid partial writes on nested queries. When using transactions, MongoDB requires you to have a [replica set](https://www.mongodb.com/docs/manual/replication/) configured — this is a group of MongoDB processes that maintain the same data set, providing redundancy and high availability. If you use MongoDB's [Atlas](https://www.mongodb.com/atlas/database) hosting service, the replica set is configured for you, but if you are running MongoDB locally you will need to set this up yourself. For more information, see MongoDB's [guide to deploying a replica set](https://www.mongodb.com/docs/manual/tutorial/deploy-replica-set/).
+
+## How to implement Prisma with MongoDB
```
dpetrick: `implement` strikes me as odd. `How to use Prisma with MongoDB` seems more generic but apt at the same time, skimming over the rest of the section.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+Prisma still returns `name: null`, but if you look in the database directly you will see that the record has no `name` field defined at all:
+
+```terminal
+_id: 6242c4af032bc76da250b208
+email: "user2@prisma.io"
+```
+
+Prisma returns the same result in both cases, because we currently don't have a way to specify this difference in MongoDB between fields that are `null` in the underlying database, and fields that are not defined at all — see [this Github issue](https://github.com/prisma/prisma/issues/12555) for more information.
+
+This means that you currently have to be careful when filtering for `null` and undefined fields. Filtering for records with `name: null` will only return the first record, with the `name` explicitly set to `null`:
```
dpetrick: Reading over this again I feel this should include a reason _why_ only the first one is returned: because `name: null` is checking for equality, and a non-existing field isn't *equal* to `null`, hence why it's not returned.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+title: 'Using Prisma with MongoDB'
+metaTitle: 'Using Prisma with MongoDB'
+metaDescription: 'Guide to using Prisma with MongoDB'
+tocDepth: 2
+toc: true
+---
+
+<TopBlock>
+
+This guide discusses the concepts behind using Prisma and MongoDB, explains the commonalities and differences between MongoDB and other database providers, and leads you through the process for configuring your application to integrate with MongoDB.
```
janpio:
```suggestion
This guide discusses the concepts behind using Prisma and MongoDB, explains the commonalities and differences between MongoDB and other database providers, and leads you through the process for configuring your application to integrate with MongoDB using Prisma.
```
or
```suggestion
This guide discusses the concepts behind using Prisma and MongoDB, explains the commonalities and differences between MongoDB and other database providers, and leads you through the process for configuring your application to integrate with Prisma using MongoDB.
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+</TopBlock>
+
+## What is MongoDB?
+
+MongoDB is a NoSQL database that stores data in [BSON](https://bsonspec.org/) format, a JSON-like document format designed for storing data in key-value pairs.
```
janpio: The early mention of BSON threw me off, but hard to find out what else to say after the first half of the sentence.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+## Commonalities with other database providers
+
+Some aspects of using Prisma with MongoDB are the same as when using Prisma with a relational database. You can still:
+
+- model your database with the [Prisma Schema Language](/concepts/components/prisma-schema)
+- connect to your database, using the [`mongodb` database connector](/concepts/database-connectors/mongodb)
+- use [Introspection](/concepts/components/introspection) for existing projects if you already have a MongoDB database
+- use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to the database
+- use [Prisma Client](/concepts/components/prisma-client) in your application to query your database
```
janpio:
```suggestion
- use [Prisma Client](/concepts/components/prisma-client) in your application to query your database in a type safe way based on your Prisma Schema
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio:
```suggestion
- **Defining IDs**: MongoDB documents have an `_id` field (that often contain an [ObjectID](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid)). This needs to be mapped to a Prisma field. For more information, see [Defining IDs in MongoDB](/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).
```

janpio: I think this is also one of the least significant differences, so we could move that down a bit.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+## Differences to consider
+
+MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences that you need to be aware of:
+
+- **Defining IDs**: MongoDB documents have an `_id` field that contain an [ObjectID](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid). This needs to be mapped to a Prisma field. For more information, see [Defining IDs in MongoDB](/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).
+
+- **Migrating old data to new schemas**: In relational databases, all your data must match your schema. If you change the type of a particular field in your schema when you migrate, all the data must also be updated to match. In contrast, MongoDB does not enforce any particular schema, so you need to take care when migrating. For more information, see [How to migrate old data to new schemas](#how-to-migrate-old-data-to-new-schemas).
```
janpio: I would not necessarily say that this is about "old" and "new", more about that there can be data that does not adhere to the schema you defined in your Prisma schema (because different entries in the same collection can be of a different structure). If that is the case you need to migrate the data somehow to match the schema or things will go bad.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: This is relevant, because MongoDB does not haave a concept of relations (via foreign keys). Only if you have data that has a "foreign key" like column, that has the same value as the ID field of another collection you can model it like that and Prisma lets you emulate relations on that that behave like real ones.

Nuance, but important.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **Introspection**: When you introspect an existing database, you will get a schema with no relations and will need to add the missing relations in manually. For more information, see [How to add in missing relations after Introspection](#how-to-add-in-missing-relations-after-introspection).
+
+- **Filtering for `null` and undefined fields**: MongoDB makes a distinction between setting a field to `null` and not setting it at all, which is not present in relational databases. Prisma currently does not express this distinction, which means that you need to be careful when filtering for `null` and undefined fields. For more information, see [How to filter for `null` and undefined fields](#how-to-filter-for-null-and-undefined-fields).
```
janpio:
```suggestion
- **Filtering for `null` and not defined fields**: MongoDB makes a distinction between setting a field to `null` and not setting it at all, which is not present in relational databases. Prisma currently does not express this distinction, which means that you need to be careful when filtering for `null` and not defined fields. For more information, see [How to filter for `null` and undefined fields](#how-to-filter-for-null-and-undefined-fields).
```
Suggestion to use "not defined" instead of "undefined", as `undefined` is also a "value" that a property can have in JavaScript. Expressing it that way, makes it clearer. Alternative would be to say "not set".

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: Same for the term "replica set". For us now pretty common, I think it might not be for all users. This really just means that your MongoDB needs to be set up in a specific way that enables Replication in theory as there the guarantees are present.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: +1

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **Using a replica set**: Prisma uses [MongoDB transactions](https://www.mongodb.com/docs/manual/core/transactions/) internally to avoid partial writes on nested queries. When using transactions, MongoDB requires you to have a [replica set](https://www.mongodb.com/docs/manual/replication/) configured — this is a group of MongoDB processes that maintain the same data set, providing redundancy and high availability. If you use MongoDB's [Atlas](https://www.mongodb.com/atlas/database) hosting service, the replica set is configured for you, but if you are running MongoDB locally you will need to set this up yourself. For more information, see MongoDB's [guide to deploying a replica set](https://www.mongodb.com/docs/manual/tutorial/deploy-replica-set/).
+
+## How to implement Prisma with MongoDB
+
+This section provides instructions for how to carry out tasks that require steps specific to MongoDB.
+
+### How to migrate old data to new schemas
```
janpio: For the headline and the full section: https://github.com/prisma/docs/pull/3017/files#r840575516

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: 🐸 
```suggestion
Migrating your database over time is an important part of the development cycle. During development, you will need to update your schema file (for example, to add new fields), then update the data in your your development environment’s database, and eventually push both the updated schema and the new data to the production database.
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+  city   String
+  zip    String
+}
+```
+
+There are a number of strategies you could use for migrating this schema:
+
+- **"On-demand" updates**: with this strategy, you and your team have agreed that updates can be made to the schema as needed. However, in order to avoid migration failures due to inconsistencies between the data and schema, there is agreement in the team that any new fields added are explicitly defined as optional.
+
+  In our scenario above, you can go ahead and add the new `billingAddress` field to the `User` model in your Prisma schema, and define it as optional:
```
janpio: There is no `User` model. I assume this means:
```suggestion
  In our scenario above, you can go ahead and add the new `billingAddress` field to the `Order` model in your Prisma schema, and define it as optional:
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+  shippingAddress Address
+  productId       String   @db.ObjectId
+}
+
+type Address {
+  street String @id @default(auto()) @map("_id") @db.ObjectId
+  city   String
+  zip    String
+}
+```
```
janpio: Simplify: Get rid of the irrelevant `Product` and the relation to it:

````suggestion
```prisma file=prisma/schema.prisma
model Order {
  id              String   @id @default(auto()) @map("_id") @db.ObjectId
  shippingAddress Address
}

type Address {
  street String @id @default(auto()) @map("_id") @db.ObjectId
  city   String
  zip    String
}
```
````

File: content/300-guides/050-database/870-using-prisma-with-mongodb.mdx [hunk 22]
```diff
@@ -0,0 +1,351 @@
...
+    billingAddress  Address?
+    productId       String   @db.ObjectId
+  }
+
+  type Address {
+    street String @id @default(auto()) @map("_id") @db.ObjectId
+    city   String
+    zip    String
+  }
+  ```
```
janpio: Simplify same as above:
````suggestion
  ```prisma file=prisma/schema.prisma highlight=4;add
  model Order {
    id              String   @id @default(auto()) @map("_id") @db.ObjectId
    shippingAddress Address
    billingAddress  Address?
  }

  type Address {
    street String @id @default(auto()) @map("_id") @db.ObjectId
    city   String
    zip    String
  }
  ```
````

File: content/300-guides/050-database/870-using-prisma-with-mongodb.mdx [hunk 23]
```diff
@@ -0,0 +1,351 @@
...
+
+When using MongoDB, be aware that the “coupling” between your schema and the database is purposefully designed to be less rigid than with with SQL databases; MongoDB will not enforce the schema, so you have to verify data integrity.
+
+</Admonition>
+
+These iterative tasks of updating the schema and the database can result in inconsistencies between your schema and the actual data in the database. Let’s look at one scenario where this can happen, and then examine several strategies for you and your team to consider for handling these inconsistencies.
+
+**Scenario**: you need to add a new field to your schema to accommodate a billing address that is separate from the currently stored shipping address. You currently have the following schema in your `schema.prisma` file:
+
+```prisma file=prisma/schema.prisma
```
janpio, line 63: Note that the relation between `Order` and `Address` is not correctly defined in a Prisma way.
You actually need a `Order.addressId` field of type `String @db.ObjectId`, the `Order.shippingAddress` needs `@relation ...` and `Address` needs a back relation.

I suggest we fix this after accepting the simplification of the whole example (See other comments)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  type Address {
+    street String @id @default(auto()) @map("_id") @db.ObjectId
+    city   String
+    zip    String
+  }
+  ```
+
+  Then regenerate your Prisma Client using the `npx prisma generate` command. Next, update your application to reflect the new field, and redeploy your app. The records in the database will be updated as the application's users begin to enter their billing address in the new field.
```
janpio: This might require some explanation why this works:
- New field was added that is allowed to be null - so the existing data fits to the new schema anyway.

(Is this correct @dpetrick?)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **"No breaking changes" updates**: this strategy builds on the first one, with further consensus amongst your team that you don't rename or delete fields, only add new fields, and always define the new fields as optional. This policy can be reenforced by adding checks in the CI/CD process to verify that there are no backwards-incompatible changes to the schema.
```
janpio, line 100: Isn't this what the first example also does?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+- **"All-at-once" updates**: this strategy is similar to traditional migrations in relational databases, where all data is updated to reflect the new schema. In the scenario above, you would create a script to add a value for billing addresses to all existing users in your database. You can then make the field a required field in the application because the schema and the data are consistent.
```
janpio: I think this one could mention a case here this is actually necessary.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+### How to add in missing relations after Introspection
+
+After Introspection of a MongoDB database, you will need to manually add in relations between databases. As an example, take a MongoDB database with two collections, `User` and `Post`. The data in these collections has the following format, with a `userId` field linking users to posts:
```
janpio:
```suggestion
After Introspection of an existing MongoDB database, you will need to manually add in relations between models. As an example, take a MongoDB database with two collections, `User` and `Post`. The data in these collections has the following format, with a `userId` field linking users to posts:
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+`User` collection:
+
+- `_id` field with a type of `ObjectID`
+- `email` field with a type of `String`
+
+`Post` collection:
+
+- `_id` field with a type of `ObjectID`
+- `name` field with a type of `String`
+- `userId` with a type of `ObjectID`
```
janpio: The MongoDB types are lowercase (https://www.mongodb.com/docs/manual/reference/bson-types/, column "Alias"):
```suggestion
- `_id` field with a type of `objectId`
- `email` field with a type of `string`

`Post` collection:

- `_id` field with a type of `objectId`
- `title` field with a type of `string`
- `userId` with a type of `objectID`
```
(Also fix the wrong field name)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
```
janpio, line 120: It might be worth to add a simple BSON/JSON example here with 1 entry per collection. Should be simple to click together in MongoDB Compass (ping me if you need help).

(This will also allow us to confirm Introspection)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+  userId String @db.ObjectId
+}
+
+model User {
+  id    String @id @default(auto()) @map("_id") @db.ObjectId
+  email String
+}
+```
+
+This is missing the relation between the `User` and `Post` models. To fix this, manually add a `user` field to the `Post` model with a `@relation` attribute linking it to the `User` model, and a `posts` field to the `User` model:
```
janpio:
```suggestion
This is missing the relation between the `User` and `Post` models. To fix this, manually add a `user` field to the `Post` model with a `@relation` attribute using `userId` as the `fields` value, linking it to the `User` model, and a `posts` field to the `User` model as the back relation:
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+model User {
+  id    String @id @default(auto()) @map("_id") @db.ObjectId
+  email String
+  posts Post[]
+}
+```
+
+For more information on how to use relations in Prisma, see [our documentation](/concepts/components/prisma-schema/relations).
+
+### How to filter for <inlinecode>null</inlinecode> and undefined fields
```
janpio: See above: https://github.com/prisma/docs/pull/3017/files#r840578270

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+<cmd>
+
+```ts
+  const createUndefined = await prisma.user.create({
+    data: {
+       email: "user2@prisma.io",
+     }
+   })
+   console.log(createUndefined)
+}
```
janpio:
```suggestion
const createUndefined = await prisma.user.create({
  data: {
    email: "user2@prisma.io",
  }
})
console.log(createUndefined)
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+</cmd>
+
+<cmdResult>
+
+```
+{
+    id: '6242c4ae032bc76da250b207',
+    email: 'user1@prisma.io',
+    name: null
```
janpio:
```suggestion
  id: '6242c4ae032bc76da250b207',
  email: 'user1@prisma.io',
  name: null
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+<cmdResult>
+
+```
+{
+    id: '6242c4ae032bc76da250b208',
+    email: 'user2@prisma.io',
+    name: null
+},
+
```
janpio:
```suggestion
{
  id: '6242c4ae032bc76da250b208',
  email: 'user2@prisma.io',
  name: null
}
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+<cmdResult>
+
+```terminal
+[
+  {
+    id: '6242c4ae032bc76da250b207',
+    email: 'user1@prisma.io',
+    name: null
+  }
+]
```
janpio, line 277: This is the first one where the terminal output is in an array and not just the object - should _all_ the ones above also have that actually?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,351 @@
...
+      {
+        name: {
+          isSet: false,
+        },
+      },
+    ],
+  },
+})
+
+console.log(findNullOrUndefined)
```
janpio:
```suggestion
const findNullOrUndefined = await prisma.user.findMany({
  where: {
    OR: [
      {
        name: null,
      },
      {
        name: {
          isSet: false,
        },
      },
    ],
  },
})
console.log(findNullOrUndefined)
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
tanberry: Yes, it is basically same, but the strategy is slightly diff in that it includes the sue of CI/CD checks.. seems nice to call that out as an additional approach. It's a little more formal than the "on demand" strategy.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: I'd also consider the final link to link to how to set it up yourself (e.g. that MongoDB config snippet on MacOS) rather than point to MongoDB's documentation.

That would also be more consistent with the links above.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: In line with simplification, on demand migrations could also apply to a single field. We could say that you find out you want to add a phoneNumber to the user model. You'd add it as optional, then as users fill it in from their profile, it gets updated. This might be slightly simpler to explain than using a composite. Totally up to you though.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: _Responding to @janpio_ 

Wrote some words on this for @tanberry 

Here's the scenario:
1. I have an existing app with existing data in my MongoDB database and an existing Prisma Schema
2. A feature we're implementing needs us to add a new phoneNumber field to the User model

In SQL, this field must be added via add column. Usually added as an optional field to avoid needing to backfill all old phone numbers. SQL won't let you add a required phoneNumber when the previous records are null.

MongoDB is different. There is nothing enforcing the schema. It's up to you as the developer to match the schema with your data.

- If you make the phoneNumber required, nothing's going to stop you. However, when you query users that don't have the phone number, you're going to get a runtime error, because the Prisma Schema says it's required, but the data shows differently.
- If you make the phoneNumber optional, what prisma (and mongoose) do is allow you to query the old users, where the phoneNumber is just undefined (or null, not sure). But say you're updating that phoneNumber, when you write the data back, you've set the phoneNumber

That's what I mean by migrating On-Demand (a term I made up). There's no specific migration step. Instead, your migrating old users over time as your application runs. It turns out this is how most people deal with changing data in MongoDB.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: _Responding to @janpio_ 

^ can you clarify this point. I'm not sure what you mean. Are you looking for an example or suggesting that there are times when it's actually necessary to do this kind of update?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller:
```suggestion
After introspecting a MongoDB database, you will need to manually add in relations between databases. As an example, take a MongoDB database with two collections, `User` and `Post`. The data in these collections has the following format, with a `userId` field linking users to posts:
```

matthewmueller (commented):
Great start @keerlu! I especially love the introduction and the organization of this guide.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: > If you make the phoneNumber required, nothing's going to stop you. However, when you query users that don't have the phone number, you're going to get a runtime error, because the Prisma Schema says it's required, but the data shows differently.

Because of this, I would only consider this strategy valid when the phoneNumber is optional. Otherwise, if you apply this strategy, you end up with an application that can not handle any existing data.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: For me this is the same as the first, as the first one actually only works if all the things described here are the case.

How would a CI check for such "backwards-incompatible changes" look like?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: The first two are possible in the specific case where you add an optional field - old data will still be valid with the new schema. When you do a change that does not adhere to that, and e.g. old data would make Prisma Client explode, you will _need_ to use this approach here. A specific example for a change that makes this required could help people to understand when this is the case.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Done

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: I've now accepted the example simplification.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Done

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Done.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Changed subsection title and clarified this in text of "how to..." section

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: @matthewmueller it might be better to eventually describe how to set it up yourself but we need a tested solution that we know works on Windows, Mac etc, see thread here: https://prisma-company.slack.com/archives/C013RMLGE72/p1648805979446329?thread_ts=1648723271.925979&cid=C013RMLGE72

keerlu: @dpetrick @janpio I've updated the text to clarify that you can have a replica set of size 1, but would be worth checking that the way I've phrased it makes sense.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Done.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Added this explanation in.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: My memory is that the two `findMany`s should return an array, whereas the `create`s above shouldn't

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Made it somewhat better.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Yep, I'm open to rephrasing but haven't thought of anything very good yet.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Changed to 'Migrating existing data to match your Prisma schema'

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: I've updated the example to use a phone number instead of a billing address, and tried to clarify what 'on demand' means using your comment here https://github.com/prisma/docs/pull/3017#discussion_r840940479

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
tanberry: Yikes! This worries me @janpio , as I too thought that we were talking about the value `undefined`. The topic that we link to is definitely about the value `undefined`, correct? If so, we should not link to that topic.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
tanberry: @janpio We are presenting these three options as "strategies"... as approaches that our users can think about and consider if they want to use. These are not instructions or processes, but rather strategies they might want to employ. SO I think it is very valuable to keep this second one; maybe readers wouldn't even consider CI/CD checks unless they read this. And we can't tell them exactly how to incorporate any of the checks into their CI/CD build, but I could imagine that they read this and think "Oh, yeah, we are a big team, we should totally add a rule in our build that checks for new fields that have not been defined as optional, and stop the build if any are found."

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: The link is internal, so I would also guess (I do not know!) this is about "not defined".

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: Then I do not understand why we gave strategy 2 the name that describes what we do in strategy 1, but add "use CI/CD" to the description (without adding it to the name of the strategy). 

To me this was super confusing and I could not differentiate strategy 1 and 2 and the names to not related directly to what they describe and how they are different from each other. I kept looking for the point here that I was missing.

To me seems we describe only 2 different strategies, and a few nuances for strategy 1.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+toc: true
+---
+
+<TopBlock>
+
+This guide discusses the concepts behind using Prisma and MongoDB, explains the commonalities and differences between MongoDB and other database providers, and leads you through the process for configuring your application to integrate with MongoDB using Prisma.
+
+</TopBlock>
+
+## What is MongoDB?
```
janpio, line 15: OT: A MongoDB logo aligned right next to this would be neat.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+MongoDB is a NoSQL database that stores data in [BSON](https://bsonspec.org/) format, a JSON-like document format designed for storing data in key-value pairs. It is commonly used in JavaScript application development because the document model maps easily to objects in application code, and there is built in support for high availability and horizontal scaling.
```
janpio:
```suggestion
[MongoDB](https://www.mongodb.com/) is a NoSQL database that stores data in [BSON](https://bsonspec.org/) format, a JSON-like document format designed for storing data in key-value pairs. It is commonly used in JavaScript application development because the document model maps easily to objects in application code, and there is built in support for high availability and horizontal scaling.
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+- connect to your database, using the [`mongodb` database connector](/concepts/database-connectors/mongodb)
+- use [Introspection](/concepts/components/introspection) for existing projects if you already have a MongoDB database
+- use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to the database
+- use [Prisma Client](/concepts/components/prisma-client) in your application to query your database in a type safe way based on your Prisma Schema
+
+## Differences to consider
+
+MongoDB's document-based structure and flexible schemas means that using Prisma with MongoDB differs from using it with a relational database in a number of ways. These are some areas where there are differences that you need to be aware of:
+
+- **Defining IDs**: MongoDB documents have an `_id` field (that often contain an [ObjectID](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid)). This needs to be mapped to a Prisma field. For more information, see [Defining IDs in MongoDB](/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).
```
janpio:
```suggestion
- **Defining IDs**: MongoDB documents have an `_id` field (that often contains an [ObjectID](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid)). This needs to be mapped to a Prisma field. For more information, see [Defining IDs in MongoDB](/concepts/components/prisma-schema/data-model#defining-ids-in-mongodb).
```

janpio: The actual information information is that Prisma does not support fields starting with `_` and thus the `@map` to map is necessary. Not sure this is explicit enough here.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+- **Migrating existing data to match your Prisma schema**: In relational databases, all your data must match your schema. If you change the type of a particular field in your schema when you migrate, all the data must also be updated to match. In contrast, MongoDB does not enforce any particular schema, so you need to take care when migrating. For more information, see [How to migrate old data to new schemas](#how-to-migrate-existing-data-to-match-your-prisma-schema).
```
janpio, line 37: > all the data must also be updated to match

Usually the database takes care of that automatically, by casting the data. (Correct @tomhoule?)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+- **Introspection and Prisma relations**: When you introspect an existing database, you will get a schema with no relations and will need to add the missing relations in manually. For more information, see [How to add in missing relations after Introspection](#how-to-add-in-missing-relations-after-introspection).
```
janpio:
```suggestion
- **Introspection and Prisma relations**: When you introspect an existing MongoDB database, you will get a schema with no relations and will need to add the missing relations in manually. For more information, see [How to add in missing relations after Introspection](#how-to-add-in-missing-relations-after-introspection).
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+</CodeWithResult>
+
+If you check your MongoDB database directly, you will also see a new record with `name` set to `null`:
+
+```terminal
+_id: 6242c4af032bc76da250b207
+email: "user1@prisma.io"
+name: null
+```
```
janpio: I don't think this is actually terminal output, but probably part of a Json document that should show what is in the database? (Same is true for the following `terminal` ones)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: I checked the linked to content: It is _not_ about any data that is returned as `undefined`, but data that is not defined and then returned as `null`. 

(But the GitHub issue we link has the exact same problem, and there undefined is even formatted as `undefined` :/ Started a discussion about how to name this here: https://prisma-company.slack.com/archives/CUXLS0Z6K/p1649096500983929)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+## How to use Prisma with MongoDB
+
+This section provides instructions for how to carry out tasks that require steps specific to MongoDB.
+
+### How to migrate existing data to match your Prisma schema
+
+Migrating your database over time is an important part of the development cycle. During development, you will need to update your schema file (for example, to add new fields), then update the data in your your development environment’s database, and eventually push both the updated schema and the new data to the production database.
```
janpio:
```suggestion
Migrating your database over time is an important part of the development cycle. During development, you will need to update your Prisma schema file (for example, to add new fields), then update the data in your your development environment’s database, and eventually push both the updated schema and the new data to the production database.
```
(only if this actually already refers to the Prisma schema file)

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,326 @@
...
+  model User {
+    id          String  @id @default(auto()) @map("_id") @db.ObjectId
+    email       String
+    phoneNumber String?
+  }
+  ```
+
+  Then regenerate your Prisma Client using the `npx prisma generate` command. Next, update your application to reflect the new field, and redeploy your app.
+
+  As the `phoneNumber` field is optional, you can still query the old users where the phone number has not been defined. The records in the database will be updated "on demand" as the application's users begin to enter their phone number in the new field.
```
janpio, line 86: Isn't this also where the `@default` support would kick in @matthewmueller @dpetrick? Might be worth another paragraph here so it is at least documented somewhere.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: Much better to make the point here.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: Sounds ok to me now 👍

janpio (approved):

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
tanberry: Hmmm maybe the names need to be tweaked. I do think it is worth calling out all three, though.

Perhaps this or something like it:
1. **On-demand, passive updates** (I use the word passive because the data is passively updated as the app is used, but that might be a misleading word for other reasons)
2. **CI/CD-enforced** (and still explain that this strategy builds on #3, but if enforced/verified)
3. **All-at-once**

We have a little more time... will get it cleaned up/decided before tomorrow's release, and can even iterate afterwards. @keerlu of course feel free to edit as you want; I am happy with you and @matthewmueller being the final decision makers on this. :-).

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: Your wish is my command 🧞 

Mostly just taking the opportunity to throw a `float: right` in there 😌

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: Thanks so much for mentioning this, I completely forgot about this feature. It also prompted: https://github.com/prisma/prisma/issues/12665.

I just wrote some words about this. See below and let me know.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: I like the term passive updates.

I see 2. as more of an organizational rule or process. For example, the MongoDB Atlas team themselves applies this rule when administering their own MongoDB database.

I see it as different, but not sure the best way to describe it.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: +1, I would probably just show the whole document as JSON:

```
{
  "_id": "..."
  ...
}
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller:
```suggestion
Migrating your database over time is an important part of the development cycle. During development, you will need to update your Prisma Schema file (for example, to add new fields), then update the data in your your development environment’s database, and eventually push both the updated schema and the new data to the production database.
```

In the other guide, @tanberry corrected me with **Prisma Schema**

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,344 @@
...
+```
+
+First, try creating a record with the `name` field explicitly set to `null`. Prisma will return `name: null` as expected:
+
+<CodeWithResult expanded={true}>
+
+<cmd>
+
+```ts
+const createNull = await prisma.user.create({
```
matthewmueller, line 174:
```suggestion
const user = await prisma.user.create({
```

Unless depended on elsewhere, `createNull` doesn't really tell what's being returned here.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -0,0 +1,344 @@
...
+  data: {
+    email: 'user1@prisma.io',
+    name: null,
+  },
+})
+console.log(createNull)
```
matthewmueller, line 180:
```suggestion
console.log(user)
```

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
matthewmueller: Same with the `createUndefined` below

matthewmueller: And findNulls

matthewmueller (approved):

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: Old school 🧓 

But unfortunately Gatsy or something does weird things to it:
```
<img src="/docs/static/a4546ec668a59cd396325fff01e316d9/mongodb-logo.svg" style="css-float:right" width="195">
```

(Image is also automatically linked)

janpio: Would look _great_ though: 
image

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
tomhoule: I don't think it's general enough to be worth mentioning, casting is very limited and there is no casting for objects of the wrong shape, missing fields, etc.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: The positioning of this that I'm seeing on the preview currently looks very strange to me: 

image

Also we don't normally have logos in these, can you explain why this is being added?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Done

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Done, agree this is much better

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Is this a Prisma convention, or some other convention? I'm happy to do it but need to know why and what the other `find`s and `create`s should be changed to

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Yes, this is the terminology I got from @dpetrick . Happy to change this later, but note that we are already using "undefined" for MongoDB in this sense so would have to change this too: https://github.com/prisma/docs/pull/3010

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: May add this later, leaving for now

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: The positioning is off, as the code that Matt added is mangled by Gatsby somehow and does not work. I emulated what he wanted to achieve manually by fixing the code in the developer console of Chrome.

Matt added this, as I had left a comment here that this would be "neat". 
I think it makes the page look much nicer and also visually able to connect back to MongoDB.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: The discussion in Slack lead to a mention of Flavian to use "missing", which I think would solve this problem properly.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
janpio: A variable called `createNull` does not contain any information in its name what is in it.

keerlu: Yes but it does contain the more salient information for the purposes of this example, that you're creating something with a `null` value. Otherwise, what do we call them? `user` and `user2`? What about the `find`s?

keerlu: Will revisit after the release.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
```diff
@@ -38,7 +38,7 @@ MongoDB's document-based structure and flexible schemas means that using Prisma
 
 - **Introspection and Prisma relations**: When you introspect an existing MongoDB database, you will get a schema with no relations and will need to add the missing relations in manually. For more information, see [How to add in missing relations after Introspection](#how-to-add-in-missing-relations-after-introspection).
 
-- **Filtering for `null` and undefined fields**: MongoDB makes a distinction between setting a field to `null` and not setting it at all, which is not present in relational databases. Prisma currently does not express this distinction, which means that you need to be careful when filtering for `null` and undefined fields. For more information, see [How to filter for `null` and undefined fields](#how-to-filter-for-null-and-undefined-fields).
```
janpio: Doesn't each of these list items need a link down to where it is explained in more detail?

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Removed logo for now, will include again once positioning is sorted.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Sorry, yes, brain fart. Put it back in.

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Have changed to 'missing'

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Will revisit improving the wording after release

content/300-guides/050-database/870-using-prisma-with-mongodb.mdx
keerlu: Will revisit improving this after the release.

PR #3019, opened 2022-03-28
Move Migrations guides to their own folder

janpio:
...

### TODO

- [ ] Prepare redirects


tanberry:
Well hello @janpio this is rather a massive PR! :scream-cry-laugh:

I do like the re-organization, but, first, ... several discussion points:

1. Most importantly, we are still (ever-so-slowly) planning a major restructuring of all content, so this might be either premature or get changed again later.

2. Is the top-level title of "Migration Guides" appropriate? Do we want to focus that hard on the Migrate component and related commands (`dev`, 'diff`, `deploy`, etc)? 

3. Would a reader think to look under Migration Guides for info like seeding, (which can be done via `dp push` so isn't Migrate-specific), or Schema prototyping with `db push`?

Let's chat more about this PR, and also for sure let's hold off until after next week's release. ;-)

janpio:
This is a draft PR where I was playing with a few things, and it is _not_ ready for review - so thanks for taking a look, but I hope you did not spend to much time on this for now.

PR #3032, opened 2022-04-02
docs: Rename global prisma variable for next.js

jaapaurelio:
## Describe this PR

Global `prisma` variable was preventing lint from validating the missing import from `./db`, meaning lint would not complain about undefined/unimported `prisma` variable.

Renaming the global variable to `prismaInstance` works better since it's not overriding `prisma` export name.

## Changes

Rename global `prisma` to `prismaInstance`.


jaapaurelio:
My bad. Lint does not care about this global definition.

PR #3033, opened 2022-04-03
Fix wrong variable name for MySQL example

arthurfalcao:
## Describe this PR

The MySQL example is using the wrong type and variable usage in the `Deleting all data with raw SQL` section. (`tablename` instead of `TABLE_NAME`)

## Changes

It fixes the example in the `Deleting all data with raw SQL` section


janpio (approved):

tanberry:
Thank you @arthurfalcao for the fix!

PR #3041, opened 2022-04-06
fix(production-troubleshooting): Fix anchor link for migrate diff

janpio:
Currently uses wrong anchor to link to.


tanberry (approved):

tanberry:
ooooofff glad you caught that, thanks @janpio !

PR #3048, opened 2022-04-07
Add free limits notification for PDP

martzoukos:
We are planning to add a limit to the number of free projects a user can create in the Data Platform. While we don't have our pricing plans communicated yet in our marketing website, we would like to still communicate this limitation in places other than our application (where users will still get a notification from the system that they can't add more than 2 free projects).

image


tanberry:
Thanks @martzoukos, this looks fine to me. If you're ready, we can merge it, or if you had other things you wanted to add we can leave it as a draft.

martzoukos:
Thank you @tanberry ! I'm merging it now.

PR #3047, opened 2022-04-07
A new parameter for SQL Server in 3.13.0.

pimeys:
Parameter `trustServerCertificateCA` should point to a certificate authority, which we use to validate the server certificate.

PR #3050, opened 2022-04-07
A new parameter for SQL Server in 3.13.0: `trustServerCertificateCA`

janpio:
Reverts prisma/docs#3049 which reverted accidental merge of https://github.com/prisma/docs/pull/3047

Original message from @pimeys:

> Parameter `trustServerCertificateCA` should point to a certificate authority, which we use to validate the server certificate.


pimeys:
Tsk tsk tsk

tanberry:
OK, so more context might need to be added to the descriptions here... should #3047 not have been merged? I think it was approved by @janpio so I took a quick look and then merged it.

Or, is it that we DO want 3047 merged, and you are reverting the accidental reversion?

janpio:
#3047 should not have been merged, I just approved the content change in there so it is known that this can be merged when the feature was released. So I reverted the change to not have misleading info on the website, and then reverted the revert so we have a PR for later to be able to merge.

tanberry:
Ah, thanks @Jan! I didn't clue in to the very obvious signal in the tile, of 3.13, sorry. :-(

Jolg42 (approved):

PR #3055, opened 2022-04-11
fix model name

boemekeld:
## Describe this PR

wrong model name


janpio (approved):

keerlu (approved):
Thanks for spotting this, @boemekeld!

PR #3070, opened 2022-04-15
Add redirects from website-v8 here directly instead

janpio:
This PR adds the docs redirects that were all in https://github.com/prisma/website-v8/blob/main/public/_redirects.
It makes removing them there later possible: https://github.com/prisma/website-v8/pull/519

(The PR also makes it easier to run the docs locally without having to define Algolia env vars)

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -144,7 +144,11 @@ let plugins = [
 ]
 
 if (process.env.INDEX_ALGOLIA === 'true') {
-  plugins = [...plugins, algoliaPlugin]
+  if (process.env.GATSBY_ALGOLIA_APP_ID) {
+    plugins = [...plugins, algoliaPlugin]
+  } else {
+    console.warn('INDEX_ALGOLIA === true, but GATSBY_ALGOLIA_APP_ID is undefined.')
+  }
 }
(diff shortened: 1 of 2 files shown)


tanberry:
This looks awesome, @Jan. Let's set up a quick call with you and/or @nilubava to go over this and make sure we (@Lucy and I) understand the changes, and how it impacts our current process of creating redirects. At first glance it looks like it means we would no longer need to edit the file in the website-v8 repo, but rather here in this gatsby-config.js file?

I'll look for a time on our calendar that works.

janpio:
No meeting necessary: Instead of editing a `_redirects` file in website-v8 repository, you edit the `_redirects` file in this repository. Nothing else changes for you.

(@nilubava of course has to verify my changes either before or after merging via what is described in https://github.com/prisma/website-v8/pull/519)

tanberry:
OK, will be nice to skip a trip to the website-v8 repo, for sure. Thanks @janpio, after Nilu tests and verifies I will update our internal documentation about how to create redirects. And I'll ask Nilu too explain the bit about Algolia.

nilubava (approved):

PR #3077, opened 2022-04-21
Adding Rallly to list of real-world examples

lukevella:
Adding [rallly](https://github.com/lukevella/rallly) to the list of real-world examples.


tanberry:
Very cool, @lukevella , thanks so much for adding Rallly to our list of real-world apps built using Prisma!

PR #3083, opened 2022-04-22
fix(migrate): Remove last scary "not supported on MongoDB" warning admonition with nicer info

janpio:
Less scary


tanberry (commented):
Thanks @janpio , does indeed look less scary.

PR #3085, opened 2022-04-23
Update Prisma type DateTime from "Datetime" to "DateTime"

neicore:
## Describe this PR
DateTime type is written as 'Datetime '

## Changes
- Changed it to 'DateTime'

## What issue does this fix?
I haven't found issues related to this


tanberry:
Great catch, thanks so much @neicore ! We really appreciate your help.

PR #3086, opened 2022-04-23
Update index.mdx

ElierHG:
## Describe this PR

The way the docs recommend to import prisma causes the following error when running build with Vite on a Svelte Kit procject:

import { PrismaClient } from '@prisma/client';
         ^^^^^^^^^^^^
SyntaxError: Named export 'PrismaClient' not found. The requested module '@prisma/client' is a CommonJS module, which may not support all module.exports as named exports.

## Changes

Import prisma like this instead:

```
import type { PrismaClient } from '@prisma/client'
import Prisma from '@prisma/client'
```

And the rest of the example: 
```
// add prisma to the NodeJS global type
interface CustomNodeJsGlobal extends NodeJS.Global {
    prisma: PrismaClient
}

// Prevent multiple instances of Prisma Client in development
declare const global: CustomNodeJsGlobal

const prisma = global.prisma || new Prisma.PrismaClient()

if (process.env.NODE_ENV === 'development') {
    global.prisma = prisma
}
```

### DIFF
--- a/content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
+++ b/content/300-guides/100-performance-and-optimization/150-connection-management/index.mdx
@@ -87,7 +87,8 @@ Frameworks like [Next.js](https://nextjs.org/) support hot reloading of changed 
 As a workaround, you can store `PrismaClient` as a global variable in development environments only, as global variables are not reloaded:
 
 ```ts file=client.ts
-import { PrismaClient } from '@prisma/client'
+import type { PrismaClient } from '@prisma/client'
+import Prisma from '@prisma/client'
 
 // add prisma to the NodeJS global type
 interface CustomNodeJsGlobal extends NodeJS.Global {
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks so much, @ElierHG ! We'll get a quick review on this, then merge.

tanberry:
OK, thanks @dpetrick for reviewing, I'll go ahead and merge now. Your help is appreciated @ElierHG !

janpio (changes requested):
This is not what we want.

tanberry:
@janpio can you elaborate? Are changes needed to the proposal, or do we not want to change the current docs at all?

janpio:
Internal Slack thread: https://prisma-company.slack.com/archives/C016KUHB1R6/p1651069938761599?thread_ts=1651010449.801909&cid=C016KUHB1R6

Gist: 
1. We have a PR open to fix the problem in general, which might make a change to this code not necessary any more.
2. Also, this code is an extended version of the code we suggest to our users when they `prisma generate`, so if we change this, then we would also need to update that and that is not desireable.

tanberry:
Thanks so much for digging into this @janpio and finding https://github.com/prisma/prisma/pull/12907 .

@ElierHG I'd like to close this PR and rely on #12907. If later you feel we still need to make changes to the docs, please open a new one. Thanks!

PR #3094, opened 2022-04-26
Add info about Apple M1-specific prerequisites

zachtil:
## Describe this PR

On a Macbook with an M1 chip running macOS Monterey, cloning the docs/ repo and running `npm install` results in the following error during the installation of core-js@2.6.12:

...
npm ERR! ../src/common.cc:24:10: fatal error: 'vips/vips8' file not found
npm ERR! #include <vips/vips8>
...

`npm install` is successful after libvips is installed with homebrew.

## Changes

Added a **Prerequisites** section to the README listing the installation of libvips on macOS Monterey.

## What issue does this fix?

No issue.

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,11 @@ Read through the [contributing guide](CONTRIBUTING.md) to learn how you can cont
 ## Prerequisites for macOS Monterey (Apple M1 chip)
 
 1. Install [homebrew](https://brew.sh/).
-2. Install libvips via homebrew.
+2. Install [`libvips`](https://github.com/libvips/libvips) via homebrew.
 
-```
-brew install libvips
-```
+    ```
+    brew install libvips
+    ```
 
 
 ## Run locally
--- a/README.md
+++ b/README.md
@@ -13,6 +13,16 @@ This repository contains the [source code](./src) and the [content](./content) f
 
 Read through the [contributing guide](CONTRIBUTING.md) to learn how you can contribute to the Prisma documentation.
 
+## Prerequisites for macOS Monterey (Apple M1 chip)
+
+1. Install [homebrew](https://brew.sh/).
+2. Install libvips via homebrew.
+
+```
+brew install libvips
+```
+
+
 ## Run locally
 
 Clone this repository and get started by running the following commands:
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ This repository contains the [source code](./src) and the [content](./content) f
 
 Read through the [contributing guide](CONTRIBUTING.md) to learn how you can contribute to the Prisma documentation.
 
-## Prerequisites for macOS Monterey (Apple M1 chip)
+## Prerequisites for Apple M1 chip (macOS BigSur and later)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,16 @@ This repository contains the [source code](./src) and the [content](./content) f
 
 Read through the [contributing guide](CONTRIBUTING.md) to learn how you can contribute to the Prisma documentation.
 
+## Prerequisites for macOS Monterey (Apple M1 chip)
+
+1. Install [homebrew](https://brew.sh/).
+2. Install libvips via homebrew.
```
tanberry: To make it clear that the command applies to the second pre-req, let's add a phrase like "Install livips via homebrew, using the following command:"

We have some decisions to make and add to our Style Guide... for example, I would bold **livips** to make it stand out as a product, others might use courier font, you left it normal font (which reads fine), etc. No rush, but eventually let's decide and add to Style Guide.

zachtil: How about nesting the codeblock? A single indentation (tab) didn't work, but a double-indentation nests the codeblock under the second step, so they look connected.

It makes sense to format product names somehow. One consideration that comes to mind with bold is that it is mostly used for UI elements (buttons, checkbox names, tab names, etc.).

zachtil: Joel posted [this](https://github.com/prisma/prisma/blob/main/CONTRIBUTING.md#general-prerequisites) as an example, so adding package names as code snippets with a link to their project seems like a convention we could use.

README.md
```diff
@@ -13,6 +13,16 @@ This repository contains the [source code](./src) and the [content](./content) f
 
 Read through the [contributing guide](CONTRIBUTING.md) to learn how you can contribute to the Prisma documentation.
 
+## Prerequisites for macOS Monterey (Apple M1 chip)
```
janpio: Is this a prereq for Monterey or running on M1?

README.md
tanberry: Yep I like how Joël added a link to their projects... and I agree we can use code font if it is a true package name, nice solution. One day we will add that to the Style Guide. ;-)

README.md
tanberry: The M1 chip itself... yeah so we could just not mention the OS version... though it is helpful to remind people of which OS that chi[p started appearing in.

@zachtil the heading could be something like ## Prerequisites for Apple M1 chip (macOS Monterey and later)

zachtil: M1s first started on Big Sur (Nov 2020), and Monterey was released in 2021.

I've just tested on my old Macbook (2013 model) which runs the now 2-year old Catalina and `npm install` just worked, so the Intel machines seem clear.

Is it worth it test on an M1 machine that still runs Big Sur just to exclude the OS variable?

zachtil: Seems to be definitely related to M1 and the fact that sharp requires vips - https://stackoverflow.com/questions/54409953/cant-install-sharp/56108335.

I'll modify as suggested, just indicate "macOS BigSur and later".

PR #3100, opened 2022-04-27
chore: add undocumented `BYTES` type to SQLite

### LINKED ISSUE #10537: Support for SQLite binary BLOB type
## Problem

I am using SQLite for a few projects due to its simplicity and no need to maintain any infrastructure, but it seems that Prisma does not support its `BLOB` (binary) type.

I see that issue https://github.com/prisma/prisma/issues/1747 added support for the binary types in other database engines, but not SQLite. The [docs](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#bytes) also mention not supported.

sqlite not supported

## Suggested solution

Not sure what the solution would look like, but similar to other engines supporting `BLOB` would be nice.

cstrnt:
## Describe this PR

This PR updates the documentation to show that the SQLite adapter actually supports the `BYTES` type as stated in  [#10537](https://github.com/prisma/prisma/issues/10537)

## Changes
- Updated the database-connector docs for SQLite by updating the compatibility table
- Update prisma-schema-reference docs

## What issue does this fix?

- [#1747](https://github.com/prisma/prisma/issues/1747)
- closes [#10537 ](https://github.com/prisma/prisma/issues/10537)

### DIFF
--- a/content/200-concepts/200-database-connectors/05-sqlite.mdx
+++ b/content/200-concepts/200-database-connectors/05-sqlite.mdx
@@ -42,7 +42,7 @@ The SQLite connector maps the [scalar types](/concepts/components/prisma-schema/
 | `Float`    | `REAL`        |
 | `Decimal`  | `DECIMAL`     |
 | `Json`     | Not supported |
-| `Bytes`    | Not supported |
+| `Bytes`    | `BYTES`       |
 
 ## Connection details
 
--- a/content/200-concepts/200-database-connectors/05-sqlite.mdx
+++ b/content/200-concepts/200-database-connectors/05-sqlite.mdx
@@ -42,7 +42,7 @@ The SQLite connector maps the [scalar types](/concepts/components/prisma-schema/
 | `Float`    | `REAL`        |
 | `Decimal`  | `DECIMAL`     |
 | `Json`     | Not supported |
-| `Bytes`    | `BYTES`       |
+| `Bytes`    | `BLOB`       |
 
 ## Connection details
 
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -928,7 +928,7 @@ Not supported
 | Microsoft SQL | `varbinary`     |
 | MySQL         | `LONGBLOB`      |
 | MongoDB       | `BinData`       |
-| SQLite        | `BYTES`         |
+| SQLite        | `BLOB`         |
 | CockroachDB   | `BYTES`         |
 
 #### PostgreSQL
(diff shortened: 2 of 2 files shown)


janpio:
Referenced wrong issue? Seems that one is a closed unrelated one.

cstrnt:
> Referenced wrong issue? Seems that one is a closed unrelated one.

Sorry, I copied the issue ids from the main repo thinking this was same, updated the links

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -928,7 +928,7 @@ Not supported
 | Microsoft SQL | `varbinary`     |
 | MySQL         | `LONGBLOB`      |
 | MongoDB       | `BinData`       |
-| SQLite        | Not supported   |
+| SQLite        | `BYTES`         |
```
janpio: Isn't this `BLOB` here then (per the SQL in https://github.com/prisma/prisma/issues/10537#issuecomment-985052256)

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -928,7 +928,7 @@ Not supported
...
 | CockroachDB   | `BYTES`         |
```
janpio, line 933: Line 969 probably also needs to be updated - I would expect a native type `@db.Blob` to exist then according to the usual pattern. We need to test and confirm that first of course.

janpio:
Ah, now things make sense 👍 Left a few comments for details.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
cstrnt: Prisma doesn't support any custom type mapping for SQLite. I found this test here (https://github.com/prisma/prisma/blob/8f6b7c7c99c1c720cf5bfed5d563423e71c1b84f/packages/migrate/src/__tests__/fixtures/nativeTypes-sqlite/prisma/schema.prisma) where it actually uses `@db.VarChar`but It doesn't work locally for me. Also tried other options that should be supported in SQLite (e.g. DateTime as `db.Integer`) but it doesn't work either

janpio: You are of course right, on https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types we usually just list the main type then - so `BLOB` in this case.

janpio (approved):
👍 

@millsp? @tomhoule?

content/200-concepts/200-database-connectors/05-sqlite.mdx
```diff
@@ -42,7 +42,7 @@ The SQLite connector maps the [scalar types](/concepts/components/prisma-schema/
 | `Float`    | `REAL`        |
 | `Decimal`  | `DECIMAL`     |
 | `Json`     | Not supported |
-| `Bytes`    | Not supported |
+| `Bytes`    | `BYTES`       |
```
tomhoule: I think this should be `BLOB` too

cstrnt: You're right!

millsp (approved):

tomhoule (approved):

PR #3115, opened 2022-04-29
Clarify that  _null enums_ require use of `equals`

### LINKED ISSUE #3103: Filtering on Null requires use of `equals`
Per [this Slack thread](https://prisma-company.slack.com/archives/C016KUHB1R6/p1651006022206219) a user got errors trying to use `anyNull` based on the [code examples in our documentation](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#filtering-by-null-values). 

We need to change the code example in docs FROM this:
```
import { Prisma } from '@prisma/client'

prisma.log.findMany({
  where: {
    data: {
      meta: Prisma.AnyNull,
    },
  },
})
```

TO this (using `equals`):
```
prisma.log.findMany({
  where: {
    data: {
      meta: {
        equals: Prisma.AnyNull
      }
    },
  },
})

```

Let's add a note in the "Filtering on null values" section of the doc (to the existing Notes), and move the whole block of Notes up higher, right before the "For example" line.

New Notes block draft:
```Notes:
* When filtering using any of the `null enums` you must use the `equals` operator, as shown in the example below.
* These null enums do not apply to MongoDB because there the difference between a JSON null and a database NULL does not exist.
*They also do not apply to the array_contains operator in all databases because there can only be a JSON null within a JSON array. Since there cannot be a database NULL within a JSON array, { array_contains: null } is not ambiguous.

### PR DESCRIPTION
**Pull request by tanberry**
Resolves https://github.com/prisma/docs/issues/3103

### CODE REVIEW

File: content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx [hunk 1]
```diff
@@ -874,7 +874,15 @@ To differentiate between these possibilities, we've introduced three _null enums
...
-For example:
+<Admonition type="info">
+
+  * When filtering using any of the _null enums_ you must use the `equals` operator, as shown in the example below.
+  * These _null enums_ do not apply to MongoDB because there the difference between a JSON `null` and a database `NULL` does not exist.
+  * The _null enums_ do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
+
+</Admonition>
+  
+  For example:
```
janpio:
```suggestion
For example:
```

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -874,7 +874,15 @@ To differentiate between these possibilities, we've introduced three _null enums
 - `DbNull`: Selects the `NULL` value in the database.
 - `AnyNull`: Selects both `null` JSON values and `NULL` database values.
 
-For example:
+<Admonition type="info">
+
+  * When filtering using any of the _null enums_ you must use the `equals` operator, as shown in the example below.
```
janpio: I don't think it is always `equals` - it is just that generally no shorthand is possible (which means `equals`).

```suggestion
  * When filtering using any of the _null enums_ you can not use a shorthand and leave the `equals` operator off.
```

@dpetrick?

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -889,7 +897,9 @@ import { Prisma } from '@prisma/client'
 prisma.log.findMany({
   where: {
     data: {
-      meta: Prisma.AnyNull,
+      meta: {
+        equals: Prisma.AnyNull
+      }
```
austincrim, line 905: need to update the line highlighting for this code block

```
```ts highlight=7;normal
```

tanberry: I think the highlight issues are related to a bug that @Nilu is working.  I'd like to merge this and then look into the highlighting problems, ok? Or are you saying there is a one-off issue and know how to fix it?

austincrim: i think we can just change line 894 to what i suggested above^^.

i've not been involved with the highlighting issues, though, so you can move forward however you think is best

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -874,6 +874,14 @@ To differentiate between these possibilities, we've introduced three _null enums
 - `DbNull`: Selects the `NULL` value in the database.
 - `AnyNull`: Selects both `null` JSON values and `NULL` database values.
 
+<Admonition type="info">
+
+  * When filtering using any of the _null enums_ you can not use a shorthand and leave the `equals` operator off.
```
janpio, line 880: I am not sure if this is exactly correct. @dpetrick?

dpetrick: Can be formulated more clearly, but the statement is correct: There are no shorthands for json filtering. The reason is simple: We can't construct a TypeScript API that allows arbitrary JSON and have shorthands at the same time, because you can't differentiate a shorthand from a JSON object if you don't have a fixed depth of the query. E.g. there's no way for us to differentiate if the user wants to use a shorthand or if the user wants to store an object with `{"equals": ...}`.

dpetrick (approved):

PR #3120, opened 2022-05-02
Note CockroachDB "char" intended for internal use

kernfeld-cockroach:
Note that CockroachDB "char" is intended for internal use in system catalogs. CockroachDB docs: https://www.cockroachlabs.com/docs/stable/string.html#related-types

@all-contributors please add @kernfeld-cockroach for doc

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -454,15 +454,15 @@ model Model {
 
 #### CockroachDB (Preview)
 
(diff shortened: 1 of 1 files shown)


tomhoule:
Thanks! Looks good to me.

kernfeld-cockroach:
The Netlify Deploy "Create Comment" step is giving me a "Resource not accessible by integration," which seems to be a permissions error, maybe because I'm not a contributor. Is there something I can do to fix this?

janpio:
No, that is all good - we need to figure out a way to fix that sooner or later. We'll be able to merge this PR just fine.

PR #3123, opened 2022-05-04
PostgreSQL inverted indices

### LINKED ISSUE #3102: PostgreSQL: GiST, GIN, SP-GiST and BRIN Indices
In Prisma version 3.14.0, together with `extendedIndexes` preview feature we start supporting more index types for PostgreSQL.

Here's a short example how we extend our syntax in the PSL. Let's figure the following DDL

```sql
CREATE TABLE "A" (id int PRIMARY KEY, val inet);
CREATE INDEX ON "A" USING GIST (val inet_ops);
```

Currently in 3.13.0 we introspect this as following:

```prisma
model A {
  id  Int    @id
  val String @db.Inet

  @@index([val])
}
```

If the user would push this to a new database, the underlying index would be BTree, not Gist.

With 3.14.0 using the `extendedIndexes` preview this becomes:

```prisma
model A {
  id  Int    @id
  val String @db.Inet

  @@index([val(ops: InetOps)], type: Gist)
  //                                 ^ the USING value in the DDL
  //                ^ this maps the inet_ops in the DDL
}
```

The value of `ops` depends on the `type` argument of the `index` attribute, and the native type of the `val` field. Prisma will validate the correct combinations, but the validations will not do the correct thing with the (now deprecated) version 9 of the database, and some of the operators in Brin indices are only available on version 14 of the database.

In general we should communicate to use these types with version 10 onwards, and tell our users the older version might work, but we don't validate and test them.

Some of the operators are considered the default, and can be taken out from the PSL and so forth the DDL we generate. An example:

```sql
CREATE TABLE "A" (id int PRIMARY KEY, val jsonb);
CREATE INDEX ON "A" USING GIN (val jsonb_ops);

-- is the same as

CREATE TABLE "A" (id int PRIMARY KEY, val jsonb);
CREATE INDEX ON "A" USING GIN (val);
```

And therefore the PSLs:

```prisma
model A {
  id  Int   @id
  val Json

  @@index([val], type: Gin)
}

// is the same as

model A {
  id  Int   @id
  val Json

  @@index([val(ops: JsonbOps)], type: Gin)
}
```

All operators, their index and native types are available in the [design documentation](https://www.notion.so/prismaio/Proposal-More-PostgreSQL-index-types-GiST-GIN-SP-GiST-and-BRIN-e27ef762ee4846a9a282eec1a5129270).

# CockroachDB

CDB supports a small subset of inverted indices, support was [merged](https://github.com/prisma/prisma/issues/13065) to be released in 3.14.0. The linked pull request lists all supported types to be documented.

pimeys:
## Describe this PR

We support new index types from version 3.14.0 forwards.

## Changes

How PostgreSQL inverted indices work.

## What issue does this fix?

Fixes #3102

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
@@ -175,7 +175,7 @@ CREATE INDEX "Example_value_idx" ON "Example" USING HASH (value);
 
 #### Generalized Inverted Index (GIN)
 
-The GIN index stores composite values, such as arrays or `JsonB`, speeding up queries such as is an object part of another object. Also commonly used for full-text searches.
+The GIN index stores composite values, such as arrays or `JsonB`, speeding up queries like "is an object part of another object". Commonly used for full-text searches.
 
 An indexed field can define the operator class, which defines the operators handled by the index.
 
--- a/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
@@ -339,7 +339,7 @@ Now a query, such as `value = 2` is using the index, which uses a fraction of a 
 
 ##### Supported Operator Classes for BRIN
 
-Prisma generally supports operator classes delivered with PostgreSQL version 10 forward, and the bloom operators with PostgreSQL version 14 forward. If the operator class requires field type to be of a type Prisma yet does not support, using the `raw` function with a string input allows using these operator classes without validation.
+Prisma generally supports operator classes delivered with PostgreSQL version 10 forward; and some of the supported operators are only available from PostgreSQL version 14 forward. If the operator class requires field type to be of a type Prisma yet does not support, using the `raw` function with a string input allows using these operator classes without validation.
 
 The default operator class (marked with ✅) can be omitted from the index definition.
 
--- a/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
@@ -274,6 +274,9 @@ model Example {
 Default operator classes are not required to be explicitly written in the data model for the given native types.
 
 The operator classes marked to work with CockroachDB are the only ones allowed on that database. On CockroachDB the operator class must be emitted from the field definition.
+
+More details on operator and index types can be read from the official documentation: [PostgreSQL](https://www.postgresql.org/docs/14/indexes-types.html), [CockroachDB](https://www.cockroachlabs.com/docs/v21.2/inverted-indexes.html)
+
 </details>
 
 ### Configuring if indexes are clustered or non-clustered with `clustered` (SQL Server)
--- a/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
@@ -251,7 +251,7 @@ CREATE TABLE "Example" (
 CREATE INDEX "Example_value_idx" ON "Example" USING GIST (value inet_ops);
 ```
 
-Now queries comparing IP addresses, such as `value > '10.0.0.2'` are captured by the index.
+Now queries comparing IP addresses, such as `value > '10.0.0.2'` is using the index.
 
 ##### Supported Operator Classes for GiST
 
@@ -311,7 +311,7 @@ The default operator class (marked with ✅) can be omitted from the index defin
 
 BRIN index is useful if having lots of data that does not change after inserted, such as a date and time values. If the data fits well to the index, it can store large datasets into a minimal space.
 
-As an example, the following model adds a `Brin` index to the `value` field with `Int4BloomOps` as the operators that will be captured by the index:
+As an example, the following model adds a `Brin` index to the `value` field with `Int4BloomOps` as the operators that will be using the index:
 
 ```prisma file=schema.prisma
 model Example {
@@ -335,7 +335,7 @@ CREATE TABLE "Example" (
 CREATE INDEX "Example_value_idx" ON "Example" USING BRIN (value int4_bloom_ops);
 ```
 
-Now a query, such as `value = 2` is captured by the index, which uses a fraction of a space compared to the `BTree` or a `Hash` indexexes.
+Now a query, such as `value = 2` is using the index, which uses a fraction of a space compared to the `BTree` or a `Hash` indexexes.
 
 ##### Supported Operator Classes for BRIN
 
--- a/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
@@ -208,14 +208,16 @@ model Example {
 }
 ```
 
-Operator classes currently supported in the Prisma Schema Language for Gist indexes:
+Default operator classes are not required to be explicitly written in the data model for the given native types.
+
+Operator classes currently supported in the Prisma Schema Language for **Gist** indexes:
 
 | operator class | allowed field type (allowed native types) |
 |----------------|-------------------------------------------|
 | `InetOps`      | `String` (`@db.Inet`)                     |
 | `raw("other")` |                                           |
 
-Operator classes currently supported in the Prisma Schema Language for Gin indexes:
+Operator classes currently supported in the Prisma Schema Language for **Gin** indexes:
 
 | operator class | allowed field type (native types) | default | other            |
 |----------------|-----------------------------------|---------|------------------|
@@ -224,7 +226,7 @@ Operator classes currently supported in the Prisma Schema Language for Gin index
 | `JsonbPathOps` | `Json` (`@db.JsonB`)              |         |                  |
 | `raw("other")` |                                   |         |                  |
 
-Operator classes currently supported in the Prisma Schema Language for SpGist indexes:
+Operator classes currently supported in the Prisma Schema Language for **SpGist** indexes:
 
 | operator class | allowed field type (native types)    | default | supported PostgreSQL versions |
 |----------------|--------------------------------------|---------|-------------------------------|
@@ -232,7 +234,7 @@ Operator classes currently supported in the Prisma Schema Language for SpGist in
 | `TextOps`      | `String` (`@db.Text`, `@db.VarChar`) | ✅      |                               |
 | `raw("other")` |                                      |         |                               |
 
-Operator classes currently supported in the Prisma Schema Language for Brin indexes:
+Operator classes currently supported in the Prisma Schema Language for **Brin** indexes:
 
 | operator class              | allowed field type (native types)    | default | supported PostgreSQL versions |
 |-----------------------------|--------------------------------------|---------|-------------------------------|
@@ -289,8 +291,6 @@ Operator classes currently supported in the Prisma Schema Language for Brin inde
 | `UuidMinMaxOps`             | `String` (`@db.Uuid`)                | ✅      |                               |
 | `UuidMinMaxMultiOps`        | `String` (`@db.Uuid`)                |         | 14+                           |
 | `raw("other")`              |                                      |         |                               |
-
-Default operator classes are not required to be explicitly written in the data model for the given native types.
 
 The operator classes marked to work with CockroachDB are the only ones allowed on that database. On CockroachDB the operator class must be emitted from the field definition.
 
--- a/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
@@ -122,7 +122,7 @@ model CompoundUnique {
   unique_1 Int
   unique_2 Int
 
-  @@unique([unique_1(sort: Desc),unique_2])
+  @@unique([unique_1(sort: Desc), unique_2])
 }
 ```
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,116 @@ This translates to the following SQL commands:
 ```sql
 CREATE TABLE "Example" (
   id INT PRIMARY KEY,
-  val INT NOT NULL
+  value INT NOT NULL
+);
+
+CREATE INDEX "Example_value_idx" ON "Example" USING HASH (val);
```
janpio:
```suggestion
CREATE INDEX "Example_value_idx" ON "Example" USING HASH (value);
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,135 @@ This translates to the following SQL commands:
...
   id INT PRIMARY KEY,
-  val INT NOT NULL
+  value INT NOT NULL
+);
+
+CREATE INDEX "Example_value_idx" ON "Example" USING HASH (value);
+```
+
+#### Gist, Gin, SpGist and Brin
+
```
janpio, line 177: TODO General description of the 4 types, similar to how the Hash one is explained above.

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,135 @@ This translates to the following SQL commands:
...
+CREATE TABLE "Example" (
+  id INT PRIMARY KEY,
+  value INET NOT NULL
 );
 
-CREATE INDEX "Example_val_idx" ON "Example" USING HASH (val);
+CREATE INDEX "Example_value_idx" ON "Example" USING GIST (value inet_ops);
 ```
 
+The allowed value to the operator class (`ops`) argument depends on the native type of the field, and the type of the index. Prisma Schema Language provides all operator classes provided natively by PostgreSQL version 14.0, if the native type is supported by Prisma. For cases where we don't yet support the operator class, or the native type of the indexed field, the `raw` function can be used as an escape hatch:
```
janpio:
```suggestion
The allowed value to the operator class (`ops`) argument depends on the native type of the field, and the type of the index. Prisma Schema Language provides all operator classes provided natively by PostgreSQL version 14.0, if the native type is supported by Prisma. 

For cases where we don't yet support the operator class, or the native type of the indexed field, the `raw` function can be used as an escape hatch:
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,135 @@ This translates to the following SQL commands:
...
+| `raw("other")` |                                           |
+
+Operator classes currently supported in the Prisma Schema Language for **Gin** indexes:
+
+| operator class | allowed field type (native types) | default | other            |
+|----------------|-----------------------------------|---------|------------------|
+| `ArrayOps`     | any array                         | ✅      | also CockroachDB |
+| `JsonbOps`     | `Json` (`@db.JsonB`)              | ✅      | also CockroachDB |
+| `JsonbPathOps` | `Json` (`@db.JsonB`)              |         |                  |
+| `raw("other")` |                                   |         |                  |
```
janpio: Switch these as we usually list Gin before Gist and Gist is closer to SpGist this way.

```suggestion
Operator classes currently supported in the Prisma Schema Language for **Gin** indexes:

| operator class | allowed field type (native types) | default | other            |
|----------------|-----------------------------------|---------|------------------|
| `ArrayOps`     | any array                         | ✅      | also CockroachDB |
| `JsonbOps`     | `Json` (`@db.JsonB`)              | ✅      | also CockroachDB |
| `JsonbPathOps` | `Json` (`@db.JsonB`)              |         |                  |
| `raw("other")` |                                   |         |                  |

Operator classes currently supported in the Prisma Schema Language for **Gist** indexes:

| operator class | allowed field type (allowed native types) |
|----------------|-------------------------------------------|
| `InetOps`      | `String` (`@db.Inet`)                     |
| `raw("other")` |                                           |
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,135 @@ This translates to the following SQL commands:
...
+| `TimeMinMaxMultiOps`        | `DateTime` (`@db.Time`)              |         | 14+                           |
+| `TimeTzBloomOps`            | `DateTime` (`@db.Timetz`)            |         | 14+                           |
+| `TimeTzMinMaxOps`           | `DateTime` (`@db.Timetz`)            | ✅      |                               |
+| `TimeTzMinMaxMultiOps`      | `DateTime` (`@db.Timetz`)            |         | 14+                           |
+| `UuidBloomOps`              | `String` (`@db.Uuid`)                |         | 14+                           |
+| `UuidMinMaxOps`             | `String` (`@db.Uuid`)                | ✅      |                               |
+| `UuidMinMaxMultiOps`        | `String` (`@db.Uuid`)                |         | 14+                           |
+| `raw("other")`              |                                      |         |                               |
+
+The operator classes marked to work with CockroachDB are the only ones allowed on that database. On CockroachDB the operator class must be emitted from the field definition.
```
janpio: > On CockroachDB the operator class must be emitted from the field definition.

What does this mean?

pimeys: You cannot set `ops` in the index field definition. Under the surface it uses the operator classes from the table, but you cannot explicitly tell them in the DDL and the information schema does not return them in the query.

janpio: Then maybe:
```suggestion
The operator classes marked to work with CockroachDB are the only ones allowed on that database. The operator class can further not be changed from the one possible value, `ops` is not necessary or allowed for CockroachDB.
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,137 @@ This translates to the following SQL commands:
...
+| `TimeMinMaxMultiOps`        | `DateTime` (`@db.Time`)              |         | 14+                           |
+| `TimeTzBloomOps`            | `DateTime` (`@db.Timetz`)            |         | 14+                           |
+| `TimeTzMinMaxOps`           | `DateTime` (`@db.Timetz`)            | ✅      |                               |
+| `TimeTzMinMaxMultiOps`      | `DateTime` (`@db.Timetz`)            |         | 14+                           |
+| `UuidBloomOps`              | `String` (`@db.Uuid`)                |         | 14+                           |
+| `UuidMinMaxOps`             | `String` (`@db.Uuid`)                | ✅      |                               |
+| `UuidMinMaxMultiOps`        | `String` (`@db.Uuid`)                |         | 14+                           |
+| `raw("other")`              |                                      |         |                               |
+
+The operator classes marked to work with CockroachDB are the only ones allowed on that database. The operator class can further not be changed from the one possible value, `ops` is not necessary or allowed for CockroachDB.
```
tanberry:
```suggestion
Note the following information:

* The operator classes marked to work with CockroachDB are the only ones allowed on that database
* The operator class cannot be changed from the value specified in the table above
* `ops` is not necessary nor allowed for CockroachDB
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,137 @@ This translates to the following SQL commands:
...
+```sql
+CREATE TABLE "Example" (
+  id INT PRIMARY KEY,
+  value INET NOT NULL
+);
+
+CREATE INDEX "Example_value_idx" ON "Example" USING GIST (value inet_ops);
+```
+
+The allowed value to the operator class (`ops`) argument depends on the native type of the field, and the type of the index. Prisma Schema Language provides all operator classes provided natively by PostgreSQL version 14.0, if the native type is supported by Prisma.
```
tanberry:
```suggestion
The allowed value of the operator class (`ops`) argument depends on the native type of the field, and the type of the index. Prisma Schema Language supports all operator classes provided natively by PostgreSQL version 14.0, if the native type is supported by Prisma. 
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,137 @@ This translates to the following SQL commands:
...
+  value INT NOT NULL
 );
 
-CREATE INDEX "Example_val_idx" ON "Example" USING HASH (val);
+CREATE INDEX "Example_value_idx" ON "Example" USING HASH (value);
+```
+
+#### Gist, Gin, SpGist and Brin
+
+These four index types are special in how they let you in some cases to define the operators catched by the index. Depending on the field type, it is sometimes required to write the operators to the indexed field. As an example, the following model adds an index with a `type` of `Gist` to the `value` field with `InetOps` as the operators that will be using the index:
```
tanberry:
```suggestion
These four index types are special in how they let you, in some cases, define the operators cached by the index. Depending on the field type, it is sometimes required to write the operators to the indexed field. As an example, the following model adds an index with a `type` of `Gist` to the `value` field with `InetOps` as the operators that will be using the index:
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -165,12 +167,137 @@ This translates to the following SQL commands:
...
   id INT PRIMARY KEY,
-  val INT NOT NULL
+  value INT NOT NULL
 );
 
-CREATE INDEX "Example_val_idx" ON "Example" USING HASH (val);
+CREATE INDEX "Example_value_idx" ON "Example" USING HASH (value);
+```
+
+#### Gist, Gin, SpGist and Brin
```
tanberry:
```suggestion
#### Gist, Gin, SpGist and Brin index types
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -30,7 +30,7 @@ The following arguments can be specified:
 
 - The `sort` argument is available for all databases on the `@unique`, `@@unique` and `@@index` attributes in version `3.5.0` and later. Additionally, SQL Server also allows it on `@id` and `@@id`.
 
-- The `type` argument is also available **in PostgreSQL only** on the `@@index` attribute in version `3.6.0` and later. This argument allows Prisma to support the `Hash` index access method as well as the default `BTree` access method.
+- The `type` argument is also available **in PostgreSQL only** on the `@@index` attribute in version `3.6.0` and later. This argument allows Prisma to support the `Hash` index access method as well as the default `BTree` access method. Version `3.14.0` adds additional access methods `Gist`, `Gin`, `SpGist` and `Brin`.
```
tanberry, line 33:
```suggestion
- The `type` argument is also available **in PostgreSQL only** on the `@@index` attribute in version `3.6.0` and later. This argument allows Prisma to support the `Hash` index access method as well as the default `BTree` access method. 
- Prisma version `3.14.0` adds additional access methods with the index types `Gist`, `Gin`, `SpGist` and `Brin`.
```

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
```diff
@@ -145,7 +145,9 @@ model Post {
 
 ### Configuring the access type of indexes with `type` (PostgreSQL)
 
-The `type` argument is available for configuring the index type in PostgreSQL in version `3.6.0` and later, using the `@@index` attribute. This allows you to use `Hash` as the index access method, instead of the default `BTree` access method.
+The `type` argument is available for configuring the index type in PostgreSQL in version `3.6.0` and later, using the `@@index` attribute. This allows you to use `Hash` as the index access method, instead of the default `BTree` access method. Versions `3.14.0` and later allow additionally the `Gist`, `Gin`, `SpGist` and `Brin` access methods.
```
tanberry: Ideally we would not duplicate the exact same content as on line #33. We could leave line 33 as it is, and here simply summarize the options, like this:

The `type` argument is available for configuring the index type in PostgreSQL, with the `@@index` attribute (version `3.6.0` and later). This allows you to use `Hash` as the index access method, instead of the default `BTree` access method. The `Gist`, `Gin`, `SpGist` and `Brin` index types provide additional access methods (version `3.14.0` and later).

tanberry (commented):
thank so very much @pimeys for writing this up, and especially for wrestling with that massive table!

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
janpio: Second sentence is also CockroachDB specific - if we turn this into a list we should probably in the intro already mention that this is all for CockroachDB.

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
janpio: Did you mean "cache" or "catch" @pimeys? Maybe "caught" is the better replacement here? Or some completely different word?

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
janpio: Each item in this list is one different argument to `@@index`. This item you broke up is the one for `type` (vs `sort` and `length` before for example). Instead of breaking this into another bullet we might want to rewrite the sentence. (Note that with the preview feature going away soon this will need to be rewritten anyway though and the version information can probably be moved somewhere else)

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
tanberry: Ooops yeah I assumed "cached" but maybe "caught" is what was intended...

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
tanberry: And @pimeys I am not certain that "supports" is the correct word here, though I think it is and is sounds better than repeating "provides".

content/200-concepts/100-components/01-prisma-schema/09-indexes.mdx
tanberry: Ah, yes, then let's not break it up! Let's leave it how it was but remove the version info. I'll do that as a suggestion.

pimeys:
Split up the index types into their own sections.

PR #3129, opened 2022-05-05
CockroachDB Sequences

### LINKED ISSUE #3124: Update docs to reflect new `@default(sequence())` option for CockroachDB columns
Per work done by the PSL Working Group, we need to update the docs to explain how CRDB users can implement `@default(sequence())` with two options for `sequences` cached sequences and virtual sequences. Adding these to the PSL will also mean that migrations and introspections work better.

See details in the PSL WG document "[Proposal: CockroachDB autoincrement columns](https://www.notion.so/prismaio/Proposal-CockroachDB-autoincrement-columns-2d0a0d46a350429da503ca29824dd7c8)".

pimeys:
-------

## Describe this PR

Changes needed for upcoming (3.14.0) sequence work for CockroachDB.

## Changes

The `@default(autoincrement())` can now only be used together with the `BigInt` field type on CockroachDB. Additionally the `@default(sequence())` can be used to define the charasteristics of the sequence the field value is based on.

## What issue does this fix?

Fixes #3124


content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+**Only supported by CockroachDB** <br />
+The sequence function is for now only supported by [CockroachDB connector](/concepts/database-connectors/cockroachdb).
+
+</Admonition>
+
+Create a sequence of integers in the underlying database and assign the incremented values to the ID values of the created records based on the sequence.
+
+#### Optional arguments
+
+| Argument    | Example                                               | Description                                                                                                                                                                                                                                           |
```
pimeys: this table doesn't look very good rendered...

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2907,7 +2907,7 @@ Create a sequence of integers in the underlying database and assign the incremen
 
 #### Remarks
 
-- Compatible with `Int`
+- Compatible with `BigInt` on CockroachDB, or with `Int` on other databases.
```
janpio: Maybe list more common case first: 
```suggestion
- Compatible with `Int` on most databases (`BigInt` on CockroachDB)
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
 </TabbedContent>
 
+### <inlinecode>sequence()</inlinecode>
+
+<Admonition type="warning">
+
+**Only supported by CockroachDB** <br />
+The sequence function is for now only supported by [CockroachDB connector](/concepts/database-connectors/cockroachdb).
+
+</Admonition>
```
janpio: Info only (less scary), and no future facing statement:
```suggestion
<Admonition type="info">

**Only supported by CockroachDB** <br />
The sequence function is only supported by [CockroachDB connector](/concepts/database-connectors/cockroachdb).

</Admonition>
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+
+Create a sequence of integers in the underlying database and assign the incremented values to the ID values of the created records based on the sequence.
```
janpio: Not ID specific, is it?
```suggestion
Create a sequence of integers in the underlying database and assign the incremented values to the values of the created records based on the sequence.
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+
+#### Optional arguments
+
+| Argument    | Example                                               | Description                                                                                                                                                                                                                                           |
+|-------------|-------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `virtual`   | `@default(sequence(virtual))`                         | Virtual sequences are sequences that do not generate monotonically increasing values and instead produce values like those generated by the built-in function `unique_rowid()`. They are intended for use in combination with `SERIAL`-typed columns. |
```
janpio: What does "`SERIAL`-typed columns" mean here?

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+| `cache`     | `@default(sequence(cache: 20))`                       | The number of sequence values to cache in memory for reuse in the session. A cache size of `1` means that there is no cache, and cache sizes of less than `1` are not valid.                                                                          |
```
janpio: Is 1 the default here @tomhoule?

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+| `minValue`  | `@default(sequence(minValue: 10, maxValue: 3030303))` | The new minimum value of the sequence.                                                                                                                                                                                                                |
+| `maxValue`  | `@default(sequence(minValue: 10, maxValue: 3030303))` | The new maximum value of the sequence.                                                                                                                                                                                                                |
+| `start`     | `@default(sequence(start: 2))`                        | The value the sequence starts at if you `RESTART` or if the sequence hits the `maxValue` and `noCycle` is not set.                                                                                                                                    |
```
janpio: `RESTART` probably has no meaning in Prisma context, but SQL only. Do we still want to mention it?

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+| `noCycle`   | `@default(sequence(noCycle))`                         | The sequence will wrap around when the sequence value hits the maximum or minimum value. If `noCycle` is set, the sequence will not wrap.                                                                                                             |
```
janpio:
```suggestion
| `noCycle`   | `@default(sequence(noCycle))`                         | By default the sequence will wrap around when the sequence value hits the maximum or minimum value. If `noCycle` is set, the sequence will not wrap.                                                                                                             |
```

janpio: > the sequence will not wrap.

What happens instead? Error?

janpio (approved):
Great start, looks correct to me.

There are probably further locations where this should be mentioned. (Where do we use `autoincrement`? Where do we maybe use `autoincrement` with CockroachDB?) Def on the Cockroach Connector page probably.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
pimeys: Maybe we should rewrite these to fit better to our context...

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
tomhoule: Yes, that's what cockroach has in its catalog if you create a sequence without specifying the cache parameter.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
tomhoule: I haven't tested that, but that would be the expectation, yes.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
pimeys: What is the actual concrete effect of the cache for the user?

tomhoule: I assume it should be invisible to the user, but it may have performance implications.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2907,7 +2907,7 @@ Create a sequence of integers in the underlying database and assign the incremen
 
 #### Remarks
 
-- Compatible with `Int`
+- Compatible with `Int` on most databases (`BigInt` on CockroachDB)
```
tomhoule, line 2910: :+1:

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
tomhoule: We can drop that sentence, because it's one way to define autoincrementing columns on cockroachdb, but once the big PR is merged, we won't use it at all anymore.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2936,6 +2936,48 @@ model User {
...
+| `maxValue`  | `@default(sequence(minValue: 10, maxValue: 3030303))` | The new maximum value of the sequence.                                                                                                                                                                                                                |
+| `start`     | `@default(sequence(start: 2))`                        | The value the sequence starts at if you `RESTART` or if the sequence hits the `maxValue` and `noCycle` is not set.                                                                                                                                    |
+| `noCycle`   | `@default(sequence(noCycle))`                         | By default the sequence will wrap around when the sequence value hits the maximum or minimum value. If `noCycle` is set, the sequence will not wrap.                                                                                                             |
```
tomhoule: To be double-checked in cockroachdb docs, but I think they just ignore that setting. It's still configurable, but it's ignored.

tomhoule (approved):

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: Indeed, our design can not handle that very well :/ 

2 ideas:

1. Remove the "unrelated" arguments from the examples, will make some ~10 characters shorter and make the second column less wide that way.
2. We could combine "Example" and "Description" in one column, and then have the "Example" as the first line always with code highlighting, then another section of the "Description". That would make both readable at the cost of less clear separation - but then at least the description was readable.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio:
```suggestion
| `virtual`   | `@default(sequence(virtual))`                         | Virtual sequences are sequences that do not generate monotonically increasing values and instead produce values like those generated by the built-in function `unique_rowid()`. |
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: https://www.cockroachlabs.com/docs/stable/create-sequence.html#cache-sequence-values-in-memory

> For improved performance, use the CACHE keyword to cache sequence values in memory.

Asked for details though: https://prisma-company.slack.com/archives/C01CSQXTNJ3/p1651856853662709

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: https://prisma-company.slack.com/archives/C02K246T9RQ/p1651857425362169

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: And here is the response: https://glennfawcett.wpcomstaging.com/2021/08/25/seqcache/

> By allowing each session to obtain a large range of values, hot-spots to obtain values a minimized. Additionally, if a large enough cache is used, single-range hot-spots will be minimized as well.
> image

PR #3138, opened 2022-05-10
Update 03-postgresql.mdx

bindermuehle:
I think POST is a mix between HOST  and PASSWORD, my understanding is that POST should be PASSWORD.


Jolg42 (approved):
Nice one, thanks for the PR!

tanberry:
Oooof, that is a nice catch, thanks so much @bindermuehle !

PR #3153, opened 2022-05-14
Update 04-data-model.mdx

emadunan:
## Describe this PR
While I was studing the documentation, I noticed a typo mistake

## Changes
From: `@map` a column name or enum value, and `@@map` an enum.
To: `@map` a column name or enum value, and `@@map` a table name.

## What issue does this fix?
data model page in the documentation

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
@@ -298,7 +298,7 @@ model Comment {
 
 With this model definition, Prisma automatically maps the `Comment` model to the `comments` table in the underlying database.
 
-> **Note**: You can also [`@map`](/reference/api-reference/prisma-schema-reference#map) <span class="api"></span> a column name or enum value, and `@@map` a table name.
+> **Note**: You can also [`@map`](/reference/api-reference/prisma-schema-reference#map) <span class="api"></span> a column name or enum value, and `@@map` an enum name.
 
 `@map` and `@@map` allow you to [tune the shape of your Prisma Client API](/concepts/components/prisma-client/working-with-prismaclient/use-custom-model-and-field-names#using-map-and-map-to-rename-fields-and-models-in-the-prisma-client-api) by decoupling model and field names from table and column names in the underlying database.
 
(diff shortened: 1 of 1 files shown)


keerlu:
Thanks for the PR @emadunan! We'll get a review of this from the dev team and get back to you

keerlu:
I'm sorry this response is so late - I was checking with our developers to get a second opinion and then didn't follow up.

The current version of the note is trying to say that `@@map` can _also_ be used to map an enum, as well as a table name (as in the example given above). So it's technically correct currently, but looks like we could do a better job of wording it. Suggestions for a version of the text that would be less confusing are very welcome. Would something like "You can also use `@map` to map a column name or enum value, or use `@@map` to map an enum." help at all?

content/200-concepts/100-components/01-prisma-schema/04-data-model.mdx
```diff
@@ -298,7 +298,7 @@ model Comment {
 
 With this model definition, Prisma automatically maps the `Comment` model to the `comments` table in the underlying database.
 
-> **Note**: You can also [`@map`](/reference/api-reference/prisma-schema-reference#map) <span class="api"></span> a column name or enum value, and `@@map` an enum.
+> **Note**: You can also [`@map`](/reference/api-reference/prisma-schema-reference#map) <span class="api"></span> a column name or enum value, and `@@map` a table name.
```
nikolasburk:
```suggestion
> **Note**: You can also [`@map`](/reference/api-reference/prisma-schema-reference#map) <span class="api"></span> a column name or enum value, and `@@map` an enum name.
```

nikolasburk:
Hey @emadunan, thank you so much for opening this PR. 

I actually believe the current version is correct (but could definitely be phrase more clearly), let me elaborate:

- `@map` works for:
  - column names
  - enum values
-  `@@map` works for:
   - table names
   - enum names

The section we're talking about here on the "Data model" page gives an example of mapping a **table name** using `@@map`. 

Afterwards, the note adds that you can also use `@map` for **column names** and **enum values** as well as `@@map` for **enum names**. 

I made a slight adjustments to your PR and will merge it!

Thanks for taking the time to improve our documentation 🎉

PR #3155, opened 2022-05-15
Update 04-data-model.mdx

emadunan:
## Describe this PR
Edit the paragraph to match the expample it refers to:
```prisma
model User {
  firstName String
  lastName  String
  email     String  @unique
  isAdmin   Boolean @default(false)
  @@unique([firstName, lastName])
}
```

## Changes
I have changes the number of attributes from two to three, and added the missed `@unique` attribute and its reference.

## What issue does this fix?
The documentation content


tanberry:
Thanks so much @emadunan !

PR #3164, opened 2022-05-23
Update README.md

andrew-walford-prisma:
Added an instruction to install `wget` before you install `mdtool`, because wget is a prerequisite.


tanberry:
Thank you @andrew-walford-prisma !!

PR #3166, opened 2022-05-24
Update 400-self-relations.mdx

zdila:
relations of `Follows` in `User` should be swapped because they represent "the other end" of the relation.

Otherwise this will return nothing:
```
const id = getMyUserId();

const usersFollowingMe = await prisma.user.findMany({
    where: {
      following: {
        some: {
          followingId: id,
        },
      },
    },
  });
```


tanberry:
Thanks @zdila for this PR, good catch. The relation is still a little tough to follow, and we are considering how we might reformat this sample schema to make it easier to understand, but in the meantime I'll merge your PR. I'm also going to make the same change in the tabbed content right above... thank again!

PR #3171, opened 2022-05-24
fix(using-prisma-with-cockroachdb): Add language to code blocks

janpio:
Noticed instantly _after_ you merged @keerlu - sorry for not noticing earlier.


keerlu:
Oops... thanks for spotting @janpio!

PR #3172, opened 2022-05-24
(fix) Remove copy button on cmdResult code block

### LINKED ISSUE #2945: Copy buttons on code blocks with "CLI results" are weird
image

1. Does it ever make sense to copy the "CLI results"?
2. Why is one "Copy" button dark, the other light?

@nilubava's reply:

> - All the code blocks were given the copy option a few months back. So we need to have an option for removing the copy button in such cases.
> - This looks like a bug. The hidden results are not supposed to have the copy button at all.

carlagn:
## Describe this PR

Removing the copy button from the `<cmdResult />` block.
## Changes

Added `hasCopy` mask on the `<AbsoluteCopyButton />` render call.

## What issue does this fix?

[Fixes #2945](https://github.com/prisma/docs/issues/2945)


janpio:
Via the Percy run, I think this is a bit far reaching unfortunately as it also removes it for normal commands:
image
Looking further, also schema files etc where copying definitely makes sense.

(I sent you an invite as I think you did not have an account yet there @carlagn)

content/200-concepts/100-components/03-prisma-migrate/index.mdx
```diff
@@ -34,122 +34,125 @@ If you are prototyping, consider using the [`db push` command](db-push) - see [S
...
+    model User {
+      id       Int    @id @default(autoincrement())
+      jobTitle String
+      name     String
+      posts    Post[]
+    }
+    ```
 
-1. Create the second migration:
+1.  Create the second migration:
```
nilubava: @carlagn Looks like the issue arises from the `<cmd>` tag here. It doesn't look like it has a corresponding closing one

content/200-concepts/100-components/03-prisma-migrate/index.mdx
```diff
@@ -34,122 +34,125 @@ If you are prototyping, consider using the [`db push` command](db-push) - see [S
...
-<cmdResult>
+    ```terminal copy
+    prisma migrate dev --name added_job_title
+    ```
 
-   ```sql
-     -- AlterTable
-   ALTER TABLE "User" ADD COLUMN     "jobTitle" TEXT NOT NULL;
-   ```
+          </cmd>
```
carlagn: The closing tag is here though, right? @nilubava

content/200-concepts/100-components/03-prisma-migrate/index.mdx
carlagn: Refer to: https://github.com/prisma/docs/pull/3172/files/de862a94418974ad32650d111b908ec2c48d0447#r880747691

janpio: Lists and these components never play well together, especially with the added auto formatting we have. I lost quite a bit of time on PRs in the past unfortunately trying to fight this.

Usually my conclusion was: If it worked before, I usually just found a way to keep the same formatting/indentation of the code as before, and made my changes manually in the least impactful way to keep the rendering intact.

With the whitespace cases I see here, this could be one of those cases.

(Making the components compatible with Markdown lists would be amazing of course)

nilubava:
@carlagn Since having the 'copy' option is the default behavior, we can make that the defalult for the code blocks - this will remove `copy` from all the code blocks. 
For code blocks which need no 'copy' option, we can do a `no-copy` flag. Can you explore this option ?

PR #3186, opened 2022-05-25
Revert "Add redirects from website-v8 here directly instead"

nilubava:
Reverts prisma/docs#3070


janpio:
I think this is/was not needed and can be closed now @nilubava?

PR #3187, opened 2022-05-25
Data proxy: initial changes

andrew-walford-prisma:
Initial changes to the docs for the Data Proxy work

## Describe this PR

This is a working PR where @millsp  and I can collaborate on the Data Proxy docs.


content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -85,14 +78,56 @@ datasource db {
...
+prisma generate --data proxy
 ```
 
+The Data Proxy does not need binaries. To keep Prisma Client lightweight, when you use this command the [query engine](/concepts/components/prisma-engines/query-engine) does not copy any binaries to Prisma Client.
+
+#### Step 5: Import and instantiate Prisma Client in your code
+
+If you want to use the Data Proxy with Node.js, then use `@prisma/client` as described in [Generating the client](/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client).
+
+If you want to use the Data Proxy with CloudFlare Workers or the Vercel Edge Network, you will need to import and instantiate Prisma Client in your code as follows.
```
millsp:
```suggestion
If you want to use the Data Proxy with CloudFlare Workers or the Vercel Edge Functions, you will need to import and instantiate Prisma Client in your code as follows.
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -85,14 +78,56 @@ datasource db {
...
+#### Step 4: Generate the client
+
+To generate the client with the Data Proxy, call `prisma generate` with the option `--data=proxy`:
 
 ```terminal
-PRISMA_CLIENT_ENGINE_TYPE='dataproxy' npx prisma generate
+prisma generate --data proxy
 ```
 
+The Data Proxy does not need binaries. To keep Prisma Client lightweight, when you use this command the [query engine](/concepts/components/prisma-engines/query-engine) does not copy any binaries to Prisma Client.
```
millsp:
```suggestion
The Data Proxy does not need binaries. To keep your Prisma Client lightweight, when you use this command, the [Query Engine](/concepts/components/prisma-engines/query-engine) aren't copied to the generated folder.
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -85,14 +78,56 @@ datasource db {
...
+If you want to use the Data Proxy with Node.js, then use `@prisma/client` as described in [Generating the client](/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client).
+
+If you want to use the Data Proxy with CloudFlare Workers or the Vercel Edge Network, you will need to import and instantiate Prisma Client in your code as follows.
+
+ <TabbedContent tabs={[<FileWithIcon text="TypeScript" icon="code"/>, <FileWithIcon text="JavaScript" icon="code"/>]}>
+
+ <tab>
+
+```ts
+const { PrismaClient } = require('@prisma/client/edge')
```
millsp:
```suggestion
import { PrismaClient } from '@prisma/client/edge'
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -85,14 +78,56 @@ datasource db {
...
+const prisma = new PrismaClient()
+// use `prisma` in your application to read and write data in your DB
+```
+
+  </tab>
+
+  <tab>
+
+```js
+import { PrismaClient } from '@prisma/client/edge'
```
millsp:
```suggestion
const { PrismaClient } = require('@prisma/client/edge')
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -130,7 +165,7 @@ This request must be retried
 The datasource provided is invalid:
 
 - Could not parse the URL
-- Datasource URL should use the `prisma://` protocol
+- Could not parse URL of the datasource. Datasource URL must use `prisma://` protocol when `--data-proxy` is used. No valid API key found in the datasource URL
```
millsp:
```suggestion
- Datasource URL must use `prisma://` protocol when `--data-proxy` is used
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -130,7 +165,7 @@ This request must be retried
 The datasource provided is invalid:
 
 - Could not parse the URL
```
millsp:
```suggestion
- Could not parse the URL of the datasource
```

content/400-reference/200-api-reference/200-command-reference.mdx
```diff
@@ -338,9 +338,10 @@ generator client {
 
 #### Options
 
-| Option    | Required | Description                                                                                                           | Default |
-| --------- | -------- | --------------------------------------------------------------------------------------------------------------------- | ------- |
-| `--watch` | No       | The `generate` command will continue to watch the `schema.prisma` file and re-generate Prisma Client on file changes. |
+| Option         | Required | Description                                                                                                                  | Default |
+| -------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `--data proxy` | No       | The `generate` command will generate Prisma Client for use with the [Prisma Data Proxy](/concepts/data-platform/data-proxy). |
```
millsp:
```suggestion
| `--data-proxy` | No       | The `generate` command will generate Prisma Client for use with the [Prisma Data Proxy](/concepts/data-platform/data-proxy). |
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
janpio:
```suggestion
The Data Proxy does not need local engine files. To keep Prisma Client lightweight, when you use this command the [Query Engine](/concepts/components/prisma-engines/query-engine) is not included in the generated Prisma Client.
```

andrew-walford-prisma: @millsp @janpio  To avoid passive voice, I'd like to say which module doesn't include the query engine in the generated Prisma Client. So I suppose my question is: which Prisma module or component usually includes the query engine in the Prisma Client?

janpio: That would be the Prisma Client. There are not further modules.

janpio: If you want to avoid passive voice, you can say something like "The generation process or `prisma generate` does not include the Query Engine in the generated Prisma Client".

andrew-walford-prisma: Ah OK, thanks. That's a bit awkward to state ("Prisma Client does not do X in the generated Prisma Client"), so I will say the following. What do you think?

The Data Proxy does not need local engine files. To keep Prisma Client lightweight, when you use this command the generated Prisma Client does not include the [query engine](/concepts/components/prisma-engines/query-engine).

Note that "query engine" is lower case in the query engine doc, so for now, I've gone with that to be consistent. For the future, we are working on terminology standardization in the docs, and deciding capitalization for modules and products is part of that.

andrew-walford-prisma: @janpio I just saw your suggestion - thanks, I'll go with that.

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -85,14 +78,56 @@ datasource db {
...
+#### Step 4: Generate the client
+
+To generate the client with the Data Proxy, call `prisma generate` with the option `--data=proxy`:
 
 ```terminal
-PRISMA_CLIENT_ENGINE_TYPE='dataproxy' npx prisma generate
+prisma generate --data proxy
 ```
 
+The Data Proxy does not need local engine files. To keep Prisma Client lightweight, when you use this command the generation process does not include the [query engine](/concepts/components/prisma-engines/query-engine) in the generated Prisma Client.
```
tanberry:
```suggestion
The Data Proxy does not need local engine files. To keep Prisma Client lightweight, use the above command because the generation process does not include the [query engine](/concepts/components/prisma-engines/query-engine) in the generated Prisma Client.
```

andrew-walford-prisma: You don't run the command to keep the client lightweight. You run the command to use the Data Proxy. The lightweight-ness is a positive side effect of using the Data Proxy. However, I can see the confusion, so I'll clarify this paragraph.

andrew-walford-prisma: Thanks. I've reworked the paragraph to remove this confusion.

janpio (commented):
I treated this review as a full review of the "Data Proxy" page (`content/200-concepts/150-data-platform/08-data-proxy.mdx`), not just the changes here, under the lens that Data Proxy will be GA and should be "complete" for our users:

1. The "Not Supported" admonition for Interactive Transactions should optimally link to a section that explains what this means (you can not generate a Data Proxy with `generate --data-proxy` while the `interactiveTransactions` feature is enabled, so you will get an error message and have to remove the preview feature instead (which will make the method uses type errors in TS).
1. "We assume ..." in line 35 could probably be expaned into a new "Step 1" that explains you should already have an account and a Prisma project. Right now this is a bit off.
1. Line 39 onwards taked about the Data Proxy location. How should I make that decision? Should it be close to my database server or to my application server? Or something completely different? Might even jsutify a "How to choose a Data Proxy location" section further down that we link to here.
1. Line 55 should be updated to mention version `3.15.0`.
1. Admonition in lines 64 to 68 about `only supported datasource` should be removed as this will not be the case any more on release. ~~If this is not merged yet, we should of course wait until it actually is the case and prepare a PR for when that is the case.~~
1. Current "Step 3" could use an introduction what we are doing. If I understand correctly it explains how to store the connection string locally in your project. Generally this step is pretty confusing for now and I am not sure what would be there for Vercel and Cloudflare. (Vercel should potentially not be there at all, and certainly come _after_ Cloudflare)
1. Step 4 could mention how the `generate` output includes "Data Proxy" somewhere to indicate that it indeed did create a Data Proxy Client, and not a normal one.
1. Step 5 should probably have an alternative tab or section for how to use the new Edge Client if you want to deploy to Worker/Edge environments instead of only having a code example for _that_ - that was initially super confusing that there is no example for the Node.js usage of Data Proxy.
1. Step 6 is a bit weird, as the connection string does not magically change - but you changed it yourself above. The `For guidance on how ...` section after the code block on the other hand feels like it is similar to what is also defined in Step 3. Somehow there seems to be a mix of local configuration and deployment configuration here in Step 6 and 3 that I am not super clear about. Could/should be clarified.
1. Is the admonition for the error codes still valid that this "may be changed at any time without notice"? Should not really be the case for a GA product.
1. The command suggested for `migrate` usage will not work on Windows: `    "migrate": "DATABASE_URL=\"$MIGRATE_DATABASE_URL\" prisma migrate deploy",`. We should probably make sure we have a better story here. (We probably have similar cases in other places in the docs)

Things that should be done in another PR:

1. The "Early Access Feature" Admonition (lines 19 to 24) will need t obe removed when PDP goes GA on June 15th.

content/300-guides/200-deployment/110-deployment-guides/450-deploying-to-cloudflare-workers.mdx
```diff
@@ -150,7 +150,7 @@ Now you're ready to add Prisma to the project.
 npx prisma init
 ```
 
-This will create a Prisma Schema in `prisma/schema.prisma` and an `.env` file.
+This will create a Prisma Schema in `prisma/schema.prisma`.
```
janpio: This does actually create an `.env` file. Are you removing the mention because it is not always useful for Data Proxy Client @millsp?

millsp: Yes, it does create one, removing as it is not useful here. Should we add an admonition here too?

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -26,15 +26,15 @@ Please note that the Prisma Data Proxy is an early-stage product and should not
 <Admonition type="warning">
 
 **Not supported** <br />
-[Interactive Transactions](/concepts/components/prisma-client/transactions#interactive-transactions-in-preview) (Preview Feature) are not supported at this time.
+You cannot use the Prisma Data Proxy at the same time as you use [Interactive Transactions](/concepts/components/prisma-client/transactions#interactive-transactions-in-preview) or Metrics preview features. To use Prisma Data Proxy, you must remove `interactiveTransactions` and `metrics` from the `previewFeatures` section of your schema.
```
janpio, line 29: You can always have placeholders with internal anchor links like this one:
```suggestion
You cannot use the Prisma Data Proxy at the same time as you use [Interactive Transactions](/concepts/components/prisma-client/transactions#interactive-transactions-in-preview) or [Metrics](#TODO) preview features. To use Prisma Data Proxy, you must remove `interactiveTransactions` and `metrics` from the `previewFeatures` section of your schema.
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -26,15 +26,15 @@ Please note that the Prisma Data Proxy is an early-stage product and should not
 <Admonition type="warning">
 
 **Not supported** <br />
```
janpio, line 28: I think we could turn this super prominent warning admonition into a "Limitations" headline further down. That would make it much less scary, but still correct. I think our error messages are good enough (from tomorrow) to not having to have this up front and top here.

Agree @albertoperdomo?

content/300-guides/200-deployment/110-deployment-guides/450-deploying-to-cloudflare-workers.mdx
andrew-walford-prisma: @millsp  To make sure I understand this: it does create an .env file, but that file isn't relevant to the Data Proxy. Is that correct? My instinct would be not to mention it if the file isn't relevant and doesn't impact anything else.

content/200-concepts/150-data-platform/08-data-proxy.mdx
andrew-walford-prisma: The docs team are revising our procedures, and we hopefully will soon have a way to avoid broken links to existent pages elsewhere that are being added as part of the same release. For this release, I think I'll add an actual link to the metrics page. We'll be doing the merges very close together, so the link will be broken for a very short time.

content/300-guides/200-deployment/110-deployment-guides/450-deploying-to-cloudflare-workers.mdx
andrew-walford-prisma: Having liaised with Pierre, I'm going to add a note saying that the file does get created, but that you cannot use it in conjunction with Cloudflare Workers.

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -48,125 +48,162 @@ Connection strings always have the following format:
...
+
+In Vercel Edge Functions, you can define environment variables as you would in `Next.js`. For more information, see [Environment Variables](https://nextjs.org/docs/basic-features/environment-variables) in the Vercel Edge documentation.
+
+### Step 4: Generate the client
+
+To generate Prisma Client with the Data Proxy, call `prisma generate` with the option `--data-proxy`:
 
 ```terminal
-PRISMA_CLIENT_ENGINE_TYPE='dataproxy' npx prisma generate
+prisma generate --data-proxy
```
ruheni:
```suggestion
npx prisma generate --data-proxy
```

content/300-guides/200-deployment/110-deployment-guides/450-deploying-to-cloudflare-workers.mdx
andrew-walford-prisma: And while I'm at it, I'll improve the wording slightly.

content/200-concepts/150-data-platform/08-data-proxy.mdx
```diff
@@ -48,125 +48,162 @@ Connection strings always have the following format:
...
+You can use `@prisma/client/edge` only with the Data Proxy. Make sure to use the `--data-proxy` flag when you [generate the client](#step-4-generate-the-client).
 
 </Admonition>
 
-#### P5000
+### Step 6: Start your app
+
+You are now ready to try the Data Proxy. Go ahead and start your app!
+
+## Error codes
```
ruheni: I think the **Error codes** should come after **Important considerations about the Data Proxy**.

Or better yet, the Error codes could live on a separate page at some point – similar to https://www.prisma.io/docs/reference/api-reference/error-reference#error-codes

content/300-guides/200-deployment/110-deployment-guides/450-deploying-to-cloudflare-workers.mdx
```diff
@@ -248,26 +248,20 @@ You're all setup and ready to generate a Prisma Client!
 
 ## 9. Generate a Prisma Client
 
-Next you'll generate a Prisma Client that connects through the Data Proxy over HTTP.
+Next you'll generate a Prisma Client that connects through the [Data Proxy](/concepts/data-platform/data-proxy) over HTTP.
 
 ```terminal
-PRISMA_CLIENT_ENGINE_TYPE=dataproxy npx prisma generate
+prisma generate --data-proxy
```
ruheni:
```suggestion
npx prisma generate --data-proxy
```

content/200-concepts/150-data-platform/08-data-proxy.mdx
andrew-walford-prisma: Fair point - I'll move the error codes section downwards.

andrew-walford-prisma: We're going to do a docs re-org in the near future, and having a separate page is something we'll consider.

content/200-concepts/150-data-platform/08-data-proxy.mdx
andrew-walford-prisma: Agree, I'll change this later in the week.

PR #3201, opened 2022-05-30
Metrics docs

andrew-walford-prisma:
## Draft of new Metrics docs

Docs for the new Metrics feature in Prisma Client. This includes a new main doc about Metrics, and an update to the API reference doc.


content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
+---
+title: 'Metrics'
+metaTitle: 'Metrics'
+metaDescription: 'Diagnose application performance with insights into Prisma Client database activity'
+tocDepth: 3
+---
+
+<TopBlock>
+
+Prisma metrics give you a detailed insight into how Prisma Client interacts with your database. You can use this insight to help diagnose performance issues with your application. If you send the data to an external metrics system, then you can view the data over time. For example, you might visualize how your application's number of idle and active connections change.
```
garrensmith: > visualize how your application's number of idle and active connections change.

It might be better to say how they change over time

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
...
+
+</TopBlock>
+
+## About metrics
+
+Metrics provide information on the following aspects of Prisma Client:
```
garrensmith: I think you could say, The Prisma Client provides the following metrics:

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
...
+
+- The number of Prisma operations
+
+  A Prisma operation is a statement like `let user = prisma.user(include: {posts: true}})`, which Prisma Client translates into one or more SQL queries.
```
garrensmith: Nice. Thank you

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
...
+
+- Query response times
+- Active transactions
+- Connection pool information
+
+You can export metrics in JSON or Prometheus formats and view them in a console log, or integrate them into an external metrics system, such as StatsD or Prometheus. If you integrate them into an external metrics system, then you can view the metrics data over time, for example to help diagnose changes in your application's performance.
```
garrensmith: > for example to help diagnose changes in your application's performance.

This feels a bit repetitive. Anyone familiar with Metrics should know this and we explained it above

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
...
+
+## About metrics
+
+Metrics provide information on the following aspects of Prisma Client:
+
+- The number of Prisma operations
+
+  A Prisma operation is a statement like `let user = prisma.user(include: {posts: true}})`, which Prisma Client translates into one or more SQL queries.
+
+- Query response times
```
garrensmith: I'm not quite following what this section is about? You only mention some of the metrics/aspects. What is your aim here?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
...
+  - `pool_active_connections`: The number of active connections in use.
+  - `pool_idle_connections`: The number of connections that are not being used.
+  - `pool_wait_count`: The number of workers waiting for a connection.
+  - `query_active_transactions`: The number of active transactions.
+- Histograms (distributions of values categorized into "buckets"):
+  - `pool_wait_duration_ms`: The wait time for a worker to get a connection.
+  - `query_total_elapsed_time_ms`: The execution time for all database queries executed.
+  - `query_operation_total_elapsed_time_ms`: The execution time for all operations.
+
+You can group your metrics data with [global labels](#global-labels).
```
garrensmith: This isn't quite true. The metrics will be grouped in the metrics collection system by label. We won't group them ourselves but we will add the global labels to all the metrics

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,335 @@
...
+
+## Get metrics in JSON format
+
+When you get metrics in JSON format, you can use them as returned, or [send them to StatSD](#use-prisma-metrics-with-statsd).
```
garrensmith: One thing I've forgotten to mention is that because metrics is a preview feature. They will need to add `"metrics"` to the list of `previewFeatures` in their schema before it will work

garrensmith (changes requested):
This is a really good start. I've left a bit of feedback but I really like what you have done. Thank you

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: Fair point - to reduce the repetition, I've moved some of the top-of-page information into this section, and also reworked this material.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: So is it more accurate to say this? "You can label your metrics data to help you analyze them by category, for example to view separate sets of metrics for each database server that you operate."

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: OK, I've removed the section, because we cover the metrics in detail slightly below it anyway.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: Ah OK - I'll add some material about this, and update the preview feature doc.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,411 @@
...
+app.get("/metrics", async (\_req, res: Response) => {
+let metrics = await prisma.$metrics.prometheus();
+res.end(metrics);
+});
+
+app.listen(port, () => {
+console.log(`Example app listening on port ${port}`);
+})
+
+Combining it with another Prometheus client library:
```
garrensmith: Can you just check how this will render? I think you need to end the code block on 308 and start a new code block on 310

andrew-walford-prisma: Good catch @garrensmith . I've fixed this, and tweaked the wording of the associated non-code text.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,414 @@
...
+      return [bucket, diff]
+    })
+    metric.value.buckets = diffBuckets
+    return metric
+  })
+}
+
+let previousHistograms: any = null
+let statsdSender = async () => {
+  let res = await prisma.$metrics.json()
```
garrensmith: Sorry final change. Can you change these lines:
```
let res = await prisma.$metrics.json()
let metrics = res.data
```

into
```
let metrics = await prisma.$metrics.json()
```

garrensmith (approved):
These look really good. Thanks for this Andrew. I have one small code change based on a mistake in the example I gave you. Otherwise looks good to me.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: Done!

content/200-concepts/100-components/250-preview-features/050-client-preview-features.mdx
```diff
@@ -14,6 +14,7 @@ The following [Preview](/about/prisma/releases#preview) feature flags are availa
 - [`dataProxy`](/concepts/data-platform/data-proxy) since release [3.3.0](https://github.com/prisma/prisma/releases/tag/3.3.0) - [Submit feedback](https://github.com/prisma/prisma/issues/9841)
 - [`extendedIndexes`](/concepts/components/prisma-schema/indexes#index-configuration) since release [3.5.0](https://github.com/prisma/prisma/releases/tag/3.5.0) - [Submit feedback](https://github.com/prisma/prisma/issues/10540)
 - [`fullTextIndex`](/concepts/components/prisma-schema/indexes#full-text-indexes-mysql-and-mongodb) since release [3.6.0](https://github.com/prisma/prisma/releases/tag/3.6.0) - [Submit feedback](https://github.com/prisma/prisma/issues/10539)
+- [`metrics`](/guides/performance-and-optimization/metrics) since release [3.15.0](https://github.com/prisma/prisma/releases/tag/3.15.0) - [Submit feedback](https://github.com/prisma/prisma/issues/13579)
```
tanberry: @andrew-walford-prisma let's move this to be below the feature introduced in 3.14, just to keep this list chronological.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+<TopBlock>
+
+Prisma metrics give you a detailed insight into how Prisma Client interacts with your database. You can use this insight to help diagnose performance issues with your application.
+
+</TopBlock>
+
+## About metrics
+
+You can export metrics in JSON or Prometheus formats and view them in a console log, or integrate them into an external metrics system, such as StatsD or Prometheus. If you integrate them into an external metrics system, then you can view the metrics data over time. For example, you can use metrics to help diagnose how your application's number of idle and active connections changes.
```
tanberry: As someone new to metrics in Prisma, my first question is "What all can I measure and pull metrics on?" Consider moving the last sentence, that provides an example, up to the front of this paragraph. and maybe add some more examples, or list the types of metrics?

tanberry: Oh duh, that is the very next section! :-)

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+  - `query_total_operations`: The total number of Prisma operations executed.
+    - A Prisma operation is a statement like `let user = prisma.user(include: {posts: true}})`, which Prisma Client translates into one or more SQL queries.
+    - The value returned by `query_total_operations` can be less than `query_total_queries`, because some operations create multiple queries.
+
+- Gauges (can increase or decrease):
+  - `pool_active_connections`: The number of active connections in use.
+  - `pool_idle_connections`: The number of connections that are not being used.
+  - `pool_wait_count`: The number of workers waiting for a connection.
+  - `query_active_transactions`: The number of active transactions.
+- Histograms (distributions of values categorized into "buckets"):
```
tanberry: Need a blank space between Gauges and Histograms, and consistent space (or not) between header and bullet items.

tanberry: Not sure what this phrase means? "distributions of values categorized into "buckets"... are histograms about graphical/visual displays of data?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+  - `pool_wait_duration_ms`: The wait time for a worker to get a connection.
+  - `query_total_elapsed_time_ms`: The execution time for all database queries executed.
+  - `query_operation_total_elapsed_time_ms`: The execution time for all operations.
+
+You can [add global labels to your metrics data](#global-labels) to help you group and separate your metrics, for example by infrastructure region or server.
```
tanberry: I am not sure that just using the anchor tag will resolve.. might have to use `/[content/guides/performance-and-optimization/metrics#global-labels`).

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+## Prerequisites
+
+To use metrics, you must do the following:
```
tanberry:
```suggestion
To use Prisma metrics, you must do the following:
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+1. Install the appropriate dependencies.
```
tanberry: For jumpy, impetuous, demanding readers like me, I'd suggest you link this to the next section, that explains what those dependencies are.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+Use version `3.15.0` or higher of the `prisma` and `@prisma/client` npm packages.
+
+```terminal
+npm install prisma@latest --save-dev
+npm install @prisma/client@latest --save
+```
+
+### Step 2: Enable the feature flag in the Prisma schema file
+
+In the `generator` block of your `schema.prisma` file, enable the feature flag:
```
tanberry:
```suggestion
In the `generator` block of your `schema.prisma` file, enable the `previewFeatures` flag:
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+```prisma
+generator client {
+  provider        = "prisma-client-js"
+  previewFeatures = ["metrics"]
+}
+```
+
+## Get metrics in JSON format
+
+When you get metrics in JSON format, you can use them as returned, or [send them to StatSD](#use-prisma-metrics-with-statsd).
```
tanberry:
```suggestion
When you get metrics in JSON format, you can use them immediately, just as they are returned, or [send them to StatSD](#use-prisma-metrics-with-statsd) in order to visualize your metrics data.
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+To get metrics in JSON format, add the following lines to your application code:
```
tanberry: Can we replace "get" with "retrieve" here? 
Also, is there typically a specific file that devs are building their app code in, and if so, can we provide its name?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+      },
+      "description": "The wait time for a worker to get a connection."
+    }
+  ]
+}
+```
+
+### Use Prisma metrics with StatsD
+
+You can send JSON-formatted metrics to StatsD to visualize your metrics data over time.
```
tanberry:
```suggestion
You can send JSON-formatted metrics to [StatsD](https://github.com/statsd/statsd) to visualize your metrics data over time.
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+<Admonition type="info">
+  Note: When you send data to StatsD, you must send Prisma counter and gauge
```
tanberry:
```suggestion
  Note: When you send data to StatsD, you must send both Prisma counter and gauge
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+  metrics as gauge data. This is because StatsD expects counter metrics to be a
+  series of incremented or decremented values, but Prisma counter metrics return
+  actual values.
+</Admonition>
+
+In the following example, we send metrics to StatsD every 10 seconds. This is aligned with the default 10s flush rate of StatsD.
```
tanberry:
```suggestion
In the following example, we send metrics to StatsD every 10 seconds. This timing aligns with the default 10s flush rate of StatsD.
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+  previousHistograms = diffHistograms
+}
+
+setInterval(async () => await statsdSender(), 10000)
+```
+
+## Get metrics in Prometheus format
+
+When you get Prisma metrics in Prometheus format, you can use them as returned, or [send them to the Prometheus metrics system](#use-prisma-metrics-with-the-prometheus-metrics-system).
```
tanberry:
```suggestion
When you retrieve Prisma metrics in Prometheus metrics format, you can use them exactly as they are returned, or [send them to the Prometheus metrics system](#use-prisma-metrics-with-the-prometheus-metrics-system).
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+
+To get Metrics in a Prometheus format, add the following lines to your application code:
```
tanberry:
```suggestion
To retrieve Metrics in a Prometheus metrics format, add the following lines to your application code:
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+let metrics = await prisma.$metrics.prometheus();
+res.end(metrics);
+});
+
+app.listen(port, () => {
+console.log(`Example app listening on port ${port}`);
+})
+```
+
+The following example shows how to combine Prisma metrics with another Prometheus client library:
```
tanberry: Should we specify _which_ Prometheus client library...? Or do we make them read the code sample below to see?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,413 @@
...
+});
+```
+
+## Global labels
+
+You can add global labels to your metrics to help you group and separate your metrics. Each instance of Prisma Client will add these labels to all the metrics that it generates. For example, you can group your metrics by infrastructure region, or by server, with a label like `{server: "us-server1", "app_version": "1"}`.
+
+Global labels work with JSON and Prometheus-formatted metrics.
+
+For example:
```
tanberry: Where are they adding this code snippet? Maybe say something like, "For example, to add global labels to the metrics generated from a specific application that is running, add the following code to your application:"

The above sentence might be trash-garbage and completely wrong, but you get the idea... we need a bit more explanation of _why_ and _where_ for the code snippet that follows. :-)

tanberry (changes requested):
Great stuff, I learned a ton about Prisma metrics. Let me know if you have any questions about my edits/suggestions/questions.

content/200-concepts/100-components/250-preview-features/050-client-preview-features.mdx
andrew-walford-prisma: Good idea!

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: It doesn't matter which library. This shows how to combine Prisma metrics with more one or more other libraries. I'll reword to make that clearer.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: Good idea. I've added links to StatsD and Prometheus at their first mentions on the page.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: 1. Sure!
2. No, this could go into any source code file where the user needs it.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: They're not about graphing per se, any more than the other categories. But the data are split into a collection of values, each item within the collection known as a "bucket". I've clarified the sentence to emphasize that we're talking about data, not graphs, and that the data is split into the distributions.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: TIL that it does work! Thanks @Andrew.

andrew-walford-prisma: I've checked, and I can confirm that anchor tags do work :-)

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: @andrew-walford-prisma I'm curious if you plan to accept this suggestion or..? Are you reluctant to link to a third-party site, or what reason do you see for not linking to the StatsD repo?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: Ditto, curious if you think the addition of "both" doesn't clarify...?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: The reason for using "this timing" is to avoid using only the word "this", which can be ambiguous. Let me know you thoughts on this, please.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+      "description": "The number of active transactions"
+    }
+  ],
+  "histograms": [
+    {
+      "key": "pool_wait_duration_ms",
+      "labels": {},
+      "value": {
+        "buckets": [
+          [0, 0],
```
SevInf: Should we document what those number mean? First one is the upper bound for the bucket, second one is number of values that fall into that bucket.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+Prisma metrics give you a detailed insight into how Prisma Client interacts with your database. You can use this insight to help diagnose performance issues with your application.
+
+</TopBlock>
+
+## About metrics
+
+You can export metrics in JSON or Prometheus formats and view them in a console log, or integrate them into an external metrics system, such as [StatsD](https://github.com/statsd/statsd) or [Prometheus](https://prometheus.io/). If you integrate them into an external metrics system, then you can view the metrics data over time. For example, you can use metrics to help diagnose how your application's number of idle and active connections changes.
+
+We provide the following metrics:
```
janpio:
```suggestion
Prisma Client provides the following metrics:
```
see https://www.prisma.io/docs/about/prisma-docs/style-guide#when-to-use-we

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+      "description": "The number of workers waiting for a connection"
+    },
+    {
+      "key": "query_active_transactions",
+      "labels": {},
+      "value": 0,
+      "description": "The number of active transactions"
+    }
+  ],
+  "histograms": [
```
janpio: Is it intentional that the list of returned histograms does not match what is documented above?
If we made this shorter intentionally, we maybe should indicate with `[...]` or something like that.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+### Use Prisma metrics with StatsD
+
+You can send JSON-formatted metrics to StatsD to visualize your metrics data over time.
+
+<Admonition type="info">
+  Note: When you send data to StatsD, you must send both Prisma counter and
+  gauge metrics as gauge data. This is because StatsD expects counter metrics to
+  be a series of incremented or decremented values, but Prisma counter metrics
+  return actual values.
```
janpio: I do not understand this explanation. Can this be simplified or have examples?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+</Admonition>
+
+In the following example, we send metrics to StatsD every 10 seconds. This timing aligns with the default 10s flush rate of StatsD.
```
janpio: This code would be part of my application and just run endlessly to get the metrics every 10 seconds and transfer them to statsd?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+This returns metrics as follows:
+
+```c
+# HELP query_total_operations The total number of operations executed
+
+# TYPE query_total_operations counter
+
+query_total_operations 0
+
+# HELP query_total_queries The total number of queries executed
```
janpio: Does this really have all these empty lines in the output?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+# HELP query_active_transactions The number of active transactions
+
+# TYPE query_active_transactions gauge
+
+query_active_transactions
+
+# HELP pool_wait_duration_ms The wait time for a worker to get a connection.
+
+# TYPE pool_wait_duration_ms histogram
```
janpio: Same question for histograms here as for the JSON output: https://github.com/prisma/docs/pull/3201/files#r888818145

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+const prisma = new PrismaClient();
+
+app.get("/metrics", async (\_req, res: Response) => {
+let metrics = await prisma.$metrics.prometheus();
+res.end(metrics);
+});
+
+app.listen(port, () => {
+console.log(`Example app listening on port ${port}`);
+})
```
janpio:
```suggestion
app.get("/metrics", async (\_req, res: Response) => {
  let metrics = await prisma.$metrics.prometheus();
  res.end(metrics);
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`);
})
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+```js
+import { PrismaClient } from "@prisma/client";
+import express, { Request, Response } from "express";
+
+const app = express();
+const port = 4000;
+const prisma = new PrismaClient();
+
+app.get("/metrics", async (\_req, res: Response) => {
```
janpio: Is this backslash correct before `_req`?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+let prismaMetrics = await prisma.$metrics.prometheus();
+let appMetrics = await registry.metrics();
+res.end(prismaMetrics + appMetrics);
+});
+
+app.listen(port, () => {
+console.log(`Example app listening on port ${port}`);
+});
```
janpio:
```suggestion
app.get("/metrics", async (\_req, res: Response) => {
let prismaMetrics = await prisma.$metrics.prometheus();
  let appMetrics = await registry.metrics();
  res.end(prismaMetrics + appMetrics);
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`);
});
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+let metrics = await prisma.$metrics.prometheus();
+res.end(metrics);
+});
+
+app.listen(port, () => {
+console.log(`Example app listening on port ${port}`);
+})
+```
+
+The following example shows how to combine Prisma metrics with other Prometheus client libraries:
```
janpio:
```suggestion
The following example shows how to combine Prisma metrics with other Prometheus client libraries, also served via an `Express.js` endpoint:
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+});
+
+app.listen(port, () => {
+console.log(`Example app listening on port ${port}`);
+});
+```
+
+## Global labels
+
+You can add global labels to your metrics to help you group and separate your metrics. Each instance of Prisma Client will add these labels to all the metrics that it generates. For example, you can group your metrics by infrastructure region, or by server, with a label like `{server: "us-server1", "app_version": "1"}`.
```
janpio: Formatting of the label JSON is a bit weird. Is that necessary, or could it also be this:
```suggestion
You can add global labels to your metrics to help you group and separate your metrics. Each instance of Prisma Client will add these labels to all the metrics that it generates. For example, you can group your metrics by infrastructure region, or by server, with a label like `{ server: "us-server1", "app_version": "1" }`.
```
?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+Global labels work with JSON and Prometheus-formatted metrics.
+
+For example, to add global labels to JSON-format metrics, add the following code to your application:
+
+```ts
+let metrics = prisma.$metrics.json({ global: 'label', app: 'one' })
```
janpio: In this and the example output below, the global label value is just `label`, right? Might be worth calling out, as above the label value was a complicated JSON string - and now just a super simple and short one. I did not make that connection initially.

janpio: Ahhh, I think actually the full `{ global: 'label', app: 'one' }` is the global label? So there is only one param to the function `json` and it is the global label?

If so, the label should probaly not contain `global` in its value as this is highly confusing with the "global label" concept.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+console.log(metrics)
+```
+
+This returns information in the following format:
+
+```json
```
janpio: Might be worth highlighting the added lines in the diff vs. not supplying a global label - so all the lines `"labels": { "global": "label", "app": "one" },`.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
SevInf: This is incorrect, here is the correct snippet

```suggestion
let metrics = prisma.$metrics.json({ globalLabels: { global: 'label', app: 'one' } })
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -5296,6 +5296,13 @@ See: [Raw database access (`$runCommandRaw()`)](/concepts/components/prisma-clie
 
 See: [Transactions](/concepts/components/prisma-client/transactions).
 
+<!-- prettier-ignore -->
+### <inlinecode>$metrics()</inlinecode>
```
SevInf: This is a readonly property, not a function

```suggestion
### <inlinecode>$metrics</inlinecode>
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: Oh, nice. I agree this would be good to add. I can draft a suggestion, then Andrew can review and tweak/merge on Monday. I'll ping you @SevInf with some specific follow-up questions. ;-)

content/300-guides/100-performance-and-optimization/090-metrics.mdx
janpio: Aha, that would already have helped me in my understanding indeed: https://github.com/prisma/docs/pull/3201#discussion_r888884700

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
janpio: We should probably then add sub sections for the actual functions? The reference should be complete.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: @janpio , in your phrase does "also" mean that these "other Prometheus libraries" are served via an Express.js endpoint? (I just want to make it very clear what the "also" refers to. :-)

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: Attempted edit, @SevInf please review for accuracy:

  Note: When you send data to StatsD, you must send _both_ Prisma counter and
  gauge metrics as gauge data. This is because StatsD expects counter metrics to
  be a series of incremented or decremented values, but Prisma returns counter metrics
  as actual values. Therefore, sending all counter metrics as gauge metrics allows StatsD to process Prisma counter metrics.
  
  Ugh, not sure if that helps...

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: @janpio you mean we should highlight the lines for each metric, right? I think this will probably have to wait until after the first release; we are having some challenges with highlighting in Docs.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+pool_wait_duration_ms_bucket{le="10"} 1
+pool_wait_duration_ms_bucket{le="50"} 1
+pool_wait_duration_ms_bucket{le="100"} 1
+pool_wait_duration_ms_bucket{le="500"} 1
+pool_wait_duration_ms_bucket{le="1000"} 1
+pool_wait_duration_ms_bucket{le="5000"} 1
+pool_wait_duration_ms_bucket{le="50000"} 1
+pool_wait_duration_ms_bucket{le="+Inf"} 1
+pool_wait_duration_ms_sum 0.050875000000000004
+pool_wait_duration_ms_count 1
```
SevInf: Here is an actual sample which takes care of whitespace issues and adds missing histograms

```suggestion
# HELP query_total_operations
# TYPE query_total_operations counter
query_total_operations 2

# HELP query_total_queries
# TYPE query_total_queries counter
query_total_queries 5

# HELP pool_active_connections The number of active connections in use.
# TYPE pool_active_connections gauge
pool_active_connections 0

# HELP pool_idle_connections The number of connections that are not being used
# TYPE pool_idle_connections gauge
pool_idle_connections 21

# HELP pool_wait_count The number of workers waiting for a connection.
# TYPE pool_wait_count gauge
pool_wait_count 0

# HELP query_active_transactions
# TYPE query_active_transactions gauge
query_active_transactions 0

# HELP pool_wait_duration_ms The wait time for a worker to get a connection.
# TYPE pool_wait_duration_ms histogram
pool_wait_duration_ms_bucket{le="0"} 0
pool_wait_duration_ms_bucket{le="1"} 3
pool_wait_duration_ms_bucket{le="5"} 3
pool_wait_duration_ms_bucket{le="10"} 3
pool_wait_duration_ms_bucket{le="50"} 3
pool_wait_duration_ms_bucket{le="100"} 3
pool_wait_duration_ms_bucket{le="500"} 3
pool_wait_duration_ms_bucket{le="1000"} 3
pool_wait_duration_ms_bucket{le="5000"} 3
pool_wait_duration_ms_bucket{le="50000"} 3
pool_wait_duration_ms_bucket{le="+Inf"} 3
pool_wait_duration_ms_sum 0.023208
pool_wait_duration_ms_count 3

# HELP query_operation_total_elapsed_time_ms
# TYPE query_operation_total_elapsed_time_ms histogram
query_operation_total_elapsed_time_ms_bucket{le="0"} 0
query_operation_total_elapsed_time_ms_bucket{le="1"} 1
query_operation_total_elapsed_time_ms_bucket{le="5"} 2
query_operation_total_elapsed_time_ms_bucket{le="10"} 2
query_operation_total_elapsed_time_ms_bucket{le="50"} 2
query_operation_total_elapsed_time_ms_bucket{le="100"} 2
query_operation_total_elapsed_time_ms_bucket{le="500"} 2
query_operation_total_elapsed_time_ms_bucket{le="1000"} 2
query_operation_total_elapsed_time_ms_bucket{le="5000"} 2
query_operation_total_elapsed_time_ms_bucket{le="50000"} 2
query_operation_total_elapsed_time_ms_bucket{le="+Inf"} 2
query_operation_total_elapsed_time_ms_sum 3.197624
query_operation_total_elapsed_time_ms_count 2

# HELP query_total_elapsed_time_ms
# TYPE query_total_elapsed_time_ms histogram
query_total_elapsed_time_ms_bucket{le="0"} 0
query_total_elapsed_time_ms_bucket{le="1"} 5
query_total_elapsed_time_ms_bucket{le="5"} 5
query_total_elapsed_time_ms_bucket{le="10"} 5
query_total_elapsed_time_ms_bucket{le="50"} 5
query_total_elapsed_time_ms_bucket{le="100"} 5
query_total_elapsed_time_ms_bucket{le="500"} 5
query_total_elapsed_time_ms_bucket{le="1000"} 5
query_total_elapsed_time_ms_bucket{le="5000"} 5
query_total_elapsed_time_ms_bucket{le="50000"} 5
query_total_elapsed_time_ms_bucket{le="+Inf"} 5
query_total_elapsed_time_ms_sum 1.8407059999999997
query_total_elapsed_time_ms_count 5
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,416 @@
...
+          [500, 0],
+          [1000, 0],
+          [5000, 0],
+          [50000, 0]
+        ],
+        "sum": 0,
+        "count": 0
+      },
+      "description": "The wait time for a worker to get a connection."
+    }
```
SevInf: Missing histograms:

```suggestion
    },
    {
      "key": "query_operation_total_elapsed_time_ms",
      "labels": {},
      "value": {
        "buckets": [
          [0,0],
          [1,1],
          [5,1],
          [10,0],
          [50,0],
          [100,0],
          [500,0],
          [1000,0],
          [5000,0],
          [50000,0]
        ],
        "sum": 1.367708,
        "count": 2
      },
      "description": ""
    },
    {
      "key": "query_total_elapsed_time_ms",
      "labels": {},
      "value": {
        "buckets": [
          [0,0],
          [1,5],
          [5,0],
          [10,0],
          [50,0],
          [100,0],
          [500,0],
          [1000,0],
          [5000,0],
          [50000,0]
        ],
        "sum": 1.113832,
        "count": 5
      },
      "description": ""
    }

```

I've noticed that they also miss description, not sure if that's intentional @garrensmith

content/300-guides/100-performance-and-optimization/090-metrics.mdx
SevInf: Yes. Not quite the usual way to work StatsD to be honest, but prometheus and statsd have very different quite incompatible ways of operations, so I guess one of them is bound to look suboptimal.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: @SevInf fixed below

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: @SevInf fixed

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,477 @@
...
+
+```js
+import { PrismaClient } from "@prisma/client";
+import express, { Request, Response } from "express";
+
+const app = express();
+const port = 4000;
+const prisma = new PrismaClient();
+
+app.get("/metrics", async (\_req, res: Response) => {
```
tanberry:
```suggestion
app.get("/metrics", async (_req, res: Response) => {
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,477 @@
...
+import { PrismaClient } from "@prisma/client";
+import express, { Request, Response } from "express";
+import prom from "prom-client";
+
+let registry = new prom.Registry();
+const app = express();
+const port = 4000;
+const prisma = new PrismaClient();
+
+app.get("/metrics", async (\_req, res: Response) => {
```
tanberry:
```suggestion
app.get("/metrics", async (_req, res: Response) => {
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: removing

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
tanberry: @SevInf suggested these exact methods:
$metrics.json()
$metrics.prometheus()

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,477 @@
...
+```prisma
+generator client {
+  provider        = "prisma-client-js"
+  previewFeatures = ["metrics"]
+}
+```
+
+## Retrieve metrics in JSON format
+
+When you retrieve metrics in JSON format, you can use them in the format they are returned, or [send them to StatSD](#use-prisma-metrics-with-statsd) to visualize how they change over time.
```
tanberry:
```suggestion
When you retrieve metrics in JSON format, you can use them in the format they are returned, or [send them to StatsD](#use-prisma-metrics-with-statsd) to visualize how they change over time.
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
tanberry: Serghii fixed

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,477 @@
...
+    - The value returned by `query_total_operations` can be less than `query_total_queries`, because some operations create multiple queries.
+
+- Gauges (can increase or decrease):
+
+  - `pool_active_connections`: The number of active connections in use.
+  - `pool_idle_connections`: The number of connections that are not being used.
+  - `pool_wait_count`: The number of workers waiting for a connection.
+  - `query_active_transactions`: The number of active transactions.
+
+- Histograms (metrics data divided into a collection of values; we call each container in the collection a "bucket"):
```
tanberry:
```suggestion
- Histograms (metrics data divided into a collection of values; we call each container in the collection a "bucket"). A bucket of metric data is returned in the format of `[_x_, _x_]`. The first number is the upper bound for the bucket, second number is number of instances that fall into that bucket. For example, the bucket `[5,1]` indicates that there was 1 instance of a query returning a value between 0 and 5.
```

tanberry: Ooooffff, @andrew-walford-prisma you might want to try your hand at rewording this, and lean on @SevInf as needed. I am not confident that my wording explains the parts of the bucket as well as it could. :-(

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,477 @@
...
+app.get("/metrics", async (\_req, res: Response) => {
+let prismaMetrics = await prisma.$metrics.prometheus();
+  let appMetrics = await registry.metrics();
+  res.end(prismaMetrics + appMetrics);
+});
+
+app.listen(port, () => {
+  console.log(`Example app listening on port ${port}`);
+});
+```
```
TasinIshmam: I tried running this code, but the `appMetrics` from the other registry is empty, which was a bit confusing. 

Suggestion: perhaps it makes sense to show an example that uses actual metrics besides ones from Prisma? 

The example can be tweaked a bit to expose the default metrics from `prom-client`. 

```suggestion
```ts
import { PrismaClient } from "@prisma/client";
import express, { Request, Response } from "express";
import prom from "prom-client";

const app = express();
const port = 4000;
const prisma = new PrismaClient();

const register = new prom.Registry()
prom.collectDefaultMetrics({ register })

app.get("/metrics", async (_req, res: Response) => {
  let prismaMetrics = await prisma.$metrics.prometheus();
  let appMetrics = await register.metrics();
  res.end(prismaMetrics + appMetrics);
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`);
});
```
```

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 21 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: I've accepted this suggestion.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 25 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: In this case, there is no specific file or place where the user adds the code. It's dependent on their application. All we can really say is that they add it to their application code.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 20 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: This change has been made in the doc.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 29 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: I've reworked the paragraph, and referred specifically to the code example below.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 39 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: Good idea - I've added this highlighting

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 26 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: Agreed - I'll add an explanation of this.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 47 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: Thanks - I can't commit this suggestion, because the file has changed too much, but I've brought it in manually.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 38 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: @janpio  Is this question cleared up by the correction to the code ` globalLabels: { global: 'label', app: 'one' },`, or is there still confusion here?

File: content/400-reference/200-api-reference/050-prisma-client-reference.mdx (hunk 40 above)
Comment (andrew-walford-prisma) [reply content/400-reference/200-api-reference/050-prisma-client-reference.mdx]: Thanks both. I'll add details for the two methods.

**Review by tanberry — CHANGES_REQUESTED** [review]

File: content/400-reference/200-api-reference/050-prisma-client-reference.mdx (hunk 40 above)
Comment (tanberry) [reply content/400-reference/200-api-reference/050-prisma-client-reference.mdx]: We can choose to either list both methods, with their relevant info, or as Jan suggests, have `metrics` as a title, and then have sub-sections for the two methods. There are pros and cons to each.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 46 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: As discussed earlier, I've written a very similar description, but in the JSON section, where this applies. I've amalgamated our two wordings - for example, you are right that these are counts, not values.

**Review by andrew-walford-prisma — COMMENTED** [review]
I've responded to the various reviewer comments, actioned most of them, and where I haven't, I've left an explanatory note.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 38 above)
Comment (janpio) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: I think it is still confusing to actually use an example global label that has the strings "global" and "label" as values. Using something else would just remove this whole confusion.

Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: Thanks - I've actually just clarified the example, using server names, as in the paragraph above. It should be a lot clearer now.

File: content/400-reference/200-api-reference/050-prisma-client-reference.mdx (hunk 40 above)
Comment (andrew-walford-prisma) [reply content/400-reference/200-api-reference/050-prisma-client-reference.mdx]: I've added details of the two methods. I don't think there's a huge amount to add, but they're there for the sake of completeness.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 45 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: I've added $metrics.json() and $metrics.prometheus() to the API reference page.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 44 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: I can't commit this suggestion, but I've enacted it separately.

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx (hunk 43 above)
Comment (andrew-walford-prisma) [reply content/300-guides/100-performance-and-optimization/090-metrics.mdx]: I can't commit this, but I've made the fix manually.

**Review by tanberry — APPROVED** [review]

File: content/300-guides/100-performance-and-optimization/090-metrics.mdx [hunk 48]
```diff
@@ -0,0 +1,497 @@
+---
```
janpio, line 1: Why was this page moved to the Guides section of the docs? It is the Concepts->Components documentation of the Metrics feature. The "Guides" section of the docs usually does not contain this (without also having at least an equivalent in Concepts=>Components)

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,497 @@
...
+Each histogram "bucket" has two values. The first one is the upper bound of the bucket, and the second one is the count (the number of data values that fall into that bucket). In the following example, there are two instances of a query returning a value between 11 and 20, and five instances of a query returning a value between 21 and 30:
+
+```json
+...
+[20, 2],
+[30, 5],
+...
+```
+
+</Admonition>
```
janpio: The "info" Admonition visually formats this was it was something very important:
image

(I suggest at least to move it _under_ the example, but definitely _after_ the basic example on the method to call.)

janpio: This might be a weird bug, the other info Admonition does not get the bold text: 
image
This looks much more adequate.

andrew-walford-prisma: Good point. There appears to be a bug with admonitions and code blocks. To work around this, I've moved the admonition content to below the code sample (I'll push this change later this morning) and made it into a sub-section, rather than a note.

I don't _love_ this solution, but I think it's the least-worst workaround we can have. I did try recasting the section to have an inline example, rather than a code block, but that makes the explanation less clear.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: The docs team had a conversation about this, and we weren't sure where to put it. In the end, I decided that usage ("how do I do this?") trumps feature lists; in that view, it belongs with the other performance and optimization pages. However, as a compromise, I propose that I put a page in components>prisma-client with a brief description, that then links to the page in performance and optimization. I'll add this shortly.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,493 @@
...
+
+<TopBlock>
+
+Prisma metrics give you a detailed insight into how Prisma Client interacts with your database. You can use this insight to help diagnose performance issues with your application.
+
+</TopBlock>
+
+## About metrics
+
+You can export metrics in JSON or Prometheus formats and view them in a console log, or integrate them into an external metrics system, such as [StatsD](https://github.com/statsd/statsd) or [Prometheus](https://prometheus.io/). If you integrate them into an external metrics system, then you can view the metrics data over time. For example, you can use metrics to help diagnose how your application's number of idle and active connections changes.
```
ruheni:
```suggestion
You can export metrics in JSON or Prometheus formats and view them in a console log, or integrate them into an external metrics system, such as [StatsD](https://github.com/statsd/statsd) or [Prometheus](https://prometheus.io/). If you integrate them into an external metrics system, then you can view the metrics data over time. 

For example, you can use metrics to help diagnose how your application's number of idle and active connections changes.
```

Adding the extra line break to make it easier to read

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,493 @@
...
+
+Prisma Client provides the following metrics:
+
+- Counters (always increase):
+
+  - `query_total_queries`: The total number of database queries executed.
+  - `query_total_operations`: The total number of Prisma operations executed.
+    - A Prisma operation is a statement like `let user = prisma.user(include: {posts: true}})`, which Prisma Client translates into one or more SQL queries.
```
ruheni:
```suggestion
    - A Prisma operation is a statement like `await prisma.user({ include: { posts: true }})`, which Prisma Client translates into one or more SQL queries.
```

I think the variable assignment is not necessary here.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,493 @@
...
+import { PrismaClient } from "@prisma/client";
+import express, { Request, Response } from "express";
+
+const app = express();
+const port = 4000;
+const prisma = new PrismaClient();
+
+app.get("/metrics", async (_req, res: Response) => {
+  let metrics = await prisma.$metrics.prometheus();
+  res.end(metrics);
```
ruheni: Should this be `res.end` or `res.send`?

content/300-guides/100-performance-and-optimization/090-metrics.mdx
```diff
@@ -0,0 +1,493 @@
...
+});
+
+app.listen(port, () => {
+  console.log(`Example app listening on port ${port}`);
+})
+```
+
+The following example shows how to combine Prisma metrics with other Prometheus client libraries, also served with an `Express.js` endpoint:
```
ruheni:
```suggestion
The following example shows how to combine Prisma metrics with other Prometheus client libraries, also served with a REST API endpoint using Express.js:
```

content/300-guides/100-performance-and-optimization/090-metrics.mdx
janpio: Make sure to create an issue for the webdev people so this can be fixed - I think no one before could isolate the problem that well. Might also be worth creating a fake PR that shows the problem in multiple code blocks so it is reproducible.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
janpio: Although that is still inconsistent with all the other documentation we have, it is definitely better than not having a page there at all.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: Will do!

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: I'm closing the conversation, but I'll bring in this suggestion, with slightly tweaked wording.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: I think you're right

andrew-walford-prisma: According to @SevInf , res.end will work here and is more lightweight, so I'll leave it as-is.

content/300-guides/100-performance-and-optimization/090-metrics.mdx
andrew-walford-prisma: This is probably reasonably moot, because we're re-orging the docs soon, but I'll switch the content of the two pages, so that the main content is in components, and we link to it from the guides section.

PR #3216, opened 2022-06-01
Update 100-data-modeling.mdx

andrew-walford-prisma:
Minor grammar fix

### DIFF
--- a/content/200-concepts/050-overview/100-what-is-prisma/100-data-modeling.mdx
+++ b/content/200-concepts/050-overview/100-what-is-prisma/100-data-modeling.mdx
@@ -24,7 +24,7 @@ Data modeling typically needs to happen on (at least) two levels:
 - On the **database** level
 - On the **application** level (i.e., in your programming language)
 
-The way how the application models are represented on both levels might differ due to a few reasons:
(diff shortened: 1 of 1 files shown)


content/200-concepts/050-overview/100-what-is-prisma/100-data-modeling.mdx
```diff
@@ -24,7 +24,7 @@ Data modeling typically needs to happen on (at least) two levels:
 - On the **database** level
 - On the **application** level (i.e., in your programming language)
 
-The way how the application models are represented on both levels might differ due to a few reasons:
+The way that the application models are represented on both levels might differ due to a few reasons:
```
janpio, line 27: Can this not be even simpler?
```suggestion
The way the application models are represented on both levels might differ due to a few reasons:
```

andrew-walford-prisma: Simpler is nearly always better, but in this case "that" clarifies it. We plan to introduce a style guideline to this effect.

janpio: Yeah my English is to third language for these nuances :D

PR #3223, opened 2022-06-02
(fix) Remove copy button on cmdResult code block

### LINKED ISSUE #2945: Copy buttons on code blocks with "CLI results" are weird

carlagn:
## Describe this PR

Copy button on the `<cmdResult>` code blocks need to be removed, as the UI looks odd with it.

## Changes

Added a `no-copy` prop to mdx files with `<cmdResult>` tags. Conditionally rendering the copy button according to the `no-copy` prop.

## What issue does this fix?

[Fixes #2945](https://github.com/prisma/docs/issues/2945)

### DIFF
--- a/content/100-getting-started/01-quickstart.mdx
+++ b/content/100-getting-started/01-quickstart.mdx
@@ -98,7 +98,7 @@ npm run dev
 
 <cmdResult>
 
-```json5 no-lines
+```json5 no-lines no-copy
 [
   { id: 1, email: 'sarah@prisma.io', name: 'Sarah', posts: [] },
   {
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
@@ -51,7 +51,7 @@ const getAnimals = await prisma.animal.findMany({
 </cmd>
 <cmdResult>
 
-```js
+```js no-copy
 {
       "id": 7,
       "name": "Salmon",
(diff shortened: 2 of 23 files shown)


tanberry:
@carlagn and @nilubava this fix looks good to me, I would be fine with it getting merged. I went through many of the pages, and see that the code snippets that are output (and this should not have a copy icon) do not have one.

The only odd thing I noticed was on [this page](https://deploy-preview-3223--prisma2-docs.netlify.app/docs/concepts/components/prisma-client/select-fields), two of the output snippets are collapsed by default, so you have to click "Show CLI results"... only two, while the others are expanded by default. Not too big a problem, up to you al if you want to fix that.

tanberry (approved):

zachtil:
The only potential issue I am seeing is that you have to add `no-copy` every time to a `<cmdResult>` component.

To me these would make more sense:

(**Note 0**: There might be no way to achieve these, in which case you can disregard my comment)
(**Note 1**: Disregard suggested option `enable-copy`, I've added that for illustration purposes)

**1. Command/code block with a copy button**

````
```terminal
```
````

**2. Command/code block without a copy button**

````
```terminal no-copy
```
````

**3. Command result without a copy button**

````
<cmdResult>
```terminal
```
</cmdResult>
````

**4. Command result with a copy button**

````
<cmdResult>
```terminal enable-copy
```
</cmdResult>
````

PR #3224, opened 2022-06-03
MySQL SetDefault is not allowed

pimeys:
The docs part of: https://github.com/prisma/prisma/issues/11498


janpio (changes requested):
Probably not a good idea ad discussed

pimeys:
closing because we need to rephrase this...

PR #3226, opened 2022-06-03
feat:cookie consent

nilubava:


janpio:
This overlaying all pages on the Percy screenshots will be a problem.
We already have some Percy specific CSS to hide some things (images), that should probably also be applied for that banner or the Percy diffs will be much harder to use.

PR #3233, opened 2022-06-06
Edit Client methods intro sentence

tanberry:
Edited to try to make more clear that "Client-level" means any methods that interact with the Prisma Client.


keerlu (commented):
'Interact' isn't quite the right word, it's the client that is doing the method. E.g. for `client.$disconnect` the client is doing the disconnecting. Not sure how you want to phrase that...

janpio:
The current sentence differentiates "Client level" from "model level" I think. You have stuff that is about the models from your schema, and you have stuff that is disconnected from that and lives one level higher - the connection lifecycle things, but also Client things like middlewares and logging, or the raw methods or transactions.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -5166,7 +5166,7 @@ const getUsers = await prisma.user.findMany({
 
 ## Client methods
 
-Client-level methods are prefixed by `$`.
+Methods that interact with the Prisma Client are prefixed by `$`.
```
tanberry:
```suggestion
Client-level methods (that is, methods executed by the Prisma Client) are prefixed by `$`.
```

janpio (changes requested):
See comment

tanberry:
> 

@janpio  I saw your comment, which is why I left the phrase "Client-level", then explained what that means in the parentheses. Let me know if you have exact preferred wording; I'd like to close this PR. ;-)

janpio:
It does not mean what you added in the parens.

From my previous comment:

> You have stuff that is about the models from your schema, and you have stuff that is disconnected from that and lives one level higher - the connection lifecycle things, but also Client things like middlewares and logging, or the raw methods or transactions.

That is meant as "Client level" here vs. the model level functionality, that lives under `prisma.modelname.*`.

PR #3236, opened 2022-06-07
docs: Minor update to CRUD docs

hendrasan:
## Describe this PR

Right now, in the CRUD docs, the comparison between Relational databases and MongoDB example schema seems like the order of datasource definition is different between the two. This PR changes that to make it easier to see the difference between the two.

## Changes

Move the datasource definition for Postgres example schema to make it the same as with the Mongo example schema


tanberry:
Thanks so much, @hendrasan ! Your change also makes the order match what we document on the [Schema page](https://www.prisma.io/docs/concepts/components/prisma-schema), with the order being 1. datasource, and 2. generator block.

Your help is much appreciate!

PR #3243, opened 2022-06-08
Fix mistake in the description of `prisma migrate deploy`

aradalvand:
The first paragraph mistakenly states that the `prisma migrate deploy` command is meant to be used in "development" environments, while it should in fact say "production". 
This command is meant to be used in non-development environments, such as production, as clearly stated [here](https://www.prisma.io/docs/concepts/components/prisma-migrate#production-and-testing-environments) and [here](https://www.prisma.io/blog/prisma-migrate-ga-b5eno5g08d0b#applying-migrations-in-production-and-other-environments).


tomhoule (approved):
Looks good to me!

tanberry:
Thanks for catching this, @aradalvand !!

PR #3261, opened 2022-06-12
graphcool: Mention that shut down was 2021

janpio:


tanberry:
@janpio , since this change seems pretty unharmful and truthful, I'll go ahead and merge it... Thanks!

PR #3262, opened 2022-06-13
fix command in Moving backwards and reverting all changes

Jolg42:
Removing the extra `**` at the end


janpio (approved):

tanberry:
Thanks so much @Jolg42 !!

PR #3276, opened 2022-06-15
3179 document mongodb composite default filling on required fields prisma 4

### LINKED ISSUE #3179: Document MongoDB composite default filling on required fields [ Prisma 4 ]
Part of the Prisma 4 effort: Docs should mention that default values on non-existing required fields in Mongo embedded documents (our composite types) are read back just like model defaults for required fields.

Related issue: https://github.com/prisma/prisma/issues/12665

Andrew's notes:

Status: I've asked Matt and Flavian to review. Matt has reviewed; I've asked him a question on one point (see the PR conversation for details).

SWE: Flavian

Possible docs to update:

- [x] https://www.prisma.io/docs/concepts/components/prisma-client/composite-types (there is no mention of this issue on the page, but we might want to mention it now)
- [x] https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#composite-type-methods (no need to change)
- [x]  https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#define-an-index-on-a-composite-type-field - (again, no need to change IMO)

We don't consider this a breaking change: in the migration guide we could mention that if you rely on the error returned prior to v4.0.0, then you will need to change your code. However, we've decided not to do this because it seems an unlikely case.

Source info:
* https://www.notion.so/prismaio/Composite-read-default-on-undefined-1688123bac0e4d6db8da25241e0cf4d8
* https://github.com/prisma/prisma/issues/12665

Done and merged

andrew-walford-prisma:
In the composite types page, I added a description of the new behavior for default values in composite types. I also updated the preview feature notice, and moved it down the page, because this notice is somewhat less important when the feature is GA.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
@@ -9,7 +9,7 @@ tocDepth: 2
 
 <Admonition type="info">
 
-Composite types are available in Preview in versions `3.10.0` and later.
+Composite types were available in Preview in versions `3.10.0` and later. They were made Generally Available in v3.12.0.
 
 </Admonition>
 
@@ -92,7 +92,54 @@ There are currently some limitations when using composite types in the Prisma Cl
 
 ## Default values in required composite types
 
-From v4.0.0, if a a [required](/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields) composite type has a [default value](/concepts/components/prisma-schema/data-model#defining-a-default-value) defined and the value read back from the database is `undefined`, then Prisma inserts the default value into the read result. This is the same behavior as with model fields.
+From v4.0.0, if you carry out a database read on a composite type when the following criteria are true, then Prisma Client inserts the default value into the result:
+
+- In your schema, the composite type is defined as [required](/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields) and has a [default value](/concepts/components/prisma-schema/data-model#defining-a-default-value).
+- The database read operation returns the value `undefined`.
+
+Note:
+
+- This is the same behavior as with [model fields](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types).
+- Prisma Client inserts the default value into the result, but it does not insert the default value into the database.
+
+### Example of default values in required composite types
+
+Add the following line to your schema:
+
+```prisma file=schema.prisma
+type Photo {
+  height Int    @default(200)
+  width  Int    @default(100)
+  url    String
++ description String @default("none")
+}
+```
+
+For example, if you run the following application code:
+
+```ts
+const orders = await prisma.order.findMany({
+  where: {
+    shippingAddress: {
+      is: {
+        street: '555 Candy Cane Lane',
+      },
+    },
+  },
+})
+```
+
+Then you get the following output:
+
+```
+field: "foo"
+```
+
+```
+await prisma.product.findMany({}), { depth: Infinity })
+```
+
+**Earlier versions:** Before v4.0.0, Prisma returned an error message similar to the following one: **!!! did it return an error message or a null? the source info is conflicting !!!\***
 
 ## Finding records that contain composite types with <inlinecode>find</inlinecode> and <inlinecode>findMany</inlinecode>
 
--- a/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
@@ -94,12 +94,12 @@ Conditions:
 
 - The composite type is [required](/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields), and
 - the composite type has a [default value](/concepts/components/prisma-schema/data-model#defining-a-default-value), and
-- the database read operation returns the value `undefined`.
+- the composite type is not present in the returned document or documents.
 
 Note:
 
 - This is the same behavior as with [model fields](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types).
-- Prisma Client inserts the default value into the result, but does not insert the default value into the database.
+- On read operations, Prisma Client inserts the default value into the result, but does not insert the default value into the database.
 
 In the example schema on this page, `photo` has two required fields with default values: `height` and `width`:
 
@@ -113,7 +113,7 @@ type Photo {
 ...
 ```
 
-The following application code returns the default value if any records have `undefined` in `height` or `width`:
+The following application code returns the default value if any records in the returned document have either `height` or `width` missing:
 
 ```ts
 console.dir(await prisma.product.findMany({}), { depth: Infinity })
--- a/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
@@ -9,7 +9,7 @@ tocDepth: 2
 
 <Admonition type="info">
 
-Composite types were available in Preview in versions `3.10.0` and later. They were made Generally Available in v3.12.0.
+We made composite types [Generally Available](/about/prisma/releases#generally-available-ga) in v3.12.0. Before this, they were available in [preview](concepts/components/preview-features) in v3.10.0.
 
 </Admonition>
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -90,6 +86,47 @@ There are currently some limitations when using composite types in the Prisma Cl
 - [`findUnique`](/reference/api-reference/prisma-client-reference#findunique) can't filter on composite types
 - [`aggregrate`](/concepts/components/prisma-client/aggregation-grouping-summarizing#aggregate), [`groupBy`](/concepts/components/prisma-client/aggregation-grouping-summarizing#group-by), [`count`](/concepts/components/prisma-client/aggregation-grouping-summarizing#count) don’t support composite operations
 
+## Default values in required composite types
```
matthewmueller:
```suggestion
## Default values on required composite types
```

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -90,6 +86,47 @@ There are currently some limitations when using composite types in the Prisma Cl
...
+
+From v4.0.0, if you carry out a database read on a composite type when all of the following conditions are true, then Prisma Client inserts the default value into the result.
+
+Conditions:
+
+- The composite type is [required](/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields), and
```
matthewmueller: I don't think this true, it can also provide defaults if the composite type is optional and the value is undefined.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -21,6 +15,8 @@ Composite types are only available with MongoDB.
 
 [Composite types](/concepts/components/prisma-schema/data-model#defining-composite-types), known as [embedded documents](https://docs.mongodb.com/manual/core/data-model-design/#std-label-data-modeling-embedding) in MongoDB, allow you to embed records within other records.
 
+We made composite types [Generally Available](/about/prisma/releases#generally-available-ga) in v3.12.0. They were previously available in [preview](concepts/components/preview-features) from v3.10.0.
```
matthewmueller:
```suggestion
We made composite types [Generally Available](/about/prisma/releases#generally-available-ga) in v3.12.0. They were previously available in [Preview](concepts/components/preview-features) from v3.10.0.
```

Not sure about the capitalization here, I feel like if Generally Available is capitalized so should preview?

matthewmueller (commented):
Nice work!

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: OK thanks, I'll amend the text

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: We're not currently consistent in our docs with capitalization of things like preview. I feel strongly that we should use capitalization very sparingly, and as such I've gone with lower case, except for terms where we appear to be consistent with upper case. For General Availability, it maps to the abbreviation GA, so my instinct is that this should be capitalized for now, until we've decided on a policy for this. In the near future, the docs team will be expanding our style guide to will spell out our guidelines on things like concept/component capitalization.

andrew-walford-prisma: However, you're right that this jumps out in this paragraph, so I'll capitalize preview for now.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: @matthewmueller According to the description [here](https://www.notion.so/prismaio/Composite-read-default-on-undefined-1688123bac0e4d6db8da25241e0cf4d8), and the [issue description](https://github.com/prisma/prisma/issues/12665), the composite type does have to be required. At least if I understand this correctly.

Weakky: Sorry for the late review. Andrew is correct, only required fields will be filled with default values in case they're undefined. Otherwise, we might end up filling values for optional fields when they're intended to be null/undefined by the user

Weakky (approved):
All good 🙌

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: Thanks - in this case I'll close the conversation.

matthewmueller: That makes sense and sorry for the confusion!!

matthewmueller (approved):

floelhoeffel (approved):
Nicely explained!

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -90,6 +86,47 @@ There are currently some limitations when using composite types in the Prisma Cl
...
 
+## Default values on required composite types
+
+From v4.0.0, if you carry out a database read on a composite type when all of the following conditions are true, then Prisma Client inserts the default value into the result.
+
+Conditions:
+
+- The composite type is [required](/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields), and
+- the composite type has a [default value](/concepts/components/prisma-schema/data-model#defining-a-default-value), and
+- the database read operation returns the value `undefined`.
```
dpetrick: This is technically incorrect, mongo can't return `undefined`, the field is simply not present on the returned document(s) (aside: `undefined` is a state property of the data, not a data value even if JS looks like it treats it as one).

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -90,6 +86,47 @@ There are currently some limitations when using composite types in the Prisma Cl
...
+
+Note:
+
+- This is the same behavior as with [model fields](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types).
+- Prisma Client inserts the default value into the result, but does not insert the default value into the database.
```
dpetrick: This is confusing because Prisma _does_ insert the defaults into the database, but only on a create operation. Prisma doesn't write the default on read operations, which is what you're referring to here I assume. I know the whole section only really mentions this in the context of read operations, but I'd like to be very explicit about this before someone thinks otherwise.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -90,6 +86,47 @@ There are currently some limitations when using composite types in the Prisma Cl
...
+...
+type Photo {
+  height Int    @default(200)
+  width  Int    @default(100)
+  url    String
+}
+...
+```
+
+The following application code returns the default value if any records have `undefined` in `height` or `width`:
```
dpetrick: Same problem as above: If the documents have either height or width _missing_ on the document.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -90,6 +86,47 @@ There are currently some limitations when using composite types in the Prisma Cl
...
+```ts
+console.dir(await prisma.product.findMany({}), { depth: Infinity })
+```
+
+** Earlier versions **
+
+Before v4.0.0, Prisma returned an error message similar to the following:
+
+```
+Invalid `prisma.product.findMany()` invocation ...
```
dpetrick: This doesn't really help, does it? Is it possible to get a more concrete error in here?

dpetrick (changes requested):

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: Thanks - I'll address this

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: Thanks - I'll address this

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: Fair point! I'll update the text.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -94,12 +94,12 @@ Conditions:
 
 - The composite type is [required](/concepts/components/prisma-schema/data-model#optional-and-mandatory-fields), and
 - the composite type has a [default value](/concepts/components/prisma-schema/data-model#defining-a-default-value), and
-- the database read operation returns the value `undefined`.
+- the composite type is not present in the returned document or documents.
```
dpetrick: I'm just seeing this, but it's not the composite type that has a default or is really interesting here, it's a _field on the composite_. Consider:
```
models Test {
  // ...
  composite Composite?
}

type Composite {
  field String @default("test")
}
```

The composiste is optional, but that doesn't matter. What matters is that the field `field` on the composite is required and has a default. The "composite is x" is wrong in multiple places here.

Sorry for not catching that earlier.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
```diff
@@ -113,7 +113,7 @@ type Photo {
 ...
 ```
 
-The following application code returns the default value if any records have `undefined` in `height` or `width`:
+The following application code returns the default value if any records in the returned document have either `height` or `width` missing:
```
dpetrick: Decide if you want to use records or documents, but not both. A record is a document in mongo. I'd leave it at `if any records have either height or width missing on the composite :`

Additionally, the example seems off. `prisma.product.findMany({})` shoudn't fill in `height` or `width`, the entire composite is unset, which doesn't fill in anything. Only if the composite exists somehow, say as an empty embedded document like `photo: {}` on the document _then_ it fills in fields on the composite.

dpetrick (dismissed):
Looks like I missed some fundamental problems in my first review round, raised them now - sorry!

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: Thanks, I've fixed this now

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: I've added details of the specific error message now.

content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
andrew-walford-prisma: I've made the first of these changes.

I'm investigating change 2.

andrew-walford-prisma: I've updated the example to address this point.

PR #3277, opened 2022-06-15
Add unique attribute to 1:1 relations

### LINKED ISSUE #3213: Update docs to only allow explicit `@unique` constraints on one-to-one relations
Currently this syntax is allowed on one-to-one relations:

```
model A {
  id   Int @id
  b_id Int
  b    B   @relation(fields: [b_id], references: [id])
}

model B {
  id Int @id
  a A?
}
```

From Prisma 4 this will not be allowed, and the unique constraint has to be made explicit:

```
model A {
  id   Int @id
  b_id Int @unique // <---
  b    B   @relation(fields: [b_id], references: [id])
}

model B {
  id Int @id
  a A?
}
```

PSL proposal: https://www.notion.so/prismaio/Proposal-make-unique-constraints-on-1-1-relations-explicit-717f5d67094243ecbfa662d1f148c4cf

Affected docs:
- https://www.prisma.io/docs/concepts/components/prisma-schema/relations/one-to-one-relations - need to make this clear and also update examples to match this
- https://www.prisma.io/docs/concepts/components/prisma-schema/relations/self-relations#one-to-one-self-relations check this example
- https://www.prisma.io/docs/concepts/components/prisma-schema/relations check examples
- https://www.prisma.io/docs/guides/database/developing-with-prisma-migrate/customizing-migrations#changing-the-direction-of-a-1-1-relation check example
- Lots more that use a `User` <-> `Profile` one-to-one relation

keerlu:
## Describe this PR

Update docs to only allow explicit @unique constraints on one-to-one relations

## What issue does this fix?

Fixes #3213


content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -1339,7 +1339,7 @@ In the following example, `authorId` is a both a relation scalar and the ID of `
 
 ```prisma
 model Profile {
-  authorId Int    @id
+  authorId Int    @id @unique
```
keerlu: Want to check this one - do we still add the `@unique` attribute when the relation scalar already has an `id` attribute?

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
```diff
@@ -23,13 +23,13 @@ techMetaDescriptions:
 
 ## Overview
 
-Each section on this page describes a potential problem when upgrading from Prisma 1 to Prisma 2.*x* and later and explains the available workarounds.
+Each section on this page describes a potential problem when upgrading from Prisma 1 to Prisma 2._x_ and later and explains the available workarounds.
```
andrew-walford-prisma:
```suggestion
Each section on this page describes a potential problem when upgrading from Prisma 1 to Prisma 2._x_ and later, and explains the available workarounds.
```

IMO the additional comma makes the sentence easier to parse.

andrew-walford-prisma (commented):
One comment; otherwise LGTM!

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
keerlu: Hm, it looks like these changes were actually introduced by the autoformatter and slipped in by mistake. So thanks for commenting on them!

I'll revert them to avoid muddling the PR, so I'll leave the comma suggestion (agree that it does read better)

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
do4gr: No, @id already implies uniqueness.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -1357,7 +1357,7 @@ model User {
 
 ```prisma
 model Profile {
-  authorId String @id @map("_id") @db.ObjectId
+  authorId String @id @unique @map("_id") @db.ObjectId
```
do4gr: Same here, no need for the additional @unique due to the presence of the @id

floelhoeffel (commented):
This looks good from my side (excluding Matthias comments)

Jolg42 (approved):
I think we could merge it after review comments are addressed, no need to wait for Prisma 4 release

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: So fine without:
```suggestion
  authorId Int    @id
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio:
```suggestion
  authorId String @id @map("_id") @db.ObjectId
```

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
```diff
@@ -328,7 +328,7 @@ model User {
 
 model Profile {
   id   String  @id
-  user String?
+  user String? @unique
```
janpio: From the text below, this sounds like this is an intentional 1:n and not a 1:1 via Introspection. (As the underlying table does not have a unique constraint on that field)

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
```diff
@@ -386,7 +386,7 @@ model User {
 
 model Profile {
   id     String @id
-  userId String
+  userId String @unique
   User   User   @relation(fields: [userId], references: [id])
 }
 ```
```
janpio: I am not sure this whole block makes sense any more. The original introspected schema is a 1:n, without the `@unique`. Here the description tells you to just remove `[]` to fix one of the sides - but that is not correct as you _also_ need to add the `@unique` manually now. If you do that, the database is still not matching that and a next migration would add unique or a next introspection would remove the @unique again and produce an invalid schema.

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/03-upgrading-the-prisma-layer.mdx
```diff
@@ -435,7 +435,7 @@ model Post {
 model Profile {
   id   String  @id @default(cuid())
   bio  String?
-  user String?
+  user String? @unique
```
janpio, line 438: Are we sure that is actually the introspection result of the mentioend P1 schema? Just us changing our P4 implementation does not necessarily change the introspection result of such P1 schema and database. 

(The SQL is in a tab above, so we can test this)

content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
```diff
@@ -22,7 +22,7 @@ The following example models a one-to-one self-relation:
 model User {
   id          Int     @id @default(autoincrement())
   name        String?
-  successorId Int?
+  successorId Int?    @unique
```
janpio, line 25: I think lines 54 to 60 describe what is needed for a one-to-one self relation. And now it seems we also require the `@unique` on the scalar relation field, so maybe we should add that to the list?

Maybe an addition to the third list item:
> - One relation field must be backed by a foreign key. The `successor` field is backed by the `successorId` foreign key, which references a value in the `id` field. The `successorId` scalar relation field needs to be `@unique` to guarantee a 1:1 relation.

content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
```diff
@@ -68,7 +68,7 @@ Either side of the relation can be backed by a foreign key. In the following exa
 model User {
```
janpio, line 68: Unrelated: When checking this change, I noticed how werid the wording in line 62 is. This example here is the same we used in lines 22-28, but that is not acknowledged or mentioend so I kept checking if I was missing something. Maybe we can add something along the lines of "(this is the same example that was used further up the page)" or similar.

content/300-guides/400-migrate-to-prisma/02-migrate-from-sequelize.mdx
```diff
@@ -403,7 +403,7 @@ model Profiles {
   bio       String
   createdAt DateTime
   updatedAt DateTime
-  userId    Int?
+  userId    Int?     @unique
```
janpio, line 406: Are we sure the SQL here is actually introspected with `@unique` and is not invalid? That might require some mention that this needs to be fixed here then.

content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
do4gr: I like the suggestion.

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
do4gr: I'm getting a 404 for the preview here

do4gr: Same for the currently

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
do4gr: Yeah, I agree. We list two workarounds for the situation. The first one for me is rather a solution to make it correct (add the unique). This second one is a hack that fundamentally leaves you in an incorrect state, I don't think we should recommend that.

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
keerlu: Should be https://deploy-preview-3277--prisma2-docs.netlify.app/docs/guides/upgrade-guides/upgrade-from-prisma-1/schema-incompatibilities-postgres ... link creator doesn't like numbers apparently

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/03-upgrading-the-prisma-layer.mdx
do4gr: I think this is correct.The unique constraint is missing from the initial full schema sql, but as part of the update process the unique constraint is added manually. So if you introspect after having gone through the process this should be the result.

content/300-guides/400-migrate-to-prisma/02-migrate-from-sequelize.mdx
do4gr: Yeah, since we're starting from a Sequelize project and I can't see the SQL being introspected I'm also not sure here. Would need to run the actual example.

do4gr (approved):

content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
keerlu: Thanks, done

content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
keerlu: Agree this is a bit weird. Hopefully worded better now.

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
keerlu: Thanks, I've removed this example

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/03-upgrading-the-prisma-layer.mdx
keerlu: Thanks, will leave as it is

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/02-schema-incompatibilities.mdx
keerlu: Agree that this looks to be an intentional 1-n, I've removed the `@unique` attribute

content/300-guides/400-migrate-to-prisma/02-migrate-from-sequelize.mdx
keerlu: Leaving this as is for now

PR #3278, opened 2022-06-15
3265 dbnull jsonnull and anynull as symbols prisma 4

### LINKED ISSUE #3265: DbNull, JsonNull, and AnyNull as symbols [ Prisma 4 ]
Part of the Prisma 4 effort: Docs should mention that DbNull, JsonNull, AnyNull are now symbols (previously strings).

Related issue: https://github.com/prisma/client-planning/issues/18
Relevant docs page: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields

=> Coordinate with @aqrln on what needs to change in the docs, if anything.

Andrew's update:

SME: Alexey

Spec:
* https://www.notion.so/prismaio/Make-DbNull-and-JsonNull-symbols-instead-of-string-constants-b29bb0fa77a941178009b8923a98efdf

We only mention this in two docs:
- [x] https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#filtering-by-null-values - I've added a brief explanation - see below; this is with Matt for comment
- [x] https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3 - no change needed here

From Matt and Frederik's notes:
We can touch on this in the docs and explain that if you encounter errors then it is because we are now catching mistakes.

Update: 
* I think we should have a minimal, quite sober, mention of this in the docs, then a bit more detail (mentioning potential issues) in the upgrade guide. Matt concurs.
* I've asked Alexey and Matt to review the PR: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#filtering-by-null-values

**Breaking change**
Alexey has put some useful notes in [his PR](https://github.com/prisma/prisma/pull/13783). These will be useful for the upgrade guide. 

Done and merged

andrew-walford-prisma:
I've added a brief mention of this in the docs. The detail for this belongs in the upgrade guide, because in the unlikely event it affects users, it will be when they upgrade to Prisma 4. Matt concurs with this approach. I'll work with Lucy on the upgrade guide material.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
@@ -878,6 +878,12 @@ To differentiate between these possibilities, we've introduced three _null enums
 - `DbNull`: Selects the `NULL` value in the database.
 - `AnyNull`: Selects both `null` JSON values and `NULL` database values.
 
+<Admonition type="info">
+
+From v4.0.0, Prisma returns `JsonNull`, `DbNull`, and `AnyNull` as [JavaScript symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). Before v4.0.0, Prisma returned them as [magic strings](https://en.wikipedia.org/wiki/Magic_string).
+
+</Admonition>
+
 For example:
 
 ```prisma
@@ -927,12 +933,6 @@ prisma.log.create({
 <Admonition type="info">
 
 These _null enums_ do not apply to MongoDB because there the difference between a JSON `null` and a database `NULL` does not exist. They also do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
-
-</Admonition>
-
-<Admonition type="info">
-
-Before v4.0.0, `JsonNull` and `DbNull` were stored as string values. From v4.0.0, they are stored as symbols.
 
 </Admonition>
 
--- a/content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
@@ -932,7 +932,7 @@ prisma.log.create({
 
 <Admonition type="info">
 
-These _null enums_ do not apply to MongoDB because there the difference between a JSON `null` and a database `NULL` does not exist. They also do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
+These _null enums_ do not apply to MongoDB because MongoDB does not differentiate between a JSON `null` and a database `NULL`. They also do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
 
 </Admonition>
 
--- a/content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
@@ -880,7 +880,7 @@ To differentiate between these possibilities, we've introduced three _null enums
 
 <Admonition type="info">
 
-From v4.0.0, `JsonNull`, `DbNull`, and `AnyNull` are [JavaScript symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). Before v4.0.0, they were strings.
+From v4.0.0, `JsonNull`, `DbNull`, and `AnyNull` are objects. Before v4.0.0, they were strings.
 
 </Admonition>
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -922,9 +932,7 @@ prisma.log.create({
 
 <Admonition type="info">
 
-These _null enums_ do not apply to MongoDB because there the difference between a JSON `null` and a database `NULL` does not exist.
-
-They also do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
+These _null enums_ do not apply to MongoDB because there the difference between a JSON `null` and a database `NULL` does not exist. They also do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
```
matthewmueller, line 935:
```suggestion
These _null enums_ do not apply to MongoDB because MongoDB does not differentiate between a JSON `null` and a database `NULL`. They also do not apply to the `array_contains` operator in all databases because there can only be a JSON `null` within a JSON array. Since there cannot be a database `NULL` within a JSON array, `{ array_contains: null }` is not ambiguous.
```

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -922,9 +932,7 @@ prisma.log.create({
 
 <Admonition type="info">
 
-These _null enums_ do not apply to MongoDB because there the difference between a JSON `null` and a database `NULL` does not exist.
```
matthewmueller, line 925: Not necessarily specific to this PR, but I'm asking myself what happens when you have a MongoDB database and you try using `Prisma.DbNull`?

Do we accept and just ignore it? Is it a type error? Do we return no results?

matthewmueller (approved):
LGTM!

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -874,6 +878,12 @@ To differentiate between these possibilities, we've introduced three _null enums
 - `DbNull`: Selects the `NULL` value in the database.
 - `AnyNull`: Selects both `null` JSON values and `NULL` database values.
 
+<Admonition type="info">
+
+From v4.0.0, Prisma returns `JsonNull`, `DbNull`, and `AnyNull` as [JavaScript symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). Before v4.0.0, Prisma returned them as strings.
```
aqrln: We don't actually return them anywhere, they are only used as inputs. Both kinds of null values are returned as JavaScript `null`s in the results.

Maybe something like:

```suggestion
From v4.0.0, `JsonNull`, `DbNull`, and `AnyNull` are [JavaScript symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). Before v4.0.0, they were strings.
```

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
aqrln: Without checking, I believe previously we would silently interpret as a string and store a literal "DbNull" string or filter by it. Now it is a type error (and a runtime validation error if you ignore the types or don't use TypeScript).

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
```diff
@@ -874,6 +878,12 @@ To differentiate between these possibilities, we've introduced three _null enums
 - `DbNull`: Selects the `NULL` value in the database.
 - `AnyNull`: Selects both `null` JSON values and `NULL` database values.
 
+<Admonition type="info">
+
+From v4.0.0, `JsonNull`, `DbNull`, and `AnyNull` are [JavaScript symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). Before v4.0.0, they were strings.
```
matthewmueller: Just a heads up that this may change. We could abstract this out and just say "constants" or wait on a decision from @aqrln on his approach.

Relevant discussion: https://prisma-company.slack.com/archives/C02FZENHD4N/p1655923522177859?thread_ts=1655821379.664189&cid=C02FZENHD4N

matthewmueller (dismissed):

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
andrew-walford-prisma: Thanks. Yep, don't worry, I have my eye on this. I'll keep in touch with @aqrln  about this.

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
dpetrick: Context: Mongo JSON is different from the other DBs. Json is actually parsed as extJson and transformed to BSON, meaning that DBNull and JsonNull are functionally the same.

Mongo doesn't have the enums on the API level and only accepts pure JSON.

aqrln (approved):

content/200-concepts/100-components/02-prisma-client/051-working-with-fields/100-working-with-json-fields.mdx
andrew-walford-prisma: As far as I understand this, we don't need a docs change for this, so I'm resolving the conversation.

PR #3284, opened 2022-06-15
Fix code snippet in raw-database-access

manassra:
## Describe this PR

Snippet under the `$queryRaw` section is referencing the wrong variable (`desc`), which should be `ordering` instead.
## Changes
Changed `desc` to `ordering` in the code snippet.

## What issue does this fix?
N/A

N/A


keerlu:
Thanks for spotting that @manassra!

PR #3286, opened 2022-06-15
fix(data-proxy): Expand section on direct connection URL requirement to Introspection and list commands

janpio:
This expands the existing section to cover all affected commands (via https://github.com/prisma/prisma/pull/13627) and also shows how to manually use it, instead of just via npm script.

---

- Before: https://www.prisma.io/docs/concepts/data-platform/data-proxy#providing-the-connection-string-to-prisma-migrate
- After: https://deploy-preview-3286--prisma2-docs.netlify.app/docs/concepts/data-platform/data-proxy#providing-the-connection-string-to-prisma-migrate-and-introspection

### DIFF
--- a/content/200-concepts/150-data-platform/02-data-proxy.mdx
+++ b/content/200-concepts/150-data-platform/02-data-proxy.mdx
@@ -173,7 +173,7 @@ Migrations or Introspection can only be run on a direct (non-proxied) database c
 - `db execute`
 - `studio`
 
-Assuming that your Prisma schema uses `env("DATABASE_URL")` as the `url` of the `datasource` block, you'll need to temporarily override the `DATABASE_URL` environment variable before running any of these commands. You can store the direct database connection string in another environment variable, for example `NONDATAPROXY_DATABASE_URL`, and use that to override the connection before you execute one of the affected commands:
+If your Prisma schema uses `env("DATABASE_URL")` as the `url` of the `datasource` block, you need to temporarily override the `DATABASE_URL` environment variable before running any of the listed commands. You can store the direct database connection string in another environment variable, such as `NONDATAPROXY_DATABASE_URL`, and use that to override the connection before you run one of the listed commands. For example:
 
 ```js
 NONDATAPROXY_DATABASE_URL="postgresql://..."
--- a/content/200-concepts/150-data-platform/02-data-proxy.mdx
+++ b/content/200-concepts/150-data-platform/02-data-proxy.mdx
@@ -213,7 +213,7 @@ Our Vercel Deployment sample repository has a [special branch](https://github.co
 If you want to adapt your project manually:
 
 - Add an environment variable `PRISMA_GENERATE_DATAPROXY` to your [Vercel project settings](https://vercel.com/docs/concepts/projects/environment-variables) and set its value to `true`.
-- If you want to deploy migrations via Vercel, follow our instructions on how to ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate-and-introspection)
+- If you want to deploy migrations via Vercel, see ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate-and-introspection)
 
 ### Deploying to Cloudflare Workers
 
--- a/content/200-concepts/150-data-platform/02-data-proxy.mdx
+++ b/content/200-concepts/150-data-platform/02-data-proxy.mdx
@@ -155,11 +155,41 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
 
 The Prisma Data Proxy does not support the preview features [Interactive Transactions](/concepts/components/prisma-client/transactions#interactive-transactions-in-preview) or [Metrics](/concepts/components/prisma-client/metrics) yet. To use the Prisma Data Proxy, you must first remove `interactiveTransactions` and `metrics` from the `previewFeatures` section of your `prisma.schema` file. When you run `prisma generate --data-proxy`, you will receive an error if you have either of these preview features enabled.
 
-### Providing the connection string to Prisma Migrate
+### Providing the connection string to Prisma Migrate and Introspection
 
 Currently, the Data Proxy can only be used to _query_ your database with [Prisma Client](/concepts/components/prisma-client).
 
-Migrations can be only be run on a direct (non-proxied) database connection. Assuming that your schema uses `env("DATABASE_URL")`, you'll need to override the `DATABASE_URL` environment variable before running `prisma migrate`. You can declare another environment variable, for example `MIGRATE_DATABASE_URL`, and use that to override the connection:
+Migrations or Introspection can only be run on a direct (non-proxied) database connection for now. This affects the following CLI commands:
+
+- `db pull`
+- `migrate deploy`
+- `migrate dev`
+- `migrate diff` (some parameters only)
+- `migrate reset`
+- `migrate resolve`
+- `migrate status`
+- `db push`
+- `db drop`
+- `db execute`
+- `studio`
+
+Assuming that your Prisma schema uses `env("DATABASE_URL")` as the `url` of the `datasource` block, you'll need to temporarily override the `DATABASE_URL` environment variable before running any of these commands. You can store the direct database connection string in another environment variable, for example `NONDATAPROXY_DATABASE_URL`, and use that to override the connection before you execute one of the affectes commands:
+
+```js
+NONDATAPROXY_DATABASE_URL="postgresql://..."
+```
+
+and then
+
+```js
+DATABASE_URL="$NONDATAPROXY_DATABASE_URL" prisma migrate deploy
+
+// or
+
+DATABASE_URL="$NONDATAPROXY_DATABASE_URL" prisma db pull
+```
+
+You can also store the command in your npm scripts of `package.json` and then run them via `npm run migrate` or `npm run introspect`:
 
 ```js
 // package.json
@@ -167,7 +197,8 @@ Migrations can be only be run on a direct (non-proxied) database connection. Ass
   ...,
   "scripts": {
     "generate-client": "prisma generate --data-proxy",
-    "migrate": "DATABASE_URL=\"$MIGRATE_DATABASE_URL\" prisma migrate deploy",
+    "migrate": "DATABASE_URL=\"$NONDATAPROXY_DATABASE_URL\" prisma migrate deploy",
+    "introspect": "DATABASE_URL=\"$NONDATAPROXY_DATABASE_URL\" prisma db pull",
     ...
   }
 }
--- a/content/200-concepts/150-data-platform/02-data-proxy.mdx
+++ b/content/200-concepts/150-data-platform/02-data-proxy.mdx
@@ -213,7 +213,7 @@ Our Vercel Deployment sample repository has a [special branch](https://github.co
 If you want to adapt your project manually:
 
 - Add an environment variable `PRISMA_GENERATE_DATAPROXY` to your [Vercel project settings](https://vercel.com/docs/concepts/projects/environment-variables) and set its value to `true`.
-- If you want to deploy migrations via Vercel, follow our instructions on how to ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate)
+- If you want to deploy migrations via Vercel, follow our instructions on how to ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate-or-introspection)
 
 ### Deploying to Cloudflare Workers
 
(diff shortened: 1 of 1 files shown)


content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -155,19 +155,50 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
...
+- `db push`
+- `db drop`
+- `db execute`
+- `studio`
+
+Assuming that your Prisma schema uses `env("DATABASE_URL")` as the `url` of the `datasource` block, you'll need to temporarily override the `DATABASE_URL` environment variable before running any of these commands. You can store the direct database connection string in another environment variable, for example `NONDATAPROXY_DATABASE_URL`, and use that to override the connection before you execute one of the affectes commands:
+
+```js
+NONDATAPROXY_DATABASE_URL="postgresql://..."
+```
```
janpio, line 180: This should probably be changed to how we usually show how to define an env var (and maybe even link to an explanation if we have any). It should probably be done via the `.env` file as well for simplicity.

content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -155,19 +155,50 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
...
+
+and then
+
+```js
+DATABASE_URL="$NONDATAPROXY_DATABASE_URL" prisma migrate deploy
+
+// or
+
+DATABASE_URL="$NONDATAPROXY_DATABASE_URL" prisma db pull
+```
```
janpio, line 190: This needs to be tested if this actually works manually.

content/200-concepts/150-data-platform/02-data-proxy.mdx
tanberry: @zachtil can you work on this PR, please and thanks?!

content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -155,19 +155,50 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
...
+
+You can also store the command in your npm scripts of `package.json` and then run them via `npm run migrate` or `npm run introspect`:
 
 ```js
```
zachtil, line 194: How about `file=package.json`?

content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -155,19 +155,50 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
 
 The Prisma Data Proxy does not support the preview features [Interactive Transactions](/concepts/components/prisma-client/transactions#interactive-transactions-in-preview) or [Metrics](/concepts/components/prisma-client/metrics) yet. To use the Prisma Data Proxy, you must first remove `interactiveTransactions` and `metrics` from the `previewFeatures` section of your `prisma.schema` file. When you run `prisma generate --data-proxy`, you will receive an error if you have either of these preview features enabled.
 
-### Providing the connection string to Prisma Migrate
+### Providing the connection string to Prisma Migrate and Introspection
```
zachtil, line 158: ...to Prisma CLI commands?

Mentioning this because the heading says "...Prisma Migrate and Introspection", but below we list also `db push`, `db drop`, `studio`. Are these considered part of Prisma Migrate, for example?

I am not sure so just using this as an opportunity to raise the question and understand.

content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -155,19 +155,50 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
...
 
 Currently, the Data Proxy can only be used to _query_ your database with [Prisma Client](/concepts/components/prisma-client).
 
-Migrations can be only be run on a direct (non-proxied) database connection. Assuming that your schema uses `env("DATABASE_URL")`, you'll need to override the `DATABASE_URL` environment variable before running `prisma migrate`. You can declare another environment variable, for example `MIGRATE_DATABASE_URL`, and use that to override the connection:
+Migrations or Introspection can only be run on a direct (non-proxied) database connection for now. This affects the following CLI commands:
```
zachtil: "This _impacts_..."

"_affects_" is best avoided because of its emotional connotation.

zachtil:
```suggestion
Currently, you can run migrations and introspection only on a direct (non-proxied) database connection. This impacts the Prisma CLI commands listed below.
```

content/200-concepts/150-data-platform/02-data-proxy.mdx
zachtil: @tanberry Sure (although a rather late reply).

@janpio How about picking the easiest to document here (setting an environment variable) and linking to the better alternative (using dotenv).

content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -155,19 +155,50 @@ You are now ready to try the Data Proxy. Go ahead and start your app!
...
+- `migrate diff` (some parameters only)
+- `migrate reset`
+- `migrate resolve`
+- `migrate status`
+- `db push`
+- `db drop`
+- `db execute`
+- `studio`
+
+Assuming that your Prisma schema uses `env("DATABASE_URL")` as the `url` of the `datasource` block, you'll need to temporarily override the `DATABASE_URL` environment variable before running any of these commands. You can store the direct database connection string in another environment variable, for example `NONDATAPROXY_DATABASE_URL`, and use that to override the connection before you execute one of the affected commands:
```
zachtil:
```suggestion
If your Prisma schema uses `env("DATABASE_URL")` as the `url` of the `datasource` block, you need to temporarily override the `DATABASE_URL` environment variable before running any of the listed commands. You can store the direct database connection string in another environment variable, such as `NONDATAPROXY_DATABASE_URL`, and use that to override the connection before you run one of the listed commands. For example:
```

content/200-concepts/150-data-platform/02-data-proxy.mdx
zachtil: So this is just in theory? :)

content/200-concepts/150-data-platform/02-data-proxy.mdx
```diff
@@ -182,7 +213,7 @@ Our Vercel Deployment sample repository has a [special branch](https://github.co
 If you want to adapt your project manually:
 
 - Add an environment variable `PRISMA_GENERATE_DATAPROXY` to your [Vercel project settings](https://vercel.com/docs/concepts/projects/environment-variables) and set its value to `true`.
-- If you want to deploy migrations via Vercel, follow our instructions on how to ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate)
+- If you want to deploy migrations via Vercel, follow our instructions on how to ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate-and-introspection)
```
zachtil:
```suggestion
- If you want to deploy migrations via Vercel, see ["Providing the connection string to Prisma Migrate"](#providing-the-connection-string-to-prisma-migrate-and-introspection)
```

zachtil: It is not a good idea to work a title/section name/heading into a sentence. Especially here where you say "on how to..." which requires that the next verb is in infinitve form, but the heading is in gerund form.

 The easiest is to use ", see ..." and cite the page or heading name.

content/200-concepts/150-data-platform/02-data-proxy.mdx
zachtil: I got this error when I tried to run `npm run migrate`:

```
> hello-next@1.0.0 migrate
> DATABASE_URL="$NONDATAPROXY_DATABASE_URL" prisma migrate deploy

Error: Get config: Schema Parsing P1012

error: Error validating datasource `db`: You must provide a nonempty URL. The environment variable `DATABASE_URL` resolved to an empty string.
  -->  schema.prisma:7
   | 
 6 |   provider          = "postgresql"
 7 |   url               = env("DATABASE_URL")
   | 

Validation Error Count: 1


Prisma CLI Version : 3.15.2
```

zachtil: It looks like for some reason `NONDATAPROXY_DATABASE_URL` wasn't set even though I have it in the .env file...

janpio: Argh right, the `.env` file is only a _file_ until some tool reads it into the environment.
Prisma CLI does that, but the command `DATABASE_URL="$NONDATAPROXY_DATABASE_URL"` is executed earlier.
We could use https://www.npmjs.com/package/dotenv-cli to load the `.env` file earlier.

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio: I don't know how we usually deal with this. It is a complex problem.

The "easiest" variant does not work on Windows for example unfortunately. There you have different commands depending on if you are in `Cmd` (`set FOO=bar`) or Powershell (`$env:FOO="bar"`).

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio: Yes, well informed one. 

As you figured out below the `NONDATAPROXY_DATABASE_URL` env var _Actually_ needs to be set, not just a file that contains the definition of that, which would be loaded by Prisma CLI - the copy into `DATABASE_URL` happens before the CLI even executes.

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio: Everything but `studio` is, yes.

I like the suggestion "Prisma CLI commands". That actually covers it nicely. I overindexed on the existing "Migrate" mention here.

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio:
```suggestion
Migrations or Introspection can only be run on a direct (non-proxied) database connection for now. This impacts the following CLI commands:
```

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio: 🤷 👍

content/200-concepts/150-data-platform/02-data-proxy.mdx
zachtil: If `dotenv` works across the board (Win, macOS, Linux), then let's always point to `dotenv` and also indicate why we suggest to use that.
We already have a good page that describes how that works and why it's a good solution. https://www.prisma.io/docs/guides/development-environment/environment-variables#how-does-prisma-use-environment-variables

zachtil: Actually, the other option is to provide tabs for this. But I have some qualms about tabs.

A **Window** tab would show how to set the variables on windows, a **Unix (macOS, Linux)** tab would show the the relevant command for that.

Yet another option is to just have a step that lists all commands for each environment.

Set the `NONDATAPROXY_DATABASE_URL` environment variable.
- In Unix environments (macOS, Linux):
   ```
  NONDATAPROXY_DATABASE_URL="postgresql://..."
  ```
- In Windows Command Prompt:
  ```
  set NONDATAPROXY_DATABASE_URL=postgresql://...
  ```
- In Windows PowerShell:
  ```
  $env:NONDATAPROXY_DATABASE_URL="postgresql://..."
  ```

content/200-concepts/150-data-platform/02-data-proxy.mdx
zachtil: Need to test this and see how it works. 

I am a bit wary of adding yet another prerequisite. I was initially thinking about using immediately available tools (such as system env vars), but I also see how `dotenv` makes more sense based on the fact that for Windows there's a difference.

Let me check.

janpio: dotenv-cli has the benefit of also using `.env`, there is also `crossenv` which does not use the file but enable to use them on the terminal in a consistent way.

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio:
```suggestion
### Providing the connection string to Prisma CLI commands
```

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio: I think Linux standalone needs to be `export FOO="bar"` actually.

content/200-concepts/150-data-platform/02-data-proxy.mdx
janpio: ping @Jolg42

janpio:
This needs to be finished with the usage of `dotenv` to make it work with `.env` files and then would improve the document and instructions substantially.

PR #3270, opened 2022-06-14
Remove Support menu option in Docs 

tanberry:
Removing the Support menu option from the Docs top menu.

Fixes https://github.com/prisma/docs/issues/3289

PR #3300, opened 2022-06-20
redirects for moving support page and child pages

tanberry:
Adds redirects for the content that as under Support 
Fixes #3270  -->

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -366,7 +366,7 @@
 /docs/about/prisma/platform-releases     /docs/concepts/data-platform/about-platform/platform-releases  301!
 
 ## Moving Support section to Troubleshooting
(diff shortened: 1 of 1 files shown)


static/_redirects
```diff
@@ -365,6 +365,10 @@
 /docs/concepts/components/prisma-data-platform   /docs/concepts/data-platform  301!
 /docs/about/prisma/platform-releases     /docs/concepts/data-platform/about-platform/platform-releases  301!
 
+## Moving Support section to Troubleshooting
+/docs/support   /docs/guides/database/troubleshooting-orm/
```
janpio:
```suggestion
/docs/support   /docs/guides/database/troubleshooting-orm
```

tanberry: thanks, just changed it!

PR #3301, opened 2022-06-21
Fix typo in Custom Models documentation

vramana:
## Describe this PR

`prisma` variable doesn't exist in the scope of the function. `prismaUser` is probably what the author intended here.


meletj:
This was fixed by a separate update to the Docs. 
Thanks @vramana for flagging this up.

PR #3309, opened 2022-06-22
Update index.mdx

### LINKED ISSUE #3308: Style guide: add deprecation note
In our public style guide, add a note about deprecation notices.

When we're explaining that a feature is deprecated, we should NOT mention the planned version that it will be removed. This is to help keep the docs maintainable.

Update: this is now done and merged.

andrew-walford-prisma:
Style guide: added note about version numbers in deprecation notices


tanberry (approved):
Thanks so much Andrew!

PR #3310, opened 2022-06-22
Update cascading deletes docs

pscheid2:
Some changes to the cascading deletes docs to make the note about using referential actions more prominent.

When adding cascading deletes to my prisma project I scanned past the existing warning, I'm hoping this change will make the new referential actions option more prominent to other users.


keerlu:
Thank you for the PR, and sorry for the very delayed reply.

We have now removed the outdated cascading deletes guides and redirected to referential actions, and added links to referential actions where relevant in our other docs (#3445 and #3989)

PR #3315, opened 2022-06-23
Update 05-sqlite.mdx

### LINKED ISSUE #3314: Add note about rounding errors in SQLite
Add a note about the following. The best place for this is probably in the [connector page](https://www.prisma.io/docs/concepts/database-connectors/sqlite). We can add an "other limitations" section, as with the other connector pages.

---

SQLite is a loosely-typed database. If the Prisma Schema has a field with `count Int`, while Prisma will prevent you from inserting a value larger than an integer, nothing prevents the database from directly accepting a bigger number. These manually inserted big numbers cause rounding errors when queried. 

Now we’ll check numbers on the way out of the database to verify that this fit within the boundaries of an integer.

This might come out - there's some discussion in the [Prisma 4 release guide](https://www.notion.so/prismaio/Prisma-4-Release-document-e0864f1024154377bce9f0330d1523ad)

Done and merged

andrew-walford-prisma:
Draft written, but there are some questions to resolve

I've added a note to the SQLite connector doc about rounding errors, and how these are avoided in Prisma 4. However, there are a couple of questions to be resolved, which I've marked with !!! characters.

### DIFF
--- a/content/200-concepts/200-database-connectors/05-sqlite.mdx
+++ b/content/200-concepts/200-database-connectors/05-sqlite.mdx
@@ -48,7 +48,7 @@ The SQLite connector maps the [scalar types](/concepts/components/prisma-schema/
 
 SQLite is a loosely-typed database. If your Schema has a field with count `Int`, Prisma prevents you from inserting a value larger than an integer. However, nothing prevents the database from directly accepting a bigger number. These manually inserted big numbers cause rounding errors when queried. To avoid this problem, Prisma 4.0.0 and later checks numbers on the way out of the database to verify that they fit within the boundaries of an integer.
 
-**!!!What happens if they don't - what error does it throw? !!!**
+**!!!What happens if a number doesn't fit within the boundaries of an integer -  does it throw an error, or does it return a specific type of number, etc. - what happens? !!!**
 **!!! When we say "on the way out" do we mean when queried? !!!**
 
 ## Connection details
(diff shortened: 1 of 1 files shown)


content/200-concepts/200-database-connectors/05-sqlite.mdx
```diff
@@ -42,7 +42,14 @@ The SQLite connector maps the [scalar types](/concepts/components/prisma-schema/
...
 | `Decimal`  | `DECIMAL`     |
 | `Json`     | Not supported |
-| `Bytes`    | `BLOB`       |
+| `Bytes`    | `BLOB`        |
+
+## Rounding errors on big numbers
+
+SQLite is a loosely-typed database. If your Schema has a field with count `Int`, Prisma prevents you from inserting a value larger than an integer. However, nothing prevents the database from directly accepting a bigger number. These manually inserted big numbers cause rounding errors when queried. To avoid this problem, Prisma 4.0.0 and later checks numbers on the way out of the database to verify that they fit within the boundaries of an integer.
+
+**!!!What happens if they don't - what error does it throw? !!!**
```
andrew-walford-prisma: What I mean here is: 

What happens if a big number doesn't fit within the boundaries of an integer

content/200-concepts/200-database-connectors/05-sqlite.mdx
```diff
@@ -74,3 +86,7 @@ datasource db {
   url      = "file:/Users/janedoe/dev.db"
 }
 ```
+
+```
+
+```
```
janpio, line 92: This seems leftover and should probably not be here.

andrew-walford-prisma: Thanks. I'm doing some post-release fixes, so I'll address this.

PR #3316, opened 2022-06-23
Update styling on AdmotionWrapper

### LINKED ISSUE #3241: Please investigate admonition bolding bug - DO NOT MERGE
Note: At first, I thought this issue was due to the presence of code blocks in the admonition. However, my experiments show that this is triggered by multiple paragraphs in the admonition. I'm guessing that this is so common that this has already been investigated! Apologies if that's the case.

**Problem**
Unwanted bold text in admonitions

**Example**
See the two admonitions under the heading [Example of problem](https://deploy-preview-3242--prisma2-docs.netlify.app/docs/concepts/data-platform/data-proxy#example-of-problem).

**Steps to reproduce**
1. Add an admonition to a docs page.
2. Inside the admonition, include more than one paragraph.

All of the text in the admonition is bolded.

carlagn:
## Describe this PR

When giving an `Admotion` component 2 lines of text, text turns bold when it shouldn't.

## Changes

- Update css styling on `AdmotionWrapper`

## What issue does this fix?

[Fixes #3241](https://github.com/prisma/docs/issues/3241)

N/A

### DIFF
--- a/src/components/customMdx/admonition.tsx
+++ b/src/components/customMdx/admonition.tsx
@@ -50,7 +50,6 @@ const ChildContainer = styled.div`
 const AdmonitionWrapper = styled.span<{ type?: string }>`
   font-family: Inter;
   font-style: normal;
-  font-weight: 600;
   font-size: 16px;
   line-height: 24px;
   color: ${theme.colors.gray600} !important;


janpio (approved):
Great change. The Percy only showed me in how many location that was popping up. Much more consistent now.

PR #3321, opened 2022-06-24
Move PDP to a top-level section

zachtil:
## Describe this PR

Move PDP to a top-level section (in the second-level header of Prisma Docs)

## Changes

* Move all PDP pages from /docs/concepts/data-platform to /docs/data-platform
* Redirects for the above change


janpio:
(We could probably go ahead and remove the P1 docs link to make this feel less cramped. If that is desirable, let me know and I will see if this is ok already before we turn off P1 later this year.)

(I assume you are aware of the mismatch in "level" between PDP vs. other links in that navigation)

zachtil:
> (We could probably go ahead and remove the P1 docs link to make this feel less cramped. If that is desirable, let me know and I will see if this is ok already before we turn off P1 later this year.)
> 
> (I assume you are aware of the mismatch in "level" between PDP vs. other links in that navigation)

What are P1 links?

Yes, @tanberry and I talked about the mistmatch and the goal here is to make this less of a painful link migration as opposed to when the ORM gets there (and the rest of the info arch changes happen). By that time, PDP docs would have more content and therefore more links to migrate and redirect.

And also, it's important to make PDP more prominent so my understanding is that this is ok at the moment.

janpio:
> What are P1 links?

P1 = Prisma 1

image

zachtil:
Ok, got it.

I've shortened the bucket name to "Data Platform". It should look better now without having to consider removing the P1 link.

Does not resolve the mismatch, but it reduces the clutter.

2022-06-27-shortened-pdp-top-level

tanberry:
This is great, @Zach! I need a bit more time to step through each page; will do that today for sure, and then approve. Will be great to get this merged. I do agree with Jan that we can get rid of the link to Prisma 1 docs. I met with Design team last week, and we discussed that as well. So it will happen eventually.

On the page with the Errors, let's do try to provide a bot more info on each, link to docs where we can, etc. Low priority, though, I guess.

tanberry (approved):
All looks good to me, as far as moving the content. LGTM!

content/200-concepts/100-components/250-preview-features/050-client-preview-features.mdx
```diff
@@ -10,7 +10,7 @@ The following [Preview](/about/prisma/releases#preview) feature flags are availa
 - [`interactiveTransactions`](/concepts/components/prisma-client/transactions#interactive-transactions-in-preview) since release [2.29.0](https://github.com/prisma/prisma/releases/tag/2.29.0) - [Submit feedback](https://github.com/prisma/prisma/issues/8664)
 - [`fullTextSearch`](/concepts/components/prisma-client/full-text-search) since release [2.30.0](https://github.com/prisma/prisma/releases/tag/2.30.0) - [Submit feedback](https://github.com/prisma/prisma/issues/8877)
 - [`referentialIntegrity`](/concepts/components/prisma-schema/relations/referential-integrity) since release [3.1.1](https://github.com/prisma/prisma/releases/tag/3.1.1) - [Submit feedback](https://github.com/prisma/prisma/issues/9380)
-- [`dataProxy`](/concepts/data-platform/data-proxy) since release [3.3.0](https://github.com/prisma/prisma/releases/tag/3.3.0) - [Submit feedback](https://github.com/prisma/prisma/issues/9841)
+- [`dataProxy`](/data-platform/data-proxy) since release [3.3.0](https://github.com/prisma/prisma/releases/tag/3.3.0) - [Submit feedback](https://github.com/prisma/prisma/issues/9841)
```
janpio: This is GA, so should technically already not be there any more, right?

static/_redirects
```diff
@@ -369,6 +369,10 @@
 /docs/support   /docs/guides/database/troubleshooting-orm
 /docs/support/*   /docs/guides/database/troubleshooting-orm/:splat 301!
 
+## Moving PDP as a top-level section
+
+/docs/concepts/data-platform     /docs/data-platform
```
janpio: I think the first line might be missing the `301!`.
Line 369 probably also has the same problem (where I guess this is modelled after)

zachtil: Right. I've added `301!` to both now.

content/200-concepts/100-components/250-preview-features/050-client-preview-features.mdx
zachtil: It was just added with the push of P4 docs but I agree. It is now removed.

content/200-concepts/100-components/03-prisma-migrate/index.mdx
```diff
@@ -67,14 +67,14 @@ To get started with Prisma Migrate in a development environment:
 
 1. Create the first migration:
 
-   <CodeWithResult showText="Show migration SQL" hideText="Hide migration SQL">
+      <CodeWithResult showText="Show migration SQL" hideText="Hide migration SQL">
```
janpio, line 70:
```suggestion
   <CodeWithResult showText="Show migration SQL" hideText="Hide migration SQL">
```

PR #3322, opened 2022-06-24
fix mongodb composite types example schema

TasinIshmam:
## Changes

- Change type of `price` from `Decimal` (unsupported) to `Float` in the example schema. 

## What issue does this fix?

The MongoDB connector does not support the `Decimal` type. So this example would previous raise an error. [Reference](https://www.prisma.io/docs/concepts/database-connectors/mongodb#mapping-from-prisma-to-mongodb-types-on-migration)

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/056-composite-types.mdx
@@ -38,7 +38,7 @@ We’ll use this schema for the examples that follow:
 model Product {
   id     String  @id @default(auto()) @map("_id") @db.ObjectId
   name   String  @unique
-  price  Decimal
+  price  Float
   colors Color[]
   sizes  Size[]
   photos Photo[]


tanberry:
Thanks @TasinIshmam , good catch, our docs clearly state that we don't currently support `Decimal`. And it sounds like we want to go with `Float` instead of `string`.. @matthewmueller can you give this a quick review, and then let's get it merged?

janpio:
Using `String` is our official suggestion: https://github.com/prisma/prisma/issues/12637#issue-1190313305 (also linked from the reference)

tanberry:
Thanks @janpio , I saw in the comment that `string` was the initital proposal, but further down in thread it seemed as though `Float` is more widely used...?

TasinIshmam:
To elaborate further, `Float` seems a bit more intuitive compared to `String` for a `price` field.

 For this particular page, I guess it's more about just having a functional schema to demonstrate composite types, perhaps it doesn't really matter which one we use. 

I can change it to `String` though, if you feel otherwise @janpio @matthewmueller @tanberry

tanberry:
I vote we leave it as `Float for now, til we hear from Matt. Go ahead and merge it if you are still awake, @Tasin... or I can in a bit if I see you haven't yet. I just want to get rid of the `Decimal` asap. ;-)

tanberry:
Let's go ahead and merge this @TasinIshmam ! If you are still up please do so, otherwise I'll check back in a bit and do it. Thank you!

janpio:
From a technical perspective it is dangerous to represent prices as Floats as doing math with them introduces nice bugs that make you loose money. Random Google result on the topic: https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio

PR #3326, opened 2022-06-25
fix: add missing awaits in pagination page

TaiwoO:
"await" is missing from several of the example code blocks on this page: https://www.prisma.io/docs/concepts/components/prisma-client/pagination#example-filtering-and-cursor-based-pagination

## Changes

- added awaits to missing example blocks


janpio (approved):

tanberry:
Thanks so much @TaiwoO for catching that... the `await` part is kinda important! :-)

PR #3341, opened 2022-06-29
3333 p4 post publish hotfixes

### LINKED ISSUE #3333: Andrew's Prisma 4 docs hotfixes - things to fix post publish
In database connectors>sqlite:

"If the Prisma Schema has a field with count Int". 

What to fix:

From Alexey:
"I think count is the name of an example field here? I agree it’s confusing, it should be either “field like count Int", or, better yet, “field of type Int"."

Update: on checking, I've found that I already fixed this myself before Alexey replied, and I went with the wording that he suggested, so no change is needed here.

andrew-walford-prisma:
This is a couple of post-publish hotfixes:

- SqLite page: removed a stray code block
- Prisma 4 upgrade guide: Removed the explicit reference to our minimum Node.js supported version, and emphasized the link to our minimum requirements page

This is done and merged

### DIFF
--- a/content/300-guides/300-upgrade-guides/200-upgrading-versions/030-upgrading-to-prisma-4.mdx
+++ b/content/300-guides/300-upgrade-guides/200-upgrading-versions/030-upgrading-to-prisma-4.mdx
@@ -52,9 +52,7 @@ This section includes changes that affect both the Prisma Schema and the Prisma 
 
 #### Node.js minimum version change
 
-From Prisma version 4.0.0, the minimum version of Node.js that we support is 14.17.x. If you use an earlier version of Node.js, you will need to update it.
-
-See our [system requirements](/reference/system-requirements) for all minimum version requirements.
+In Prisma version 4.0.0, we've changed the [minimum version of Node.js that we support](/reference/system-requirements). If you use an earlier version of Node.js, you will need to update it.
 
 ### Schema changes
 
--- a/content/200-concepts/200-database-connectors/05-sqlite.mdx
+++ b/content/200-concepts/200-database-connectors/05-sqlite.mdx
@@ -86,7 +86,3 @@ datasource db {
   url      = "file:/Users/janedoe/dev.db"
 }
 ```
-
-```
-
-```
--- a/content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/08-upgrade-from-mongodb-beta.mdx
+++ b/content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/08-upgrade-from-mongodb-beta.mdx
@@ -20,8 +20,8 @@ We unfortunately can't cover all possible scenarios or changes required, but thi
 ## Requirements
 
 - Must be running MongoDB 4.2+ as a replica set (MongoDB Atlas does this for you automatically)
-- Node.js 14.17.x or later, or Node.js 16.x or later
-- TypeScript 4.1.x or later
+- Node.js: see [system requirements](/reference/system-requirements)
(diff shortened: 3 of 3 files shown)


keerlu:
I think I'm fine with this, but just want to understand the reason for removing the specific version number? Is that so we can change it later without having to remember to change it here too?

content/300-guides/300-upgrade-guides/200-upgrading-versions/030-upgrading-to-prisma-4.mdx
```diff
@@ -52,9 +52,7 @@ This section includes changes that affect both the Prisma Schema and the Prisma
 
 #### Node.js minimum version change
 
-From Prisma version 4.0.0, the minimum version of Node.js that we support is 14.17.x. If you use an earlier version of Node.js, you will need to update it.
```
keerlu, line 55: I think I'm fine with this, but just want to understand the reason for removing the specific version number? Is that so we can change it later without having to remember to change it here too?

andrew-walford-prisma: Yes. There's also another doc that I forgot to change - I'll update it a bit later.

keerlu: Ok. It adds a bit of indirection, and I don't think we would be changing this outside of major releases anyway, as it would be a breaking change. But I'm ok to approve this if you think it's standard practice.

keerlu (approved):

content/300-guides/300-upgrade-guides/200-upgrading-versions/030-upgrading-to-prisma-4.mdx
janpio: Oh sorry if I triggered that, I really just meant in the `content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/08-upgrade-from-mongodb-beta.mdx` (Edit below, so that one is fine). Here the specific version that this was updated to in Prisma 4.0.0 is ok to mention and would stay unchanged in future major versions.

andrew-walford-prisma: Yeah, I was in two minds about whether to apply it here. In the end I went with docs maintainability. However, I doubt we would let version numbers go stale, because when they change, we can do a global find in the docs source files.

janpio: The version number that was made a requirement for Prisma 4 will not change, that is my point. We are also not adding all the additional P4 upgrade steps to the P3 upgrade guide etc as that really just describes the step from 2.x to 3.x. And this one here describe from 2.x or 3.x to 4.x - and for that the requirement will always and forever be `14.17.x` as we will only be able to change that requirement to a higher version with Prisma 5.x.

andrew-walford-prisma: Ah OK - so if the requirement won't change before Prisma 5 then I think I should reverse this change. It would be better if the upgrade guide is self-contained.

andrew-walford-prisma: I've reverted this change in PR [3374](https://github.com/prisma/docs/pull/3374)

PR #3366, opened 2022-07-01
Move PDP URLs from 'about-platform' to 'about'

zachtil:
## Describe this PR

The initial data-platform/about-platform URLs use twice the term platform which is easily avoided. However, it includes 3 changes:

* Changing the MDX file names
* Fixing links to /about-platform to now use /about
* Adding redirects (sigh)

## Changes

As described above.


janpio:
Technically the "platforms-releases" and "platform-limits-and-status" pages still duplicate "platform" 😆

zachtil:
> Technically the "platforms-releases" and "platform-limits-and-status" pages still duplicate "platform" 😆

Correct. So I've fixed those too. And added yet more redirects.

static/_redirects
```diff
@@ -374,6 +374,14 @@
 /docs/concepts/data-platform     /docs/data-platform   301!
 /docs/concepts/data-platform/*   /docs/data-platform/:splat 301!
 
+## Rename PDP About section
+
+/docs/data-platform/about-platform     /docs/data-platform/about   301!
+/docs/data-platform/about-platform/*   /docs/data-platform/about/:splat 301!
+
+/docs/data-platform/about-platform/platform-limits-and-status     /docs/data-platform/about/limits-and-status   301!
+/docs/data-platform/about-platform/platform-releases              /docs/data-platform/about/releases   301!
```
janpio: These two should never become active, as line 380 will catch these and redirect to `/docs/data-platform/about/:splat`.

So probably you want:
```suggestion
/docs/data-platform/about/platform-limits-and-status     /docs/data-platform/about/limits-and-status   301!
/docs/data-platform/about/platform-releases              /docs/data-platform/about/releases   301!
```

zachtil: Thanks for catching this, Jan. That makes perfect sense.

tanberry (approved):

janpio (approved):

PR #3372, opened 2022-07-05
Prisma schema => Prisma Schema Language

janpio:


tanberry:
@janpio I do agree we need to add a sentence or two about PSL near the top of this topic, and possibly a whole paragraph or even child page about PSL. But here in the topic, I don't think we should change the title. The topic is really about the schema file, and the various parts of it; all of which of course are using the PSL... but the topic is about the file itself. 

Let me know your thoughts about us adding a section specifically about the PSL, and helping people grok that the PSL is what is used to create the schema file?

janpio:
Again, just a draft PR not ready for review or comments 😆 

The plan here was indeed to test out and potentially propose to make the main page (this one) about PSL, and split the file stuff into a sub page "Prisma Schema file". Only "Naming" and "Prisma schema file location" are really about the file - everything else is about the "Schema" as a higher level concept, which is better described with PSL in my opinion.

Wdyt?

content/200-concepts/100-components/01-prisma-schema/index.mdx
```diff
@@ -1,7 +1,7 @@
 ---
-title: 'Prisma schema'
-metaTitle: 'Prisma schema (Reference)'
-metaDescription: 'The Prisma schema is the main configuration file when using Prisma. It is typically called schema.prisma and contains your database connection and data model.'
+title: 'Prisma Schema Language'
+metaTitle: 'Prisma Schema Language (Reference)'
+metaDescription: 'TODO The Prisma schema is the main configuration file when using Prisma. It is typically called schema.prisma and contains your database connection and data model.'
```
janpio, line 4: TODO Update to match new title

janpio:
I think making PSL more explicit might help understanding some things here better, and also keep away the things that are actually about the Schema file and similar.

nikolasburk:
I would actually keep it as "Prisma schema" but don't have a strong opinion. Let's defer this one.

PR #3386, opened 2022-07-06
fix(referential-actions): Rewrite section about old behavior

janpio:
Seems we forgot to rewrite this section when Prisma 3.0.0 happened.


content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -476,13 +476,13 @@ main()
   })
 ```
 
-## Behavior without using <inlinecode>referentialActions</inlinecode>
+## Referential Actions Behavior before Prisma 3.0.0
```
tanberry:
```suggestion
## Referential actions behavior before Prisma 3.0.0
```

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -476,13 +476,13 @@ main()
...
 
-<a id="current"></a>
+> Warning: The following section is only relevant if you are using Prisma older than 3.0.0 (to be more precise: For all uses of Prisma 2.25.0 and earlier, and 2.26.0 until 3.0.0 _without_ the `referentialActions` preview flag). You can completely disregard it if you are using Prisma 3.0.0 or newer.
```
tanberry:
```suggestion
> Warning: The following section is only relevant if you are using Prisma older than 3.0.0 (to be more precise: For all uses of Prisma 2.25.0 and earlier, and 2.26.0 until 3.0.0 _without_ the `referentialActions` preview flag). You can completely disregard it if you are using Prisma 3.0.0 or later. 
```

tanberry (changes requested):
Thanks so much for catching this @janpio !

janpio:
Note that I started the PR, but it is still in draft mode so no need to waste time on reviewing technically - will try to incorporate the suggestions though when I next take a look at this.

janpio:
If the old information is still here as in the before of this PR, then it should be changed similar to how this PR did it.

nikolasburk:
This one seems hairy to resolve, just spent a bit of time comparing old and new versions. I'd actually close it because it it's about behaviour of Prisma 2.X which I don't think we should spent time on figuring out. (I'd also be ok deleting that entire section, I think in other places we've removed references to how things worked in v2.X already.)

nikolasburk:
Closing this since we decided to delete the entire section about behaviour pre v3: https://github.com/prisma/docs/pull/5497

Issue #380, opened 2020-05-26
Linkable search results


janpio:
Sometimes I want to link to the search results page in Slack. It would nice if the URL would be updated to reflect the current search term.

Jolg42:
This would be great for SEO as we could let Google know that we have the feature like this: 
```html
<script type="application/ld+json">
  {
    "@context": "http://schema.org",
    "@type": "WebSite",
    "url": "https://www.prisma.io/docs/",
    "potentialAction": {
      "@type": "SearchAction",
      "target": "https://www.prisma.io/search?q={search_term_string}",
      "query-input": "required name=search_term_string"
    }
  }
</script>
```

janpio:
Still very much needed.

nilubava:
@janpio Check out this new PR: #3556

PR #3398, opened 2022-07-08
feat: Linkable search results

nilubava:
#380

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "prism-react-renderer": "1.2.1",
     "prismjs": "1.25.0",
     "prop-types": "15.7.2",
+    "qs": "^6.11.0",
     "react": "17.0.2",
     "react-copy-to-clipboard": "5.0.4",
     "react-dom": "17.0.2",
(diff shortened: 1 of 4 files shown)


janpio:
Turned into draft as currently does not work, so not reviewable.

(Using search redirects to https://deploy-preview-3398--prisma2-docs.netlify.app/docs/docs/?query=test&page=1 which is a white page for me)

nilubava:
@janpio The link you mentioned above has 2 `/docs` strings. Try this instead: https://deploy-preview-3398--prisma2-docs.netlify.app/docs/?query=test&page=1

nilubava:
@janpio Can you please review the search behaviour again? The results wont be available since we do not index on branch PRs

janpio:
- Searching now changes the URL to something that could be linked to 👍 
- Opening that URL in a new tab also starts the search again 🥳 
- Clsoing the search windows does not change the URL back though
- I also see no search results at all (which I am not sure should be the case - last I tested this I just saw the same results as on the live site but search generally worked)

nilubava:
@janpio Can you test again and let me know if its good to go?

janpio:
Search does not work for me at all on https://docs-git-linkable-search-prisma.vercel.app/ unfortunately.
The header also looks broken.

PR #3402, opened 2022-07-09
Adding a note about using different port for pgbouncer=true which connects to pooling endpoint

Korayem:
## Describe this PR

Adding a note about checking port number with pgbouncer
The lack of info cost me some downtime because I thought just adding `?pgbouncer=true` will magically work with edge functions connected to supabase so I wish others won't face the same issue

## Changes

Mention checking port number for changes

https://supabase.com/docs/guides/database/connecting-to-postgres#finding-the-connection-pool-config


tanberry:
Hi @Korayem thank you so. much for adding this helpful Note. I'm sorry you had to learn this the hard way, but we very much appreciate you saving your fellow community users from the same pain. 💜

PR #3405, opened 2022-07-11
Document enforced `@unique` attribute on 1:1 and 1:n relations

### LINKED ISSUE #3319: Enforce `@unique` attribute for MySQL and MongoDB 1:1 and 1:n relations
MySQL’s InnoDB has an extension to ANSI standard to allow foreign keys with the referenced side only holding an index, but not a unique constraint. We’ve allowed it previously due to it being possible to migrate, but it breaks our assumption of having the singular side always returning at most one record when joining.

We want to set the MySQL and MongoDB connectors to follow the validations we have for other connectors. This means the referenced side has to be (@)@unique or an (@)@id for the data model to validate.

keerlu:
## Describe this PR

- Add example where referenced field isn't the id
- Include admonition about MySQL behaviour

## What issue does this fix?

Fixes #3319

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
@@ -6,7 +6,13 @@ tocDepth: 2
 
 <TopBlock>
 
-One-to-one (1-1) relations refer to relations where at most **one** record can be connected on both sides of the relation. In the example below, there's one 1-1-relation between `User` and `Profile`:
+This page introduces one-to-one relations and explains how to use them in your Prisma schema.
+
+</TopBlock>
+
+## Overview
+
+One-to-one (1-1) relations refer to relations where at most **one** record can be connected on both sides of the relation. In the example below, there is a one-to-one relation between `User` and `Profile`:
 
 <TabbedContent tabs={[<FileWithIcon text="Relational databases" icon="database"/>, <FileWithIcon text="MongoDB" icon="database"/>]}>
 <tab>
@@ -43,12 +49,10 @@ model Profile {
 </tab>
 </TabbedContent>
 
-The `userId` relation scalar is a direct representation of the foreign key in the underlying database.This 1-1-relation expresses the following:
+The `userId` relation scalar is a direct representation of the foreign key in the underlying database. This one-to-one relation expresses the following:
 
 - "a user can have zero or one profiles" (because the `profile` field is [optional](/concepts/components/prisma-schema/data-model#type-modifiers) on `User`)
 - "a profile must always be connected to one user"
-
-</TopBlock>
 
 ## Multi-field relations in relational databases
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
@@ -95,7 +95,7 @@ model Profile {
 
 <Admonition type="warning">
 
-In MySQL, it is possible to create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
+In MySQL, you can create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
 
 </Admonition>
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
@@ -95,7 +95,7 @@ model Profile {
 
 <Admonition type="warning">
 
-In MySQL, you can create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
+In MySQL, you can create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation of this type it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
 
 </Admonition>
 
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
@@ -54,7 +54,7 @@ The `userId` relation scalar is a direct representation of the foreign key in th
 - "a user can have zero or one profiles" (because the `profile` field is [optional](/concepts/components/prisma-schema/data-model#type-modifiers) on `User`)
 - "a profile must always be connected to one user"
 
-In the previous example, the `user` relation field of the `Profile` model references the `id` field of the `User` model. It is also possible to reference another field, as long as that field is marked with the `@unique` attribute. This is necessary to guarantee that there is only a single `User` connected to each `Profile`. In the following example, the `user` field references an `email` field in the `User` model, which is marked with the `@unique` attribute:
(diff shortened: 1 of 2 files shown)


content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
```diff
@@ -43,12 +49,55 @@ model Profile {
...
 </TabbedContent>
 
-The `userId` relation scalar is a direct representation of the foreign key in the underlying database.This 1-1-relation expresses the following:
+The `userId` relation scalar is a direct representation of the foreign key in the underlying database. This one-to-one relation expresses the following:
 
 - "a user can have zero or one profiles" (because the `profile` field is [optional](/concepts/components/prisma-schema/data-model#type-modifiers) on `User`)
 - "a profile must always be connected to one user"
 
-</TopBlock>
+In this example, the `user` field of the `Profile` model references the `id` field of the `User` model. It is also possible to reference another field, as long as that field is marked with the `@unique` attribute. This is necessary to guarantee that there is only a single `User` connected to each `Profile`. In the following example, the `user` field references an `email` field in the `User` model, which is marked with the `@unique` attribute:
```
janpio:
```suggestion
In the previous example, the `user` relation field of the `Profile` model references the `id` field of the `User` model. It is also possible to reference another field, as long as that field is marked with the `@unique` attribute. This is necessary to guarantee that there is only a single `User` connected to each `Profile`. In the following example, the `user` field references an `email` field in the `User` model, which is marked with the `@unique` attribute:
```

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
```diff
@@ -45,12 +51,55 @@ model Post {
...
 > **Note** The `posts` field does not "manifest" in the underlying database schema. On the other side of the relation, the [annotated relation field](/concepts/components/prisma-schema/relations#annotated-relation-fields-and-relation-scalar-fields) `author` and its relation scalar `authorId` represent the side of the relation that stores the foreign key in the underlying database.
 
-This 1-n-relation expresses the following:
+This one-to-many relation expresses the following:
 
 - "a user can have zero or more posts"
 - "a post must always have an author"
 
-</TopBlock>
+In this example, the `author` field of the `Post` model references the `id` field of the `User` model. It is also possible to reference another field, as long as that field is marked with the `@unique` attribute. This is necessary to guarantee that there is only a single `User` connected to each `Post`. In the following example, the `author` field references an `email` field in the `User` model, which is marked with the `@unique` attribute:
```
janpio:
```suggestion
In the previous example, the `author` relation field of the `Post` model references the `id` field of the `User` model. It is also possible to reference another field, as long as that field is marked with the `@unique` attribute. This is necessary to guarantee that there is only a single `User` connected to each `Post`. In the following example, the `author` field references an `email` field in the `User` model, which is marked with the `@unique` attribute:
```

janpio (approved):

content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
```diff
@@ -43,12 +49,55 @@ model Profile {
 </tab>
 </TabbedContent>
 
-The `userId` relation scalar is a direct representation of the foreign key in the underlying database.This 1-1-relation expresses the following:
+The `userId` relation scalar is a direct representation of the foreign key in the underlying database. This one-to-one relation expresses the following:
```
andrew-walford-prisma, line 52:
```suggestion
The `userId` relation scalar is a direct representation of the foreign key in the underlying database. This one-to-one relation expresses the following:
```
```suggestion
The `userId` relation scalar is a direct representation of the foreign key in the underlying database. This one-to-one relation expresses the following:
```

In light of our team discussion earlier, consider adding a link to the definition of scalar (data model page) on the first mention of the word on this page.

content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
andrew-walford-prisma: Consider making "It is also possible to" more active. E.g. "You can also"

For "as long as" consider to changing to "when" or "if". "As long as" might suggest a duration.

content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
```diff
@@ -43,12 +49,55 @@ model Profile {
...
+  userId String @unique @db.ObjectId // relation scalar field (used in the `@relation` attribute above)
+}
+```
+
+</tab>
+</TabbedContent>
+
+<Admonition type="warning">
+
+In MySQL, it is possible to create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
```
andrew-walford-prisma:
```suggestion
In MySQL, you can create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
```

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
andrew-walford-prisma: See my comment above about the similar paragraph on the other page.

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
```diff
@@ -45,12 +51,55 @@ model Post {
...
+  author   User @relation(fields: [authorId], references: [email])
+}
+```
+
+</tab>
+</TabbedContent>
+
+<Admonition type="warning">
+
+In MySQL, it is possible to create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
```
andrew-walford-prisma:
```suggestion
In MySQL, you can create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
```

andrew-walford-prisma:
```suggestion
In MySQL, it is possible to create a foreign key with only an index on the referenced side, and not a unique constraint. In Prisma versions 4.0.0 and later, if you introspect a relation like this it will trigger a validation error. To fix this, you will need to add a `@unique` constraint to the referenced field.
```

By "introspect a relation like this", do you mean "introspect a relation of this type"? Or do you mean "introspect a relation in this way"? Can this be clarified?

andrew-walford-prisma (commented):
A few suggestions and questions, but it otherwise LGTM!

content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
keerlu: Thanks, made both these changes

content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
keerlu: Hm... I don't think the general data model definition of scalar fields would be a very helpful link, it would need to be to the relation scalar field specifically. The best place for that is [the overall Relations page](https://www.prisma.io/docs/concepts/components/prisma-schema/relations) that this is a subpage of, which is unfortunately not amazingly organised at the moment, so there isn't a clear link point. Will leave for now.

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
keerlu: Done

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
keerlu: Thanks, changed to 'of this type' here and on the other page. Agree this is much better!

PR #3421, opened 2022-07-14
Encourage using UUID primary keys in CockroachDB

kernfeld-cockroach:
Randomly-generated identifiers like UUIDs provide more even distribution of data to prevent hotspots in distributed databases like CockroachDB.

### DIFF
--- a/content/300-guides/050-database/860-using-prisma-with-cockroachdb.mdx
+++ b/content/300-guides/050-database/860-using-prisma-with-cockroachdb.mdx
@@ -74,7 +74,7 @@ For a full list of type mappings, see our [connector documentation](/concepts/da
 
 When generating unique identifiers for records in a distributed database like CockroachDB, it is best to use randomly generated IDs such as UUIDs rather than sequential IDs – for more information on this, see CockroachDB's [blog post on choosing index keys](https://cockroachlabs.com/blog/how-to-choose-db-index-keys).
 
-To generate random UUIDs in Prisma, Prisma's [`dbgenerated()`](/reference/api-reference/prisma-schema-reference#dbgenerated) attribute function can be used in conjunction with CockroachDB's [`gen_random_uuid()` function](https://www.cockroachlabs.com/docs/stable/serial.html). For example, the following `User` model has an `id` primary key, generated using the `dbgenerated()` function:
+To generate random UUIDs in Prisma, Prisma's [`dbgenerated()`](/reference/api-reference/prisma-schema-reference#dbgenerated) attribute function can be used in conjunction with CockroachDB's [`gen_random_uuid()` function](https://www.cockroachlabs.com/docs/v22.1/uuid.html). For example, the following `User` model has an `id` primary key, generated using the `dbgenerated()` function:
 
 ```prisma file=schema.prisma
 model User {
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks so much @kernfeld-cockroach for raising this. What would you think of leaving your changes as they are (i.e. making that the default suggestion), but also adding a sentence or two below explaining that either can be used, but that for CockroachDB specifically, UUIDs scatter the keys and prevent the hotspots? In other words, taking the opportunity to teach a bit more.
I found this post interesting, and it supports your statement: https://news.ycombinator.com/item?id=14524174

Let me know your thoughts, and also in the meantime I'll ask for a technical review on this PR. Thanks again!

tanberry:
Hi again @kernfeld-cockroach. After a chat with our dev team, it seems we are indeed working on a unified approach for using UUIDs. We aren't keen on suggesting the use of  `dbgenerated()` here, and using integers does work, so for now we plan to only edit our doc to add a link to the CockroachDB documentation on using UUIDs and let users learn more there. As soon as we have defined more of how we want to suggest using UUIDs (hopefully within next several sprints) we can edit this doc to reflect those decisions.

I'm going to go ahead and close this PR, but feel free to reopen it if you need... or, if after we edit the doc, open a new PR with any additional suggestions. Also, please consider joining our shared channel, [#shared-cockroach](https://prisma-company.slack.com/archives/C01CSQXTNJ3); would be great to have your thoughts there. Thanks again for raising this issue; and good that it is already on our dev team's radar.

content/300-guides/050-database/860-using-prisma-with-cockroachdb.mdx
```diff
@@ -72,18 +72,18 @@ For a full list of type mappings, see our [connector documentation](/concepts/da
 
 ### How to use database keys with CockroachDB
 
-When generating unique identifiers for records in a distributed database like CockroachDB, it is best to avoid using sequential IDs – for more information on this, see CockroachDB's [blog post on choosing index keys](https://cockroachlabs.com/blog/how-to-choose-db-index-keys).
+When generating unique identifiers for records in a distributed database like CockroachDB, it is best to use randomly generated IDs such as UUIDs rather than sequential IDs – for more information on this, see CockroachDB's [blog post on choosing index keys](https://cockroachlabs.com/blog/how-to-choose-db-index-keys).
```
tanberry:
```suggestion
When generating unique identifiers for records in a distributed database like CockroachDB, it is best to avoid using sequential IDs – for more information on this, see CockroachDB's [blog post on choosing index keys](https://cockroachlabs.com/blog/how-to-choose-db-index-keys). Consider using UUIDs if you want to explicitly plan for avoiding hotspots, where all writes are trying to go the same place.
```

tanberry: OK, sorry for my thrashing on this @kernfeld-cockroach . I think a good solution wold be for you to please accept the one suggestion I made (a slight tweak on your edit), but then remove your other changes. Then we can merge this, leaving our current code examples as they were, but adding in your prompt about using UUIDs. 

Let me know if that works for you; I am going to step away from this now, for a while. ;-)

kernfeld-cockroach: Advising users to do the correct thing is a good start, but we should strive to make it easy for developers to use UUIDs correctly by providing sample code that they can copy and paste.

Although it's true that many applications will never reach the scale where they need to care about UUIDs, the applications that do reach that scale tend to be the most important and influential.

andrew-walford-prisma:
@tanberry  From your previous comment, it looks like we can this PR. Shall we go ahead and do that?

PR #3426, opened 2022-07-15
Clarify migrate resolve can only be used with failed migrations

### LINKED ISSUE #3420: Clarify `migrate resolve --rolled-back` only works with failed migrations
API Reference: https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-resolve

The `prisma migrate resolve --rolled-back <migration_name>` only works with failed migrations. When run against a database with no migration history issues, it leads to the following error:

```
❯ npx prisma migrate resolve --rolled-back 20220714121901_add_profile
Datasource "db": PostgreSQL database "schema-prototyping", schema "public" at "localhost:5435"
Error: P3012

Migration `20220714121901_add_profile` cannot be rolled back because it is not in a failed state.
```

keerlu:
## What issue does this fix?

Fixes #3420

### DIFF
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -913,7 +913,9 @@ prisma migrate deploy
 
 ### <inlinecode>migrate resolve</inlinecode>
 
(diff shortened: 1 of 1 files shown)


janpio (approved):

tanberry (approved):
thanks @keerlu !

PR #3429, opened 2022-07-17
Fix fragment for `seed` command

baldyeagle:
## Describe this PR

The link fragment is currently `db-seed-preview`, which does not exist on the command reference page.

## Changes

This updates the fragment to `db-seed`.

## What issue does this fix?

N/A, this was such a small change an issue seems like too much

N/A


baldyeagle:
@all-contributors please add @baldyeagle for doc

tanberry:
Thanks so much @baldyeagle for catching that! Looks like we failed to update the url when when we moved `db seed` from Preview to GA.

PR #3445, opened 2022-07-19
Cascading deletes: clarify the version differences and direct users on newer versions to referential actions

### LINKED ISSUE #2644: Consider removing the "Cascading deletes"-guides
We have these guides in our docs:
- https://www.prisma.io/docs/guides/database/advanced-database-tasks/cascading-deletes/postgresql
- https://www.prisma.io/docs/guides/database/advanced-database-tasks/cascading-deletes/mysql
- https://www.prisma.io/docs/guides/database/advanced-database-tasks/cascading-deletes/sqlite

I would consider removing them (or maybe moving them to the [Data Guide](https://www.prisma.io/dataguide/)) for these reasons:
- These guides were written because Prisma didn't support cascading deletes natively so we needed a resource to point people to if they still wanted to use them that explained a workaround with plain SQL, now that we have proper docs, these workarounds don't seem needed any more.
- Most users won't think of this feature as "referential actions" but as "_cascading_ deletes/updates"; when searching for _cascading_ in the docs, the first results that come up are the guides and not the new docs pages

Andrew:
I'm in favour of completely removing the pages, from the principal of lean docs. If Prisma now does this natively, then I don't think we need to document workarounds.

But we do have links to them that we'll need to remove. I need to check that it's OK to remove each of these. In particular, it might be that this workaround is still relevant to earlier versions of Prisma, so we might actually want to keep the pages for now. We could also move the pages to the data guide, as Nikolas suggests, and change the links to point there. I will need to think about/investigate this.

References to these pages in our docs:
* [concepts/components/prisma-schema/relations/referential-actions/_index (two mentions)](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions#using-introspection
) - links from delete() and deleteAll() - can we remove?
   * Also links to: https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions#how-to-use-cascading-deletes

* https://www.prisma.io/docs/concepts/components/prisma-migrate/legacy-migrate#supported-operations
        Link to workaround - can we remove it?

* https://www.prisma.io/docs/guides/performance-and-optimization/prisma-client-transactions-guide#are-cascading-deletes-supported-by-the-prisma-client-nested-writes

* https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3/referential-actions#using-introspection - same links as in first item in this list

Unambiguously OK to remove:
* _redirects (two mentions)

Current status: done and merged. I concluded that we couldn't remove these docs, but I've mitigated the issue by adding better signposting to the various pages involved.

andrew-walford-prisma:
Emphasize new method of doing cascading deletes

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
@@ -15,8 +15,8 @@ From version 2.26.0, you can define referential actions on the relation fields i
 
 **Version differences**
 
-- If you use version 3.0.0 or later, you can use referential actions as described on this page.
-- If you use a version between 2.26.0 and 2.30.3, you can use referential actions as described on this page, but you must [enable the preview feature flag](/concepts/components/preview-features/client-preview-features#enabling-a-prisma-client-preview-feature) `referentialActions`.
+- If you use version 3.0.1 or later, you can use referential actions as described on this page.
+- If you use a version between 2.26.0 and 3.0.0, you can use referential actions as described on this page, but you must [enable the preview feature flag](/concepts/components/preview-features/client-preview-features#enabling-a-prisma-client-preview-feature) `referentialActions`.
 - If you use version 2.25.0 or earlier, you can do cascading deletes [manually in the database](/guides/database/advanced-database-tasks/cascading-deletes).
 
 </Admonition>
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
@@ -7,7 +7,16 @@ tocDepth: 2
 
 <TopBlock>
 
-Referential actions determine what happens to a record when a related record is deleted or updated. In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later (before 3.0.0 with a preview feature flag `referentialActions`, afterwards without), you can define referential actions on your relation fields in your Prisma schema.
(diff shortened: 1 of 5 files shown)


content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
```diff
@@ -7,7 +7,19 @@ tocDepth: 2
 
 <TopBlock>
 
-Referential actions determine what happens to a record when a related record is deleted or updated. In [2.26.0](https://github.com/prisma/prisma/releases/tag/2.26.0) and later (before 3.0.0 with a preview feature flag `referentialActions`, afterwards without), you can define referential actions on your relation fields in your Prisma schema.
+Referential actions determine what happens to a record when your application deletes or updates a related record.
+
+From version 2.26.0, you can define referential actions on the relation fields in your Prisma schema. This allows you to do cascading deletes and cascading updates.
```
keerlu: Nitpicks, can be ignored:

Maybe better to point out that these are examples of what you can do with referential actions (as they are used for other things too)

Also should maybe be something like 'define cascading deletes and updates at a Prisma level', as they could already be done manually.

keerlu (approved):
LGTM!

content/200-concepts/100-components/01-prisma-schema/06-relations/425-referential-actions/index.mdx
andrew-walford-prisma: Thanks - I brought your suggestions into the doc.

andrew-walford-prisma: Thanks, I've brought these suggestions into the doc.

PR #3457, opened 2022-07-19
fix: first test case of 'with-singleton.ts'

Octobug:
The `expect()` part of test case 'should create new user' lacked `acceptTermsAndConditions: true` within the user object for comparison, which stopped it from passing the case.

## Describe this PR

The singleton demo of [Example unit tests](https://www.prisma.io/docs/guides/testing/unit-testing#example-unit-tests) cannot pass all test cases. And my modification fixes it.

## Changes

- Supplement `acceptTermsAndConditions: true,` to the comparison object of user.

## What issue does this fix?

I don't see any related issue. Should I open an issue first?


tanberry:
Ah, good catch, thanks so much @Octobug !

PR #3471, opened 2022-07-21
Add workaround for vercel pkg ENOENT error

### LINKED ISSUE #3455: Use root cwd if starting query engine from...
Details TBW

SME: Daniel

Eng ticket: https://github.com/prisma/prisma/pull/8634

Status: this is done and can be merged. Daniel and Lucy have both reviewed it.

Notes from Daniel:

This issue is to do with users who would like to use:
https://github.com/vercel/pkg

PKG will bundle your application into an exe to be used on Vercel. [People are having issues with the bundle step and the engines](https://github.com/prisma/prisma/issues/8449).  The docs bit should be to include the comment/workaround mentioned [here](https://github.com/prisma/prisma/issues/8449#issuecomment-1147743988). 

I suggest adding a header on this page: https://www.prisma.io/docs/guides/deployment/deployment-guides/deploying-to-vercel

Something like: ‘Using Vercel PKG’
Then add a small note that you need to add the workaround code to your package.json.

andrew-walford-prisma:
Added workaround to the Vercel deployment guide for ENOENT error


danstarns (approved):

content/300-guides/200-deployment/110-deployment-guides/300-deploying-to-vercel.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: 'Deploying to Vercel'
 metaTitle: 'How to deploy a Prisma app to Vercel'
-metaDescription: 'Learn how to deploy a Next.js app with a serverless REST API using Prisma Client to Vercel.'
+metaDescription: 'Learn how to deploy a Next.js app with a serverless REST API, with Prisma Client to Vercel.'
```
keerlu: I see the point of avoiding 'using', but this reads very oddly to me and took me a few goes to make sense of... sounds like 'Prisma Client to Vercel' is a product

Maybe 'Learn how to use Prisma Client to...' instead?

content/300-guides/200-deployment/110-deployment-guides/300-deploying-to-vercel.mdx
```diff
@@ -138,10 +138,28 @@ In order to enable Preview deployment for pull requests, create a separate datab
 
 > **Note**: because each pull request shares the same preview database, if you have multiple pull requests with diverging schema migrations, you will encounter problems applying them.
 
+## <inlinecode>vercel/pkg</inlinecode> <inlinecode>ENOENT</inlinecode> error
+
+If you use [vercel/pkg](https://github.com/vercel/pkg) to package your Node.js project, then you might encounter an error like the following at the bundle stage:
```
keerlu: Is the 'bundle stage' discussed elsewhere? Can't see it higher in the page...

keerlu (approved):

content/300-guides/200-deployment/110-deployment-guides/300-deploying-to-vercel.mdx
andrew-walford-prisma: Good call!

content/300-guides/200-deployment/110-deployment-guides/300-deploying-to-vercel.mdx
andrew-walford-prisma: Good point. Thinking about this some more, we just mean "when you use vercel/pkg to package your project". That's implicit in the text already, so I've trimmed it down.

janpio:
Where was this problem reported?

content/300-guides/200-deployment/110-deployment-guides/300-deploying-to-vercel.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: 'Deploying to Vercel'
 metaTitle: 'How to deploy a Prisma app to Vercel'
-metaDescription: 'Learn how to deploy a Next.js app with a serverless REST API using Prisma Client to Vercel.'
+metaDescription: 'Learn how to use Prisma Client to deploy a Next.js app with a serverless REST API to Vercel.'
```
janpio, line 4: This change sounds incorrect.

andrew-walford-prisma: Good spot. I've corrected this in the following PR:

[https://github.com/prisma/docs/pull/3572](https://github.com/prisma/docs/pull/3572)

content/300-guides/200-deployment/110-deployment-guides/300-deploying-to-vercel.mdx
```diff
@@ -138,10 +138,28 @@ In order to enable Preview deployment for pull requests, create a separate datab
...
+
+To avoid this error, add your Prisma query engine binary path to the `pkg/assets` section of your `package.json` file, as follows:
+
+```json file=package.json copy
+{
+  "pkg": {
+    "assets": ["node_modules/.prisma/client/*.node"]
+  }
+}
+```
```
janpio, line 157: What does this have to do with Vercel deployment?

`pkg` is a library that is created by the company Vercel. It has 0 connection with Vercel the deployment platform.

andrew-walford-prisma: As I understand it, `pkg` bundles your app into an exe to be used on Vercel. Isn't that part of the task of deploying to Vercel?

janpio: No, Vercel can not even run executable/binaries. They run Node apps directly.

This is a general purpose tool (only coincidentally from Vercel) to package any Node app into an executable/binary. We use(d) it to package our Prisma CLI into an executable so Go users (for our Prisma Client Go) can use the CLI without having to install Node. Some of our users are doing similar things with their Node apps, so the users can just download 1 file and run their app locally.

Nothing to do with Vercel deployment at all. It would fit more into a "Weird usage and deployment options of Prisma" page, and could sit next to a section about "Using Prisma in Electron apps" for example. I don't think we have something like that yet though 😢

PR #3503, opened 2022-07-23
Update 064-custom-models.mdx

muddledluck:
## Describe this PR

In extending Prisma client through function 
```return Object.assign(prisma, {``` should be ```return Object.assign(prisma.user, {```
because when we try to use `Users` like

```
const prisma = new PrismaClient();

const user = Users(prisma.user)

async function getUser(email){
     const userExists = user.findFirst({
        where: { email: email }
     })
}
```

gives error 
says 
```
Property 'findFirst' does not exist on type 'PrismaClient<PrismaClientOptions, never, RejectOnNotFound | RejectPerOperation | undefined> & { ...; }'
```

So, for this we need to update 
```return Object.assign(prisma, {```
to
```return Object.assign(prisma.user, {```

Otherwise, we should go like this

```
const prisma = new PrismaClient();

const user = Users(prisma.user)

async function getUser(email){
     const userExists = user.user.findFirst({
        where: { email: email }
     })
}

async function signUp(data){
     const userExists = user.signUp(data)
}
```

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/064-custom-models.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/064-custom-models.mdx
@@ -64,7 +64,7 @@ type Signup = {
 }
 
 function Users(prismaUser: PrismaClient['user']) {
-  return Object.assign(prisma, {
+  return Object.assign(prisma.user, {
     /**
      * Signup the first user and create a new team of one. Return the User with
      * a full name and without a password
--- a/content/200-concepts/100-components/02-prisma-client/064-custom-models.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/064-custom-models.mdx
@@ -64,7 +64,7 @@ type Signup = {
 }
 
 function Users(prismaUser: PrismaClient['user']) {
-  return Object.assign(prisma.user, {
+  return Object.assign(prismaUser, {
     /**
      * Signup the first user and create a new team of one. Return the User with
      * a full name and without a password

### CODE REVIEW

**Comment by muddledluck** [comment]
@all-contributors please add @muddledluck for doc

**Comment by tanberry** [comment]
Thanks so much for catching this @muddledluck ! We are going to run a quick tech review on it and get it merged (along with your contributors PR!).

**Review by millsp — APPROVED** [review]

File: content/200-concepts/100-components/02-prisma-client/064-custom-models.mdx [hunk 1]
```diff
@@ -64,7 +64,7 @@ type Signup = {
 }
 
 function Users(prismaUser: PrismaClient['user']) {
-  return Object.assign(prisma, {
+  return Object.assign(prisma.user, {
```
millsp, line 67:
```suggestion
  return Object.assign(prismaUser, {
```

millsp:
```suggestion
  return Object.assign(prismaUser, {
```

millsp (dismissed):
It will be better to keep the variable (thanks @janpio)

tanberry:
Thanks again @muddledluck !

muddledluck:
My pleasure @tanberry 😁

muddledluck:
one question, when does this change visible on https://www.prisma.io/docs/concepts/components/prisma-client/custom-models ?
@tanberry

PR #3505, opened 2022-07-24
Fix incorrect filename in getting-started/quickstart

exclowd:
## Describe this PR
There is an inconsistency in the filenames between the code-block and the text above in the getting started page. This pr fixes that. Since this is in the getting started page, this might throw off newbies.

## Changes
Changed the filename in the code-block to `script.ts`


tanberry:
Hi @exclowd thanks so much for catching this, and apologies that I took so long to get to it that a team member already fixed it, with PR https://github.com/prisma/docs/pull/3507.

Thank you, though, we appreciate the help!

PR #3519, opened 2022-07-28
feat: Update banner

nilubava:


src/components/banner.tsx
```diff
@@ -36,7 +36,7 @@ const Banner = () => (
   <BannerWrapper>
     <BannerText>
       <a href="/dataplatform">
```
nilubava: @michellegreer The link should be changed from `dataplatform` to `docs/data-platform/data-proxy` here in line 38

michellegreer (approved):

PR #3535, opened 2022-07-29
Fixes Links in "Constraint names in relational databases" section

nurul3101:
The current links does not open the correct section and instead open the footer of the page.

## Changes

Links are corrected to open the correct section.


tanberry:
Thank so much @nurul3101 !!

PR #3539, opened 2022-07-29
MongoDB connection strings

tanberry:
presented two diff connection string options (one for atlas and one for local instance of Mongo); added note about using `authSource=admin` with local instance

related to https://github.com/prisma/prisma/discussions/9994#discussioncomment-1562283


janpio:
We do not want this.

Issue #3453, opened 2022-07-19
Allow use of `findUnique()` with MongoDB composite types


andrew-walford-prisma:
Update 05 Sep 2022:

I accidentally merged this, so I reverted it, then made a new PR. The new PR is here: https://github.com/prisma/docs/pull/3722
This docs work is done, but we need to wait until the feature actually foes out. **We will need to update the version numbers in the text to [whatever version this goes into]**

SME: Cristian, Daniel

Cristian and Dan have reviewed it; however, we can't merge this, because there are issues with it. Also note that we still mention 4.3, and this didn't go out with that version, so we'll need to update it.

This is NOT a preview feature

- Eng ticket: https://github.com/prisma/prisma/issues/12409
- Notion discussion: https://www.notion.so/prismaio/Proposal-Composite-Field-Indices-03baf2ea1865417684fdab7fda2ee007
- See this section on Docs notes: https://www.notion.so/prismaio/Proposal-Composite-Field-Indices-03baf2ea1865417684fdab7fda2ee007 (the naming there is pretty convuluted)

- This change relates to MongoDB only
- It allows findUniques inside MongoDB embedded types

Notes:

Our docs [here](https://www.prisma.io/docs/concepts/components/prisma-schema/names-in-underlying-database#related-naming-indexes-and-primary-keys-for-prisma-client) currently giv an example of composite findUniques as they work now: 

```
prisma
  .user
  .findUnique({
    where: {
      firstName_lastName: {
      fullName: {
        firstName: "Paul",
        lastName: "Panther"
    }
  })
```

Incidentally, we should ensure we link to these examples from our main findUnique entry (if we don't already), because we don't go into this amount of detail there.

MongoDB has embedded documents. We currently don't support findUniques with them. With this new feature, composite findUniques work with MongoDB embedded types.

With this feature, you can findUnique on embedded documents, as follows:

In the schema:

```
@@unique([name, location.address])
@@unique([cost, location.address])
```

So if we want to find records with a unique attribute that is a combination of name, location.address and cost (a bit of a wacky example - we might be able to come up with a better one for the doc), our app code might be as follows:

````
findUniqueA(where: {
  name_location_address: {
    name: "",
    location: {
      address: ""
    }
  },
  cost_location_address: {
   cost: 1,
   location: {
     address: ""
    }
  }
})

wlans:
Is this PR going to go in? Considering Prima client for a project.

jharrell:
[...]

PR #3541, opened 2022-08-01
Fix broken example `findUnique` snippets on 'Names in the underlying database' page

keerlu:
## Describe this PR

I found a broken code snippet (mismatched brackets) while looking into #3453. Also tidied up the one below it. May as well fix these now instead of waiting for the release.


cprieto (approved):

content/200-concepts/100-components/01-prisma-schema/07-names-in-underlying-database.mdx
```diff
@@ -204,29 +203,25 @@ model User {
...
 Specifying `@@id([firstName, lastName], name: "fullName")` will change the Prisma Client API to this instead:
 
-
-```ts highlight=6;add|5;delete
-prisma
-  .user
-  .findUnique({
-    where: {
-      firstName_lastName: {
-      fullName: {
```
janpio, line 227: Was it intentional that this removed the add/delete syntax with a edit one?

PR #3567, opened 2022-08-05
Fix explicit m:n SQL

janpio:
The SQL was missing the actual primary key of `CategoriesOnPosts` and was using a unique instead.
Some of the fields on the models were also missing.

Additionally the SQL was not like what current Prisma Migrate would create.
Note that this is more verbose and might not be what we actually want to show here.
But at least this is _correct_ now.


tanberry:
Thanks @janpio . Do we need to simultaneously clean up the [diagram](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#:~:text=For%20relational%20databases%2C%20the%20following%20entity%20relationship%20diagram%20represents%20the%20database%20that%20corresponds%20to%20the%20sample%20Prisma%20schema%3A) you mentioned in [Slack](https://prisma-company.slack.com/archives/C5Z9TH6N9/p1659726540023929)? Or can they be separate improvements?

janpio:
This is completely independent from any messages in Slack.

tomhoule (approved):
I think the changes are correct :+1:

andrew-walford-prisma:
@keerlu This is a PR I'd otherwise merge, but it has a merge conflict with one of your docs, so I thought I'd leave this for you to decide on!

PR #3568, opened 2022-08-05
docs: clarify restriction on template variables in strings

LewisJEllis:
## Describe this PR

Adds clarification that this bit about "strings" means "SQL string literals".

## Changes

"strings" -> "SQL string literals"

## What issue does this fix?

I had to reread the line + code example a few times to infer/understand what it meant exactly, since the code contains both TS strings and SQL strings. The example just a little bit before this one appeared to use template variables inside a string, so the restriction appeared to be potentially contradictory at first.


matthewopave (approved):
nice change!

ObamaTheLlama114:
Thank you @LewisJEllis for the PR

PR #3574, opened 2022-08-08
Prepend npx to CLI reference commands

### LINKED ISSUE #3538: Prepend `npx` to our example commands
Since most of our users install Prisma locally (not globally), they will need to use `npx` with their commands.

Most of the commands in our CLI Reference section need `npx` prepended to them, and there are probably many scattered around elsewhere.

For details, see this Slack thread: https://prisma-company.slack.com/archives/C02JV0K2Q31/p1654595286707199

Let's also add more info, where appropriate (like in the CLI Reference?) about Prisma installed locally (need to prepend `npx`) versus globally (no need to use `npx`). We can link to this topic as appropriate: https://nodejs.dev/learn/npm-global-or-local-packages

ObamaTheLlama114:
## Describe this PR

Prepends npx to Prisma commands in CLI reference

## Changes

- Prepends npx to Prisma commands in CLI reference

## What issue does this fix?

Fixes #3538

-


janpio (commented):
Technically this is not correct. The command itself really is just `prisma *`. Many people just use `npx ...` to execute said command. With a global install of Prisma it is just `prisma *`, with yarn it is `yarn prisma` and so on - there are many options.

tanberry:
@janpio the Docs team dug up the metrics around what percentage of people run Prisma globally (meaning they do not need to prepend `npx`), and it is quite small, [according to your analysis 2.5%.](https://prisma-company.slack.com/archives/C02JV0K2Q31/p1654597252152349?thread_ts=1654595286.707199&cid=C02JV0K2Q31).

I do think we need to add a Note to the top of this page, explaining the use (or non-use), of `npx`, but adding it to the Docs is based on the usage stats. 

We *could* leave it off the reference pages (leave the reference commands as bare-boned as possible), add the Note I mention above, and only add `npx` to the Get Started Guides and elsewhere as appropriate. But we do for sure want to add it to the Get Started, anywhere where users might trip up using our commands simply because they didn't know they needed to prepend `npx`.

janpio:
Yes, this has nothing to do in the end with local vs. global - but that the reference lists the "commands" and not "the thing you execute in your terminal". That is one a different level.

We could have a "Executing Prisma CLI commands on your machine" section somewhere that explains that what you type in your terminal depends on your packager you are using (Npm, Yarn, Pnpm etc) and wether you have a local or a global installation of the CLI - and that this has to consistently be prepended to the actual commands in the command reference. But that is usually assumed in most documentation of Node based tools I think.

(Getting Started etc where we really just tell people "This is the usual thing to type into your terminal" are of course, as you say, different from the reference. There we should be consistent and chose one of the options, and `npx prisma *` makes sense there.)

tanberry:
Yep, two topics, intersecting. @ObamaTheLlama114 , want to share your proposal on handling this page, and for opening a second PR for the non-Reference topics where we do want to add `npx`?

ObamaTheLlama114:
I sure would.  I can remove all of the npx prepends, since there already were some, and note at the top of the page the that if prisma is installed locally, the user needs to prepend with their chosen package manager (npx, pnpx, yarn...)

content/400-reference/200-api-reference/200-command-reference.mdx
```diff
@@ -9,6 +9,8 @@ tocDepth: 2
 
 This document describes the Prisma CLI commands, arguments, and options.
 
+Command should be prepended by npx, pnpx, or yarn if prisma is installed locally with npm, pnpm, or yarn respectivly.
```
tanberry: Thanks @ObamaTheLlama114 for adding this note. Let's put it into an <Admonition> and perhaps make it slightly less terse... maybe something like "Be aware that Prisma commands are typically prepended by `npx`, `pnpx`, or `yarn`, if Prisma is installed locally with npm, pnpm, or yarn respectively.

ObamaTheLlama114: fixed

tanberry:
Looks good. I had suggested making it an Admonition, so that it stands out more, but this is fine too. Thanks, let's merge it!

tanberry (approved):

PR #3596, opened 2022-08-11
Did not instruct to install @prisma/client

imMatheus:
## Describe this PR

All the code uses the @prisma/client package to interact with prisma but it never gets said to install it from this specific docs

## Changes

Added a single instruction to install @prisma/client

## What issue does this fix?

No issue from the issues open on GitHub, but will help new devs get up and running with Prisma

No

### DIFF
--- a/content/100-getting-started/01-quickstart.mdx
+++ b/content/100-getting-started/01-quickstart.mdx
@@ -51,6 +51,12 @@ Then, install the Prisma CLI as a development dependency in the project:
 npm install prisma --save-dev
 ```
 
+Then, install the Prisma Client:
+
+```terminal
+npm install @prisma/client
+```
+
 Finally, set up Prisma with the `init` command of the Prisma CLI:
 
 ```terminal


janpio (changes requested):
Do you specifically need to run this command? 
Usually it should happen under the hood when using `prisma generate`, which is why we do not document its usage.

(Using "Request Changes" to indicate that this should not be merged until discussed)

nikolasburk:
Hey @imMatheus, thanks a lot for opening this PR.

The `@prisma/client` package is installed for you under the hood, so we don't need to explicitly install it.

It works like this in the Quickstart tutorial:
1. run `npx prisma migrate dev`
2. `prisma migrate dev` runs `prisma generate`
3. `prisma generate` checks whether `@prisma/client` is installed; if not, it runs `npm install @prisma/client`


 We want to keep the Quickstart minimal and focused on the _actual_ steps that are needed to set up Prisma, so I'm going to close this PR.

However, please feel free to open issues and PRs in the `docs` repo in the future if anything doesn't make sense to you!

PR #3630, opened 2022-08-19
Update 100-prisma-schema-reference.mdx

andrewicarlson:
Currently the wrong line is highlighted in the code sample which leads to user confusion as we include examples of both a relationship (`Post[]`) and a Scalar list (`String[]`)

## Changes

- Changed the highlighted line

## What issue does this fix?

Screen Shot 2022-08-19 at 09 09 37

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -1046,10 +1046,9 @@ Makes a field a list.
 <TabbedContent tabs={[<FileWithIcon text="Relational databases" icon="database"/>, <FileWithIcon text="MongoDB" icon="database"/>]}>
 <tab>
 
-```prisma highlight=4;normal
+```prisma highlight=3;normal
 model User {
   id             Int      @id @default(autoincrement())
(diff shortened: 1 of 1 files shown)


janpio (approved):
Indeed. The section in general is about lists, but the example is explicitly about a "scalar list". 

We could also update the "Remarks" to use a scalar. and maybe even remove the relation from the examples - and then add another example just for relations (although I am not sure how those "lists" are called then to be honest.)

andrewicarlson:
Yeah I thought about removing the relation from these examples to simplify them but assumed the relation had been added for an explicit purpose such as demonstrating that there are relation lists and scalar lists, and they're different things.

I'm down to remove it but don't want to if it was intentional.

keerlu (commented):
Thanks for finding this @andrewicarlson!

I would also agree with removing the confusing `posts` relation field from both of these.

A separate relation field example would also be good, but don't feel you have to, we've already annoyingly expanded the scope of your two-character PR 😆

keerlu (approved):

andrewicarlson:
Done! Thanks for the quick responses folks :)

PR #3634, opened 2022-08-22
Add missing fields to example schemas

### LINKED ISSUE #3622: Update implicit many-to-many example schema
The example schema in this section doesn't match the queries below.

Via @garrensmith: 

> I was looking at our many-to-many-relations docs. This example here https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations#implicit-many-to-many-relations
> 
> The schema doesn't define a "title" for posts or a name for categories but the ts code has those. I think the schema might need to be updated.

Slack conversation: https://prisma-company.slack.com/archives/C5Z9TH6N9/p1660820223397969

keerlu:
## Describe this PR

Added missing `title` and `name` fields to example schemas.

## What issue does this fix?

Fixes #3622

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
@@ -65,7 +65,7 @@ CREATE TABLE "Post" (
 CREATE TABLE "CategoriesOnPosts" (
     "categoryId" integer NOT NULL,
     "postId" integer NOT NULL,
-    "assignedBy" text NOT NULL
+    "assignedBy" text NOT NULL,
     "assignedAt" timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
     FOREIGN KEY ("categoryId")  REFERENCES "Category"(id),
     FOREIGN KEY ("postId") REFERENCES "Post"(id)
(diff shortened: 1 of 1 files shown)


garrensmith (approved):

content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
```diff
@@ -436,6 +442,7 @@ model Post {
 
 model Category {
   id      String   @id @default(auto()) @map("_id") @db.ObjectId
+  name    String
```
janpio, line 445: Although this uses the same model names, I think this is a different example where this addition would not be needed. (If we want to be consistent, then `title` is also missing in `Post`  here)

keerlu: The `name` field is used in a query further down: https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations#mongodb:~:text=some%3A%20%7B-,name%3A,-%7B

I was just fixing all schemas on the page to match later queries, rather than making them consistent with each other

janpio: Ugh, ok 👍

PR #3655, opened 2022-08-24
Update 050-prisma-client-reference.mdx

mohamad-n:
the "some" filter option   should replace by "every" on sample of using "every"  filter query option


tanberry:
Thanks for catching this, @mohamad-n !

PR #3657, opened 2022-08-24
Improve `@ignore` and `@@ignore` examples

### LINKED ISSUE #3562: Separate `@ignore` and `@@ignore` in schema reference example
Via @janpio ([Slack conversation](https://prisma-company.slack.com/archives/C5Z9TH6N9/p1659337200463439)):

> It is confusing that our `@ignore` documentation example also uses `@@ignore`. It would be better if we had a standalone example that does not mix these super similar things together.
> 
> Current example:
> https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#ignore

keerlu:
## What issue does this fix?

Fixes #3562

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -2898,7 +2898,7 @@ model Post {
 }
 ```
 
-In the following example, the `Post` model is invalid because it does not have a unique identifier, and the `posts` field is invalid because it refers to the invalid `Post` model. Use `@@ignore` on the `Post` model and `@ignore` on the `posts` field to exclude them from the generated Prisma Client API:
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2891,17 +2885,38 @@ In [2.17.0](https://github.com/prisma/prisma/releases/tag/2.17.0) and later, Pri
...
+model Post {
+  id       Int  @default(autoincrement()) // no unique identifier
+  author   User @relation(fields: [authorId], references: [id])
+  authorId Int
 
   @@ignore
 }
 ```
 
+In the following example, the `Post` model is invalid because it does not have a unique identifier, and the `posts` field is invalid because it refers to the invalid `Post` model. Use `@@ignore` on the `Post` model and `@ignore` on the `posts` field to exclude them from the generated Prisma Client API:
```
janpio:
```suggestion
In the following example, the `Post` model is invalid because it does not have a unique identifier, and the `posts` relation field on `User` is invalid because it refers to the invalid `Post` model. Use `@@ignore` on the `Post` model and `@ignore` on the `posts` relation field in `User` to exclude both the model and the relation field from the generated Prisma Client API:
```

janpio (approved):

PR #3672, opened 2022-08-26
add: First step install @prisma/client on topic 4

tdrunsri:
This PR add some missing step in Quick Start Doc

## Changes
- add step install @prisma/client on topic 4


janpio (changes requested):
Installing Prisma Client is not part of the recommended steps, as `prisma generate` takes care of that when missing.

PR #3673, opened 2022-08-26
Add-caveat-about-OR-and-empty-values.mdx

### LINKED ISSUE #3650: Investigate docs misunderstanding about filtering records
Latest status: I've investigated - the user is correct about the error message, but I don't think the doc is relevant to the issue.

See Slack discussion: https://prisma-company.slack.com/archives/C02FZENHD4N/p1661242347725959

Original issue: https://github.com/prisma/prisma/issues/14426

Update: 25 Aug 2022. IMO the user's interpretation of the docs is reasonable, so the next step is to determine if the docs are wrong and if so, fix them.

Update 2: I've replicated the issue. Basically, with AND, the filter returns all records. However, with OR, it depends on whether the filter value is undefined or empty. If it's empty, the filter returns all records. If it's undefined, it returns []. IMO we should add a note to the doc below the example that when you pass nothing at all with OR, you get all the records

Details:

In the following doc...
https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined#the-effect-of-null-and-undefined-on-conditionals

It says:

> This is because passing an undefined value to an AND or NOT operator is the same as passing nothing at all, meaning the findMany query in the example will run without any filters and return all the users.

The user says:

> I'm trying to filter an array of videos, using keywords array, that if they exist, it will fetch all the records that contain some of the values (using hasSome filter) however, if I get undefined from graphql, it'll return all the records and not filter at all.

> For some reason, if prisma gets undefined in the array filter it returns an error.
**I've looked at the docs and it says that whenever a filter option gets undefined, it's as if it wasn't written and it'll ignore it.**

> I may be wrong, but factually whenever I pass an undefined value using my graphql resolver to the prisma, it ignores it, and returns the values as if there wasn't a filter, so this is the behaviour I expected throughout every filter, including array filtering.

We need to investigate the following:
* Is the doc misleading?
* If so, what do we need to change to make it clearer

andrew-walford-prisma:
Add clarification about filtering with the OR operator and empty values


content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
```diff
@@ -137,17 +137,23 @@ updateUser: (parent, args, ctx: Context) => {
...
+| Operator | 0 filters                                  | 1 filter               | n filters            |
+| -------- | ------------------------------------------ | ---------------------- | -------------------- |
+| `OR`     | return empty list (but see the note below) | validate single filter | validate all filters |
+| `AND`    | return all items                           | validate single filter | validate all filters |
+| `NOT`    | return all items                           | validate single filter | validate all filters |
 
-This example shows how an `undefined` parameter impacts the results returned by a query that uses the [`OR`](/reference/api-reference/prisma-client-reference#or) <span class="api"></span> operator.
+<Admonition type="info">
+
+Note: If you pass an empty value, instead of `null` or `undefined`, then the `OR` operator returns all items, as it does with the `AND` and `NOT` operators.
```
aqrln, line 152: What does "empty value" mean here? Does it refer to an empty array passed as the value of `OR` property itself, or an empty object passed as the value of a specific filter, or something else?

content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
```diff
@@ -174,13 +180,18 @@ const users = await prisma.user.findMany({
 // returns: []
 ```
 
-The query receives filters from a formData object, which includes an optional email property. In this instance, the value of the email property is `undefined`. When this query is run no data is returned.
+The query receives filters from a `formData` object, which includes an optional email property.
+
+- In this instance, the value of the email property is `undefined`. The query returns an empty list.
+- If the value of the email property is empty, instead of `undefined`, then the query returns all the users.
```
aqrln, line 186: A code snippet to illustrate this would've been very helpful, as I'm not sure I correctly understand what the difference should be.

content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
andrew-walford-prisma: Good point, I'll clarify it

content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
andrew-walford-prisma: Thanks - I'll investigate and add one

aqrln: Thank you!

content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
andrew-walford-prisma: @aqrln  What I meant here was an empty value passed into a filter, when the filter includes an OR expression. So for example in the code example:

```
const users = await prisma.user.findMany({
  where: {
    OR: [
      {
        email: {
          contains: formData.email,
        },
      },
    ],
  },
})
```

If `formData.name` is an empty string, then the filter returns all items. This is a different outcome from the situation where `formData.name` is `null` or `undefined`.

content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
andrew-walford-prisma: @aqrln 

So this is the same as above - we're talking about an empty string here. Do you think it would be sufficient to say:

```suggestion
- If the email property is an empty string then the query returns all the users.
```

content/200-concepts/100-components/02-prisma-client/080-null-and-undefined.mdx
andrew-walford-prisma: Therefore, how about this change?

```suggestion
Note: If you pass an empty string, instead of `null` or `undefined`, then the `OR` operator returns all items. This is the same behavior as with the `AND` and `NOT` operators.
```

aqrln: Ooh, I see. So, here's my understanding of what happens here: an empty string is not any kind of special value — it shouldn't be, at least — and this is not the kind of situation that you get with `null` or `undefined` where you effectively have 0 filters. Here it's a pretty normal well-formed query with one filter where you check that email contains some specific substring. It just so happens that this substring is an empty string in this case, and any string contains an empty string as a substring, so this condition is always true, and you get all records as a result. But there's nothing special about an empty string from the point of view of how filters work (it would be a bug if there is in fact), and you would probably get a very different result if you used a different condition instead of `contains`.

andrew-walford-prisma: Ah, I see your point. This would be expected behavior. In that case, we don't need this caveat in the docs at all.

Before I close the PR, though, there's a bit more testing I want to do to confirm or refute what the user reported. Thanks @aqrln .

andrew-walford-prisma:
Looking at this issue again, I've got too focussed on the notion of empty strings, which is irrelevant here - and, from a chat with @aqrln  - I realize that empty strings produce expected behavior.

The issue here is with the user's error message when they tried to filter on a string[] array by passing `undefined` - they got an error, instead of the behavior as described in the doc. However, the doc section in question is about filter used with conditionals, and this wasn't what the user was doing. The docs section is clear about that. Therefore, this isn't a docs error, but we could maybe signpost the user's issue better in our VSCode plugin.

PR #3674, opened 2022-08-27
fix(data-model): Remove `@createdAt` that does not exist

janpio:
re https://github.com/prisma/prisma/issues/1332


ObamaTheLlama114:
Thanks @janpio for your contribution

PR #3676, opened 2022-08-29
Update index.mdx

tanberry:
added a link to env var reference page


janpio:
The page you added a link to does not have specific details about "environment variables".

I would suggest an admonition that mentions that there is a reference of environment variables that can be used to influence and configure Prisma (CLI, Client, Studio etc.).

PR #3700, opened 2022-08-31
Contribution correct 1 crucial command starter npx

miguelgargallo:
I correct 1 crucial command for starters or begginers,
always all commands must proceed with "npx/pnpx/yarn" at the beggining.

### DIFF
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -34,9 +34,9 @@ Prisma is a modern DB toolkit to query, migrate and model your database (https:/
 
 Usage
 
-  $ prisma [command]
+  $ [npx/pnpx/yarn] prisma [command]
 
-Commands
+Commands (
 
             init   Setup Prisma for your app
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
This is the help output of Prisma CLI. In that context, mentioning that to run a Npm based CLI you need to prefix with npx/pnpx/yarn for non global installations should not be necessary.

If this should be mentioned here, it has to happen in a note next to the help output.

keerlu:
Thanks for raising this, @miguelgargallo... we're actually looking in to how to deal with these commands right now. It's kind of complicated:

- if we leave the prefix off, people copy the command and it doesn't work, which as you say is bad for beginners
- if we just add `npx` (the most common choice), that's great for those users but no good for `yarn` or `pnpx`
- if we add all three in the command, as in your PR, it's clearer but still doesn't copy correctly, and also people could use yet other options
- if we leave the prefix off but make a note about it somewhere, people might not see the note

Given this discussion we're going to close the PR, but we have made [a new issue](https://github.com/prisma/docs/issues/3715) to track this - feel free to comment there.

PR #3708, opened 2022-09-01
Update 070-case-sensitivity.mdx

lregaladohdez:
When using NestJS with TypeScript, directly putting mode: 'insensitive' throws an error because 'mode' should be of type Prisma.QueryMode, so one needs to put Prisma.QueryMode.insensitive or Prisma.QueryMode.default instead of putting the string.

## Describe this PR

I was getting an error on a NestJS project using the 'mode' option on queries. Reviewing the doc on how to add the mode: insensitive it's not correct if we are using TypeScript you should put mode: Prisma.QueryMode.insensitive

## Changes

Only add a small comment in the same line where  the example of mode: insensitive is presented

## What issue does this fix?

No issues reported, just a clarification for improving the documentation

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/070-case-sensitivity.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/070-case-sensitivity.mdx
@@ -121,7 +121,7 @@ const users = await prisma.user.findMany({
   where: {
     email: {
       endsWith: 'prisma.io',
-      mode: 'insensitive', // Default value: default
+      mode: 'insensitive', // Default value: default, use Prisma.QueryMode.insensitive from @prisma/client  when using TypeScript
     },
   },
 })


keerlu:
Thanks for raising this, @lregaladohdez! We'll get a technical review on this and get back to you.

matthewmueller:
Hmm, I wasn't able to reproduce this either as a compiler error or a runtime error:

```ts
  const results = await prisma.test.findFirst({
    where: {
      testId: {
        mode: "insensitive"
      }
    }
  })
```

Can you share a reproduction with us?

keerlu:
@lregaladohdez, in case you didn't see Matt's comment above - are you able to share a reproduction of this with us?

aqrln:
`Prisma.QueryMode.insensitive` should be the same thing as `'insensitive'` as a literal string. The only way it could be different is if you used it in the context where TypeScript wouldn't have been aware of the expected type and, in addition to that, would be forced to infer a wider type (i.e., `string`) instead of a specific literal type due to reasons like mutability.

In other words, one way to trigger the type error would be something like this:

```ts
  const query = {
    where: {
      email: {
        endsWith: 'prisma.io',
        mode: 'insensitive',
      },
    },
  }

  await prisma.user.findMany(query)
```

In that case, both

```diff
-         mode: 'insensitive',
+         mode: 'insensitive' as const,
```

and

```diff
-         mode: 'insensitive',
+         mode: Prisma.QueryMode.insensitive,
```

are equally valid ways to fix the type error — however neither of them is the best one. The proper solution, in my opinion, would be to annotate the variable itself with the expected type:

```diff
-  const query = {
+  const query: Prisma.UserFindManyArgs = {
     where: {
       email: {
         endsWith: 'prisma.io',
         mode: 'insensitive',
       },
     },
   }

   await prisma.user.findMany(query)
```

I don't think there's anything that needs to be changed in the docs here, especially in this specific place, as we have quite a lot of enums represented as unions of string literal types. However, some kind of guide explaining this part of TypeScript's type system and how it is used in Prisma might be useful, especially to developers coming to TypeScript from other languages.

nikolasburk:
Hey @lregaladohdez 👋 

based on the comments from our Engineering team, I'm going to close this PR. Please feel free to reopen if you have another argument for it 🙂 

Also, please feel free to open issues and PRs in the `docs` repo in the future if anything doesn't make sense to you!

PR #3714, opened 2022-09-02
adding some doc on on using variable expansion

obh:
for variable expansion one needs to add `dotenv.expand` or else the variables loaded with placeholders won't work. 
> cat myenv.env
> DATABASE_URL="mysql://${MYSQL_USERNAME}:${MYSQL_PASSWORD}@localhost:3306/slack?connection_limit=10&pool_timeout=20"


janpio:
What is a use case for this? That page currently already is pretty confusing (especially the section you are expanding) - so vbeing clear when and why one would want or need to use this sounds like a good idea.

obh:
Yes, the page could be structured better because few things are interconnected. The usecase is around expanding variables. For example the doc says -
```
You can also use environment variables in the expansion that are set outside of the .env file, for example a database URL that is set on a PaaS like Heroku or similar:

DATABASE_URL=postgresql://test:test@localhost:5432/test
DATABASE_URL_WITH_SCHEMA=${DATABASE_URL}?schema=public
```
Now this usecase is common because in some cases the database details might get injected from some other system. In my case it was coming from a k8 config map. So, there's an env variable which is then getting referenced in the .env file. As per the doc it should work without any issues by just doing `dotenv.config()`. But dotenv won't expand the variables and one needs to use `dotenv.expand`.

janpio:
I think in Prisma CLI and Client this should wokr out of the box, and `dotenv.config()` should not be required in the first place. If I read the current page correctly, you only need to use `dotenv` manually when you wnat to do something with these env vars in your application code from `.env` as well - correct? And your PR adds the variable expansion to that use case as well?

obh:
yes, you are correct. You mostly don't need `dotenv` if you can set env variables to node runtime and then you can even do expansion in code. You only need `dotenv` if you have some custom file which needs to get injected and then this PR is about variable expansion within `dotenv`.

content/300-guides/125-development-environment/050-environment-variables/index.mdx
```diff
@@ -130,3 +130,13 @@ import * as dotenv from 'dotenv'
...
+
+If you are using environment variables in a custom file, you can add it to your code seperately using `dotenv`. If you need variable expansion across environment files, you will have to use [`dotenv-expand`](https://github.com/motdotla/dotenv-expand):
+
+```ts
+const dotenvExpand = require('dotenv-expand')
+
+var envFile = path.resolve(join(__dirname, "myenv.env"))
+var mySqlEnv = dotenv.config({path: envFile})
+dotenvExpand.expand(mySqlEnv)
+ ```
```
janpio:
````suggestion
If you are using a custom file name for your environment variables, you can configure `dotenv` to use that filename:

```ts
import * as dotenv from 'dotenv'

var envFile = path.resolve(join(__dirname, "myenv.env"))
dotenv.config({path: envFile}) // Load the environment variables
console.log(`The connection URL is ${process.env.DATABASE_URL}`)
```

If you need variable expansion across environment files, you can additionally use [`dotenv-expand`](https://github.com/motdotla/dotenv-expand):

```ts
import * as dotenv from 'dotenv'
const dotenvExpand = require('dotenv-expand')

var envFile = path.resolve(join(__dirname, "myenv.env"))
var mySqlEnv = dotenv.config({path: envFile})
dotenvExpand.expand(mySqlEnv)
```
````

gniting:
@jharrell @nikolasburk is this still applicable? If not, close it out?

jharrell (dismissed):

jharrell (approved):

PR #3717, opened 2022-09-02
fix(prisma-client-transactions-guide): Fix highlighted lines in code example

janpio:
From:
image

To:
image


janpio:
Interesting how Percy is not picking up the change about the highlighting in the code block.

nikolasburk (approved):
nice catch!

PR #3728, opened 2022-09-06
Add GA version for `migrate diff` and `db execute` in the CLI reference

keerlu:
## Describe this PR

Add GA version for `migrate diff` and `db execute` in the CLI reference


janpio (approved):
👍 

Do we still know how one had to activate it? That might be useful information here as well.

keerlu:
> Do we still know how one had to activate it? That might be useful information here as well.

Thanks, I've added that in

PR #3734, opened 2022-09-07
update: WebsiteHeader from new lens ver

carlagn:
## Describe this PR

Update WebsiteHeader + WebsiteFooter with new, stable version from lens@0.0362.

## Changes

- Deleted "old" internal WebsiteHeader/WebsiteFooter folder
- Update WebsiteHeader and WebsiteFooter import to fetch from lens package
- Update lens package to latest @0.0.367.

N/A


janpio:
This seems to loose the highlighting of the currently active header menu item, in the case of docs the "Developer". It used to be blue, not it is not any more.
image
(via Percy obviously)

janpio:
Footer also seems to loose some of the things that were added or updated (arrows for external links, updated wording ecosystem => stack) 
image

carlagn:
@janpio Footer and current active tab have been updated and should be fixed now.

nilubava (approved):

janpio:
Percy shows me the footer is fixed, but the highlight in the header still seems to be gone.
(Dropdown arrows also get fatter/bold, but I assume that is intended?)

carlagn:
> Percy shows me the footer is fixed, but the highlight in the header still seems to be gone. (Dropdown arrows also get fatter/bold, but I assume that is intended?)

@janpio

arrows are intended. the highlight is gone because it was being manually set since we had the header developed internally on the docs repo. In the meantime, `prisma.io/docs` (docs homepage) should be highlighted (but it won't show in the preview because window location is being checked. I already have a fix for this highlighting issue for docs: https://github.com/prisma/lens/pull/450

We'd prefer to check first if this new version is working as intended as it's important to get our new header from lens above all else.

janpio:
Ok 🤷 - not keeping you from merging here 👍 

(Side note: That logic sounds scary and brittle. I hope over time we can really improve this to also work on previews, maybe even without JS etc)

PR #3735, opened 2022-09-07
Add guide to Prisma type system

### LINKED ISSUE #3352: Improve native types docs
Add a guide explaining:
- how Prisma's type system works
- how to add native type attributes to refine the basic types
- how to use native type attributes when introspecting and migrating

Currently we have some documentation of native types in the connectors and the schema reference, but nothing at top level, and many topics are not covered. Need to discuss the specifics more but would be good to improve this.

See Slack comment [here](https://prisma-company.slack.com/archives/C5Z9TH6N9/p1653311689577529).

Current Notion discussion [here](https://www.notion.so/prismaio/f84b446d64054ccb9b01b3f69438c343?v=781e623a12fc4aed858f096771c233ea&p=6556b6ccf21e46adaa1b5e816d1d3de3).

keerlu:
## Describe this PR

This is a draft of a guide to Prisma's type system, explaining how to use native type attributes when introspecting or migrating.

## What issue does this fix?

Fixes #3352


content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+<TopBlock>
+
+This guide explains Prisma's type system and walks you through using it to introspect your database and apply schema changes.
+
+</TopBlock>
+
+## How does Prisma's type system work?
+
+Prisma uses types to define the kind of data that a field can hold. To make it easy to get started, Prisma provides a small number of core [scalar types](/reference/api-reference/prisma-schema-reference#model-field-scalar-types). For example, in the blog post model below, the `title` field uses the `String` scalar type, while the `createdAt` field uses the `DateTime` scalar type:
```
janpio:
```suggestion
Prisma uses _types_ to define the kind of data that a field can hold. To make it easy to get started, Prisma provides a small number of core [scalar types](/reference/api-reference/prisma-schema-reference#model-field-scalar-types) that should cover most default use cases. For example, in the blog post model below, the `title` field uses the `String` scalar type, while the `createdAt` field uses the `DateTime` scalar type:
```
- highlight first usage of `types`
- add some color to what we think the scalar types should cover

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+See Prisma's [database connector pages](/concepts/database-connectors) for the default type mappings for a given database. For example, [this table](/concepts/database-connectors/postgresql#prisma-migrate) gives the default type mappings for PostgreSQL. To see the default type mappings for a given database, see the [model field scalar types section]() of the Prisma schema reference. For example, [this table](/reference/api-reference/prisma-schema-reference#model-field-scalar-types) gives the default type mappings for the `String` scalar type.
+
+### Native type mappings
+
+Sometimes you may need to use a more specific database type that is not one of the default type mappings for your Prisma type. For this purpose, Prisma provides [native type attributes](/concepts/components/prisma-schema/data-model#native-types-mapping) to refine the core scalar types. For example, in the `createdAt` field of your `Post` model above you may want to use a timestamp with a time zone in an underlying PostgreSQL database, by using the `timestamptz(3)` type instead of the default type mapping of `timestamp(3)`. To do this, add a `@db.Timestamptz(3)` native type attribute to the `createdAt` field:
```
janpio: This is difficult to grok, as the database type `timestamptz(3)` and the Prisma native type `@db.Timestamptz(3)` looks slightly different, but both look pretty similar to the default PostgreSQL `timestamp(3)`. I bet some people would overlook the `tz` and be left wondering what is going on.

(I unfortunately do not have a solution handy for this problem. Maybe a native type that is more different from the default?)

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+```prisma file=schema.prisma
+model Post {
+  id        Int      @id
+  title     String
+  createdAt DateTime @db.Timestamptz(3)
+}
+```
+
+The following sections give more detail on how to introspect existing native types in your database, and how to use types when you apply schema changes to your database with `migrate dev` or `db push`.
```
janpio: This appearing before a headline is a bit weird - as it does not belong under the `### Native type mappings` heading any more really.

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+## How to introspect database types
+
+When you introspect your database, Prisma will take the database type of each table column and represent it in your Prisma schema using the correct Prisma type for the corresponding model field. If the database type is not the default database type for that Prisma type, Prisma will also add a native type attribute.
```
janpio:
```suggestion
When you introspect your database, Prisma will take the database type of each table column and represent it in your Prisma schema using the correct Prisma type for the corresponding model field. If the database type is not the default database type for that Prisma scalar type, Prisma will also add a native type attribute.
```
to be super precise - not sure if necessary

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,154 @@
...
+model Post {
+  id        Int      @id
+  title     String
+  createdAt DateTime @db.Timestamptz(3)
+}
+```
+
+The following sections give more detail on how to introspect existing native types in your database, and how to use types when you apply schema changes to your database with `migrate dev` or `db push`.
+
+## How to introspect database types
```
janpio: This whole section is smooooth. Nice!

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,154 @@
...
+```
+
+This `Post` model has:
+
+- an `id` field with a Prisma type of `Int`
+- a `title` field with a Prisma type of `String`
+- a `createdAt` field with a Prisma type of `DateTime`
+- an `updatedAt` field with a Prisma type of `DateTime` and a `@db.Timestamptz(3)` native type attribute
+
+Now apply these changes to a PostgreSQL database with the following command, run from the root directory of your project:
```
janpio:
```suggestion
Now apply these changes to an empty PostgreSQL database with the following command, run from the root directory of your project:
```

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: I also considered using something like `@db.Date` instead, to be more visibly different, though it seems more contrived as a type choice? 

I wish I'd written down my thought process when I was choosing an example, but I was having difficulty picking a good one. Everything I thought of seemed to have something different wrong with it:

- `@db.citext` doesn't require much explanation, but needs a Postgres extension
- `@db.Uuid` requires explaining what UUIDs are, plus reading our docs it felt kind of confusing because we also have the `uuid()` function
- anything to do with money runs in to a lot of opinions about how you should represent money
- some other things I forgot...

Happy to have other suggestions!

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: May as well have it!

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: Ah, yes, that used to make more sense and then I split the section into subsections.

I've moved it into the TopBlock instead and rewritten it a bit

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+```prisma file=schema.prisma
+model Post {
+  id        Int      @id
+  title     String
+  createdAt DateTime
+}
+```
+
+Databases also have their own type system, which defines the type of value that a column can hold. Most databases provide a large number of data types to allow fine-grained control over exactly what a column can store. For example, a database might provide inbuilt support for multiple sizes of integers, or for XML data. The names of these types vary between databases. In PostgreSQL a variable length text column has a type of `text`, whereas in MySQL the type is `varchar`.
```
tomhoule: The example in the last sentence is less clear cut than it could be, because postgres also has `VARCHAR` and MySQL also has `TEXT`. One alternative example could be the column type for booleans: `BOOLEAN` on Postgres, but  `TINYINT(1)` is what people typically use on MySQL.

pimeys: PostgreSQL has a `varchar` type and MySQL has a `text` type accordingly. In Prisma, if the user does not define a native type for a `String` field, we use `text` on PostgreSQL and `varchar(...)` on MySQL.

The difference between `text` and `varchar` is that most of the time `text` is having an arbitrary length and is stored separately outside of the row structure, where `varchar` has a defined length and can be compressed for faster selection directly to the row. Of course SQL Server is a bit different here, and has additionally the `varchar(max)` and `nvarchar(max)` types that behave like the `text` type.

content/300-guides/050-database/830-prismas-type-system.mdx
janpio: Maybe strings could work? The default string native type could very easily be very different from some of the other options available.

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: Thanks, that's definitely a better example. I've changed it to that.

keerlu: Thanks - given this and Tom's comment above I've changed the example

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: Strings are good from the point of the view of the names looking different:

image

I don't really know enough to pick a plausible example though - do you have any ideas out of those?

janpio: A super specific case would be someone wanting to have `String` with long text (so `@db.Text`) and another `String` that just holds bit information: https://www.postgresql.org/docs/current/datatype-bit.html Very different.

keerlu: I thought about this some more, and I can't really see how to introduce a bit string type as an example without getting into some very specific use case.

I've rewritten using the `@db.Date` example I mentioned above - that's visually distinct and might be good enough, what do you think?

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
+---
```
janpio: I think we should rename the file to `prisma-type-system`, without the `s`.

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+### Default type mappings
+
+To allow you to get started with our core scalar types, Prisma provides _default type mappings_ that map each scalar type to a default type in the underlying database. For example:
+
+- by default Prisma's `String` type gets mapped to PostgreSQL's `text` type and MySQL's `varchar` type
+- by default Prisma's `DateTime` type gets mapped to PostgreSQL's `timestamp(3)` type and SQL Server's `datetime2` type
+
+See Prisma's [database connector pages](/concepts/database-connectors) for the default type mappings for a given database. For example, [this table](/concepts/database-connectors/postgresql#prisma-migrate) gives the default type mappings for PostgreSQL. To see the default type mappings for a given database, see the [model field scalar types section]() of the Prisma schema reference. For example, [this table](/reference/api-reference/prisma-schema-reference#model-field-scalar-types) gives the default type mappings for the `String` scalar type.
```
janpio: This paragraph is confusing to read. I know that one is from the connector, and the other from the type direction.

Actually, I think there is a mistake:
```suggestion
See Prisma's [database connector pages](/concepts/database-connectors) for the default type mappings for a given database. For example, [this table](/concepts/database-connectors/postgresql#prisma-migrate) gives the default type mappings for PostgreSQL.   
To see the default type mappings for all databases for a specific given Prisma type, see the [model field scalar types section](/reference/api-reference/prisma-schema-reference#model-field-scalar-types) of the Prisma schema reference. For example, [this table](/reference/api-reference/prisma-schema-reference#float) gives the default type mappings for the `Float` scalar type.
```

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+### Native type mappings
+
+Sometimes you may need to use a more specific database type that is not one of the default type mappings for your Prisma type. For this purpose, Prisma provides [native type attributes](/concepts/components/prisma-schema/data-model#native-types-mapping) to refine the core scalar types. For example, in the `createdAt` field of your `Post` model above you may want to use a date-only column in your underlying PostgreSQL database, by using the `date` type instead of the default type mapping of `timestamp(3)`. To do this, add a `@db.Date` native type attribute to the `createdAt` field:
+
+```prisma file=schema.prisma
+model Post {
+  id        Int      @id
+  title     String
+  createdAt DateTime @db.Date
```
janpio: Here it would be good to include the `datasource db` block to be able to show where `@db` comes from. People sometimes get this wrong. This is also where we know from that this is e.g. PostgreSQL and can show the adequate types. (Might even be worth adding to the explanation what is happening in this example)

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+}
+```
+
```
janpio: I think here would be good to have a summary sentence for this basic explanation. Maybe something along the lines of "With the scalar types and native type mappings you can express all the types of your database, but do not have to if the Prisma default satisfy your needs. This leads to a shorter, more readable Prisma schema".

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+```sql
+CREATE TABLE "public"."User" (
+  id serial PRIMARY KEY NOT NULL,
+  name text NOT NULL,
+  "isActive" boolean NOT NULL
+);
+```
+
+Introspect your database with the following command run from the root directory of your project:
```
janpio: This skips over the bit how you need a `datasource` in your schema, and a connection string defined somewhere, often as an env var that is used in the schema. Might be fine to skip over that - maybe we have a good place to link to though?

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+```prisma file=schema.prisma
+model User {
+  id       Int     @id @default(autoincrement())
+  name     String
+  isActive Boolean
+}
+```
+
+The `id`, `name` and `isActive` columns in the database are mapped respectively to the `Int`, `String` and `Boolean` Prisma types respectively. These are the default database types for these Prisma types, so Prisma does not add any native type attributes.
```
janpio: - 1 `respectively` too much
- The second sentence was misleading, as the list of types before are the Prisma scalar types. I think the second sentence refers to the database types though.
```suggestion
The `id`, `name` and `isActive` columns in the database are mapped respectively to the `Int`, `String` and `Boolean` Prisma types. The database types are the _default_ database types for these Prisma types, so Prisma does not add any native type attributes.
```

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+Now add a `createdAt` column to your database with a data type of `date` by running the following SQL command:
```
janpio:
```suggestion
Now add an additional `createdAt` column to your database with a data type of `date` by running the following SQL command:
```

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+```sql
+ALTER TABLE "public"."User"
+ADD COLUMN "createdAt" date NOT NULL;
+```
+
+Now introspect your database again:
```
janpio:
```suggestion
Introspect your database again:
```
get rid of repeated "now"

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+```terminal
+npx prisma db pull
+```
+
+Your Prisma schema now includes the new `createdAt` field with a Prisma type of `DateTime`. The `createdAt` field also has a `db.Date` native type attribute, because `date` is not the default type for the `DateTime` type:
```
janpio:
```suggestion
Your Prisma schema now includes the new `createdAt` field with a Prisma type of `DateTime`. The `createdAt` field also has a `@db.Date` native type attribute, because PostgreSQL's `date` is not the default type for the `DateTime` type:
```

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+  id        Int      @id @default(autoincrement())
+  name      String
+  isActive  Boolean
+  createdAt DateTime @db.Date
+}
+```
+
+## How to use types when you apply schema changes to your database
+
+When you apply schema changes to your database using `migrate dev` or `db push`, Prisma will use both the Prisma scalar type of each field and any native attribute it has to determine the correct database type for the corresponding column in the database.
```
janpio:
```suggestion
When you apply schema changes to your database using Prisma Migrate or `db push`, Prisma will use both the Prisma scalar type of each field and any native attribute it has to determine the correct database type for the corresponding column in the database.
```
It is not just `migrate dev`, but the whole Migrate.

content/300-guides/050-database/830-prismas-type-system.mdx
```diff
@@ -0,0 +1,152 @@
...
+```
+
+This `Post` model has:
+
+- an `id` field with a Prisma type of `Int`
+- a `title` field with a Prisma type of `String`
+- a `createdAt` field with a Prisma type of `DateTime`
+- an `updatedAt` field with a Prisma type of `DateTime` and a `@db.Date` native type attribute
+
+Now apply these changes to an empty PostgreSQL database with the following command, run from the root directory of your project:
```
janpio: To run this the schema is missing the `datasource` bit. Might not be relevant.

janpio (approved):
With the comments adressed, I think this will be great!

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: Done (and added some explanation)

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: Agree that's better - I've added a couple of sentences along those lines

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: I've added a link to the page on introspection further up. Not sure we need to go into more detail here, I think most of the docs implicitly assume you've worked out the connecting to your database part

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: I think this is now ok with the datasource added and explained at the top

content/300-guides/050-database/830-prismas-type-system.mdx
keerlu: Done

eviefp:
For completeness, would it be worth mentioning `enum`s as well?

nikolasburk (approved):
I think this looks great!

keerlu:
> For completeness, would it be worth mentioning `enum`s as well?

Sorry, missed this comment before - thanks, merging as is but will think about whether to add `enum`s later

PR #3739, opened 2022-09-07

TravisL12:
The many-to-many `connect` case of assignment of existing `categories` requires that the model being assigned match the schema letter case (i.e. must be capitalized).

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
@@ -109,6 +109,7 @@ The following query:
 - Creates a new `Post`
 - Creates a new category assignment, or `CategoriesOnPosts`
 - Connects the category assignment to existing categories (with IDs `9` and `22`)
+- The `Category` connect must match the case of the schema (i.e. `Category` should be capitalized)
 
 ```ts
 const assignCategories = await prisma.post.create({
@@ -119,7 +120,7 @@ const assignCategories = await prisma.post.create({
         {
           assignedBy: 'Bob',
           assignedAt: new Date(),
-          category: {
+          Category: {
             connect: {
               id: 9,
             },
@@ -128,7 +129,7 @@ const assignCategories = await prisma.post.create({
         {
           assignedBy: 'Bob',
           assignedAt: new Date(),
-          category: {
+          Category: {
             connect: {
               id: 22,
             },


janpio (commented):
I can't imagine that to be correct or necessary. There is no field `Category`, only a field `category` of (relation) type `Category`.

What is your indication that this change would be necessary?

TravisL12:
I just ran into this in my own project. I have a joins table of `MealsOnMenus` which join a `Meal` and `Menu` model. In this example where I was creating a `Menu` with associated (existing) `Meal` objects. The query was creating the `Menu` object but I was getting an error on the `Meal` connect query. The issue was that originally I was using `meal`, once I capitalized the connect query to `Meal` the query worked as expected.

```
async function createMenu({ userId, menu, meals }) {
  return db.menu.create({
    data: {
      userId,
      name: menu.name,
      description: menu.description,
      meals: {
        create: meals.map((meal) => ({
          Meal: { // <--- capitalized this `Meal`.
            connect: {
              id: +meal.id,
            },
          },
        })),
      },
    },
  });
}
```

Relevant Schema as well:
```
model Meal {
  id           Int      @id @default(autoincrement())
  createdAt    DateTime @default(now())
  updatedAt    DateTime @updatedAt
  name         String
  description  String   @default("")
  prepTime     Int      @default(0) // total time can be summed (prep + cook)
  cookTime     Int      @default(0)
  isFavorite   Boolean  @default(false)
  importUrl    String   @default("")
  userId       String
  ingredients  Ingredient[]
  instructions Instruction[]
  menus        MealsOnMenus[]
  User         User     @relation(fields: [userId], references: [id])
}

model Menu {
  id             Int      @id @default(autoincrement())
  createdAt      DateTime @default(now())
  updatedAt      DateTime @updatedAt
  name           String
  description    String   @default("")
  userId         String
  meals          MealsOnMenus[]
  scheduledMenus ScheduledMenu[]
  User           User     @relation(fields: [userId], references: [id])
}

model MealsOnMenus {
  Meal         Meal     @relation(fields: [mealId], references: [id])
  mealId       Int
  Menu         Menu     @relation(fields: [menuId], references: [id])
  menuId       Int
  createdAt    DateTime @default(now())
  updatedAt    DateTime @updatedAt
   
  @@id([mealId, menuId])
}
```

TravisL12:
I actually do see my error in raising this ticket. The examples I was following on the Prisma docs has lowercase columns, where in my code I am using capitalized columns. So I am incorrect in this. Closing.

janpio:
Yep, easy mistake to make unfortunately. In your example the schema seems to come from Introspection, which just uses the model name as the field name (because it is hard to make a decision what the correct name should be).

Good you figured it out!

PR #3742, opened 2022-09-08
Update the postgres truncation strategy

emilos:
Update the example of the truncation strategy with a single query that runs faster.

## Describe this PR

The truncation strategy from the docs does multiple queries in a loop. It's possible to do it one query instead, which is faster and can e.g. help make test suites run faster.

## Changes

Changing the postgres example of multiple truncation queries to a single truncation query.

It might be possible to do the same with the MySQL example below.


garrensmith (approved):
That looks good.

PR #3747, opened 2022-09-08
Update: add await before prisma update promise

immorez:
Yet another fix to documents.


keerlu:
Thanks @immorez - we'll include this explicit `await` to make things clearer.

keerlu (approved):

PR #3753, opened 2022-09-09
chore: update to more recent version of jaeger docker

millsp:
## Describe this PR

There is a more recent command to run `jaeger` via `docker`

## Changes

Updates the command for the new, maintained docker container.

## What issue does this fix?

The other container has not received updates for two years.


andrew-walford-prisma:
@millsp  Sorry, I've only just seen this PR. Serhii suggested a different edit to the same line, which I've just merged. I'll open a Slack convo for the three of us!

PR #3756, opened 2022-09-12
Better name for relation scalar in examples

keerlu:
## Describe this PR

Update examples to use `userEmail` instead of `userId`. See [Slack thread](https://prisma-company.slack.com/archives/C5Z9TH6N9/p1662658762707029).


janpio (changes requested):
I think some `userEmail` fields are still `Int` although they should be `String`.

keerlu:
> I think some `userEmail` fields are still `Int` although they should be `String`.

Oops, thanks, should be fixed now.

janpio (approved):

PR #3759, opened 2022-09-12
Add note for P1012 about schema compatibility

### LINKED ISSUE #3599: Consider expanding error P1012 description
We've had a suggestion to expand our [docs for error code P1012](https://www.prisma.io/docs/reference/api-reference/error-reference#p1012) to include invalid schema errors (when parsed with Prisma 4 or later).

Wrinkle: the errors page is generated content, so we'll need to be careful about preserving any new material we add.

Slack thread: https://prisma-company.slack.com/archives/C5Z9TH6N9/p1660549192661239

Example:
* User has Prisma 3.12.0 with a valid schema
* User upgrades to Prisma 4.0.0  (or newer) which now considers their schema file to be invalid. Here is an example error:

```
Error code: P1012
error: Error parsing attribute "@relation": A one-to-one relation must use unique fields on the defining side. Either add an `@unique` attribute to the field `profileId`, or change the relation to one-to-many.
  -->  schema.prisma:14
   | 
13 |   id        Int      @id @default(autoincrement())
14 |   profile   Profile? @relation(fields: [profileId], references: [id])
15 |   profileId Int?
```

The P1012 section of the docs already contains:

`Error parsing attribute @{}: {}`

But it might be helpful to direct users to our [upgrade guide](https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4), because some of these errors are not considered to be errors with all Prisma versions.

andrew-walford-prisma:
Added a note to explain that you can get error P1012 when you upgrade to version 4..0.0 and your schema is no longer compatible.


janpio:
Why are we highlighting this one imcompatibility on the reference page like this? Many of those errors were added with a specific version, making older schemas invalid. Was this a specifically popular one that justifies the highlight?

PR #3765, opened 2022-09-12
Replace "middlewares" with "middleware functions"

### LINKED ISSUE #3750: Middleware/middlewares - make consistent
In our middleware docs, we use the terms middleware and middlewares. IMO we should standardize (I'd go for the former variation).

Page: https://www.prisma.io/docs/concepts/components/prisma-client/middleware
 
We don't use the plural form in the sub pages, so we only need to update it on this page.

andrew-walford-prisma:
I've replaced "middlewares" with "middleware functions" (I was going with "middleware items", but @nikolasburk  suggested the superior "functions"). "Middlewares" doesn't seem grammatical (we don't say "softwares", for example). I did consider "pieces of middleware", as in "pieces of software", but I think that might apply "partial middleware items" to some second language speakers, so I went with what I hope is a good alternative.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
@@ -6,7 +6,7 @@ metaDescription: 'Prisma Client middleware allows you to perform actions before 
 
 <TopBlock>
 
-Middleware items act as query-level lifecycle hooks, which allow you to perform an action before or after a query runs. Use the [`prisma.$use`](/reference/api-reference/prisma-client-reference#use) <span class="api"></span> method to add middleware, as follows:
+Middleware functions act as query-level lifecycle hooks, which allow you to perform an action before or after a query runs. Use the [`prisma.$use`](/reference/api-reference/prisma-client-reference#use) <span class="api"></span> method to add middleware, as follows:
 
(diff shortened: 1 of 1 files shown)


janpio:
"Middlewares" is a pretty commonly used variant though: https://www.google.com/search?q=%22middlewares%22
Also: https://en.wiktionary.org/wiki/middlewares

janpio:
"middleware functions" is also not the correct term. They are really commonly called "middlewares" or just "middleware (plural)".

nikolasburk:
FWIW, I'd also prefer "middlewares". I think we can sacrifice the bit of "grammatical" worry for clarity when it comes to being understood by our users.

andrew-walford-prisma:
I'm not concerned with what is grammatically correct per se, but with communicating ideas well and consistently. Usage determines meaning, so we should select words based on how they're actually used. So, in this case, if "middlewares" is widely understood, I agree that this is the term we should use.

However, there's a secondary consideration. While it might be clear what we mean, the credibility of the docs can be undermined by perceived mistakes. In this case, my concern is that there's disagreement out there on this term. For some readers, if we pluralize the term, this might be perceived as a mistake.

However, maybe there's only a least-worst option. If there isn't a clear, accurate alternative term, and the plural form is widely understood, then we should go with that. So therefore I'll undo the changes.

PR #3766, opened 2022-09-12
docs: clarify interactiveTransactions

danstarns:
I found it confusing that we were overriding the `prisma` variable in the scope of the transaction callback. This PR introduces the usage of `tx` to help showcase the difference between a user's 'top level' `prisma` variable and the 'special' transaction variable that's inside the scope of the callback.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
@@ -164,7 +164,7 @@ generator client {
 
 Then you can pass an async function into [`$transaction`](/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api). 
 
-The first argument passed into this async function is an instance of Prisma Client. Below we will call this instance `tx` and any Prisma call invoked on this `tx` instance will be encapsulated into the transaction. 
+The first argument passed into this async function is an instance of Prisma Client. Below, we will call this instance `tx`. Any Prisma call invoked on this `tx` instance is encapsulated into the transaction. 
 
 Let's look at an example:
 
--- a/content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
@@ -162,7 +162,9 @@ generator client {
 }
 ```
 
-Then you can pass an async function into [`$transaction`](/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api).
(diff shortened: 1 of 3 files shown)


janpio:
(I think we switched between these two before...)

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -162,7 +162,9 @@ generator client {
 }
 ```
 
-Then you can pass an async function into [`$transaction`](/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api).
+Then you can pass an async function into [`$transaction`](/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api). 
+
+The first argument passed into this async function is an instance of Prisma Client. Below we will call this instance `tx` and any Prisma call invoked on this `tx` instance will be encapsulated into the transaction.
```
andrew-walford-prisma:
```suggestion
The first argument passed into this async function is an instance of Prisma Client. Below, we will call this instance `tx`. Any Prisma call invoked on this `tx` instance is encapsulated into the transaction. 
```

content/300-guides/100-performance-and-optimization/100-prisma-client-transactions-guide.mdx
```diff
@@ -842,6 +842,8 @@ generator client {
 
 Then you can pass an async function into [$transaction](/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api).
 
+The first argument passed into this async function is an instance of Prisma Client. Below we will call this instance `tx` and any Prisma call invoked on this `tx` instance will be encapsulated into the transaction.
```
andrew-walford-prisma:
```suggestion
The first argument passed into this async function is an instance of Prisma Client. Below, we will call this instance `tx`. Any Prisma call invoked on this `tx` instance is encapsulated into the transaction. 
```

andrew-walford-prisma (commented):
I've made a couple of suggested wording tweaks (including a tense change to bring it in line with our new style guide). Otherwise LGTM. Thanks Dan!

millsp:
We initially decided on having shadowed variables so one could not mistakenly write non-transactional queries in a transaction. This is just a convention, that would help you not to do the mistake and realize only later that eg. data was written even though there was a rollback. The other advantage of shadowing `prisma` is that you can move code/queries in and out of the transaction easily. That was just a convention, not something users are forced to do.

PR #3784, opened 2022-09-14
changed cli prepend note

ObamaTheLlama114:
## Describe this PR
changed cli prepend note in cli reference


janpio:
You can also install Npm CLIs globally, which removes the need to use any prefix.

ObamaTheLlama114:
> You can also install Npm CLIs globally, which removes the need to use any prefix.

The docs team agreed to not mention global installs since its not recommended.

PR #3785, opened 2022-09-15
Add clarification of @ignore behavior for required fields.

### LINKED ISSUE #11283: When using `@ignore` for some fields, there is no `create` method in the model
### Bug description

When using `@ignore` for some fields, there is no `create` method in the model. I'm ignoring just two fields but the rest should be available.

### How to reproduce

Define this model:
```prisma
model User {
  email     String    @unique
  password  String
  createdAt DateTime  @default(now()) @ignore
  updatedAt DateTime  @updatedAt @ignore
  remarks   String?
}
```
Then generate it. Then you cannot create a user using this code:
```ts
await this.prismaService.user.create({ data: newUser });
```
Because of this error:
```bash
Property 'create' does not exist on type 'UserDelegate<RejectOnNotFound | RejectPerOperation | undefined>'.ts(2339)
```

### Expected behavior

The `create` method should be there!

### Environment & setup

- OS: Ubuntu 21.10
- Database: PostgreSQL
- Node.js version: 16.13.1

### Prisma Version

```
3.8.1
```

AndrewSouthpaw:
## Describe this PR

Clarifies behavior of `@ignore` when being used on required fields, and steers the user toward making a field optional for most cases (otherwise `create` is removed from the Client API). Also indicates that `@ignore` can be used to safely delete columns.

## Changes

I provided an example for both, but I think it's also reasonable to drop the example with a required field, and leave it to people reading the remarks section carefully. I'm not sure if the Prisma team cares about being terse with the documentation.

## What issue does this fix?

Fixes https://github.com/prisma/prisma/issues/11283.


tanberry:
Thanks so much, @AndrewSouthpaw and apologies for the delay in reviewing the PR. Taking a look now!

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2863,7 +2863,8 @@ In [2.17.0](https://github.com/prisma/prisma/releases/tag/2.17.0) and later, Pri
 #### Remarks
 
 - Ignored fields are excluded from the generated Prisma Client
-- You can manually add `@ignore` to a field that you want to exclude from the Prisma Client (for example, a field that you do not want Prisma users to update)
+- Required fields with no default value marked with `@ignore` removes the `create` and `createMany` methods in the Client API.
+- Can be used to prevent a field from being updated, or make it safe to delete from the database in a subsequent migration with zero downtime.
```
tanberry:
```suggestion
- Can be used to prevent a field from being updated, or to make it safe to delete a field from the database in a subsequent migration with zero downtime.
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2873,10 +2874,24 @@ The following example demonstrates manually adding `@ignore` to exclude the `ema
 model User {
   id    Int    @id
   name  String
-  email String @ignore // this field will be excluded
+  email String? @ignore // this field will be excluded
```
tanberry, line 2893:
```suggestion
  email String? @ignore // this field will be excluded from any database actions
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2873,10 +2874,24 @@ The following example demonstrates manually adding `@ignore` to exclude the `ema
...
 }
 ```
 
+When adding `@ignore` to a required field, it becomes impossible to add new records to the database, so the `create` and `createMany` methods are removed from the Client API:
```
tanberry, line 2897: Can we reword this to make the "cause and effect" more obvious? Im not sure if my interpretation is correct; is it something like:
"Adding `@ignore` to any required field makes it impossible to add new records to the database (because the lack of any required field in a record disallows the adding of the record), and thus the `create` and `createMany` methods cannot be used, as they are removed from the Client API:"??

janpio (changes requested):
The remarks make statements I am not sure are correct or explained enough (zero downtime migrations). I suggest we merge my https://github.com/prisma/docs/pull/3792 that clarifies the functionality itself first, and then resovle the conflicts here to see how we can express that adequately.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: That is correct, although the order of things is a bit unexpected in your wording @tanberry.
I made a suggestion to explain that in the initial description of `@ignore` here that should be clearer: https://github.com/prisma/docs/pull/3792

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2877,6 +2877,9 @@ Add `@ignore` to a field that you want to exclude from Prisma Client (for exampl
 
 #### Remarks
 
+- Ignored fields are excluded from the generated Prisma Client
```
nikolasburk:
```suggestion
- Ignored fields are excluded from the generated Prisma Client.
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2877,6 +2877,9 @@ Add `@ignore` to a field that you want to exclude from Prisma Client (for exampl
 
 #### Remarks
 
+- Ignored fields are excluded from the generated Prisma Client.
+- Required fields with no default value marked with `@ignore` removes the `create` and `createMany` methods in the Client API.
```
nikolasburk, line 2881: is this actually correct @AndrewSouthpaw? I've just tested this and I was still able to execute `create` without a type error but then actually got a runtime error:

1. create schema:
```prisma
model Post {
  title     String 
  published Boolean @default(false)
}
```
2. `npx prisma migrate dev --name init`
3. create record:
```ts
const r = await prisma.post.create({
  data: {
    title: "Hello World"
  }
})
```
4. update schema:
```prisma
model Post {
  title     String  @ignore
  published Boolean @default(false)
}
```
5. `npx prisma migrate dev --name add-ignore`
6. type error in query, so remove `title` from `data` object:
```ts
const r = await prisma.post.create({
  data: {  }
})
```
7. `npm run dev`
8. violated null constraint
```
PrismaClientKnownRequestError: 
Invalid `prisma.post.create()` invocation in
/Users/nikolasburk/Desktop/ignore/script.ts:8:31

  5 // A `main` function so that we can use async/await
  6 async function main() {
  7  
→ 8   const r = await prisma.post.create(
Null constraint violation on the fields: (`title`)
```

PR #3786, opened 2022-09-15
update: percy packages to cli and removed agent

carlagn:
Upgrade percy gatsby plugin


.github/workflows/ci.yml
```diff
@@ -19,4 +19,5 @@ jobs:
           cache: 'npm'
           node-version: '14'
       - run: npm ci
+      - run: npx percy exec -- gatsby build
       - run: npx gatsby build
```
janpio: Is this still needed? I assume the previous line does all the things needed for Gatsby already?

janpio (changes requested):
Noticed two things in the screenshots in Percy and commented on them - not sure if that qas equally broken before.

janpio:
(Merged `main` in here so the Percy comparison is more recent - hopefully)

janpio:
Footer seems to have a problem now, and on a single page some other CSS rendering problems :/

config.js
```diff
@@ -5,7 +5,12 @@ const config = {
     titlePrefix: '',
     titleSuffix: ' | Prisma Docs',
   },
-  redirects: [],
+  redirects: [
+    {
+      from: '/',
+      to: '/',
```
janpio: That looks like something we should not keep around.

janpio (changes requested):
Navigation the preview manually currently adds `#undefined` to any URL.

Old redirects from `config.js` are not testable:
- https://docs-pwn4wgigw-prisma.vercel.app/reference/tools-and-interfaces/prisma-schema/prisma-schema-file goes to https://docs-pwn4wgigw-prisma.vercel.app/docs/reference/tools-and-interfaces/prisma-schema#undefined which is a 404 **(Note that this URL is for a specific deployment and will not update!)**

But that might be a general problem with the previews, as the redirects always seem to include the `/docs` which does not work on the preview deployments.

janpio:
A list of the removed redirects that should keep working:
- https://docs-git-update-percy-migrate-prisma.vercel.app/reference/tools-and-interfaces/prisma-schema/prisma-schema-file
- https://docs-git-update-percy-migrate-prisma.vercel.app/reference/tools-and-interfaces/prisma-schema
- https://docs-git-update-percy-migrate-prisma.vercel.app/reference/tools-and-interfaces/prisma-client/api
- https://docs-git-update-percy-migrate-prisma.vercel.app/reference/tools-and-interfaces/prisma-schema/models
- https://docs-git-update-percy-migrate-prisma.vercel.app/concepts/components/prisma-data-platform#prisma-data-proxy
- https://docs-git-update-percy-migrate-prisma.vercel.app/concepts/overview/what-is-prisma/data-modeling#data-modeling-without-sometext

config.js
carlagn: Leaving the comment here as well for context:

Gatsby usually asks for dummy data to be written in order to go through all the frontmatter values and other query values that may have been assigned and used for filters or any type of logic, like here: https://github.com/prisma/docs/actions/runs/3159217753/jobs/5142171081

.github/workflows/ci.yml
```diff
@@ -19,4 +19,6 @@ jobs:
           cache: 'npm'
           node-version: '14'
       - run: npm ci
-      - run: npx gatsby build
+      - run: npx gatsby clean
+      - run: npx gatsby build --prefix-paths
+      - run: npx percy snapshot ./public --exclude ./public/reference/tools-and-interfaces/**/*.html
```
janpio: Can we remove the exclusion here now maybe?

gatsby-config.js
```diff
@@ -146,12 +147,6 @@ let plugins = [
       defaultDataLayer: { website: 'docs' },
     },
   },
-  {
-    resolve: `gatsby-plugin-percy`,
-    options: {
-      // see actual configuration in `percy` key in `package.json`
```
janpio, line 152: Is that configuration thing in pacakge.json still used? Is it the only way now to configure the plugin or would it make sense to add something similar around line 78?

janpio (changes requested):
We are aaaaaalmost good here I think, but one important page seems to have gone missing (although it is in the preview): 
image
https://docs-git-update-percy-migrate-prisma.vercel.app/reference/api-reference/prisma-client-reference
Did something go wrong when taking the screenshots for Percy maybe?

(The three ok ones were the weird redirects)

gatsby-config.js
carlagn: I don't think it's going through it anymore, so I deleted this package/plugin.

janpio (approved):
Great job @carlagn

PR #3792, opened 2022-09-16
fix(schema-reference): Clarify @ignore and @@ignore

### LINKED ISSUE #11283: When using `@ignore` for some fields, there is no `create` method in the model

janpio:
closes https://github.com/prisma/prisma/issues/11283

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -2858,7 +2858,7 @@ model Post {
 
 <AlgoliaTerm name="apiReference" value="@ignore" />
 
-Add `@ignore` to a field that you want to exclude from the Prisma Client (for example, a field that you do not want Prisma users to update). Ignored fields are excluded from the generated Prisma Client. The model's `create` method is disabled when doing this for _required_ fields with no `@default` (as the database will not be able to create an entry without that data).
+Add `@ignore` to a field that you want to exclude from the Prisma Client (for example, a field that you do not want Prisma users to update). Ignored fields are excluded from the generated Prisma Client. The model's `create` method is disabled when doing this for _required_ fields with no `@default` (because the database cannot create an entry without that data).
 
 #### Remarks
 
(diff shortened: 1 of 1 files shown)


janpio:
Might overlap with https://github.com/prisma/docs/pull/3785 a bit - so that one probably needs conflict resolution after this was merged.

dmmulroy:
Just stumbled onto this PR - thank you for updating the docs, this was super confusing to figure out 🙌

Weakky (approved):
LGTM 👍

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -2858,12 +2858,11 @@ model Post {
 
 <AlgoliaTerm name="apiReference" value="@ignore" />
 
-In [2.17.0](https://github.com/prisma/prisma/releases/tag/2.17.0) and later, Prisma adds `@ignore` to fields that _refer to_ invalid models when you introspect.
+Add `@ignore` to a field that you want to exclude from the Prisma Client (for example, a field that you do not want Prisma users to update). Ignored fields are excluded from the generated Prisma Client. The model's `create` method is disabled when doing this for _required_ fields with no `@default` (as the database will not be able to create an entry without that data).
```
andrew-walford-prisma:
```suggestion
Add `@ignore` to a field that you want to exclude from the Prisma Client (for example, a field that you do not want Prisma users to update). Ignored fields are excluded from the generated Prisma Client. The model's `create` method is disabled when doing this for _required_ fields with no `@default` (because the database cannot create an entry without that data).
```

andrew-walford-prisma (commented):
LGTM - I've made two minor wording suggestions

PR #3828, opened 2022-09-27
Order transaction isolation levels table

kernfeld-cockroach:
Order the table of transaction isolation levels by strength, similar to how they are listed in the bullet points above. Each transaction isolation level is now positioned to the left of all stronger transaction isolation levels.


kernfeld-cockroach:
I think the Vercel deploy failed because I do not have permissions.

tanberry:
Thanks so much @kernfeld-cockroach for reordering the columns to match the bullet points. Solid improvement.

PR #3839, opened 2022-10-03
Update treenode collapse fn

### LINKED ISSUE #3823: Bug: Tech switcher pages on docs removes sidebar highlight on change
When changing the techs on the switcher the highlight on the menu disappears.
Example:
Screenshot 2022-09-26 at 15 24 02
Screenshot 2022-09-26 at 15 24 06

carlagn:
## Describe this PR

When changing the techs on the switcher the highlight on the menu disappears.

## Changes

Update `treeNode.tsx` to collapse if current page is active, if it is last level, and has an expand button.

## What issue does this fix?

[Fixes #3823](https://github.com/prisma/docs/issues/3823)

N/A

### DIFF
--- a/src/components/sidebar/treeNode.tsx
+++ b/src/components/sidebar/treeNode.tsx
@@ -213,10 +213,14 @@ const TreeNode = ({
     }
   }, [isCollapsed])
 
+  React.useEffect(() => {
+    if (lastLevel && isCurrent && hasExpandButton) collapse()
+  }, [])
+
   const isCurrent =
     location &&
     slug &&
-    (location.pathname + '/').includes(urlGenerator(slug).replace(/\/index$/, '') + '/')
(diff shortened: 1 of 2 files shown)


src/components/sidebar/treeNode.tsx
```diff
@@ -213,10 +213,28 @@ const TreeNode = ({
     }
   }, [isCollapsed])
 
+  React.useEffect(() => {
```
nilubava, line 219: Wouldn't it be better to create an array of these special cases and keep it in `config.ts` so that we can add more special cases (if any) in the future? 
ex: 
```
const specialPaths = ['prisma-cli', 'deployment', 'sql-views']
```
then you can iterate over this array for `const specialCases`

carlagn: reviewed! thanks for the tip :)

nilubava (approved):

PR #3858, opened 2022-10-07
fix(relations): Optional/mandatory headlines for 1-1 and 1-n

janpio:


content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
```diff
@@ -199,7 +199,11 @@ model User {
 
 This restriction was introduced in 2.12.0.
 
-However, you can choose if the side of the relation _with_ a relation scalar should be optional or mandatory. In the following example, `profile` and `profileId` are mandatory. This means that you cannot create a `User` without connecting or creating a `Profile`:
+However, you can choose if the side of the relation _with_ a relation scalar should be optional or mandatory. 
+
+### Mandatory 1-1 relation
+
+In the following example, `profile` and `profileId` are mandatory. This means that you cannot create a `User` without connecting or creating a `Profile`:
```
ruheni, line 206:
```suggestion
### Non-optional 1-1 relation

In the following example, `profile` and `profileId` are non-optional. This means that you cannot create a `User` without connecting or creating a `Profile`:
```

"Mandatory" seems too aggressive. Suggestions:
- "Required"
- "Non-optional"
Feel free to select which one is accurate and sounds okay

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
```diff
@@ -246,6 +248,8 @@ model Post {
 </tab>
 </TabbedContent>
 
+### Mandatory one-to-many relation
```
ruheni, line 251:
```suggestion
### Non-optional one-to-many relation
```

https://github.com/prisma/docs/pull/3858/files#r1395684857

content/200-concepts/100-components/01-prisma-schema/06-relations/100-one-to-one-relations.mdx
jharrell: I would prefer either `Mandatory` or `Required`. `Non-optional` doesn't sound right.

EDIT: actually, after taking another look at this, I think these sections should be "Nullable" and "Non-Nullable".

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
```diff
@@ -209,6 +209,8 @@ A 1-n-relation always has two relation fields:
 
 The list side of a 1-n relation is **always mandatory**. On the other side of the relation, the annotated relation field and relation scalar can either _both_ be optional, or _both_ be mandatory.
 
+### Optional one-to-many relation
```
jharrell, line 212:
```suggestion
### Nullable one-to-many relation
```

content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
jharrell:
```suggestion
### Non-Nullable one-to-many relation
```

jharrell:
Made some comments. TL;DR I think our terminology should be "nullable" and "non-nullable"

janpio:
What I used here is established terminology in Prisma context.

- We always use "mandatory" in this context to describe them being "non optional".
- "null" is not a concept that exists in Prisma context. There things are optional or not, which is then translated to nullable on the database level.

nikolasburk:
> What I used here is established terminology in Prisma context.
> 
> * We always use "mandatory" in this context to describe them being "non optional".
> * "null" is not a concept that exists in Prisma context. There things are optional or not, which is then translated to nullable on the database level.

There doesn't seem to be any established terminology since we're using all of these terms across the docs.

For that reasons and because they're all interchangeable, I wouldn't let this discussion block the merging of this PR if it otherwise looks good.

The "real" solution here will be to determine one term (my vote would go to "required") that we want to use in this context and then update the docs everywhere to use that term consistently. I created an issue for that: https://github.com/prisma/docs/issues/5487

I would encourage to merge this PR regardless of the terms being use.

janpio:
Taking that as an approval and merging.

PR #3860, opened 2022-10-10
Update schema docs

mbicknese:
## Describe this PR

The order of declarations was out of order making the examples weird to compare.

## Changes

Updated the examples for writing a schema.


tanberry:
Thank you @mbicknese; having the same ordering for the schema components does indeed help!

PR #3874, opened 2022-10-11
Docs for `extendedWhereUnique` feature

### LINKED ISSUE #3848: 4.5.0: Document ability to use non-unique fields in where statements
**Note:**
preview issue needs to go into the PR

For 4.5.0, we're adding a preview feature that gives the ability to use non-unique fields in where statements for findUnique, upsert, update, delete.

SMEs: Flavian and Pierre

Design doc is here: https://www.notion.so/prismaio/Allow-non-unique-fields-to-be-used-with-unique-where-fddc3e52cf5d4bdfba58a970a1a55c9f?d=dfb444426ac345b68d29f0269947871f#3321403897ac4809bbb73a8940b1d4aa

andrew-walford-prisma:
Add draft documentation for new `extendedWhereUnique` feature


content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -5579,3 +5581,197 @@ await prisma.user.findMany({
...
+function main () {
+  await Promise.allSettled([updateOne(), updateTwo()]);
+}
+```
+
+<Admonition type="info">
+
+In optimistic concurrency control, you must fetch the record, relation, or nested record before you update it.
+
+**Have I mangled this with my rewrite?**
```
Weakky: No you haven't but you might want to add this in your example above (in both update functions).

```
function updateOne() {
  const user = await prisma.user.findUnique({ id: 1 });

   await prisma.user.update({
     where: { id: user.id, version: user.version },
     data: { city: "Berlin", version: { increment: 1 }
  });
}
```

Weakky: Just to make sure you understand why you need to fetch the record before:

In your example, you're querying for `version: 0`. That's ok as an example but in a real world use-case, you wouldn't know what the version is. It might work the first time you run this query but then it wouldn't anymore since the version would be 1.

Therefore, to know that version, you have to query the user first and use whatever version it's at as filter, and then increment it. That's why you need to fetch before. It's not so much that OCC needs it, just that you need to apply a filter with a value that you don't know. So the only way is to fetch that row first.

The more I read this note the more I'm starting to think we should just remove it and have the example fetch the record before doing the update just like on my example above.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -5579,3 +5581,197 @@ await prisma.user.findMany({
...
+#### One-to-one relations
+
+From version 4.5.0, you can filter on non-unique fields in the following operations on [one-to-one relations](/concepts/components/prisma-schema/relations/one-to-one-relations):
+
+- Nested update
+- Nested upsert
+- Nested disconnect
+- Nested delete
+
+Prisma Client automatically uses a unique filter to select the appropriate related record. As a result, you do not need to specify a unique filter in your `where` statement with a `WhereUniqueInput` [generated type](#generated-types-for-where). Instead, the `where` statement has a `WhereInput` generated type. You can use this to filter without the restrictions of `WhereUniqueInput`.
```
Weakky, line 5710: Great explanation 💯

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma: Thanks - that makes the example code much better.

PR #3889, opened 2022-10-12
Add section about transaction timing issues

### LINKED ISSUE #3825: Document new error code P2034
https://github.com/prisma/prisma-engines/commit/911f8bae8f9b08719066939ba0aaff82c1fe4a50

Update by Andrew, 12-oct-2022:
- Jan has added the error code itself in a separate PR (this is merged)
- We need to document the actual usage in the context of isolation level docs/transactions. See comment here: https://github.com/prisma/prisma/issues/9678#issuecomment-1259873693

andrew-walford-prisma:
Add section about transaction timing issues. This covers the causes of error code P2034 and provides a solution.


content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,73 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+### Transaction timing issues
+
+<Admonition type = "note">
+
+- This section is not available on MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
+- This section does not apply to CockroachDB and SQLite, because these databases only support the highest `Serializable` isolation level.
+
+</Admonition>
+
+When two or more transactions run concurrently, timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` then a `createMany` operation:
```
janpio:
```suggestion
When two or more transactions run concurrently in certain isolation levels, timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` then a `createMany` operation:
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,73 @@ transaction functions. We recommend you get in and out as quick as possible!
 
 </Admonition>
 
+### Transaction timing issues
+
+<Admonition type = "note">
+
+- This section is not available on MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
```
andrew-walford-prisma:
```suggestion
- This section does not apply to MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,73 @@ transaction functions. We recommend you get in and out as quick as possible!
 
 </Admonition>
 
+### Transaction timing issues
+
+<Admonition type = "note">
+
+- This section does not apply to MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
```
andrew-walford-prisma:
```suggestion
- The solution in this section does not apply to MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,73 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+- This section does not apply to CockroachDB and SQLite, because these databases only support the highest `Serializable` isolation level.
```
andrew-walford-prisma:
```suggestion
- The timing issues discussed in this section do not apply to CockroachDB and SQLite, because these databases only support the highest `Serializable` isolation level.
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,73 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+1. Transaction A: `deleteMany` removes all rows that would cause a conflict.
+1. Transaction B: `deleteMany` operation.
+1. Transaction B: `createMany` operation creates a new set of rows.
+1. Transaction B: Transaction B is committed.
+1. Transaction A: `createMany` operation. The newly-inserted rows conflict with the rows that Transaction B committed at step 4.
+
+In PostgreSQL and Microsoft SQL Server, the default [isolation level](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels) is `ReadCommited`. This isolation level does not guarantee repeatable reads. If your application commits another transaction between a `deleteMany` and a `createMany` call, then the current transaction might not observe the newly created rows.
+
+To avoid this error, you can set higher isolation levels of either `RepeatableRead` or `Serializable`. From Prisma version 4.4.0, you do not need to set the isolation level in your database configuration. You can instead set a higher isolation level on the transactions.
```
andrew-walford-prisma:
```suggestion
To avoid this problem, you can set higher isolation levels of either `RepeatableRead` or `Serializable`. From Prisma version 4.4.0, you do not need to set the isolation level in your database configuration. You can instead set a higher isolation level on the transactions.
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,73 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+When two or more transactions run concurrently in certain isolation levels, timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` then a `createMany` operation:
+
+1. Transaction A: `deleteMany` removes all rows that would cause a conflict.
+1. Transaction B: `deleteMany` operation.
+1. Transaction B: `createMany` operation creates a new set of rows.
+1. Transaction B: Transaction B is committed.
+1. Transaction A: `createMany` operation. The newly-inserted rows conflict with the rows that Transaction B committed at step 4.
+
+In PostgreSQL and Microsoft SQL Server, the default [isolation level](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels) is `ReadCommited`. This isolation level does not guarantee repeatable reads. If your application commits another transaction between a `deleteMany` and a `createMany` call, then the current transaction might not observe the newly created rows.
```
andrew-walford-prisma:
```suggestion
In PostgreSQL and Microsoft SQL Server, the default [isolation level](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels) is `ReadCommited`. This isolation level does not guarantee repeatable reads. If your application commits another transaction between a `deleteMany` and a `createMany` call, then the current transaction might not observe the newly-created rows.
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
 
 </Admonition>
 
+### Transaction timing issues
+
+<Admonition type = "note">
```
zachtil: Hm, I know that it's not a good idea to have an admonition immediately under the title / section heading. Part of the reason is that you attract readers with the title, and you immediately give them additional food for thought before they even know what it is about.

I would start with the introductory sentence below, which reads "When two or more transactions..." and then give the MongoDB note that it doesn't support isolation levels.

This way you have already introduced how transactions can run concurrently in certain isolation levels, and then back it up in the note a bit that this won't work in MongoDB.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+- The solution in this section does not apply to MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
+- The timing issues discussed in this section do not apply to CockroachDB and SQLite, because these databases only support the highest `Serializable` isolation level.
+
+</Admonition>
+
+When two or more transactions run concurrently in certain [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels), timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` and a `createMany` operation:
+
+1. Transaction A: `deleteMany` removes all rows that would cause a conflict.
```
zachtil: I am missing context here:

'_... cause a conflict._' with what?

Perhaps '_... cause a conflict with the subsequent `createMany` operation of Transaction A._'

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
 
 </Admonition>
 
+### Transaction timing issues
```
zachtil: This is a very well structured section. It's like a troubleshooting topic:

Problem
Cause
Solution

Personally, I would add these as subtitles to aid with scanning the page more quickly. I am using that in PDP Docs here:
https://www.prisma.io/docs/data-platform/troubleshooting/connection-to-db-timed-out

If you decide to go down this road, you could add them as bolded text, to avoid them appearing in the page TOC.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+When two or more transactions run concurrently in certain [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels), timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` and a `createMany` operation:
+
+1. Transaction A: `deleteMany` removes all rows that would cause a conflict.
+1. Transaction B: `deleteMany` operation.
+1. Transaction B: `createMany` operation creates a new set of rows.
+1. Transaction B: The application commits transaction B.
+1. Transaction A: `createMany` operation. The newly-inserted rows conflict with the rows that transaction B added at step 4.
+
+This conflict can occur at the isolation level `ReadCommited`, which is the default isolation level in PostgreSQL and Microsoft SQL Server. To avoid this problem, you can set a higher isolation level (`RepeatableRead` or `Serializable`).
```
zachtil: Cause section. See first comment.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+From Prisma version 4.4.0, you do not need to set the isolation level in your database configuration. You can instead set the isolation level on a transaction. This overrides the isolation level on your database server for that transaction. [Learn more about isolation levels in Prisma tranasctions](#transaction-isolation-level).
+
+To avoid transaction write conflicts and deadlocks:
```
zachtil: Solution section. See first comment.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+1. Transaction A: `deleteMany` removes all rows that would cause a conflict.
+1. Transaction B: `deleteMany` operation.
+1. Transaction B: `createMany` operation creates a new set of rows.
+1. Transaction B: The application commits transaction B.
+1. Transaction A: `createMany` operation. The newly-inserted rows conflict with the rows that transaction B added at step 4.
+
+This conflict can occur at the isolation level `ReadCommited`, which is the default isolation level in PostgreSQL and Microsoft SQL Server. To avoid this problem, you can set a higher isolation level (`RepeatableRead` or `Serializable`).
+
+From Prisma version 4.4.0, you do not need to set the isolation level in your database configuration. You can instead set the isolation level on a transaction. This overrides the isolation level on your database server for that transaction. [Learn more about isolation levels in Prisma tranasctions](#transaction-isolation-level).
```
zachtil: Can we skip 'server' here?

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+To avoid transaction write conflicts and deadlocks:
+
+1. On your transaction, use the `isolationLevel` parameter to set the transaciton to `Prisma.TransactionIsolationLevel.Serializable`.
```
zachtil: Typo in 'transaciton'

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+   This ensures that your application commits multiple concurrent or parallel transactions as if they were run serially. Where a transaction fails due to a write conflict or deadlock, Prisma returns a [P2034 error](/reference/api-reference/error-reference#p2034).
```
zachtil: _When a transaction fails due..._ instead of _Where..._?

also

_Prisma Client returns a ..._ instead of _Prisma returns ..._?

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+
+2. In your application code, build a retry around your transaction to trap any P2034 errors, as shown in this example:
```
zachtil: * _create_ or _add_ instead of _build_?
* _handle_ instead of _trap_?

zachtil (commented):
LGTM. I found this easy to follow and understand, it is very clearly written and put together.

I have 3 comments:
* separate visually with subtitles the problem, cause, and solution subsections
* start with an intro of the issue and then follow with the admonition
* small note on the transactions example as the first step might be missing some context.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -302,6 +302,74 @@ transaction functions. We recommend you get in and out as quick as possible!
...
+        retries++
+        continue
+      }
+      throw error
+    }
+  }
+}
+```
+
 ### Transaction isolation level
```
zachtil: In terms of flow, I would also move the **Transaction timing issues** under **Transaction isolation level** because the timing issues expect that you know about isolation levels.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Fair point, but I want to "filter out" people who use MongoDB, CockroachDB and SQLite. In other words, I'm telling them that they don't need to read this section.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Good point. As I understand it, for the purposes of the example it's only relevant that at this step Tranasaction A removes a series of rows, so I'll remove the second part.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Let's discuss this as a team. I don't have a view either way on whether we should use these headings, and I'd like to hear the pros and cons.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Noted - thanks!

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Noted - thanks!

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Good point - I'll remove one of the instances of this word

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Thanks, I'll make both changes

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Sure - I don't really expect anyone to be reading this page top to bottom, but you're right that they logically fall in that order. I'll switch the order.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Yes, I think so! Thanks

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: - Build - Agreed - "build" in dev docs implies compilation, so I'll change this to "add"
- Handle - Yes, I think you're right. I'll change it here. For the future, I've added this term to our style guide questions list

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -395,3 +395,71 @@ See the following resources:
...
+### Transaction timing issues
+
+<Admonition type="info">
+
+- The solution in this section does not apply to MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
+- The timing issues discussed in this section do not apply to CockroachDB and SQLite, because these databases only support the highest `Serializable` isolation level.
+
+</Admonition>
+
+When two or more transactions run concurrently in certain [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels), timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` and a `createMany` operation:
```
janpio: Is "conflict" the correct term here? 
I would naively have called this a "race condition", but not sure that is correct either.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -395,3 +395,71 @@ See the following resources:
...
+
+1. Transaction A: `deleteMany` operation removes a set of rows.
+1. Transaction B: `deleteMany` operation removes a set of rows.
+1. Transaction B: `createMany` operation creates a new set of rows.
+1. Transaction B: The application commits transaction B.
+1. Transaction A: `createMany` operation. The newly-inserted rows conflict with the rows that transaction B added at step 4.
```
janpio: Would this not only conflict when transaction A is committed in a further step?

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -395,3 +395,71 @@ See the following resources:
...
+
+- The solution in this section does not apply to MongoDB, because MongoDB does not support [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels).
+- The timing issues discussed in this section do not apply to CockroachDB and SQLite, because these databases only support the highest `Serializable` isolation level.
+
+</Admonition>
+
+When two or more transactions run concurrently in certain [isolation levels](https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels), timing issues can cause conflicts, such as the violation of unique constraints. For example, consider the following sequence of events where Transaction A and Tranasaction B both attempt to execute a `deleteMany` and a `createMany` operation:
+
+1. Transaction A: `deleteMany` operation removes a set of rows.
+1. Transaction B: `deleteMany` operation removes a set of rows.
```
janpio: Are these two steps even necessary for the problem to trigger?

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -395,3 +395,71 @@ See the following resources:
...
+1. Transaction B: `createMany` operation creates a new set of rows.
+1. Transaction B: The application commits transaction B.
+1. Transaction A: `createMany` operation. The newly-inserted rows conflict with the rows that transaction B added at step 4.
+
+This conflict can occur at the isolation level `ReadCommited`, which is the default isolation level in PostgreSQL and Microsoft SQL Server. To avoid this problem, you can set a higher isolation level (`RepeatableRead` or `Serializable`).
+
+From Prisma version 4.4.0, you do not need to set the isolation level in your database configuration. You can instead set the isolation level on a transaction. This overrides the isolation level on your database for that transaction. [Learn more about isolation levels in Prisma tranasctions](#transaction-isolation-level).
```
janpio: I think we could turn this around, and after the explanation of how to set it in the transaction via `isolationLevel` mention that up to Prisma 4.4.0 this was not possible and you had to configure this in your database directly.

janpio:
```suggestion
From Prisma version 4.4.0, you do not need to set the isolation level in your database configuration. You can instead set the isolation level on a transaction. This overrides the isolation level on your database for that transaction. [Learn more about isolation levels in Prisma transactions](#transaction-isolation-level).
```

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
janpio: Ah, below we call it "write conflict or deadlock".

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
```diff
@@ -395,3 +395,71 @@ See the following resources:
...
+    } catch (error) {
+      if (error.code === 'P2034') {
+        retries++
+        continue
+      }
+      throw error
+    }
+  }
+}
+```
```
janpio: (Out of scope for this PR)

This works, but is pretty impractical (as you would have to repeat this code for each and every transaction in your codebase.

This might be possible to do much nicer via a middleware instead that can add retry for _all_ transactions with a certain isolation level for example.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: I'll change the instance here to "write conflict or deadlock" too - that's much better than using slightly different terms for the same thing. Plus if we did decided to vary a term, the first use should be the full version.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: Good point - I'll add an extra step here.

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: You might well be right, but I think there was originally a need for these steps. I'll go back to the original notes and see if I've chopped something important out in one of the revisions. If not, I'll remove these two steps.

andrew-walford-prisma: I think you're right - I've removed these two steps

content/200-concepts/100-components/02-prisma-client/058-transactions.mdx
andrew-walford-prisma: I've reworked this section to reflect your suggestion.

PR #3919, opened 2022-10-19
Fix typo in generating-prisma-client docs

t3dotgg:
"also" was probably intended word, I find removing it entirely to be better.

Not sure if personal tone is used at all in the docs, would prefer "We recommend..."


andrew-walford-prisma (approved):
Thanks for this contribution @TheoBr . I'll merge this correction into the docs repo.

PR #3925, opened 2022-10-19
Deno Deploy guide: added a note about other database types; drive-by fixes to intro wording

andrew-walford-prisma:
I added a note at the top to clarify that users can use this feature in conjunction with any supported database type. I also brought the intro more in line with our current style guide.

### DIFF
--- a/content/300-guides/200-deployment/110-deployment-guides/550-deploying-to-deno-deploy.mdx
+++ b/content/300-guides/200-deployment/110-deployment-guides/550-deploying-to-deno-deploy.mdx
@@ -161,10 +161,13 @@ The next step is to sign up for Prisma's [Data Proxy](/data-platform/data-proxy)
    ![Import a Project](./450-03-import-project.png)
 
 4. Connect the Prisma Data Platform to your database and set up the Data Proxy:
+
    1. Paste your PostgreSQL connection string.
    1. Under **Location**, select a Data Proxy location that is geographically close to your database location.
    1. Click **Create Project** to test the connection and set up the Data Proxy.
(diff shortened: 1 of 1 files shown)


content/300-guides/200-deployment/110-deployment-guides/550-deploying-to-deno-deploy.mdx
```diff
@@ -149,19 +157,19 @@ The next step is to sign up for Prisma's [Data Proxy](/data-platform/data-proxy)
...
+
    ![Import a Project](./450-03-import-project.png)
 
 4. Connect the Prisma Data Platform to your database and set up the Data Proxy:
    1. Paste your PostgreSQL connection string.
    1. Under **Location**, select a Data Proxy location that is geographically close to your database location.
    1. Click **Create Project** to test the connection and set up the Data Proxy.
-      
-   You are greeted with a new connection string that starts with `prisma://`. 
+      You are greeted with a new connection string that starts with `prisma://`.
```
janpio: I don't think that is correct here.
image

andrew-walford-prisma: Good catch! I'll fix it.

andrew-walford-prisma: @janpio  This is fixed now.

janpio:
How about we move the admonition a bit down after the first paragraph? I think it is a bit weird to start with a warning or note.

PR #3928, opened 2022-10-20
Add exit code documentation to CLI reference

### LINKED ISSUE #3830: Document CLI exit code behavior
I'm unsure where this should go, probably:
https://www.prisma.io/docs/concepts/components/prisma-cli
https://www.prisma.io/docs/reference/api-reference/command-reference

We recently did changes/cleanup to our exit code behavior of some commands in 4.3.0, and we think it is a good idea to document exit code behavior in general which was not documented before.

What we want to document:
General behavior
- exit code 0 when a command runs successfully
- exit code 1 when a command errors
- exit code 130 (SIGINT) when the CLI receives a SIGINT or if the user cancels a prompt (PR #14979)

`migrate status`: It now exit(1) when an error occurred or if the database is not in sync (PR #14860)
Potential reasons for an error 
- if database connection error
- If databaseIsBehind (= unapplied migrations)
- If historiesDiverge (= migration history diverged)
- If we can't find a migration table (= not managed by Prisma)
- If failed migration(s) are found

`migrate diff`: has a special `--exit-code` flag to change the exit code behavior ([we already document that](https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-diff))

keerlu:
## Describe this PR

Add exit code documentation to CLI reference

## What issue does this fix?

Fixes #3830

### DIFF
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -1009,8 +1009,8 @@ The migrations from the database are not found locally in prisma/migrations:
 In versions 4.3.0 and later, `prisma migrate status` exits with exit code 1 in the following cases:
 
 - a database connection error occurs
-- there are migration files that have not been applied to the database
-- the migration history has diverged
+- there are migration files in the `migrations` directory that have not been applied to the database
+- the migration history in the `migrations` directory has diverged from the state of the database
 - no migration table is found
 - failed migrations are found
 
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -98,7 +98,7 @@ You can get additional help on any of the `prisma` commands by adding the `--hel
 
 ## Exit codes
 
-All `prisma` commands return the following codes when they exit:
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/200-command-reference.mdx
```diff
@@ -998,6 +1006,14 @@ The migrations from the database are not found locally in prisma/migrations:
 20201208100950_new_migration
 ```
 
+In versions 4.3.0 and later, `prisma migrate status` exits with exit code 1 in the following cases:
+
+- a database connection error occurs
+- there are migration files that have not been applied to the database
+- the migration history has diverged
```
keerlu: Just checking, diverged from what? The state of the database, or the migration table, or something else?

Jolg42: Between the migration history (that is calculated from the `./migrations` directory and the database)
I think many cases can happen here, any difference between the directory and the database state = fail.

Jolg42 (approved):

content/400-reference/200-api-reference/200-command-reference.mdx
keerlu: Thanks, added some words to clarify

PR #3938, opened 2022-10-24
Improve discoverability of Prisma.DbNull/Prisma.JsonNull

### LINKED ISSUE #3800: Improve discoverability of `Prisma.DbNull`/`Prisma.JsonNull`
We should make the [documentation about `Prisma.DbNull` and `Prisma.JsonNull`](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#filtering-by-null-values) more discoverable.

The title of the section refers to "Filtering by null values", but the same also applies to create, insert and upsert operations (only `Prisma.AnyNull` is only used for filtering). This is, in fact, stated in the referenced section of the documentation, however it is hard to find.

This leads to confusion by our users who report expected and intended behavior as a bug: https://github.com/prisma/prisma/issues/9247

ObamaTheLlama114:
Makes the [documentation about Prisma.DbNull and Prisma.JsonNull](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#filtering-by-null-values) more discoverable and clearer about its contents.

Fixes #3800


aqrln (approved):
Looks good to me!

PR #3953, opened 2022-10-26
CLI: add `validate` to CLI ref

Jolg42:
It's available since "a long time" (Since 2.0.0-beta.1, 2019! [see commit](https://github.com/prisma/prisma/commit/fac2b79f7d2be28b07404ddfe91454af45b1bf81)) but was never documented 

https://github.com/prisma/prisma/pull/16008 adds it to the help output

### DIFF
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -458,7 +458,7 @@ From Prisma 3.0.0 onwards, the `prisma introspect` command is deprecated and rep
 
 ### <inlinecode>validate</inlinecode>
 
-Validates the Prisma schema file.
+Validates the [Prisma Schema Language](https://www.prisma.io/docs/concepts/components/prisma-schema) of the Prisma schema file.
 
 #### Arguments
 
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -63,7 +63,8 @@ Commands
               db   Manage your database schema and lifecycle
          migrate   Migrate your database
           studio   Browse your data with Prisma Studio
-          format   Format your schema
+        validate   Validate your Prisma schema
+          format   Format your Prisma schema
 
 Flags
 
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/200-command-reference.mdx
```diff
@@ -455,6 +456,72 @@ From Prisma 3.0.0 onwards, the `prisma introspect` command is deprecated and rep
...
 
+### <inlinecode>validate</inlinecode>
+
+Validates the Prisma schema file.
+
+#### Arguments
+
+| Argument   | Required | Description                                                                                                                                         | Default                                     |
+| ---------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------- |
+| `--schema` | No       | Specifies the path to the desired `schema.prisma` file to be processed instead of the default path. Both absolute and relative paths are supported. | `./schema.prisma`, `./prisma/schema.prisma` |
```
andrew-walford-prisma, line 467: Suggestion 1 (there's a different variant below that you may or may not prefer!)

```suggestion
| `--schema` | No       | Specifies the path to a `schema.prisma` file that will be validated instead of the file at the default path. You can specify an absolute or relative path. | `./schema.prisma`, `./prisma/schema.prisma` |
```

andrew-walford-prisma, line 467: Alternative version, making it active voice:

```suggestion
| `--schema` | No       | Specifies the path to a `schema.prisma` file that Prisma will validate instead of the file at the default path. You can specify an absolute or relative path. | `./schema.prisma`, `./prisma/schema.prisma` |
```

content/400-reference/200-api-reference/200-command-reference.mdx
```diff
@@ -455,6 +456,72 @@ From Prisma 3.0.0 onwards, the `prisma introspect` command is deprecated and rep
 
 </Admonition>
 
+### <inlinecode>validate</inlinecode>
+
+Validates the Prisma schema file.
```
andrew-walford-prisma: I think it would be good to have one or two sentences that briefly explain what validation means. Can we say that if the check passes then your schema file is valid to work in a Prisma project? I presume it doesn't check the file against the current state of the database - it checks the internal syntax of the file, and the model/field naming in the file?

Also maybe we should say that the command either returns a message stating that the schema is valid or that it is invalid (with a list of the errors in the schema). I know that we cover this in the examples below, but I think it would be nice to mention it briefly in the description.

andrew-walford-prisma (commented):
I've made a few suggestions, but it otherwise LGTM. This looks like a very useful feature - I'll be using it when I work on my personal test project!

content/400-reference/200-api-reference/200-command-reference.mdx
Jolg42: Indeed, it only checks that the content of the file is valid.
If it's invalid because of syntax error or any validation error the schema file cannot be used with Prisma Client or Prisma Migrate.

content/400-reference/200-api-reference/200-command-reference.mdx
Jolg42: Not sure if we want to only change this, it's the same text all over the page (10 times)

Jolg42: Not sure if we want to only change this, it's the same text all over the page (10 times)

content/400-reference/200-api-reference/200-command-reference.mdx
Jolg42:
```suggestion
Validates the [Prisma Schema Language](https://www.prisma.io/docs/concepts/components/prisma-schema) of the Prisma schema file.
```

PR #3958, opened 2022-10-27
Fix error where in BigInt type mapping for MySQL

### LINKED ISSUE #3957: BigInt default type mappings table shows INT instead of BIGINT for MySQL
https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#bigint

image

A small error in the BigInt default type mappings that shows `INT` instead of `BIGINT` for the MySQL row. I have verified that the `BigInt` prisma type maps to `BIGINT` MySQL type.

paolotiu:
Fixes #3957


keerlu (approved):

keerlu:
Thanks @paolotiu - sorry this got held up in review

PR #3966, opened 2022-10-31
Added section about unique key constraint errors on upserts

### LINKED ISSUE #3965: Document upsert unique key constraint errors
SME: Garren

Info from Garren: https://github.com/prisma/docs/issues/3952

He recommends two additions to [prisma.io/docs/reference/api-reference/prisma-client-reference#upsert](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#upsert):

Change 1:
- information about handling Unique Key Constraint errors - this can be added to the doc at any time (it's not 4.6 specific, but we should merge before 4.6, so I've tagged it as such)

andrew-walford-prisma:
Added section about unique key constraint errors on upserts

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -954,6 +954,7 @@ const user = await prisma.user.update({
 #### Remarks
 
 - To perform arithmetic operations on update (add, subtract, multiply, divide), use [atomic updates](#atomic-number-operations) to prevent race conditions.
+- If two or more upsert operations happen at the same time, a race condition might happen. As a result, one or more of the upsert operations might throw a unique key constraint error. Your application code can trap this error and retry the operation. [Learn more](#unique-key-constraint-errors-on-upserts).
 
 #### Reference
 
@@ -980,6 +981,25 @@ const user = await prisma.user.upsert({
   create: { email: 'alice@prisma.io' },
 })
 ```
+
+#### Unique key constraint errors on upserts
+
+##### Problem
+
+If multiple upsert operations happen at the same time, then one or more of the operations might throw a [unique key constraint error](/reference/api-reference/error-reference#p2002).
+
+##### Cause
+
+When Prisma does an upsert, it first checks whether that record exists in the database. To make this check, Prisma performs a read operation with the `where` clause from the upsert operation. This has two possible outcomes, as follows:
+
+- If the record does not exist, then Prisma creates that record.
+- If the record exists, then Prisma updates it.
+
+When your application tries to perform two or more upsert operations at the same time, then there might be a race condition where more than operation does not find the record and therefore tries to create that record. In this situation, one of the operations creates the new record. The other operations then fail and return a unique key constraint error.
+
+##### Solution
+
+Trap the P2002 error in your application code. When it occurs, retry the upsert operation to update the row.
 
 ### <inlinecode>delete</inlinecode>
 
@@ -5622,27 +5642,25 @@ model User {
 
 ```ts
 function updateOne() {
+  const user = await prisma.user.findUnique({ id: 1 })
 
-   const user = await prisma.user.findUnique({ id: 1 });
-
-   await prisma.user.update({
-     where: { id: user.id, version: user.version },
-     data: { city: "Berlin", version: { increment: 1 } },
-  });
+  await prisma.user.update({
+    where: { id: user.id, version: user.version },
+    data: { city: 'Berlin', version: { increment: 1 } },
+  })
 }
 
 function updateTwo() {
+  const user = await prisma.user.findUnique({ id: 1 })
 
-   const user = await prisma.user.findUnique({ id: 1 });
-
-   await prisma.user.update({
-     where: { id: user.id, version: user.version },
-     data: { city: "New York", version: { increment: 1 } },
-  });
+  await prisma.user.update({
+    where: { id: user.id, version: user.version },
+    data: { city: 'New York', version: { increment: 1 } },
+  })
 }
 
-function main () {
-  await Promise.allSettled([updateOne(), updateTwo()]);
+function main() {
+  await Promise.allSettled([updateOne(), updateTwo()])
 }
 ```
 
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -954,7 +954,7 @@ const user = await prisma.user.update({
 #### Remarks
 
 - To perform arithmetic operations on update (add, subtract, multiply, divide), use [atomic updates](#atomic-number-operations) to prevent race conditions.
-- If two or more upsert operations happen at the same time and the record doesn't already exist, then a race condition might happen. As a result, one or more of the upsert operations might throw a unique key constraint error. Your application code can trap this error and retry the operation. [Learn more](#unique-key-constraint-errors-on-upserts).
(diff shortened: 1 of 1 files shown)


content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -954,6 +954,7 @@ const user = await prisma.user.update({
 #### Remarks
 
 - To perform arithmetic operations on update (add, subtract, multiply, divide), use [atomic updates](#atomic-number-operations) to prevent race conditions.
+- If two or more upsert operations happen at the same time, a race condition might happen. As a result, one or more of the upsert operations might throw a unique key constraint error. Your application code can trap this error and retry the operation. [Learn more](#unique-key-constraint-errors-on-upserts).
```
garrensmith: The race condition happens if the row doesn't exist. So maybe it should be:
`If two or more upsert operations run at the same time and the row doesn't exist a race condition could happen.`

garrensmith (approved):
This is really nicely explained. My only comment is maybe mention earlier that this only happens if the row doesn't exist. Awesome work Andrew. Thanks.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma:
```suggestion
- If two or more upsert operations happen at the same time and the record doesn't already exist, then a race condition might happen. As a result, one or more of the upsert operations might throw a unique key constraint error. Your application code can trap this error and retry the operation. [Learn more](#unique-key-constraint-errors-on-upserts).
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -981,6 +982,25 @@ const user = await prisma.user.upsert({
 })
 ```
 
+#### Unique key constraint errors on upserts
+
+##### Problem
+
+If multiple upsert operations happen at the same time, then one or more of the operations might throw a [unique key constraint error](/reference/api-reference/error-reference#p2002).
```
andrew-walford-prisma:
```suggestion
If multiple upsert operations happen at the same time and the record doesn't already exist, then one or more of the operations might throw a [unique key constraint error](/reference/api-reference/error-reference#p2002).
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -954,6 +954,7 @@ const user = await prisma.user.update({
 #### Remarks
 
 - To perform arithmetic operations on update (add, subtract, multiply, divide), use [atomic updates](#atomic-number-operations) to prevent race conditions.
+- If two or more upsert operations happen at the same time and the record doesn't already exist, then a race condition might happen. As a result, one or more of the upsert operations might throw a unique key constraint error. Your application code can trap this error and retry the operation. [Learn more](#unique-key-constraint-errors-on-upserts).
```
zachtil: How about using "handle" over "trap"?

"trap" is a bit overloaded on meaning as is "execute" which is frowned upon in Style Guides.

keerlu: Minor point: we don't seem to use 'trap this error' elsewhere in the docs. I'm more familiar with 'catch this error' which we do use, does this still work?

keerlu (approved):
LGTM!

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -981,6 +982,25 @@ const user = await prisma.user.upsert({
 })
 ```
 
+#### Unique key constraint errors on upserts
+
+##### Problem
+
+If multiple upsert operations happen at the same time and the record doesn't already exist, then one or more of the operations might throw a [unique key constraint error](/reference/api-reference/error-reference#p2002).
```
zachtil: How about using "show" or "return" for errors? "throw" is also quite a bit loaded with meaning.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
keerlu: Actually I think I prefer @zachtil's 'handle this error' above

andrew-walford-prisma: OK - I think "trap" is probably the term from when I learned to code, back in the days of difference engines. I'll change it.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -981,6 +982,25 @@ const user = await prisma.user.upsert({
...
+
+##### Cause
+
+When Prisma does an upsert, it first checks whether that record already exists in the database. To make this check, Prisma performs a read operation with the `where` clause from the upsert operation. This has two possible outcomes, as follows:
+
+- If the record does not exist, then Prisma creates that record.
+- If the record exists, then Prisma updates it.
+
+When your application tries to perform two or more concurrent upsert operations, then a race condition might happen where two or more operations do not find the record and therefore try to create that record. In this situation, one of the operations succesfully creates the new record but the other operations fail and return a unique key constraint error.
```
zachtil, line 998: Just wondering: do we need commas around 'therefore'?

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma: Point taken - I'll update it

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -981,6 +982,25 @@ const user = await prisma.user.upsert({
...
+
+##### Solution
+
+Trap the P2002 error in your application code. When it occurs, retry the upsert operation to update the row.
```
zachtil: Let's discuss "handle" vs "trap"

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma:
```suggestion
- If two or more upsert operations happen at the same time and the record doesn't already exist, then a race condition might happen. As a result, one or more of the upsert operations might throw a unique key constraint error. Your application code can catch this error and retry the operation. [Learn more](#unique-key-constraint-errors-on-upserts).
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma: We do use "throw" in the docs, and indeed our API uses the term. However, I think "return" is also clear, so I'll switch to that.

andrew-walford-prisma:
```suggestion
If multiple upsert operations happen at the same time and the record doesn't already exist, then one or more of the operations might return a [unique key constraint error](/reference/api-reference/error-reference#p2002).
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma: That's a stylistic choice here. I aim for fewer commas, rather than more.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
andrew-walford-prisma:
```suggestion
Handle the P2002 error in your application code. When it occurs, retry the upsert operation to update the row.
```

PR #3994, opened 2022-11-03
update pascalcase link to go to pascalcase

mrhappyma:
Updated the Wikipedia link for pascal case to go to the pascal case page. This currently redirects to camel case anyway, however it seemed more appropriate to use the pascal case link instead.


ObamaTheLlama114:
Thank you for your contribution @mrhappyma

PR #4006, opened 2022-11-04
Fix: Remove localhost in link

janpio:


keerlu:
Oops. Thanks very much for catching this!

keerlu (approved):

PR #4008, opened 2022-11-04

### LINKED ISSUE #4002: Docs state (incorrectly) that adding a full-text index in Postgres will improve performance
Title says it all, really. The docs currently state that adding a full-text search index in Postgres will improve performance, but those indexes aren't actually used, so any effort creating them is wasted right now. Probably we should just remove that section from the docs, or add a big bold caveat that following those instructions won't actually do anything to speed up your queries?

I'm happy to push a PR updating the docs, just want some guidance on which change to make (e.g remove the section altogether or add a caveat warning)

Docs link: https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search#postgresql-1
Github issue showing that indexes are not being used: https://github.com/prisma/prisma/issues/8950#issue-980473627
Slack conversation about this: https://prisma.slack.com/archives/CA491RJH0/p1666896107795639

dickfickling:
resolves #4002.

The docs currently state that adding a full-text search index in Postgres will improve performance, but those indexes aren't actually used, so any effort creating them is wasted right now. Probably we should just remove that section from the docs.

Docs link: https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search#postgresql-1
Github issue showing that indexes are not being used: https://github.com/prisma/prisma/issues/8950#issue-980473627
Slack conversation about this: https://prisma.slack.com/archives/CA491RJH0/p1666896107795639

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
@@ -170,19 +170,7 @@ For the full range of supported operations, see the [MySQL full text search docu
 
 ### PostgreSQL
 
-To speed up your full-text queries, you should add an index to your database. Prisma schema and Prisma Migrate currently don't support defining and creating PostgreSQL search indices, but you can add one yourself. The following example adds a `GIN` index to the table `posts` on the field `body`:
-
-```sql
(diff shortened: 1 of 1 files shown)


keerlu:
Thanks for removing this, @dickfickling

PR #4011, opened 2022-11-07
feat: clarify that MySQL's SetDefault referential action does not work as expected

### LINKED ISSUE #3885: Document MySQL < 8 / MariaDB < 10.3 does not support the referential action `onDelete: setDefault`
Based on https://github.com/prisma/prisma/issues/11498

MySQL < 8 / MariaDB < 10.3 (not 100% sure for that version) does not support the referential action `onDelete: setDefault`

It errors with `Error: Cannot add foreign key constraint`.

TODO:
- update https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions with version incompatibility

jkomyno:
## Describe this PR

MySQL doesn't support `SET DEFAULT` referential actions but we do mark them as valid in the Prisma schema.

## Changes

Add a clarification in the `Referential Actions` section about MySQL's weird support for the `SET DEFAULT` referential action.

## What issue does this fix?

Contributes to https://github.com/prisma/prisma/issues/11498.
Closes https://github.com/prisma/docs/issues/3885


content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
```diff
@@ -159,6 +159,9 @@ The following table shows which referential action each database supports.
 | MongoDB††  | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **❌**     |
 
 - † On [InnoDB](https://dev.mysql.com/doc/refman/5.6/en/innodb-introduction.html#:~:text=InnoDB%20is%20a%20general%2Dpurpose,clause%20creates%20an%20InnoDB%20table.), `SetDefault` is ignored and is silently replaced with `NoAction` instead.
+  On MySQL 8.0+, when a `SetDefault` referential action is triggered, a foreign key violation error is raised.  
+  On older versions, like MySQL 5.6+, creating or altering a table with a `SetDefault` referential action is not possible at all.
+  We therefore advise against using `SetDefault` at all on MySQL.
```
janpio: Hm, the wording here is a bit weird still.

I think what we _want_ to highlight is that `SetDefault` is accepted and supported, but actually is not implemented in MySQL's InnoDB and instead a) fails on DDL SQL execution in 5.6 and 5.7, and behaves like `NoAction` and `Restrict` in 8.x - correct?

(Which reminds me: We probably also want to add that Restrict and NoAction are identical here in this table via a footnote as well)

jkomyno: `Restrict` and `NoAction` are identical (aliases) in:
- `mysql`
- `cockroachdb`

In the other databases that support both actions, the're behaviorally equivalent only when executing a single statement in a transaction (but I would probably not include this last info, as it may be a source of confusion for the users).

janpio: Yep exactly.

So we need to 1) refine the `SetDefault` wording, and 2) either in this PR or a separate one add a clarification in the table about the MySQL and Cockroach situation on `NoAction` being an alias. as they behave the same.

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
```diff
@@ -150,16 +150,18 @@ The following caveats apply:
 
 The following table shows which referential action each database supports.
 
-| Database    | Cascade | Restrict | NoAction | SetNull | SetDefault |
-| :---------- | :------ | :------- | :------- | :------ | :--------- |
-| PostgreSQL  | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
-| MySQL       | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
-| SQLite      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
-| SQL Server  | **✔️**  | **❌**‡  | **✔️**   | **✔️**  | **✔️**     |
-| CockroachDB | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
```
keerlu: Is this table supposed to have had Cockroach removed?

janpio: No, that was a merge fuckup from my side I think.

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
```diff
@@ -150,16 +150,18 @@ The following caveats apply:
...
-| MongoDB††   | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **❌**     |
+| Database   | Cascade | Restrict | NoAction | SetNull | SetDefault |
+| :--------- | :------ | :------- | :------- | :------ | :--------- |
+| PostgreSQL | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
+| MySQL      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
+| SQLite     | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
+| SQL Server | **✔️**  | **❌**‡  | **✔️**   | **✔️**  | **✔️**     |
```
janpio: | CockroachDB | **✔️** | **✔️** | **✔️** | **✔️** | **✔️** |

janpio:
```suggestion
| SQL Server | **✔️**  | **❌**‡  | **✔️**   | **✔️**  | **✔️**     |
| CockroachDB | **✔️** | **✔️** | **✔️** | **✔️** | **✔️** |
```

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
janpio: Fixed

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
```diff
@@ -150,16 +150,19 @@ The following caveats apply:
...
+| CockroachDB | **✔️** | **✔️** | **✔️** | **✔️** | **✔️** |
+| MongoDB††  | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **❌**     |
```
janpio:
```suggestion
| Database    | Cascade | Restrict | NoAction | SetNull | SetDefault |
| :---------- | :------ | :------- | :------- | :------ | :--------- |
| PostgreSQL  | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
| MySQL       | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
| SQLite      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
| SQL Server  | **✔️**  | **❌**‡  | **✔️**   | **✔️**  | **✔️**     |
| CockroachDB | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**    |
| MongoDB††   | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **❌**     |
```

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
```diff
@@ -156,10 +156,13 @@ The following table shows which referential action each database supports.
 | MySQL       | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**†    |
 | SQLite      | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
 | SQL Server  | **✔️**  | **❌**‡  | **✔️**   | **✔️**  | **✔️**     |
-| CockroachDB | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
+| CockroachDB | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**    |
```
janpio:
```suggestion
| CockroachDB | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **✔️**     |
```

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
```diff
@@ -160,6 +160,9 @@ The following table shows which referential action each database supports.
 | MongoDB††   | **✔️**  | **✔️**   | **✔️**   | **✔️**  | **❌**     |
 
 - † On MySQL, `SetDefault` is ignored and is silently replaced with `NoAction` instead.
+  On MySQL 8.0+, when a `SetDefault` referential action is triggered, a foreign key violation error is raised.  
+  On older versions, like MySQL 5.6+, creating or altering a table with a `SetDefault` referential action is not possible at all.
+  We therefore advise against using `SetDefault` at all on MySQL.
```
janpio, line 165: See earlier comment thread: https://github.com/prisma/docs/pull/4011#discussion_r1015765067

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
keerlu: Do we know exactly which version of MySQL has the transition between failing and having the same behaviour as `NoAction` and `Restrict`?

I'd want to rewrite that part of the note as something like:

>  In MySQL versions X.X and later, the `SetDefault` referential action has the same behavior as `NoAction` or `Restrict`. In earlier versions, it is not possible to create or alter a table with a `SetDefault` referential action.

once we know the correct version.

janpio: 8.x as far as we know.

content/200-concepts/100-components/01-prisma-schema/06-relations/410-referential-actions/index.mdx
jkomyno: How about:

> While MySQL 8.0+ accepts table definitions with the `SET DEFAULT` referential action (i.e., `SetDefault` in a Prisma schema definition), this action is actually not supported by InnoDB - MySQL's default storage engine and the only one Prisma supports. In fact, InnoDB silently interprets `SET DEFAULT` as `NO ACTION`. This is evident when an update or deletion triggers a `SetDefault` referential action, as InnoDB would raise the same foreign key violation error you would see with `NoAction`.
> On older versions, like MySQL 5.6+, creating or altering a table with a `SetDefault` referential action is not possible at all, as it would fail with a syntax error when attempting to push the Prisma schema to the database.
> We therefore advise against using `SetDefault` at all on MySQL.

janpio: We might want to figure out with @keerlu if there is a better way to go about this. 

We need to list `SetDefault` as working in the table - but we also need to explain people that it does not _actually_ work properly. So maybe ✔️/❌? 

Also, having this full explanation in a footnote seems suboptimal. Maybe another section somewhere that explains these "weird" cases where we do not just want to put " ✔️ " in the table?

janpio:
Superseeded by https://github.com/prisma/docs/pull/4132

Issue #3813, opened 2022-09-22
Consider improving our descriptions of filter conditions.


andrew-walford-prisma:
@janpio reports that our docs are confused/confusing on some of the filter conditions. They don't clearly differentiate `not` and `NOT`, and so on.

The filter conditions material might need to be substantially reorganized, and we might need to write some definitions of some filter conditions that the docs currently don't define. There are also some specific minor issues he's highlighted, such as links going to the wrong place. See the Slack thread for details.

Slack thread: 

https://prisma-company.slack.com/archives/C5Z9TH6N9/p1663861444496519

jharrell:
👋 we're cleaning up our backlog of issues and are closing everything older than two years. If you still feel that this issue is relevant, please feel free to re-open and our team will take a look. Thank you!

PR #4015, opened 2022-11-08
fix NOT client API links in filtering-and-sorting and null-and-undefined docs

arkinmodi:
## Describe this PR

The links to the `NOT` client API are incorrect on pages https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting and https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined. 
## Changes

Link have been updated to the correct `...#not-1`.

## What issue does this fix?

I assume these are part of the incorrect links mentioned in https://github.com/prisma/docs/issues/3813


andrew-walford-prisma (approved):

andrew-walford-prisma:
Thanks for this fix, @arkinmodi

PR #4037, opened 2022-11-09
Docs: excluding fields section, code typo

18jad:
## Describe this PR

The keys parameter should be an array but a string is passed and it's possible cause it's being used with the spread operator it shouldn't cause in this case you are only able to delete 1 key and the loop is redundant, correct me if I am wrong but in my project, it didn't work deleting multiple values
## Changes

- Refactored parameters and function use example

## What issue does this fix?

Unable to exclude multiple keys

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
@@ -24,7 +24,7 @@ The following is a type-safe `exclude` function returns a user without the `pass
 // Exclude keys from user
 function exclude<User, Key extends keyof User>(
   user: User,
-  ...keys: Key[]
+  keys: Key[]
 ): Omit<User, Key> {
   for (let key of keys) {
     delete user[key]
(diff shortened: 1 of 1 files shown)


keerlu:
Thanks, @18jad, we'll get a technical review for this and get back to you.

millsp:
Hey @18jad I see that the auto-completion breaks, however typescript does not report any error if I pass `b` in the last parameter.
Screenshot from 2022-11-14 06-25-49

While it still works, I agree it is not ideal from a DX point of view. I think it would be interesting for you to raise this as an issue at the TypeScript repository directly (or if you want me to do it, let me know). Arrays and spread arrays should work the same IMO when it comes to autocompletion, especially if inference is working correctly. One workaround can be done like [so](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgMTgXzgMyhEcBEMAzgLQwQQA2ARgKaUz4BQTmArgHYDGMwEHcWgA8ulNgBNaAHgCqRWlAA0cANK0AnoKExaHcUTgBrDRExw5CgHwAKJnDht5UAFzmniu3AB0P4+qKuyF4AchAAkhyYClJq6pYA2gC6TACUrgDyIMAwsu6qGpaInpjQcNaUtPB+cKZGGkQpRfb2khU6Dk7xfsn2aJ5QlWxQAo4KTH1AA), but I think your solution is more than enough for the example.

millsp (approved):

PR #4072, opened 2022-11-16
fix(database-features): Add some more indexes

janpio:


content/400-reference/300-database-reference/01-database-features.mdx
```diff
@@ -59,8 +59,10 @@ This section describes which database features exist on the relational databases
...
+| `USING`  | ✔️         | No                   | No    | No     | ✔️          | [`type`](/concepts/components/prisma-schema/indexes#configuring-the-access-type-of-indexes-with-type-postgresql)<sup>†</sup> | ✔️            | ✔️<sup>†</sup> |
 | `WHERE`  | ✔️         | ✔️                   | ✔️    | ✔️     | ✔️          | Not yet                                                                                                                      | ✔️            | Not yet        |
-| `USING`  | ✔️         | ✔️                   | ✔️    | ✔️     | ✔️          | [`type`](/concepts/components/prisma-schema/indexes#configuring-the-access-type-of-indexes-with-type-postgresql)<sup>†</sup> | ✔️            | Not yet        |
+| `(expression)` | ✔️         | ✔️                   | ✔️    | ✔️     | ✔️         | Not yet                                                                                                                   | ✔️            | Not yet        |
```
keerlu, line 54: Does this refer to something like https://www.postgresql.org/docs/current/indexes-expressional.html?

If so maybe 'Indexes on expressions' would be clearer?

keerlu (approved):

content/400-reference/300-database-reference/01-database-features.mdx
janpio: Yes, I think that is what I meant back then. They are just called "Expression indexes" as well sometimes, but in the SQL they are marked by being in `( ... )` - that might have been my thought back in November.

PR #4074, opened 2022-11-17
Update 100-one-to-one-relations.mdx

raamponsah:


raamponsah (commented):
"a user can have zero or one profiles"
I think this statement could be a bit confusing as a user cannot have one profiles(plural) besides we are dealing with a one-to-one relation

PR #4076, opened 2022-11-17
PDP: Remove blurring from screenshots

### LINKED ISSUE #4075: PDP: Remove blurring from any screenshots
Several PDP documentation pages use blurring in screenshots. Because there are available algorithms to unblur screenshots, we should remove existing blurring and replace that with redactions.

zachtil:
## Describe this PR

As blurring in screenshots can be "reverse engineered", the goal is to apply redactions from now on in screenshots.

## Changes

Updates links to screenshot images

## What issue does this fix?

Fixes #4075


andrew-walford-prisma (commented):
I like the colour you choose - it's unobtrusive, but still leaves it clear that there was something in that field.

PR #4079, opened 2022-11-17
Improve Customising Migrations page structure

### LINKED ISSUE #4078: Improve structure for Customising Migrations page
From @janpio:

> https://www.prisma.io/docs/guides/database/developing-with-prisma-migrate/customizing-migrations the main content how to edit a migration does not have a headline and hence does not appear in the sidebar on the right. I think all other 3 headline are examples, so maybe "Example usecases" as a higher level headline for them would make the page structure clearer.
>
> The 2 SQL examples at the bottom also have the problem that the - is interpreted as a diff - so we should add an empty line before it I think.

keerlu:
## Describe this PR

Improve page structure

## What issue does this fix?

Fixes #4078

### DIFF
--- a/content/300-guides/050-database/100-developing-with-prisma-migrate/170-customizing-migrations.mdx
+++ b/content/300-guides/050-database/100-developing-with-prisma-migrate/170-customizing-migrations.mdx
@@ -1,6 +1,8 @@
 ---
-title: Customizing migrations
+title: Customize a migration file
+metaTitle: Customize a migration file
 metaDescription: How to edit a migration file before applying it to avoid data loss in production.
+tocDepth: 2
 ---
 
 <TopBlock>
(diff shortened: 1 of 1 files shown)


content/300-guides/050-database/100-developing-with-prisma-migrate/170-customizing-migrations.mdx
```diff
@@ -30,9 +38,7 @@ To edit a migration file before applying it:
    npx prisma migrate dev
    ```
 
-</TopBlock>
-
-## Renaming a field
+### Example: Rename a field
```
janpio, line 41:
```suggestion
## Examples

### Example: Rename a field
```

janpio (approved):

content/300-guides/050-database/100-developing-with-prisma-migrate/170-customizing-migrations.mdx
keerlu: I prefer these as subsections of the general 'how to edit topic'

PR #4080, opened 2022-11-17
Add explanation about cumulative histograms

### LINKED ISSUE #4068: Ensure cumulative histograms are covered in the Metrics docs
In Prisma metrics, our histogram data is cumulative (the term is "cumulative histogram"). We do mention cumulative data as a type of histogram data. We need to confirm with Garren whether all histogram data is cumulative or if it's this type.

In the docs, we need to make the explanation clear what this means: e.g. if a query takes 5 seconds, it goes into the 1 second bucket, the 2 seconds bucket, 3, 4 and 5. In other words, all buckets up to and including the ones whose highest bound applies.

Once that's cleared up, we need to ensure the examples are correct.

I've confirmed the following with Garren:
- In our [docs](https://www.prisma.io/docs/concepts/components/prisma-client/metrics), we describe three types of histogram. Is the doc correct that there is one cumulative histogram type, and two other types (total and count)? - Answer: the doc is correct
- If you have five buckets for 1, 2, 3, 4, and 5 seconds, and an incoming value is 4 seconds, then buckets for 1, 2, 3, and 4 seconds are incremented and the bucket for 5 isn’t. Is that correct? Answer: this is correct

andrew-walford-prisma:
Add a section to the bullet point on cumulative histograms that explains how the cumulative nature works

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/240-metrics.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/240-metrics.mdx
@@ -379,15 +379,14 @@ prisma_datasource_queries_duration_histogram_ms_sum 1.8407059999999997
 prisma_datasource_queries_duration_histogram_ms_count 5
 ```
 
-
 Metrics of type `histogram` expose three different class of values in the Prometheus format:
-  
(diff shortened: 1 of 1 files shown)


garrensmith (approved):
+1 nice explanation

PR #4088, opened 2022-11-20
Added new real world project sample （Fintech GraphQL API Template）

Mingyang-Li:
## Describe this PR
✅ Added new sample project built using `Prisma`

✅ Project GitHub repo: https://github.com/Mingyang-Li/nest-graphql-prisma
✅ Deployed project URL: https://nest-graphql-prisma-g5s7.onrender.com/graphql

### ✋ Reason for this PR:
Too many template/boilerplate projects are not built for real-world use-bases, especially when it comes to GraphQL APIs. No one needs a single resolver that simply returns `Hello world` (like many boilerplate and tutorials provide you with). Instead, beginners need an established way to build highly customisable & filterable `query` and `mutation` resolvers for basic CRUD operations, which is why I not only built, but also deployed [my project](https://github.com/Mingyang-Li/nest-graphql-prisma), hopefully it'll help more people to be less confused about how to optimise their CRUD logics.


nikolasburk:
Hey @Mingyang-Li, thanks a lot for opening this PR!

It seems like this project is more of a "starter project" than an actual "real-world app", so I think we'll not include it in this list at this time.

We're currently thinking about creating a list of starter projects though, so once we have that we'd be happy to add yours to that list (assuming your starter project will be maintained 🙂 ).

PR #4106, opened 2022-11-22
Update index.mdx

dumb-programmer:
Correct sentence in Disambiguating relations section, the sentence was written as "When you define two relations between two the same models", instead of "When you define two relations between the same two models"

## Describe this PR

Incorrect sentence in Disambiguating relations section, currently written as "When you define two relations between two the same models", while it should be "When you define two relations between two the same models"

## Changes

Correct sentence in Disambiguating relations section

- When you define two relations between two the same models
- ... --> When you define two relations between the same two models


janpio (approved):

keerlu:
Thanks for the fix, (not so) @dumb-programmer!

PR #4126, opened 2022-11-24
Update Prisma Client page metaTitle and metaDescription

### LINKED ISSUE #4125: Remove "(Reference)" from Prisma Client getting started page
The Prisma Client page (under Concepts > Components) incorrectly contains "(Reference)" in the metaTitle.

Size: S

zachtil:
## Describe this PR

Remove "(Reference)" from the `metaTitle` of the Prisma Client page. Also update the `metaDescription` to be a bit more precise in its description.

## Changes

Changes to `metaTitle` and `metaDescription` frontmatter attributes in `/content/200-concepts/100-components/02-prisma-client/index.mdx`

## What issue does this fix?

Fixes #4125

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/index.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/index.mdx
@@ -1,7 +1,7 @@
 ---
 title: 'Prisma Client'
-metaTitle: 'Prisma Client (Reference)'
-metaDescription: 'Prisma Client is an auto-generated, type-safe query builder generated from the tables in your database.'
+metaTitle: 'Prisma Client'
+metaDescription: 'Prisma Client is an auto-generated, type-safe query builder generated based on the models and attributes of your Prisma schema.'
 ---
 
 <TopBlock>


andrew-walford-prisma (approved):
LGTM

content/200-concepts/100-components/02-prisma-client/index.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: 'Prisma Client'
-metaTitle: 'Prisma Client (Reference)'
-metaDescription: 'Prisma Client is an auto-generated, type-safe query builder generated from the tables in your database.'
+metaTitle: 'Prisma Client'
```
janpio, line 3: Is this the only page we have with "Prisma Client" as title?
https://www.prisma.io/client has a bit more at the end, so should be an ok overlap.

zachtil: @janpio Yes, this is the only page that has `title: Prisma Client` as of this PR.



Not sure I understand the connection between this PR and the website URL for the client.

Do you mean that the web page has this <title> element?

`<title>Prisma Client - Auto-generated query builder for your data</title>`

Or something else?

janpio: Generally a domain should not have 2 pages with the same page title. As the `/client` one adds a bunch of stuff at the end we should be good. Otherwise it would have been and indication to get a bit more concrete in _this_ page title here in this page. Technically this is not "Prisma Client" but documentation of the Concept Prisma Client.

zachtil: Understood. Thanks for the explanation. Merging.

PR #4134, opened 2022-11-25
workaround when using pnpm/pnpx

delight:
## Describe this PR

When working with pnpm/pnpx there is a problem while prisma generate is run. 

```
npm ERR! Cannot read properties of null (reading 'matches')
```
workaround is to do a pnpm i  @prisma/client before running migrate (generate)

## Changes

Adds a little side note into the quickstart 

## What issue does this fix?
See PR description

don't know if that is the right place to add this information - but it certainly would help others running into this problem

### DIFF
--- a/content/100-getting-started/01-quickstart.mdx
+++ b/content/100-getting-started/01-quickstart.mdx
@@ -107,7 +107,7 @@ Congratulations, you now have your database and tables ready. Let's go and learn
 
 ### Note on the usage with pnpm
 
-if you use pnpm/pnpx instead of npm/npx, you need to install @prisma/client before running prisma migrate.
+If you use pnpm/pnpx instead of npm/npx, you need to install @prisma/client before running prisma migrate.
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
This is actually related to a bug in Prisma CLI, so we should fix that.

delight:
> This is actually related to a bug in Prisma CLI, so we should fix that.

Is there an issue for that bug ?

janpio:
Multiple: https://github.com/prisma/prisma/issues?q=is%3Aopen+label%3A%22topic%3A+pnpm%22+sort%3Aupdated-desc+

If you think your situation is not represented or could be better, feel free to open a new one though.

nikolasburk:
@delight thanks for opening this PR. Since it seems like this is a bug in the Prisma CLI, I'm closing this PR. Let me know if you still feel it's relevant and we can re-open it 🙏 

PR #4162, opened 2022-11-30
Removed unnecessary closed parentheses

mmarcon:
## Describe this PR

Looks like the extra `)` might be the result of double closing parentheses in the Markdown source, so I am suggesting removing it.

## Changes

Removed extra `)`


janpio (approved):

janpio:
Thanks, I remember how this happened - I copied an already linked string into the URL part of the link and then seemed to have messed up when fixing it.

PR #4166, opened 2022-11-30
docs: update the example to include raw queries

lirantal:
## Describe this PR

Updating the documentation to also include raw queries as a reference being documented and how to use it with an existing seed.js/seed.ts file

## Changes

A documentation change

## What issue does this fix?

No specific issue opened for this.


lirantal:
@zachtil can you take a look at this docs one too?

lirantal:
Thanks for pushing. Let me know if there are any updates needed here.

PR #4179, opened 2022-12-01
Proposed wording change

andrew-walford-prisma:
Proposed wording change (disambiguation)


janpio:
I think these examples are extra confusing because the field that points to `User` is not called `userId`, but `authorId`.

PR #4188, opened 2022-12-05
Add warning note for `shadowDatabaseUrl`.

rmariuzzo:
## Describe this PR

As a developer I made the mistake to put the same value for both `url` and `shadowDatabaseUrl`. This lead to situation where the Prisma's migration table disappeared several times resulting in an inconsistent app state.

## Changes

- Added a note.

## What issue does this fix?

Can be useful for other devs.

Probably, we could expand and tell concisely what happens when we put the same value.

### DIFF
--- a/content/200-concepts/100-components/03-prisma-migrate/200-shadow-database.mdx
+++ b/content/200-concepts/100-components/03-prisma-migrate/200-shadow-database.mdx
@@ -98,6 +98,8 @@ datasource db {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


janpio:
I created an issue that we actually forbid that, so we will not need that warning when this is implemented: https://github.com/prisma/prisma/issues/16628

andrew-walford-prisma (approved):

andrew-walford-prisma:
Even though we're going to fix the issue in the product, I've merged this docs PR, so that this information is available in the interrim. Thans @rmariuzzo

PR #4196, opened 2022-12-06
Created info for percent encoding

rigba:
## Describe this PR

It's not mentioned in the docs how, when, why someone would need to percent encode characters in their database uri

## Changes

I added a small indent to describe why and when someone would need to percent encode characters in their connection url and provided a link to explain how.

## What issue does this fix?

I ran into this issue myself and noticed other people were having this problem and posting it on the repos discussions and creating issues. Assuming they read the docs properly people wont need to encounter the issue.

undefined


andrew-walford-prisma (approved):
Thanks for your contribution @rigba . We do explain this on the [connection URLs page](https://www.prisma.io/docs/reference/database-reference/connection-urls#special-characters). However, I think it's also good to note this in the guides, so I'll merge your PR. Thanks again.

PR #4197, opened 2022-12-06
Remove duplicative verb relations

smichae:
Removes one of two verbs that make the sentence a little confusing.


janpio (approved):

keerlu:
Thanks for spotting, @smichae!

PR #4205, opened 2022-12-07
Various improvements to extensions docs

### LINKED ISSUE #4150: Client extensions - post-release docs improvements part 1
There are a few outstanding tasks to resolve that came up in the docs reviews, but we ran out of time to do.

I've tagged this as 4.8.0, but we can merge these changes - when they're done - at any time.

SMEs: Pierre, Serhii

- [x]  Add links to the extensions docs from Middlewares doc. Also, from any others?
- [x] - Add description of 'Prisma.getExtensionContext'. See [here](https://github.com/prisma/prisma/issues/16500#issuecomment-1337070719).
- [x] We need a note in the docs to cover the case where if you call `query` twice in a row, it executes the query twice. See [here](https://prisma-company.slack.com/archives/C043X6JBND6/p1669864680536019). The example, [here](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/query#add-or-modify-a-custom-query-for-a-specific-operation-in-a-specific-model), I think is misleading.  Pierre has [opened a PR](https://github.com/prisma/docs/pull/4184/files) to fix this specific issue, so I'm ticking this one off.

andrew-walford-prisma:
- Add link to client extensions docs from middleware docs
- Fix link to middleware page in extensions docs
- Model: add section about how to call a method in the same extension or another extension

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
@@ -6,6 +6,12 @@ metaDescription: 'Prisma Client middleware allows you to perform actions before 
 
 <TopBlock>
 
+<Admonition type="info">
+
+From version 4.7.0, you can use the [Prisma Client extensions `query` component type](/concepts/components/prisma-client/client-extensions/query) as an alternative to middleware. You can create independent clients and bind each client to a specific filter or user, for example to provide user isolation. Client extensions also provide end-to-end type safety.
+
+</Admonition>
+
 Middlewares act as query-level lifecycle hooks, which allow you to perform an action before or after a query runs. Use the [`prisma.$use`](/reference/api-reference/prisma-client-reference#use) <span class="api"></span> method to add middleware, as follows:
 
 ```ts highlight=4-9,12-17;normal
--- a/content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
@@ -8,7 +8,7 @@ metaDescription: 'Prisma Client middleware allows you to perform actions before 
 
 <Admonition type="info">
 
-From version 4.7.0, you can use the [Prisma Client extensions `query` component type](/concepts/components/prisma-client/client-extensions/query) as an alternative to middleware. You can create independent clients and bind each client to a specific filter or user, for example to provide user isolation. Client extensions also provide end-to-end type safety.
+From version 4.7.0, you can use the [Prisma Client extensions `query` component type](/concepts/components/prisma-client/client-extensions/query) as an alternative to middleware. You can create independent clients and bind each client to a specific filter or user. For example, you could bind clients to specific users to provide user isolation. Client extensions also provide end-to-end type safety.
 
 </Admonition>
 
--- a/content/200-concepts/100-components/02-prisma-client/260-client-extensions/100-model.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/260-client-extensions/100-model.mdx
@@ -107,11 +107,11 @@ xprisma.user.validate(someData)
 xprisma.post.validate(someData)
 ```
 
-## Call a method in the same extension or in another extension
+## Call a different custom method in the same model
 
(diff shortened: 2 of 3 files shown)


content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
```diff
@@ -6,6 +6,12 @@ metaDescription: 'Prisma Client middleware allows you to perform actions before
 
 <TopBlock>
 
+<Admonition type="info">
+
+From version 4.7.0, you can use the [Prisma Client extensions `query` component type](/concepts/components/prisma-client/client-extensions/query) as an alternative to middleware. You can create independent clients and bind each client to a specific filter or user, for example to provide user isolation. Client extensions also provide end-to-end type safety.
```
zachtil:
```suggestion
From version 4.7.0, you can use the [Prisma Client extensions `query` component type](/concepts/components/prisma-client/client-extensions/query) as an alternative to middleware. You can create independent clients and bind each client to a specific filter or user, for example, to provide user isolation. Client extensions also provide end-to-end type safety.
```

zachtil: Also, add a comma after "for example"?

zachtil: Does it make sense to introduce this more abstractly and then have the example sentence, such as:

"You can create independent clients and bind each to a specific query or model. For example, you can bind an independent client to the `user` model to provide user isolation."

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
```diff
@@ -9,7 +9,7 @@ tocDepth: 3
...
+You can use the Prisma Client extensions `query` component to create independent clients. This provides an alternative to [middlewares](/concepts/components/prisma-client/middleware). You can bind one client to a specific filter or user, and another client to another filter or user. For example, you might do this to get [user isolation](/concepts/components/prisma-client/client-extensions#extended-clients) in a row-level security (RLS) extension. In addition, unlike middlewares the `query` extension component gives you end-to-end type safety. [Learn more about `query` extensions versus middlewares](#query-extensions-versus-middlewares).
```
zachtil, line 12: General question: I've read before that we refer to `query`, `model`, `client`, and `result` as component types, and here you mention just "the `query` component".

I am not 100% in a good understanding of the topic, but "component type" feels harder to comprehend than just "component". Is there a difference?

content/200-concepts/100-components/02-prisma-client/053-middleware/index.mdx
andrew-walford-prisma: Thanks - I like your second approach, because it breaks this fairly complex sentence up.

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
andrew-walford-prisma: We say "component type" when we want to make it clear that we don't mean a component instance. However, we might use "component type" a bit too often in these docs when we could actually just say "component".

content/200-concepts/100-components/02-prisma-client/260-client-extensions/100-model.mdx
```diff
@@ -106,3 +106,30 @@ You would call `validate` in your application as follows:
 xprisma.user.validate(someData)
 xprisma.post.validate(someData)
 ```
+
+## Call a custom method from another custom method
+
+You can call a custom method from another custom method, if the two methods are declared on the same model. For example, you can call a custom method on the `user` model from another custom method on the `user` model. It does not matter if the two methods are declared in the same extension or in different extensions.
+
+To do so, use `Prisma.getExtensionContext(this).methodName`. Note that you cannot use the form `prisma.user.methodName`.
```
SevInf: Would be worth explaining, that `prisma` in this case refers to the instance that haven't been extended yet and therefore won't contain `firstMethod`

andrew-walford-prisma: Good point, I'll add this

PR #4209, opened 2022-12-07
fix: Code highlight on `--` 

### LINKED ISSUE #4138: Formatting bug with code snippets starting with a hyphen
Adding a hyphen at the beginning of a code block causes the first line to be displayed as a deleted line. For example, the following code:

```
-- Test

CREATE TABLE "User" (
    id SERIAL PRIMARY KEY
);
CREATE TABLE "Profile" (
    id SERIAL PRIMARY KEY,
    "userId" INTEGER NOT NULL UNIQUE,
    FOREIGN KEY ("userId") REFERENCES "User"(id)
);
```

looks like this:

image

This is particularly a problem with SQL snippets as hyphens are used to start comments.

nilubava:
## Describe this PR
Enable double hyphens in code blocks

## Changes
Check for double hyphens in code block line starts and exclude from highlighting

## What issue does this fix?
Fixes #4138


keerlu:
Just tested and looks good now:

image

keerlu (approved):

PR #4211, opened 2022-12-07
docs: clarify when env-vars are read

mleimer:
## Describe this PR

Clarification about when the connection URL is being read. Potentially different environment variables could be used between build and run time, as env variables get re-read from the env-file.

## Changes

Add clarification about when the connection URL is read for which purpose.

## What issue does this fix?

Make documentation clearer


content/400-reference/300-database-reference/02-connection-urls.mdx
```diff
@@ -107,6 +107,8 @@ datasource db {
 
 You can then either set the environment variable in your terminal or by providing a [dotenv](https://github.com/motdotla/dotenv) file named `.env`. This will automatically be picked up by the Prisma CLI.
 
+The connection URL is read from the env-file during build time to generate the client as well as during run time to connect to the actual database.
```
janpio: In which way is this relevant for Client during generate/build time?

mleimer: It is important to understand that the env-variable is re-read during run time and that said variable could, but shouldn't be different. This is not clear from the current documentation, as one may think that it is sufficient to pass the env variable during build time. 

Consider a Next.JS application deployed to Vercel. During build time, I can pass the DATABASE_URL, but I also need to make sure that I have a valid DATABASE_URL set for run time on Vercel itself.

If I use the CLI or GitHub Actions to deploy to Vercel, then I need to be aware that I can have at most two different DATABASE_URL configured on Vercel (preview & production), as branch-specific env variables are not supported in that case.

janpio: Yeah, but why do you need a `DATABASE_URL` value at all during `generate`/build?

mleimer: You have a point there. I have updated the documentation, by specifying that it is used while updating the prsima.schema when pulling the latest schema from the DB

content/400-reference/300-database-reference/02-connection-urls.mdx
```diff
@@ -107,6 +107,8 @@ datasource db {
 
 You can then either set the environment variable in your terminal or by providing a [dotenv](https://github.com/motdotla/dotenv) file named `.env`. This will automatically be picked up by the Prisma CLI.
 
+The connection URL is read from the env-file when updating the schema based on the actual DB as well as during run time to connect to the actual database.
```
andrew-walford-prisma:
```suggestion
Prisma reads the connection URL from the dotenv file in the following situations:
- When it updates the schema during build time
- When it connects to the database during run time
```

@mleimer 
Thanks for your contribution! I think splitting this out into bullets emphasizes your point that Prisma reads the file in both situations. I wasn't clear what you meant by "actual database", so I've removed "actual". If you think we need to convey something beyond just "database" here, could you clarify? Thanks.

mleimer: @andrew-walford-prisma sounds good to me. Thanks a lot for your support!

PR #4222, opened 2022-12-07
Do not pollute the global type when using Prisma in Next.js

dtinth:
## Describe this PR

The original code has this:

```ts
declare global {
  // allow global `var` declarations
  // eslint-disable-next-line no-var
  var prisma: PrismaClient | undefined
}
```

It breaks auto-import. Typing `pris` and hit <kbd>Tab</kbd> on `prisma` no longer generates an auto-import because it is now resolved to the global variable.

image

Following the advice in the documentation leads to an inconvenient DX.

## Changes

Instead of polluting the `global` type, we cast `global` to a type where we can safely add `prisma` to it..

```ts
const globalForPrisma = global as unknown as { prisma?: PrismaClient }
```

## What issue does this fix?

This brings back auto-import, which improves DX.

image

Also removes a need to do things like `// eslint-disable-next-line no-var`.

N/A


content/300-guides/050-database/880-troubleshooting-orm/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
```diff
@@ -23,19 +23,15 @@ The solution in this case is to instantiate a single instance `PrismaClient` and
 ```ts file=./db
 
-declare global {
```
millsp, line 26: I agree that this might not be necessary if `prisma` as a global isn't useful. Looking at the example below, it looks not so useful indeed since we're using imports.

content/300-guides/050-database/880-troubleshooting-orm/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
```diff
@@ -23,19 +23,15 @@ The solution in this case is to instantiate a single instance `PrismaClient` and
...
-  // allow global `var` declarations
-  // eslint-disable-next-line no-var
-  var prisma: PrismaClient | undefined
-}
+const globalForPrisma = global as unknown as { prisma: PrismaClient }
 
 export const prisma =
```
millsp, line 28: I would ensure that the types a still being passed down this way
```suggestion
export const prisma: PrismaClient =
```

millsp (approved):

content/300-guides/050-database/880-troubleshooting-orm/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
dtinth: Yes, in my projects Prisma is always imported. In addition, the `prisma` in the global scope is not available in production because of this `if` clause, as a result, the `prisma` is not useful in actual code. The only reason we put it in `global` is to be able to re-use the Prisma client instance despite Next.js’ hot-reloads.

```ts
if (process.env.NODE_ENV !== 'production') globalForPrisma.prisma = prisma
```

However I think it would be useful to also include `globalForPrisma.prisma = prisma` in production as well because:

1. The code would be even more simplified. e.g. the `??=` operator can be used instead of requiring an extra statement.
2. The behavior in development and production would more closely match, which I think is a good thing.
3. Should you ever need to attach a debugger to a live process running in production, providing access to `prisma` client instance via globals is very handy.

But that’s for another time 😃

PR #4225, opened 2022-12-08
Harmonize get started page title with topnav (remove the gerund)

andrew-walford-prisma:
Change "getting started" to "get started" so that when visitors click "get started" in the topnav, the page title agrees with what they clicked


zachtil (approved):
Looking good!

Issue #3614, opened 2022-08-16
Determine where we want images stored. 


tanberry:
Research, gain consensus, and then document [in our internal docs] our decision.
--Cloudinary, Yes or No?
--A single /images dir, or images in same dir as file?
--Is the image optimization that Vercel 4 provides good enough to not have to consider the complexity of Cloudinary?

zachtil:
Keep images as peer files to MDX files.
* `deployment/`
   * `deploy-to-netlify.mdx`
   * `deploy-to-cfw.mdx`
   * `image1.png`
   * `image2.png`
 
If the images grow to be more than the MDX files in a directory, create an `images/` folder and store all images there.

- `deployment/`
    - `images/`
        - `image1.png`
        - `image2.png`
        - …
        - `image20.png`
    - `deploy-to-netlify.mdx`
    - `deploy-to-cfw.mdx`

This is in contrast to having a centralized, top-level `images/` folder which would lead to having to create these kinds of image paths: `../../../images/pdp-screenshot1.png`.

zachtil:
Until we find a better alternative, we'll move all images out of Cloudinary and will be storing them on GitHub.

zachtil:
Because image file names have SEO impact (descriptive filenames are going to work better as keywords and ranking higher on those), the goal here is to also come up with a good convention for filenames.

zachtil:
The "migration" is ready in a PR: #4236.

One minor issue remains with current Gatsby builds: 
* with Cloudinary, when you click an image, it opens the original image size
* with Gatsby, when you click an image on our Docs page, it opens a lower-resolution (and blurrier) image than what appears on the page.
   * For example: https://docs-git-migrate-from-cloudinary-prisma.vercel.app/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-the-prisma-layer-postgres#511-fix-columns-with-enum-data-types

meletj:
- **Image hosting requirements**
    - Image optimization for responsiveness (based on different screen sizes, on bandwidth, etc.)
    - In Gatsby MDX, we can specify `@width` / `@max-width` on individual images
    - On click, it shows original image size (not lower quality image)
    - Image updates do not lead to the generation of a new filename / URL on upload of updated file
       - With Cloudinary, you cannot overwrite a previously uploaded image. When you upload the same image, it preserves the old image and uploads the new one, with a newly generated ID at the end of the file name. So you have to update the new URL in the image path in the MDX file.

[...]

PR #4236, opened 2022-12-09
Migrate images from Cloudinary

zachtil:
## Describe this PR

Migrate all Cloudinary-hosted images (diagrams and screenshots) to GitHub-hosted image files.

## Changes

* Download all Cloudinary-hosted images that are linked from Prisma docs (ORM + PDP)
* Add them to GitHub repository
* Change all image URLs (done with either `` or ``) to local image file paths

## What issue does this fix?

Impacts #3614


zachtil:
After discussing with @nilubava, I decided to not use the `static/` folder for images because that leads to file path confusion.

### Background

#### Images with relative paths in the `content/` folder
When you add images near the MDX files and inside the `content/` folder, it easy to figure out where an image file is by looking at the file path.

For example:

The file `content/800-data-platform/500-data-browser.mdx` references `images/data-browser/01-models-view.png`. This means that the image appears in `content/800-data-platform/images/data-browser/01-models-view.png`.

#### Images with relative paths in the `/static/` folder

When you add images in the `static/` folder, it is not easy to figure out where the image file is when you look at the file path. Instead, you need to understand the folder structure of the generated Gatsby build in the `public/` folder which we typically do not care about.

For example:

The file `content/800-data-platform/500-data-browser.mdx` references `../../images/data-browser/01-models-view.png`. 
This image is added to the `static/images/data-browser/01-models-view.png`. It is obvious from the relative path to the image that the `static` folder is not featured in the path.

The way to understand how the relative path works is by looking at the `public/` folder structure that is generated when you run `npm run build`.

img-filepath-in-public-folder

### Best practice

Therefore, as a best practice, avoid the use of the `static/` folder.

PR #4269, opened 2022-12-14
Fix text saying batch inserts are not supported

peterskim12:
Doc for create says batch inserts not supported. Does not reflect recent addition of createMany query.

## Describe this PR

Docs said batch inserts are not supported and included a link to a Github issue. createMany was added recently, so this is no longer correct.

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -813,7 +813,7 @@ const user = await prisma.user.create({
 
 ##### Create multiple new records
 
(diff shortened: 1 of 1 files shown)


peterskim12:
I was thinking it'd be nice to link specifically to the Remarks section under `createMany` that highlight some of the limitations of `createMany` and would lead people to use `create` instead. I didn't see other examples linking to a remarks section though, so I kept it simple. Feel free to suggest changes. Thanks!

andrew-walford-prisma:
Thanks @peterskim12 . I like your idea of linking to the remarks section. I'll merge your PR, then add a link to the remarks section as a follow-up.

PR #4273, opened 2022-12-15
feat(multiSchema): Definition of relation table in multi schema cases + fixes

janpio:


content/200-concepts/100-components/01-prisma-schema/06-relations/300-many-to-many-relations.mdx
```diff
@@ -350,19 +350,23 @@ Implicit m-n relations:
...
-- Then the name of the second table in alphabetical order (in this case `Post`)
+- Then the name of the second model in alphabetical order (in this case `Post`)
 
 In the example, the correct table name is `_CategoryToPost`.
 
 When creating an implicit m-n-relation yourself in the Prisma schema file, you can [configure the relation](#configuring-the-name-of-the-relation-table-in-implicit-many-to-many-relations) to have a different name. This will change the name given to the relation table in the database. For example, for a relation named `"MyRelation"` the corresponding table will be called `_MyRelation`.
 
+###### Multi-schema
+
+If your implicit many-to-many relationship spans multiple database schemas (using the [`multiSchema` preview feature](/guides/database/multi-schema)), the relation table (with the name defined directly above, in the example `_CategoryToPost`) must be present in the same database schema as the first model in alphabetical order (in this case `Category`).
```
ruheni, line 368: How likely is a user going to run into this situation? 🤔

If one uses an implicit m-n relationship, they are/ have been using Prisma from the beginning. Do we know of a user who's run into this? Do we know of other ORMs or tools that use implicit m-n relationships?

janpio: This is not very common for sure, but it needs to be documented so it is clear how this works.
(In general it is rather uncommon that our users need to know how to create implicit m-n relations manually in their database schema - but they exist.)

nikolasburk (approved):
looks good to me 👍

nikolasburk:
> looks good to me 👍

Want to merge this @janpio ?

PR #4282, opened 2022-12-19
feat(schema-reference): update Alpine binaryTargets

jkomyno:
## Describe this PR

Mention the newly available `binaryTargets` for Linux Alpine, for which only the `x86_64` architecture is supported.

## Changes

- Add `linux-musl-openssl-3.0.x` target


Jolg42 (approved):
Looks good! 
Note: only merge after the release.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -188,11 +188,12 @@ The following tables list all supported operating systems with the name of platf
...
 
-##### Linux (Alpine)
+##### Linux (Alpine) x86_64
 
-| Build OS | Prisma engine build name |
-| :------- | :----------------------- |
-| Alpine   | `linux-musl`             |
+| Build OS             | Prisma engine build name   | OpenSSL |
+| :------------------- | :------------------------- | :-----: |
+| Alpine (3.17+)       | `linux-musl-openssl-3.0.x` |  3.0.x  |
```
Jolg42: Can we somehow mention, maybe with an asterisk or so, that `linux-musl-openssl-3.0.x` was added in 4.8.0?

jkomyno: While @aqrln and I initially though about this as well, this would be a novel approach to the document, as we haven't ever mentioned the Prisma version an engine started appearing in in the past.

Jolg42: Indeed, maybe something for the docs team to think about

janpio (approved):

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
janpio: We should totally document that for things that were not available earlier.

At the same time, these should also be documented somewhere else and not just in this list here which would be a really difficult place to put all this information.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -188,11 +188,12 @@ The following tables list all supported operating systems with the name of platf
 | :------- | :----------------------- |
 | Windows  | `windows`                |
 
-##### Linux (Alpine)
+##### Linux (Alpine) x86_64
```
keerlu: Just want to understand why - x86_64 was added - is this the only architecture that Alpine supports, or the only one that we support, or something?

Jolg42: It's the only architecture we support for Alpine for now

keerlu: Suggesting something like this to spell it out. Not sure what the best option is but it definitely looks odd after the bracket:

```suggestion
##### Linux (Alpine)

Currently supported for x86_64 only.
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
keerlu: I think documenting on here with an asterisk would work for now, at least we get the information in somehow. Agree we need to think more about how to do this in general.

keerlu (approved):

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
Jolg42: Looks good 👍🏼

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
keerlu: (just noticed that it explains exactly what I asked in the issue description)

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
keerlu: (just noticed that it explains exactly what I asked in the issue description)

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
keerlu: I've added the asterisk to the table

keerlu: Also made #4287 to track documenting OpenSSL etc properly somewhere

janpio (approved):

PR #4310, opened 2022-12-23
Updated the quick start with typescript .

1shaked:
Added the Prisma client library cause without it's does not work


janpio:
This should not be necessary, as the `prisma generate` will install the dependency automatically if it is not present. If not, plesae open a bug report in `prisma/prisma`. Thanks.

PR #4342, opened 2023-01-03
fix(relations): Add subheadlines for better structure

janpio:
This PR adds a few subheadlines to the "Relations in the Prisma Client" as that one was hard to parse with the different queries and explanations flowing into each other.

The wording of the headlines is probably suboptimal. An alternative could be to also use the example model names in the headlines, or just completely use different terms. Open to suggestions.


content/200-concepts/100-components/01-prisma-schema/06-relations/index.mdx
```diff
@@ -146,6 +148,8 @@ In the underlying database, this query:
 1. Creates a `User` with an auto-generated `id` (for example, `20`)
 2. Creates two new `Post` records and sets the `authorId` of both records to `20`
 
+### Retrive a record and include related records
```
janpio:
```suggestion
### Retrieve a record and include related records
```

keerlu (approved):
This looks much better with the subheadings!

I think the current wording is fine. It would be good to get clearer generally on when/whether we use 'retrieve' in the docs vs 'get' or 'find', and when we use 'nested' vs 'connected' and 'related', but that's more of a general terminology question.

jharrell (approved):

PR #4345, opened 2023-01-04
Style guide: definite articles with product and feature names

zachtil:


content/600-about/200-prisma-docs/20-style-guide/05-prisma-product-names.mdx
```diff
@@ -5,17 +5,48 @@ metaDescription: 'This section of the style guide provides guidelines on how to
...
+
 - Data Proxy (_not_ Prisma Data Proxy)
 - Data Browser (in the Prisma Data Platform UI)
 - Query Console (in the Prisma Data Platform UI)
 - Schema Viewer (in the Prisma Data Platform UI)
 
-Do not abbreviate any of these names, except where noted above.
+## Use of articles with product names and feature names
+
+Do not use a definite article in front of official product names.
```
andrew-walford-prisma, line 29: This is not strictly true - IMO we should use the list of specific product names above. We do use a definite article for the Prisma Data Platform (but not when we abbreviate it to PDP). In future, Prisma might introduce more products that use definite articles, so I think it's easier just to refer to the list above.

content/600-about/200-prisma-docs/20-style-guide/05-prisma-product-names.mdx
```diff
@@ -5,17 +5,48 @@ metaDescription: 'This section of the style guide provides guidelines on how to
...
 - The Prisma Data Platform (or just "PDP" - no definite article - after the first mention)
 - Prisma Client (not any variation of "the Client")
 - Prisma Schema
 - Prisma Migrate
 - Prisma CLI
 - Prisma Studio
+
+## Feature names
+
 - Data Proxy (_not_ Prisma Data Proxy)
```
andrew-walford-prisma, line 22:
```suggestion
- The Data Proxy (_not_ Prisma Data Proxy)
```

See my comment below about definite articles for feature names. We might want to add definite articles to some/all of the other items in this list.

content/600-about/200-prisma-docs/20-style-guide/05-prisma-product-names.mdx
```diff
@@ -5,17 +5,48 @@ metaDescription: 'This section of the style guide provides guidelines on how to
...
+You use the Prisma Client to send queries to the database.
+
+```
+
+Where appropriate, you can use definite articles in front of feature names.
+
+```
+<!-- Good -->
+
+The Data Proxy provides database connection management and pooling.
```
andrew-walford-prisma, line 47: It seems that Prisma uses the definite article for the Data Proxy - e.g. on our [home page](https://www.prisma.io/)

When we have the results of Julie's investigation into this, we can update the list above.

content/600-about/200-prisma-docs/20-style-guide/05-prisma-product-names.mdx
zachtil: I've left the instance of "The Prisma Data Platform" above but I am not sure I wholeheartedly agree with it.

If it is a product name, I think the rule of not using "the" should apply to it. If we were to say something like "the Data Platform", then that kind of makes sense, but with the note that it is unambiguously clear from the context that we're talking about PDP and not another data platform and this is only used for brevity purposes.

From previous experience, long-form product names (Prisma Data Platform) are introduced on first mention in a document, and then their short-form version is used consistently (PDP). This means that any additional forms (the Data Platform, Data Platform, data platform, the platform) are avoided and discouraged to avoid any confusion.

content/600-about/200-prisma-docs/20-style-guide/05-prisma-product-names.mdx
zachtil: I want to only list feature names and avoid the general rule that we always use "the" in front. I've noted in the rule notes below to use that "when appropriate". I don't think it always makes sense to use "the" for these.

PR #4351, opened 2023-01-04
Update Prisma Studio with OSS note

zachtil:

### DIFF
--- a/content/200-concepts/050-overview/100-what-is-prisma/index.mdx
+++ b/content/200-concepts/050-overview/100-what-is-prisma/index.mdx
@@ -23,7 +23,7 @@ Prisma is an [open source](https://github.com/prisma/prisma) next-generation ORM
 - **Prisma Studio**: GUI to view and edit data in your database.
   <Admonition>
 
-  Prisma Studio is the only part of Prisma ORM that is not open source. It is fully integrated in our commercial offering Prisma Data Platform under the name **Data Studio**.
+  **Prisma Studio** is the only part of Prisma ORM that is not open source. It is fully integrated in our commercial offering [Prisma Data Platform](/data-platform/about) under the name **Data Browser** where you can view and edit data for each project.
 
   </Admonition>
 
--- a/content/200-concepts/050-overview/100-what-is-prisma/index.mdx
+++ b/content/200-concepts/050-overview/100-what-is-prisma/index.mdx
@@ -23,7 +23,7 @@ Prisma is an [open source](https://github.com/prisma/prisma) next-generation ORM
 - **Prisma Studio**: GUI to view and edit data in your database.
   <Admonition>
 
-  **Prisma Studio** is the only part of Prisma ORM that is not open source. It is fully integrated in our commercial offering [Prisma Data Platform](/data-platform/about) under the name **Data Browser** where you can view and edit data for each project.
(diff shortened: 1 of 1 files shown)


content/200-concepts/050-overview/100-what-is-prisma/index.mdx
```diff
@@ -20,7 +20,12 @@ Prisma is an [open source](https://github.com/prisma/prisma) next-generation ORM
 
 - **Prisma Client**: Auto-generated and type-safe query builder for Node.js & TypeScript
 - **Prisma Migrate**: Migration system
-- **Prisma Studio**: GUI to view and edit data in your database
+- **Prisma Studio**: GUI to view and edit data in your database.
+  <Admonition>
+
+  **Prisma Studio** is the only part of Prisma ORM that is not open source. It is fully integrated in our commercial offering [Prisma Data Platform](/data-platform/about) under the name **Data Browser** where you can view and edit data for each project.
```
petradonka: I thought that this sentence sounds a bit like Studio is *only* available as Data Browser, which is not the case. What do you think?

What would you think about saying that Client & Migrate are open source, instead of saying that Studio is not open source?

zachtil: Yes, you're right. I've now expanded the sentence to hopefully make it clearer.

My concern with highlighting Client & Migrate as open source and not saying anything about Studio (that's how I'm interpreting your suggestion at least) still leaves the question of the Prisma Studio status hanging, therefore, still unclear.

Just to make sure that I also cover this, here's how I would present that. Did you mean something similar?

> Prisma is an [open source](https://github.com/prisma/prisma) next-generation ORM. It consists of the following parts:
> 
> * **Prisma Client**: An open-source, auto-generated, and type-safe query builder for Node.js & TypeScript
> * **Prisma Migrate**: An open-source migration system
> * **Prisma Studio**: GUI to view and edit data in your database
> 
>

petradonka: That's something like I meant, but agree that it's not a great solution.

I like the updated version you have in the PR 👌

petradonka (approved):

content/200-concepts/050-overview/100-what-is-prisma/index.mdx
```diff
@@ -20,7 +20,12 @@ Prisma is an [open source](https://github.com/prisma/prisma) next-generation ORM
 
 - **Prisma Client**: Auto-generated and type-safe query builder for Node.js & TypeScript
 - **Prisma Migrate**: Migration system
-- **Prisma Studio**: GUI to view and edit data in your database
+- **Prisma Studio**: GUI to view and edit data in your database.
+  <Admonition>
+
+  **Prisma Studio** is the only part of Prisma ORM that is not open source. You can only run Prisma Studio locally. At the same time, Prisma Studio is also integrated in our commercial offering [Prisma Data Platform](/data-platform/about) under the name [**Data Browser**](/data-platform/data-browser) where you can view and edit data for each project and grant other team members the permission to only view or also edit data based on their [role](/data-platform/members/roles-permissions).
```
andrew-walford-prisma:
```suggestion
  **Prisma Studio** is the only part of Prisma ORM that is not open source. You can only run Prisma Studio locally. Prisma Studio is also integrated in our commercial offering [Prisma Data Platform](/data-platform/about) under the name [**Data Browser**](/data-platform/data-browser). In Data Browser, you can view and edit data for each project and grant other team members the permission to view or edit data based on their [role](/data-platform/members/roles-permissions).
```

I made a few suggestions to reduce the word count (while still preserving the sense; for example, I don't think we need both "at the same time" and "also"). I also split the last sentence, because it was quite long - I think I've preserved the sense here - we are saying that Data Browser is the tool where you can view and edit data and grant permissions?

andrew-walford-prisma (commented):
LGTM - I've made a few small suggestions to reduce the word count

janpio (approved):

PR #4353, opened 2023-01-04
Exclude "About the docs" section from the search

### LINKED ISSUE #4349: Exclude About->Prisma docs section from search results
Add the front matter tag `search: false` to our [About->Prisma docs](https://www.prisma.io/docs/about/prisma-docs) pages so that they don't appear in the docs search results. Include all the sub-pages in this section.

See: [search tag](https://www.prisma.io/docs/about/prisma-docs/docs-components/frontmatter?query=mdx%20co&page=1#search)

Exception: IMO it would be best if we leave the [root style guide page](prisma-docs/style-guide) in the results, so that a straight search for "style guide" finds it. We just don't want the content of individual style guide pages (e.g. with their code examples) to show up in general searches.

andrew-walford-prisma:
Exception: I've left the intro page for the style guide findable in search


janpio:
Google fortunately can still help me find the page where it explains how I do admonitions :D

PR #4369, opened 2023-01-09
Fixing a minor typo.

Patryk1098:
In the Computed Fields intro, there is a typo when describing the ability to create a function.

## Describe this PR

Intro to Computed Fields should have 'that' instead of 'the' when describing the ability to define a function.

## Changes

Changed the word "the" to "that" in the Computed Fields documentation intro.

## What issue does this fix?

No issue documented that relates to this typo.

Nothing of note.


janpio (approved):

andrew-walford-prisma (approved):
Thanks for the fix, @Patryk1098

PR #4375, opened 2023-01-09
Remove internal message from 010-mental-model.mdx

baptistemarchand:
While reading the docs I saw a "(link to example?)" that seems to be an internal message that was pushed to production by accident.


ruheni:
Thanks for catching that! 🙏🏽

PR #4392, opened 2023-01-11
Fix example TS interface

kafkas:
## Describe this PR

Small fix in an example.

## Changes

Fixed TS interface.

## What issue does this fix?

Example.


aqrln (approved):

keerlu:
Thanks for spotting this, @kafkas!

PR #4411, opened 2023-01-16
added missing db push command to documentation

ufukbakan:
## Describe this PR

MongoDB documentation has no mention to prisma db push.

## Changes

Added single line which tells users they should run npx prisma db push, to the end of the tweaking scheme section.


janpio:
Is this really necessary here? `db push` "only" pushed unique indexes for MongoDB, so might not be necessary here.

ufukbakan:
> Is this really necessary here? `db push` "only" pushed unique indexes for MongoDB, so might not be necessary here.

https://www.prisma.io/docs/guides/database/using-prisma-with-mongodb#commonalities-with-other-database-providers

janpio:
Yes, as the link says this will not update your database schema at all but only create any existing `unique` indexes - which the schema here does not have. Maybe we should fix that @sabinadams so that calling out `db push` makes sense. (Not sure if we call out migrate in the other database though?)

sabinadams:
@janpio I would agree that this change may not make sense for MongoDB. 

In the context of the section it is in, it shouldn't be present anyways. This section talks about introspection and then tweaking the introspected schema. If anything, this should point out that after making your tweaks you should run `npx prisma generate` again to re-generate the client with the tweaks. 

I do think in the list of commonalities @ufukbakan linked to we should re-phrase to make it obvious that while you "can" use `db push` with MongoDB, it doesn't actually do anything aside from create new unique indexes.

nikolasburk:
Thanks for opening this PR @ufukbakan 🙏 I think it's important that `db push` is called out on the ["Using Prisma with MongoDB" page](https://www.prisma.io/docs/guides/database/mongodb#commonalities-with-other-database-providers) as is currently the case. However, I agree with Jan's and Sabin's assessment that it's superfluous in the "Getting started with an existing project" guide, so I'm closing this PR.

image

PR #4451, opened 2023-01-25
feat: os information on sys requirements

### LINKED ISSUE #4303: (Serverless Initiative): Document system libraries required by Prisma
Prisma engines are dynamically linked to number of system libraries (e.g., `openssl`, `libz`, `musl`), which are considered runtime dependencies of Prisma.

We're currently only documenting the role of `openssl` in the `binaryTargets` attribute subsection [here](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#binarytargets-options).
We should however also expand the **System requirements** [page](https://www.prisma.io/docs/reference/system-requirements) to reflect the system libraries Prisma depends on, their versions, and the different expectations depending on the operating system / architecture. We have raised this concern on [Notion](https://www.notion.so/prismaio/Dynamically-linked-system-libraries-c9fa44ca779a4cf0b9fb02436a43aa43) already.

cc @aqrln @pimeys

janpio:
closes #4303


content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,20 @@ Notes:
...
+### Operating systems
+
+Prisma is supported on MacOS, Windows and most Linux distributions.
+
+#### Linux runtime dependencies
+
+Prisma requires the following system libraries installed to work:
+
+- OpenSSL 1.0.x, 1.1.x or 3.0.x
+- glibc 2.17 or newer
```
aqrln: also
- zlib (`libz.so.1`)
- libgcc (`libgcc_s.so.1`)

(for these two the exact versions aren’t extremely useful and don’t tell much about the ABI compatibility, or it’s hard to trace what the requirements in terms of versions would be, so I included the [sonames](https://en.m.wikipedia.org/wiki/Soname) instead)

janpio: This way then?
```suggestion
- glibc 2.17 or newer
- zlib (`libz.so.1`)
- libgcc (`libgcc_s.so.1`)
```

keerlu (approved):

content/400-reference/100-system-requirements.mdx
jkomyno: Note: on `Alpine Linux`, both `libgcc` (the GCC runtime library) and `glibc` aren't needed, right @aqrln?

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,20 @@ Notes:
...
+
+#### Windows runtime dependencies
+
+On [Windows Microsoft Visual C++ Redistributable 2015](https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x64.exe) or newer must be installed (which is by default the case on most modern installations).
```
jkomyno, line 79: We're missing a paragraph about the minimum support version of Mac OS. Let's also be explicit and state that Prisma works on both `Intel` CPUs (with the `x86_64` architecture) and the Silicon CPUs (M1, M2, with the `arm64` architecture)

content/400-reference/100-system-requirements.mdx
jkomyno: I would also add:
- a statement that informs the user that, most likely, `libgcc`, `zlib` are already installed in their system
- a list of popular Docker images of Linux Distros that we know support the requirements above
  - `node:lts-alpine` (Uses Linux Alpine, substitute `lts` with `14`, `16`, `18`, Prisma will still work)
  - `node:lts-bullseye-slim` (Uses Linux Debian)
- a couple of example of Docker images that don't support the requirements out of the box
  - `node:lts-buster-slim` (Uses Linux Debian, requires the user to manually install `libssl` / `openssl`)

Moreover, I have strong suspicion that we actually require a higher `glibc` version on `arm64`. Probably `glibc 2.19` or `glibc 2.24`. I'll confirm that this afternoon

jkomyno: We should a general statement about Prisma mostly supporting the `x86_64` and `arm64` CPU architectures, with other ones potentially supported in a "best effort" strategy, with a link to the existing `binaryTargets`

aqrln: @jkomyno

> Note: on Alpine Linux, both libgcc (the GCC runtime library) and glibc aren't needed, right @aqrln?

`glibc` is indeed not needed on Alpine as musl libc is used instead, but `libgcc` (which Rust uses for stack unwinding on panic) is still required on Alpine and needs to be installed, as it's not present in the vanilla Alpine image. Node.js depends on it too though, so it's already installed in the `node:alpine` image or will be pulled as the dependency when installing Node.js via the package manager.

> Moreover, I have strong suspicion that we actually require a higher glibc version on arm64. Probably glibc 2.19 or glibc 2.24.

We build with glibc 2.24 on ARM64, see https://github.com/prisma/docs/issues/4303#issuecomment-1384701662.

janpio: Figure out a way to present all this information I would say - but what is suggested here I think already improves it, so I will accept now.

content/400-reference/100-system-requirements.mdx
janpio: Please do add a suggestion for a headline and wording.

content/400-reference/100-system-requirements.mdx
```diff
@@ -42,10 +42,12 @@ Prisma is supported on MacOS, Windows and most Linux distributions.
 Prisma requires the following system libraries installed to work:
 
 - OpenSSL 1.0.x, 1.1.x or 3.0.x
-- glibc 2.17 or newer
+- On glibc-based distributions, glibc 2.17 or newer (glibc 2.24 or newer on ARM64)
 - zlib (`libz.so.1`)
 - libgcc (`libgcc_s.so.1`)
 
+Note that systems that can run the supported versions of Node.js will most likely have zlib and libgcc present on the system. One notable exception is Google's Distroless images, where `libz.so.1` needs to be copied from a compatible Debian system.
```
janpio: Really "copied from a compatible Debian system"? No command to execute?

content/400-reference/100-system-requirements.mdx
```diff
@@ -42,10 +42,12 @@ Prisma is supported on MacOS, Windows and most Linux distributions.
 Prisma requires the following system libraries installed to work:
 
 - OpenSSL 1.0.x, 1.1.x or 3.0.x
-- glibc 2.17 or newer
+- On glibc-based distributions, glibc 2.17 or newer (glibc 2.24 or newer on ARM64)
```
janpio: What does "glibc-based distributions" really mean? Will all Linux users know what that means? Can that be made more concrete? 

> (glibc 2.24 or newer on ARM64)

So what are the other ones that are not "ARM64"?

content/400-reference/100-system-requirements.mdx
```diff
@@ -52,6 +52,10 @@ Note that systems that can run the supported versions of Node.js will most likel
 
 On Windows [Microsoft Visual C++ Redistributable 2015](https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x64.exe) or newer must be installed (which is by default the case on most modern installations).
 
+#### macOS runtime dependencies
+
+Prisma supports macOS 10.15 or newer on Macs with Intel CPUs and macOS 11 or newer on Macs with Apple Silicon chips. There are no additional platform-specific requirements on macOS other than what is listed for all platforms in the [Software requirements](#software-requirements) section.
```
janpio: Why the difference depending on Intel/Apple silicon?

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,28 @@ Notes:
...
+Prisma is supported on MacOS, Windows and most Linux distributions.
+
+#### Linux runtime dependencies
+
+Prisma requires the following system libraries installed to work:
+
+- OpenSSL 1.0.x, 1.1.x or 3.0.x
+- On glibc-based distributions, glibc 2.17 or newer (glibc 2.24 or newer on ARM64)
+- zlib (`libz.so.1`)
+- libgcc (`libgcc_s.so.1`)
```
janpio, line 46: `Libgcc` instead maybe?
https://gcc.gnu.org/onlinedocs/gccint/Libgcc.html

(Unfortunately not consistent: https://gcc.gnu.org/gcc-3.0/libgcc.html)

aqrln: I think lowercase would be more standard/widespread

content/400-reference/100-system-requirements.mdx
aqrln: there's no package manager you could use to install something there

content/400-reference/100-system-requirements.mdx
aqrln: macOS 11 is the minimal version that runs on Apple Silicon, we can't target 10.15 there

content/400-reference/100-system-requirements.mdx
aqrln: no strong opinion though

content/400-reference/100-system-requirements.mdx
aqrln: we could add an example dockerfile with more detailed explanation to the "Troubleshooting" section below and link to it here

content/400-reference/100-system-requirements.mdx
aqrln: made a table instead, should now be more clear and most importantly more accurate

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,39 @@ Notes:
...
+### Operating systems
+
+Prisma is supported on MacOS, Windows and most Linux distributions.
+
+#### Linux runtime dependencies
+
+Prisma requires the following system libraries installed to work:
+
+- OpenSSL 1.0.x (except Alpine Linux), 1.1.x or 3.x
+- On most distributions except Alpine Linux, glibc 2.17 or newer on x86_64 systems and glibc 2.24 or newer on ARM64 systems
```
aqrln:
```suggestion
```

content/400-reference/100-system-requirements.mdx
janpio: Maybe this here then just links down to that section about the distroless images.

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,38 @@ Notes:
...
+
+The following table shows the required C standard library version for each supported combination of a Linux distribution, CPU architecture and OpenSSL version:
+
+| Distribution           | CPU architecture | OpenSSL version   | libc version        |
+| ---------------------- | ---------------- | ----------------- | ------------------- |
+| RHEL-like              | x86_64           | 1.0.x, 1.1.x, 3.x | glibc 2.17 or newer |
+| Debian-like            | x86_64           | 1.0.x             | glibc 2.19 or newer |
+| Debian-like            | x86_64           | 1.1.x, 3.x        | glibc 2.24 or newer |
+| RHEL-like, Debian-like | ARM64            | 1.0.x, 1.1.x, 3.x | glibc 2.24 or newer |
+| Alpine Linux           | x86_64, ARM64    | 1.1.x, 3.x        | musl libc 1.2.x     |
```
janpio: I feel this is far too much information for our standard System Requirements page in our docs. 
This looks scary, and this page should not be scary. 
Is this really needed? Could this be an "extended version" that lives in our repo somewhere and we just link to it here?

content/400-reference/100-system-requirements.mdx
janpio: If you can not run older than 11 on Apple Silicon, can this then not completely be left out as it is implied in owning a Apple Silicon laptop that you can not run older OS?
```suggestion
Prisma supports macOS 10.15 or newer. There are no additional platform-specific requirements on macOS other than what is listed for all platforms in the [Software requirements](#software-requirements) section.
```

content/400-reference/100-system-requirements.mdx
jkomyno: I actually saw this table in the preview before reading this comment, and my first impression was "wow, I like it".

However we could simplify a bit by splitting the OS support based on the CPU architecture. First we list the major 

And I would specify that we also use `Debian` and `OpenSSL 1.1.x` as fallbacks in case we're not able to detect this info. Should we display the actual warnings Prisma would show in this page as well, so that users googling them would find this page?

content/400-reference/100-system-requirements.mdx
jkomyno: Let's change `the supported versions of Node.js` into `the supported Node.js versions`, maybe with a link to where we list them

content/400-reference/100-system-requirements.mdx
jkomyno: The following two tables show the supported Linux distro' family, OpenSSL versions and C standard libraries for each CPU architecture.
Note that, in case of Prisma not being able to resolve the OpenSSL version on your system (e.g. because it is not installed), it will default to OpenSSL 1.1.x

On `amd64` (`x86_64`) architectures:

| Distro family    | OpenSSL version   | libc version | 
| ---------------- | ----------------- | ------------ |
| Alpine           | 1.1.x, 3.x        | musl 1.2.x+  |
| RHEL             | 1.0.x, 1.1.x, 3.x | glibc 2.17+  |
| Debian or others | 1.0.x             | glibc 2.19+  |
| Debian or others | 1.1.x, 3.x        | glibc 2.24+  |

On `arm64` (`aarch64`) architectures:

| Distro family    | OpenSSL version   | libc version | 
| ---------------- | ----------------- | ------------ |
| Alpine           | 1.1.x, 3.x        | musl 1.2.x+  |
| RHEL             | 1.0.x, 1.1.x, 3.x | glibc 2.24+  |
| Debian or others | 1.0.x, 1.1.x, 3.x | glibc 2.24+  |

```suggestion
The following table shows the required C standard library version for each supported combination of a Linux distribution, CPU architecture and OpenSSL version:

| Distribution           | CPU architecture | OpenSSL version   | libc version        |
| ---------------------- | ---------------- | ----------------- | ------------------- |
| RHEL-like              | x86_64           | 1.0.x, 1.1.x, 3.x | glibc 2.17 or newer |
| Debian-like            | x86_64           | 1.0.x             | glibc 2.19 or newer |
| Debian-like            | x86_64           | 1.1.x, 3.x        | glibc 2.24 or newer |
| RHEL-like, Debian-like | ARM64            | 1.0.x, 1.1.x, 3.x | glibc 2.24 or newer |
| Alpine Linux           | x86_64, ARM64    | 1.1.x, 3.x        | musl libc 1.2.x     |
```

janpio: Yes, that would be the typical "Troubleshooting" entry - first the error message, then explanation, and solution/workaround.

content/200-concepts/100-components/08-prisma-engines/index.mdx
```diff
@@ -37,7 +37,6 @@ Use the following environment variables to specify custom locations for your bin
 - [`PRISMA_QUERY_ENGINE_LIBRARY`](/reference/api-reference/environment-variables-reference#prisma_query_engine_library) <span class="api"></span> (Query engine, library)
 - [`PRISMA_QUERY_ENGINE_BINARY`](/reference/api-reference/environment-variables-reference#prisma_query_engine_binary) <span class="api"></span> (Query engine, binary)
 - [`PRISMA_MIGRATION_ENGINE_BINARY`](/reference/api-reference/environment-variables-reference#prisma_migration_engine_binary) <span class="api"></span> (Migration engine)
-- [`PRISMA_INTROSPECTION_ENGINE_BINARY`](/reference/api-reference/environment-variables-reference#prisma_introspection_engine_binary) <span class="api"></span> (Introspection engine)
```
janpio: This removal (an similar below) should not be part of this PR.

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+| Alpine           | 1.1.x, 3.x        | musl 1.2.x+  |
+| RHEL             | 1.0.x, 1.1.x, 3.x | glibc 2.17+  |
+| Debian or others | 1.0.x             | glibc 2.19+  |
+| Debian or others | 1.1.x, 3.x        | glibc 2.24+  |
+
+On `ARM64` (`aarch64`) architectures:
+
+| Distro family    | OpenSSL version   | libc version | 
+| ---------------- | ----------------- | ------------ |
+| Alpine           | 1.1.x, 3.x        | musl 1.2.x+  |
```
aqrln:
```suggestion
| Alpine           | 1.1.x, 3.x        | musl 1.2.x   |
```

We don't know yet if we are compatible with musl 1.3 before 1.3 exists, it is not necessarily backward compatible in the second digit. Musl 1.2 had a breaking change for 32-bit architectures, for example. On all architectures we officially compile for we are actually compatible with 1.1 as well as far as I tested, but better to document only 1.2.x since that's what we compile against.

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+
+- OpenSSL 1.0.x (except Alpine Linux), 1.1.x or 3.x
+- zlib (`libz.so.1`)
+- libgcc (`libgcc_s.so.1`)
+- C standard library (glibc on most Linux distributions or musl libc on Alpine Linux)
+
+The following two tables show the supported Linux distro' family, OpenSSL versions and C standard libraries for each CPU architecture.
+Note that, in case of Prisma not being able to resolve the OpenSSL version on your system (e.g. because it is not installed), it will default to OpenSSL 1.1.x
+
+On `AMD64` (`x86_64`) architectures:
```
aqrln:
```suggestion
On `AMD64` (`x86_64`) architecture:
```

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+
+| Distro family    | OpenSSL version   | libc version | 
+| ---------------- | ----------------- | ------------ |
+| Alpine           | 1.1.x, 3.x        | musl 1.2.x+  |
+| RHEL             | 1.0.x, 1.1.x, 3.x | glibc 2.17+  |
+| Debian or others | 1.0.x             | glibc 2.19+  |
+| Debian or others | 1.1.x, 3.x        | glibc 2.24+  |
+
+On `ARM64` (`aarch64`) architectures:
```
aqrln:
```suggestion
On `ARM64` (`aarch64`) architecture:
```

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+- C standard library (glibc on most Linux distributions or musl libc on Alpine Linux)
+
+The following two tables show the supported Linux distro' family, OpenSSL versions and C standard libraries for each CPU architecture.
+Note that, in case of Prisma not being able to resolve the OpenSSL version on your system (e.g. because it is not installed), it will default to OpenSSL 1.1.x
+
+On `AMD64` (`x86_64`) architectures:
+
+| Distro family    | OpenSSL version   | libc version | 
+| ---------------- | ----------------- | ------------ |
+| Alpine           | 1.1.x, 3.x        | musl 1.2.x+  |
```
aqrln:
```suggestion
| Alpine           | 1.1.x, 3.x        | musl 1.2.x   |
```

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
 See also: [Supported database versions](database-reference/supported-databases)
 
-</TopBlock>
+### Operating systems
+
+Prisma is supported on MacOS, Windows and most Linux distributions.
+
+#### Linux runtime dependencies
+
+Prisma requires the following system libraries installed to work:
```
janpio:
```suggestion
Prisma requires the following system libraries to be installed to work:
```

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+
+- OpenSSL 1.0.x (except Alpine Linux), 1.1.x or 3.x
```
janpio: Why is the `(except Alpine Linux)` in the middle of the versions here? Does that only apply to `1.0.x` somehow?

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+- zlib (`libz.so.1`)
+- libgcc (`libgcc_s.so.1`)
+- C standard library (glibc on most Linux distributions or musl libc on Alpine Linux)
+
+The following two tables show the supported Linux distro' family, OpenSSL versions and C standard libraries for each CPU architecture.
+Note that, in case of Prisma not being able to resolve the OpenSSL version on your system (e.g. because it is not installed), it will default to OpenSSL 1.1.x
```
janpio: Moved below the tables.
```suggestion
```

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+
+On `ARM64` (`aarch64`) architecture:
+
+| Distro family    | OpenSSL version   | libc version | 
+| ---------------- | ----------------- | ------------ |
+| Alpine           | 1.1.x, 3.x        | musl 1.2.x   |
+| RHEL             | 1.0.x, 1.1.x, 3.x | glibc 2.24+  |
+| Debian or others | 1.0.x, 1.1.x, 3.x | glibc 2.24+  |
+
+Note that systems that can run the supported versions of Node.js will most likely have zlib and libgcc present on the system. One notable exception is Google's Distroless images, where `libz.so.1` needs to be copied from a compatible Debian system.
```
janpio:
```suggestion
Systems that can run the supported Node.js versions will most likely have zlib and libgcc available. One notable exception is Google's Distroless images, where `libz.so.1` needs to be copied from a compatible Debian system.
```

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,48 @@ Notes:
...
+| Debian or others | 1.1.x, 3.x        | glibc 2.24+  |
+
+On `ARM64` (`aarch64`) architecture:
+
+| Distro family    | OpenSSL version   | libc version | 
+| ---------------- | ----------------- | ------------ |
+| Alpine           | 1.1.x, 3.x        | musl 1.2.x   |
+| RHEL             | 1.0.x, 1.1.x, 3.x | glibc 2.24+  |
+| Debian or others | 1.0.x, 1.1.x, 3.x | glibc 2.24+  |
+
```
janpio, line 67: Moved from before the tables:
```suggestion

When Prisma can not resolve the OpenSSL version on a system (e.g. because it is not installed), it will default to OpenSSL 1.1.x.

```

janpio:
✅ Good from my side now, except for that one open question about the parenthesis placement.

content/400-reference/100-system-requirements.mdx
jkomyno: No build for Alpine Linux relies on `openssl-1.0.x`

jkomyno: In other words, we only support `1.1.x` and `3.0.x` in Alpine, regardless of the architecture

jkomyno (approved):

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,49 @@ Notes:
...
-</TopBlock>
+### Operating systems
+
+Prisma is supported on MacOS, Windows and most Linux distributions.
+
+#### Linux runtime dependencies
+
+Prisma requires the following system libraries to be installed to work:
+
+- OpenSSL 1.0.x (except Alpine Linux), 1.1.x or 3.x
```
janpio:
```suggestion
- OpenSSL 1.0.x, 1.1.x or 3.x
```

janpio:
Thoughts @sabinadams? When you like it you can directly merge this, this is not bound to the release.

content/400-reference/100-system-requirements.mdx
```diff
@@ -1,11 +1,23 @@
...
 <TopBlock>
 
-The [latest version of Prisma](https://www.npmjs.com/package/@prisma/client) requires the following:
+This page provides an overview of the system requirements for Prisma.
+
+</TopBlock>
+
+## System requirements
+
+This section lists the software that Prisma requires, and the supported operating systems, along with runtime dependency requirements for specific operating systems.
```
sabinadams:
```suggestion
This section lists the software that Prisma requires and the supported operating systems, along with runtime dependency requirements for specific operating systems.
```

I think the punctuation here wasn't quite correct. Grammarly fixed it up for me 🙂

content/400-reference/100-system-requirements.mdx
```diff
@@ -21,7 +33,49 @@ Notes:
...
+#### Linux runtime dependencies
+
+Prisma requires the following system libraries to be installed to work:
+
+- OpenSSL 1.0.x, 1.1.x or 3.x
+- zlib (`libz.so.1`)
+- libgcc (`libgcc_s.so.1`)
+- C standard library (glibc on most Linux distributions or musl libc on Alpine Linux)
+
+The following two tables show the supported Linux distro' family, OpenSSL versions and C standard libraries for each CPU architecture.
```
sabinadams:
```suggestion
The following two tables show the supported Linux distro families, OpenSSL versions and C standard libraries for each CPU architecture.
```

sabinadams (commented):
Just a few cleanup items.

sabinadams (approved):
Looks great!

PR #4470, opened 2023-01-29
fix(docs): link to planetscale relation mode

brenodanyel:
## What issue does this fix?

The property has changed and this link is pointing to the old url


sabinadams (approved):
Looks great, thanks!

PR #4472, opened 2023-01-30
Update 400-self-relations.mdx : Fix typo, by swapping follower and following relations

redbar0n:
Simple two-liner typo fix.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/400-self-relations.mdx
@@ -321,8 +321,8 @@ If you need the relation to hold other fields, you can create an [explicit](many
 model User {
   id         Int       @id @default(autoincrement())
   name       String?
-  followedBy Follows[] @relation("following")
-  following  Follows[] @relation("follower")
+  followedBy Follows[] @relation("follower")
+  following  Follows[] @relation("following")
 }
 
 model Follows {


janpio:
I think this is the third time these are flipped around now. We might need to come up with a better example that has less similar sounding words to describe the concept at hand @sabinadams

redbar0n:
maybe replace "follower" with "followedBy" in the relations as well, then they will align perfectly with the same names used in the model.

redbar0n:
and also maybe rename "following" to "follows", since "following" might otherwise be mistaken for a user having "a following" (that's maybe where the mistake came from: since a following is an english term used to describe a group of followers).

janpio:
That is exactly the problem, I don't think it was meant like you understand the terms. The relation name and the relation field name describe different sides of the same relationship - and you can look at the relationships in 2 directions, as it is a self relation.

redbar0n:
Yeah, I know. To be clear, everything I wrote after `, since` is not how I actually understand the term "following", but was to illustrate how people could misunderstand it.

sabinadams:
Ah yeah, reading through that section it does seem the choice of example may not have been the most clear.

I think it would make sense to present a different scenario that used a bit clearer names.

@janpio Also want to note that there are a couple of other places on this page that could use some tweaks & revisions as well.

redbar0n:
@nilubava thanks, how about also renaming "following" to "follows", for the aforementioned reason?

nikolasburk:
I think there's no right or wrong here to be honest. The most important thing is that we have two relation names that match each other in both models, they could also just be called `A` and `B`.

>  We might need to come up with a better example that has less similar sounding words to describe the concept at han

The example is fine and important, it shows a _directional_ m-n-relationship.

We can consider though adding a _non-directional_ m-n-relationship, like:

```prisma
model User {
  id          Int           @id @default(autoincrement())
  email       String        @unique
  name        String?
  hasFriends1 FriendShips[] @relation("hasFriends1")
  hasFriends2 FriendShips[] @relation("hasFriends2")
}

model FriendShips {
  friend1   User @relation("hasFriends1", fields: [friend1Id], references: [id])
  friend1Id Int
  friend2   User @relation("hasFriends2", fields: [friend2Id], references: [id])
  friend2Id Int

  @@id([friend1Id, friend2Id])
}
```

With "followers" it's important who follows who, with "friends", it doesn't matter.

> maybe replace "follower" with "followedBy" in the relations as well, then they will align perfectly with the same names used in the model.

I did like this suggestion though so I've created [another PR](https://github.com/prisma/docs/pull/5455) where I make this change. Thanks for the suggestion @redbar0n 🎉 💚 

Here's the new version: 

```prisma
model User {
  id         Int       @id @default(autoincrement())
  name       String?
  followedBy Follows[] @relation("followedBy")
  following  Follows[] @relation("following")
}

model Follows {
  followedBy   User @relation("followedBy", fields: [followedById], references: [id])
  followedById Int
  following    User @relation("following", fields: [followingId], references: [id])
  followingId  Int

  @@id([followingId, followedById])
}
```

PR #4482, opened 2023-02-03
fix(algolia): Output when no indexing happens, also move setup to only happen when we will use it

janpio:
...


nilubava (approved):

janpio:
Let's wait for the Percy run to proof that this works.

PR #4487, opened 2023-02-03
Update plugin list for JetBrains IDE

jume-dev:
The two mentioned plugins are outdated and aren't maintained anymore, but JetBrains now provides an official plugin with their backing.


content/300-guides/125-development-environment/100-editor-setup.mdx
```diff
@@ -35,8 +35,9 @@ You can install the official [Prisma VS Code extension](https://marketplace.visu
 
 ### JetBrains IDE
 
-- [Prisma](https://plugins.jetbrains.com/plugin/14240-prisma) provides syntax highlighting of the Prisma Schema Language.
-- [Prisma Support](https://plugins.jetbrains.com/plugin/14282-prisma-support) provides syntax highlighting of the Prisma Schema Language.
+- [Prisma ORM](https://plugins.jetbrains.com/plugin/20686-prisma-orm) JetBrains provided plugin with support for grammar, syntax highlighting, LSP, and many more.
+- ~~[Prisma](https://plugins.jetbrains.com/plugin/14240-prisma) provides syntax highlighting of the Prisma Schema Language.~~ (Outdated)
+- ~~[Prisma Support](https://plugins.jetbrains.com/plugin/14282-prisma-support) provides syntax highlighting of the Prisma Schema Language.~~ (Outdated)
```
sabinadams: I would suggest leaving these as is. There is a warning at the top of the page that community packages are not maintained by Prisma and may be out of date.

content/300-guides/125-development-environment/100-editor-setup.mdx
```diff
@@ -35,8 +35,9 @@ You can install the official [Prisma VS Code extension](https://marketplace.visu
 
 ### JetBrains IDE
 
-- [Prisma](https://plugins.jetbrains.com/plugin/14240-prisma) provides syntax highlighting of the Prisma Schema Language.
-- [Prisma Support](https://plugins.jetbrains.com/plugin/14282-prisma-support) provides syntax highlighting of the Prisma Schema Language.
+- [Prisma ORM](https://plugins.jetbrains.com/plugin/20686-prisma-orm) JetBrains provided plugin with support for grammar, syntax highlighting, LSP, and many more.
```
sabinadams:
```suggestion
- [Prisma ORM](https://plugins.jetbrains.com/plugin/20686-prisma-orm) Provided by JetBrains. This plugin provides PSL grammar, syntax highlighting, LSP, and more.
```

sabinadams (changes requested):
Great addition!

Left some suggestions about the changes.

content/300-guides/125-development-environment/100-editor-setup.mdx
jume-dev: I can see where you are coming from, but I am not sure, what's best.
I mean, these two are already no longer maintained, don't work and will be "probably" never get updates unless JetBrains decides to abandon their own plugin, which afaik will only happen if somehow prisma would be no longer a thing.
Long story short, I'd rather remove both of them completely instead of leaving them in there without an outdated marker.

But in the end, it's your call, I am happy to follow your decision

janpio: I suggest we actually remove them completely then:
```suggestion
```
Do you agree @sabinadams?

sabinadams: For sure, I considered it myself but wasn't sure if removing community projects was the right solution.

I think if you both agree, that would be great.

jume-dev: Yeah, totally get what you mean. In my humble opinion, it helps more to have an updated list instead of keeping outdated stuff only for the sake of keeping it.

janpio (approved):

sabinadams (approved):

PR #4503, opened 2023-02-08
Added information about the official Prisma plugin in JetBrains IDEs

kricact:
The WebStorm team has added support for Prisma recently, It includes complete Prisma schema language syntax support with semantic highlighting, code completion, automatic brace, and quote insertion, and Rename refactoring. 
And it works better than the plugins mentioned here, which are not supported in recent versions


janpio:
See https://github.com/prisma/docs/pull/4487

kricact:
@janpio Thanks for the link

PR #4514, opened 2023-02-13
Blogging platform schema example

velizario:
Implicit many-to-many relation should not have references argument defined in relational databases

## Describe this PR

Adjusted code example in documentation for relational database in /prisma-schema/data-model .

## Changes

Removed @relation from the sample code where many-to-many relation is defined between Category and Post for a relational database


nikolasburk:
Thanks a lot for catching this @velizario 🎉

PR #4519, opened 2023-02-16
docs: update to best practice `migrate deploy`

dijonkitchen:
Per docs: https://www.prisma.io/docs/guides/deployment/deploy-database-changes-with-prisma-migrate


janpio:
Why the change from `dev` to `deploy`?
`deploy` does not have `--name` so this would at least also need to be changed.

dijonkitchen:
> Why the change from `dev` to `deploy`? `deploy` does not have `--name` so this would at least also need to be changed.

Based on your docs and the context, it doesn't seem to make sense to create a new migration in the test environment. Instead, it should be apply migrations by using `deploy` in that test environment. https://www.prisma.io/docs/guides/deployment/deploy-database-changes-with-prisma-migrate

janpio:
Where are you reading this in our documentation?

We advise you not to create migrations in the _production_ environment, as `migrate dev` is a tool that should be run against a dev/staging database. But if you have multiple of those, and those are reflected via branches in your code, I see no reason why you should not want to switch between them via multiple `.env` files?

What am I missing?

dijonkitchen:
In the link above, the first line says: 
> To apply pending migrations to staging, testing, or production environments, run the migrate deploy command as part of your CI/CD pipeline

janpio:
I still don't udnerstand why that implies that this package.json should not also be able to execute `migrate dev` in different environments.

But as it does not make any difference which command is used here: ✅

janpio (approved):

dijonkitchen:
> I still don't udnerstand why that implies that this package.json should not also be able to execute `migrate dev` in different environments.
> 
> But as it does not make any difference which command is used here: ✅

Sure, it works, but `migrate dev` is doing more than `migrate deploy`, which we don't want or need in other environments based on: https://www.prisma.io/docs/concepts/components/prisma-migrate/migrate-development-production

sabinadams (commented):
Looks good to me. I agree, the result either way is the same but our docs do suggest the updated way is more desirable.

PR #4532, opened 2023-02-21
Add tip about the satisfies operator

gunhaxxor:
Add info that users might prefer satisfies instead of the validator function.

## Describe this PR
Changed the docs page about the `Prisma.validator`.
Added tip about using the satisfies operator instead of the validator function.

## Changes

Small "info box" at the top of the docs page

At least for me, intellisense (vs code with a pnpm monorepe) does **NOT** work inside the validator function. This gives a somewhat cumbersome DX where i had to move the code outside to a normal
`const thingQuery: Prisma.ThingArgs = { ... }`
where i got intellisense to construct my query. Then I had to move the query object back into the validator function which narrowed the type for me. This back and forth felt really non-optimal. Trying to google the issue I stumbled upon the [blog post ](https://www.prisma.io/blog/satisfies-operator-ur8ys8ccq7zb) about using satisfies instead. Using satisfies operator I get perfect intellisense **and** narrowed types. Hurray!

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/052-advanced-type-safety/99-prisma-validator.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/052-advanced-type-safety/99-prisma-validator.mdx
@@ -6,7 +6,9 @@ metaDescription: 'The Prisma validator is a utility function that takes a genera
 
 <TopBlock>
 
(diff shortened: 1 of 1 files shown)


nikolasburk:
Hey @Dealerpriest 👋 thanks for this PR and sorry for the delay, our docs PRs were with little love for a while! In the meantime, we've actually already added a similar note to the top of the validator page, so I'm closing this PR without this message 🙂 

Thanks so much for caring about our docs and helping us improve them 💚 

image

PR #4541, opened 2023-02-23
Fixes Incorrect Prisma version mentioned for directUrl

nurul3101:
directUrl was released in Prisma version 4.10.0 and not in 4.9.0. This PR mentions the correct version for directUrl.


janpio (approved):
Was planned for 4.9.0, but then only released in 4.10.0 👍

sabinadams (approved):

PR #4543, opened 2023-02-24
docs: Updates Views page introspection warnings and limitations

sabinadams:
## Describe this PR

Updates the warnings and limitations section to include support for MySQL, SQL Server and CockroachDB when using `db pull` with views.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 # . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+# npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 # . "$(dirname "$0")/_/husky.sh"
 
-# npx lint-staged
+npx lint-staged
--- a/content/200-concepts/100-components/01-prisma-schema/08-views.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/08-views.mdx
@@ -19,10 +19,18 @@ Database views allow you to name and store queries. In relational databases, vie
 
 The `views` preview feature allows you to represent views in your Prisma schema with the `view` keyword. To use views in Prisma, follow these steps:
 
-- [Enable the `views` preview feature](#enable-the-views-preview-feature)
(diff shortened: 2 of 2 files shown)


content/200-concepts/100-components/01-prisma-schema/08-views.mdx
```diff
@@ -19,10 +19,18 @@ Database views allow you to name and store queries. In relational databases, vie
...
+- [Use views with Prisma Migrate and db push](#use-views-with-prisma-migrate-and-db-push)
+- [Add views to your Prisma schema](#add-views-to-your-prisma-schema)
+  - [Write by hand](#write-by-hand)
+    - [Define a `view` block](#define-a-view-block)
+    - [Define fields](#define-fields)
+  - [Use introspection](#use-introspection)
+  - [Limitations](#limitations)
+    - [Unique Identifier](#unique-identifier)
+    - [Introspection](#introspection)
+- [Query views in Prisma Client](#query-views-in-prisma-client)
```
janpio: Why the switch from an editorialized list of the concrete steps to a (auto-generated?) table of contents of the whole page?

janpio (approved):
availability updates ✅

content/200-concepts/100-components/01-prisma-schema/08-views.mdx
sabinadams: This was my bad, I was using a new Markdown extension in VSCode and apparently it generates a TOC on save...

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 # . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+# npx lint-staged
```
janpio: undo

content/200-concepts/100-components/01-prisma-schema/08-views.mdx
janpio: Haha ok. Interesting that it figure out where to put it. I remember having to use weird comments to make something like this work in the past.

.husky/pre-commit
sabinadams: I had accidentally commented this out and committed. The original state was it was uncommented. 
_I had commented it because the linter previously was breaking the prisma code blocks that use the `view` block._

janpio: There are comments/annotations on how to disable the linter I think, which would be the more permanent solution to that problem. (I don't know more about them, just picked up in the past that they exist)

PR #4545, opened 2023-02-24
Add example for raw query in client extensions

millsp:
## Describe this PR

Add examples for raw query in client extensions.

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
@@ -24,17 +24,17 @@ generator client {
 }
 ```
 
-## Create and modify queries
+## Extend Prisma Client query operations
 
 Use the `$extends` [client-level method](/reference/api-reference/prisma-client-reference#client-methods) to create an [extended client](/concepts/components/prisma-client/client-extensions#about-prisma-client-extensions). An extended client is a variant of the standard Prisma Client that is wrapped by one or more extensions.
 
-Use the `query` extension component to create and modify queries. You can add or modify a custom query in the following:
+Use the `query` extension component to modify queries. You can modify a custom query in the following:
 
-- [A specific operation in a specific model](#add-or-modify-a-custom-query-for-a-specific-operation-in-a-specific-model)
-- [A specific operation in all models of your schema](#add-or-modify-a-custom-query-for-a-specific-operation-in-all-models-of-your-schema)
-- [All operations in a specific model](#add-or-modify-a-custom-query-for-all-operations-in-a-specific-model)
-- [All operations in all models of your schema](#add-or-modify-a-custom-query-for-all-operations-in-all-models-of-your-schema)
-- [A specific top-level client operation](#add-or-modify-a-custom-query-for-a-specific-top-level-client-operation)
+- [A specific operation in a specific model](#modify-a-specific-operation-in-a-specific-model)
+- [A specific operation in all models of your schema](#modify-a-specific-operation-in-all-models-of-your-schema)
+- [All operations in a specific model](#modify-all-operations-in-a-specific-model)
+- [All operations in all models of your schema](#modify-all-operations-in-all-models-of-your-schema)
+- [A specific top-level raw query operation](#modify-a-top-level-raw-query-operations)
 
 To create a custom query, use the following structure:
 
@@ -52,7 +52,7 @@ The properties are as follows:
 - `name`: (optional) specifies a name for the extension that appears in error logs.
 - `query`: defines a custom query.
 
-### Add or modify a custom query for a specific operation in a specific model
+### Modify a specific operation in a specific model
 
 The `query` object can contain functions that map to the names of the [Prisma Client operations](/reference/api-reference/prisma-client-reference#model-queries), such as `findUnique`, `findFirst`, `findMany`, `count`, and `create`. The following example modifies `user.findMany` to a use a customized query that finds only users who are older than 18 years:
 
@@ -91,7 +91,7 @@ In the above example, a call to `xprisma.user.findMany` triggers `query.user.fin
 
   - You can use `await` and then mutate the result of this promise, because its value is type-safe. TypeScript catches any unsafe mutations on the object.
 
-### Add or modify a custom query for a specific operation in all models of your schema
+### Modify a specific operation in all models of your schema
 
 To extend the queries in all the models of your schema, use `$allModels` instead of a specific model name. For example:
 
@@ -110,7 +110,7 @@ const xprisma = prisma.$extends({
 })
 ```
 
-### Add or modify a custom query for all operations in a specific model
+### Modify all operations in a specific model
 
 Use `$allOperations` to extend all operations in a specific model.
 
@@ -129,7 +129,7 @@ const xprisma = prisma.$extends({
 })
 ```
 
-### Add or modify a custom query for all operations in all models of your schema
+### Modify all operations in all models of your schema
 
 Use `$allModels` and `$allOperations` to extend all operations in all models of your schema.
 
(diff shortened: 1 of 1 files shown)


sabinadams (approved):
The changes here look great to me!

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
```diff
@@ -147,6 +148,54 @@ const xprisma = prisma.$extends({
 })
 ```
 
+### Add or modify a custom query for a specific top-level client operation
+
+To apply a custom query to a specific top-level client operation:
```
janpio: What does "apply a custom query" really mean here?
Are we overwriting the existing `$queryRaw` and making it do other things instead?
Or can we do something before the actual query is sent? Can we modify the raw query that is later executed?

millsp: Yep, so I based this on previous paragraphs and assumed this was explained somewhere in the beginning or via prior examples. So yes, you can make it do different things, like wrap it in a transaction, append another query before it, or even build a cache in front of it. And yes, you can also modify the args before they are sent. So really like a middleware here.

sabinadams: I made a pretty minor wording change to that sentence to hopefully make it a bit clearer. @janpio Look alright now?

janpio: That definitely is clearer.

Does it apply only to raw queries though or all of them?
If only raw, we could mention that?
If not, why the xample only for raw ones?

sabinadams: It's only the queries if I understood correctly because as the examples show you have to extend those within the `query` key of the `.$extends` object parameter. You couldn't extend the `$connect` function.

I specified "top-level client _query_ operations" here to clarify the section refers only to query functions:

> To apply custom behavior to a specific top-level client query operation:

Is that too subtle?

janpio: query is also `findMany` - I am unclear if those also belong to "specific top-level client operation", and if so, why we only have raw queries as examples.

sabinadams: Oh sorry, I misunderstood. 

I read "top-level" as being query functions available at the top level of the Prisma Client API (e.g. `prisma.$queryRaw`) whereas non-raw query functions are nested under a model (e.g. `prisma.user.findMany`)

The API for extending the nested functions is different. You have to drill down into the model:

```ts
prisma.$extends({
  query: {
    user: {
      findMany: ...
    }
  }
})
```

The sections above this one explain how to extend model-specific functions.

See this section for example: /concepts/components/prisma-client/client-extensions/query#add-or-modify-a-custom-query-for-a-specific-operation-in-a-specific-model

janpio: Ah, that makes sense.
That I did not understand directly, might indicate that we could just be explicit and call these "raw queries" instead of " top-level client operation" if they are all raw query methods after all.

sabinadams: Agreed, I'll make that change now 👍🏻

sabinadams: Okay, I've made the change to the section title. It is now: 

> Modify top-level raw query operations:

The previous name after re-reading it a few times was also misleading as it seemed to suggest you could add your own custom queries to the top-level API which currently isn't possible. 

> Add or modify a custom query for top-level raw query operations:

For example, I couldn't create a new top-level query function named `$myCustomRawQuery`

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
```diff
@@ -34,6 +34,7 @@ Use the `query` extension component to create and modify queries. You can add or
 - [A specific operation in all models of your schema](#add-or-modify-a-custom-query-for-a-specific-operation-in-all-models-of-your-schema)
 - [All operations in a specific model](#add-or-modify-a-custom-query-for-all-operations-in-a-specific-model)
 - [All operations in all models of your schema](#add-or-modify-a-custom-query-for-all-operations-in-all-models-of-your-schema)
+- [A specific top-level client operation](#add-or-modify-a-custom-query-for-a-specific-top-level-client-operation)
```
millsp: This link is invalid now, I think.

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
```diff
@@ -34,6 +34,7 @@ Use the `query` extension component to create and modify queries. You can add or
 - [A specific operation in all models of your schema](#add-or-modify-a-custom-query-for-a-specific-operation-in-all-models-of-your-schema)
 - [All operations in a specific model](#add-or-modify-a-custom-query-for-all-operations-in-a-specific-model)
 - [All operations in all models of your schema](#add-or-modify-a-custom-query-for-all-operations-in-all-models-of-your-schema)
```
millsp: The links and titles here also are "add or modify", do you think it's worth clearing this up?

sabinadams: Yeah, looks like pretty much all of them could use some love. Just modified them  👍🏻

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
janpio: Sounds good to me. Can you confirm @millsp this is what it does?

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
```diff
@@ -24,17 +24,17 @@ generator client {
...
 
-- [A specific operation in a specific model](#add-or-modify-a-custom-query-for-a-specific-operation-in-a-specific-model)
-- [A specific operation in all models of your schema](#add-or-modify-a-custom-query-for-a-specific-operation-in-all-models-of-your-schema)
-- [All operations in a specific model](#add-or-modify-a-custom-query-for-all-operations-in-a-specific-model)
-- [All operations in all models of your schema](#add-or-modify-a-custom-query-for-all-operations-in-all-models-of-your-schema)
-- [A specific top-level client operation](#add-or-modify-a-custom-query-for-a-specific-top-level-client-operation)
+- [A specific operation in a specific model](#modify-a-specific-operation-in-a-specific-model)
+- [A specific operation in all models of your schema](#modify-a-specific-operation-in-all-models-of-your-schema)
+- [All operations in a specific model](#modify-all-operations-in-a-specific-model)
+- [All operations in all models of your schema](#modify-all-operations-in-all-models-of-your-schema)
```
janpio, line 36: I think these 4 it is actually about modifying or adding though?

sabinadams: This section only describes modifying a query.

There is another section that describes how to add methods here:

https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model

The way you add methods is a little bit different than extending as you have to go through the `model` key rather than the `query` key.

millsp: Yes, `query` technically modifies **operation** behavior and `model` adds new **methods** (or can even override existing ones).

janpio: Ah cool. Cross links might be nice.

content/200-concepts/100-components/02-prisma-client/260-client-extensions/120-query.mdx
millsp: Looks good.

PR #4550, opened 2023-02-27
Missing `await` in Prisma's Sequelize comparison

malee31:
## Slightly Inaccurate Documentation Example
The docs are missing an `await` for a `delete()`. [Here is the example with the missing `await`](https://www.prisma.io/docs/concepts/more/comparisons/prisma-and-sequelize#deleting-objects)

[Here is a correct example of `delete()` from elsewhere in the docs](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#delete-the-user-record-with-an-id-of-1):
image

## Changes
Added the missing `await` to the Markdown file.

See links from description:
[Missing `await`](https://www.prisma.io/docs/concepts/more/comparisons/prisma-and-sequelize#deleting-objects)
[Correct usage](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#delete-the-user-record-with-an-id-of-1)

### DIFF
--- a/content/200-concepts/300-more/400-comparisons/02-prisma-and-sequelize.mdx
+++ b/content/200-concepts/300-more/400-comparisons/02-prisma-and-sequelize.mdx
@@ -186,7 +186,7 @@ Sequelize [doesn't offer a dedicated API for relation filters](https://github.co
 Cursor-style pagination:
 
 ```ts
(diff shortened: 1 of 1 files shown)


sabinadams (approved):
Updated to fix a few more missing awaits on the page, but thanks for pointing out the issue!

PR #4567, opened 2023-03-04
Improve typescript types for prisma nextjs docs

### LINKED ISSUE #4566: Improve Typescript types in NextJS prisma documentation example
Hi, I'd like to recommend an improvement to the docs. 

```

const globalForPrisma = global as unknown as { prisma: PrismaClient }

export const prisma =
  globalForPrisma.prisma ||
  new PrismaClient({
    log: ['query'],
  })

if (process.env.NODE_ENV !== 'production') globalForPrisma.prisma = prisma
```

The Typescript type is misleading as it does not consider the fact that the prisma instance is `undefined` prior to instantiation. 

Additionally, the `||` operator can be replaced with `??` as the intent here is to signal a fallback value when the `prisma` instance is 'undefined' rather than 'falsy'.

The page in question is: [Best practice for instantiating PrismaClient with NextJS](https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices)

JingLi1998:
## Describe this PR

Improves the typescript types and operators in documentation to be more accurate.

This is because prior to instantiation, the global value for prisma is actually `undefined` as nothing has been instantiated.

## Changes

Add `undefined` to global type definition.

Replace `||` with `??`

## What issue does this fix?

[Link to issue](https://github.com/prisma/docs/issues/4566)

Addresses #4566 

N/A

### DIFF
--- a/content/300-guides/050-database/880-troubleshooting-orm/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
+++ b/content/300-guides/050-database/880-troubleshooting-orm/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx
@@ -23,10 +23,12 @@ The solution in this case is to instantiate a single instance `PrismaClient` and
 ```ts file=./db
 import { PrismaClient } from '@prisma/client'
 
-const globalForPrisma = global as unknown as { prisma: PrismaClient }
+const globalForPrisma = global as unknown as { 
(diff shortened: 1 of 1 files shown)


JingLi1998:
@all-contributors please add @JingLi1998  for doc and example

sabinadams (approved):

PR #4569, opened 2023-03-08
Fix CLI docs `migrate dev --create-only` example description

skplunkerin:
## Describe this PR

The **Prisma CLI Reference** docs are incorrect for the [`migrate dev --create-only` example](https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-dev);
the Docs state that using `--create-only` will "Apply all migrations", when it shouldn't (right?).

## Changes

* Remove incorrect "Apply all migrations..." text from `migrate dev --create-only` example

## What issue does this fix?

No specific issue for this, I went straight to fixing the inaccurate docs.

N/A

### DIFF
--- a/content/400-reference/200-api-reference/200-command-reference.mdx
+++ b/content/400-reference/200-api-reference/200-command-reference.mdx
@@ -951,7 +951,7 @@ Apply all migrations, then create and apply any new migrations:
 prisma migrate dev
 ```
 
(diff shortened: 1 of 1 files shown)


janpio:
I actually does apply all pending migrations, but then _not_ apply the new one it created. That is surprising and should be documented - probably even a bit more explicit than it currently is.

PR #4576, opened 2023-03-14
docs: add relation name for consistency

devpolo:
## Describe this PR

Replacing every `@relation("Name")` by `@relation(name: "Name")` for consistency.

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/index.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/index.mdx
@@ -516,17 +516,17 @@ To disambiguate these relations, you need to annotate the relation fields with t
 model User {
   id           Int     @id @default(autoincrement())
   name         String?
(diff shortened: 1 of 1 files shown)


janpio:
Why do you suggest to make `name` explicit in all relation attributes?

devpolo:
To maintain uniformity, why is the name attribute present in some instances in the documentation and absent in others?

janpio:
That might just be a mistake. The default for `name` is not to render the attribute name I think, so we should possibly unify in the other direction unless there is a good reason to be verbose in this example.

sabinadams:
> That might just be a mistake. The default for `name` is not to render the attribute name I think, so we should possibly unify in the other direction unless there is a good reason to be verbose in this example.

I agree here. Explicitly adding the `name` attribute in a real-world schema could (and likely would) negatively impact the size and readability of the document. The default should be to not have the `name` attribute in the docs to match what a real schema should contain. 

I can make the switch here. Thanks @devpolo for the change suggestion here!

Issue #4414, opened 2023-01-16
Link checker does not seem to catch certain broken anchor links


zachtil:
After upgrading to Node `v18.13.0`, the latest link checker improvements started working locally again. So right now, I am seeing the following broken anchor links:

```
warn 1 broken links found on /data-platform/projects/create/
warn 162:13   /data-platform/data-proxy#using-the-data-proxy-in-a-prisma-application
⠀
warn 1 broken links found on /guides/database/developing-with-prisma-migrate/generating-down-migrations/
warn  10:86   /concepts/components/prisma-migrate#migration-history
⠀
warn 2 broken links found on /concepts/components/prisma-schema/relations/one-to-many-relations/
warn  52:124  /concepts/components/prisma-schema/relations#annotated-relation-fields-and-relation-scalar-fields
warn 208:7    /concepts/components/prisma-schema/relations#annotated-relation-fields-and-relation-scalar-fields
⠀
warn 1 broken links found on /guides/upgrade-guides/upgrade-from-prisma-1/schema-incompatibilities/
warn 421:71   /concepts/components/prisma-schema/relations#conventions-for-relation-tables-in-implicit-m-n-relations
⠀
warn 1 broken links found on /concepts/components/prisma-migrate/migrate-development-production/
warn  40:41   /concepts/components/prisma-migrate#do-not-edit-or-delete-migrations-that-have-been-applied
⠀
warn 1 broken links found on /guides/database/developing-with-prisma-migrate/squashing-migrations/
warn   9:45   /concepts/components/prisma-migrate#migration-history
⠀
warn 1 broken links found on /guides/deployment/deploy-database-changes-with-prisma-migrate/
warn  33:3    /concepts/components/prisma-migrate#production-and-testing-environments
⠀
warn 1 broken links found on /reference/api-reference/prisma-client-reference/
warn 5554:5    /concepts/components/prisma-client/middleware#examples
⠀
warn 1 broken links found on /concepts/components/prisma-schema/relations/self-relations/
warn  57:32   /concepts/components/prisma-schema/relations#annotated-relation-fields-and-relation-scalar-fields
⠀
warn 4 broken links found on /concepts/components/introspection/
warn 199:15   /concepts/components/prisma-schema/relations#one-to-one-relations
warn 229:27   /concepts/components/prisma-schema/relations#one-to-many-relations
warn 259:1    /concepts/components/prisma-schema/relations#many-to-many-relations
[...]

Issue #4084, opened 2022-11-18
Link checker not running consistently on Vercel


keerlu:
We're now seeing problems with the link checker running inconsistently on Vercel builds, so that broken links are making it into `main` without us realising.

Example:
- link checker fails: https://vercel.com/prisma/docs/9dji4SKgw3L8VKYMMTfpPWTPBnLB
- no link checker failure: https://vercel.com/prisma/docs/CkBXKHxNKhphs46RXLcmwSyGJtyb

Only difference between these builds is printing a different string when the link checker is run.

See [Slack thread](https://prisma-company.slack.com/archives/C02JV0K2Q31/p1668762943369889) for more details.

nilubava:
Also, a new issue:

The PR `https://github.com/prisma/docs/pull/4236/files` caused a failure in the github action `ci.yml` due to link checker - https://github.com/prisma/docs/actions/runs/4017060499

Temp fix: make it report the broken links as a warning and continue with the next steps of the workflow. 

Todo: Make the link checker fail the build. Resolve the broken links reported


Context: https://prisma-company.slack.com/archives/C02JV0K2Q31/p1674722924837509

Issue #3330, opened 2022-06-27
Disallow links starting with `https://www.prisma.io/docs/`


janpio:
Links starting with `https://www.prisma.io/docs/` should be relative and not use the full, qualified URL with domain and full path - so we should just let the link checker fail for these and output a helpful error message instead.

Issue #2734, opened 2022-01-17
Warning about links that use `https://www.prisma.io/docs`


janpio:
Generally we want links to not use the domain and path, so they also all work on preview deployments.

We could even consider making these links invalid.

meletj:
Related to #3330

nilubava:
Related to #2512

PR #4602, opened 2023-03-27
fix: bugs and improvements in link checker

nilubava:
Fixes:
- #4414
- #4084 
- #2512 
- #2500 
- #2650 
- #2734 
- #3330 
- #2723 
- #4628


janpio:
Nice:

```
warning /300-guides/200-deployment/110-deployment-guides/600-deploying-migrations-from-a-local-environment contains links with trailing slashes: ./ - Please remove trailing slashes.
warning /200-concepts/100-components/01-prisma-schema/06-relations/420-relation-mode contains links with domain name prisma.io: https://www.prisma.io/docs/guides/database/using-prisma-with-planetscale#how-to-emulate-relations-in-prisma-client - Please use relative urls instead.
```
I think "use relative urls instead" is not correct though. We don't want relative links necessarily, but links without the domain that are still fully qualified in the path and not relative to the current page. Not sure how those are called actually.


Does this really cover https://github.com/prisma/docs/issues/2500 or https://github.com/prisma/docs/issues/2650 though? Would those be already "applied" when the linkchecker runs

janpio:
A few things that might be false positives:
```
warning /200-concepts/100-components/01-prisma-schema/04-data-model contains links with trailing slashes: ./ - Please remove trailing slashes.
```
Does that actually cause a link with a trailing slash and not just the URL of the current page?

```
warning /200-concepts/100-components/02-prisma-client/058-transactions contains links with domain name prisma.io: https://www.prisma.io/dataguide/intro/database-glossary#isolation-levels - Please use relative urls instead.
```
Data guide should be ok - only docs internal links are problematic as it makes the previews not work.

Also:
```
Error: 0 broken (or redirected) internal links and 42 broken anchors found
```

Do these 2 new categories of broken links (trailing slash, and external URLs for prisma.io) count to "broken anchors"?

nilubava:
@janpio 
> I think "use relative urls instead" is not correct though: 

I can change the wording to "Remove the domain name prisma.io from the link"?

> A few things that might be false positives

I can exclude such links from the warnings.

> Data guide should be ok 

I included dataguide and blog since they can also be written as relative links without domain names. 

> Do these 2 new categories of broken links (trailing slash, and external URLs for prisma.io) count to "broken anchors"?

Nope, they are caught in another plugin which transforms the urls (`gatsby-remark-to-absolueturl`). The broken anchors and links are caught in `gatsby-remark-check-links-numberless`. Hence the numbers are not for all the broken links

janpio:
> I can change the wording to "Remove the domain name prisma.io from the link"?

👍 

> I included dataguide and blog since they can also be written as relative links without domain names.

I would only apply that to things that will work in the preview deployment.

> Nope, they are caught in another plugin which transforms the urls (gatsby-remark-to-absolueturl). The broken anchors and links are caught in gatsby-remark-check-links-numberless. Hence the numbers are not for all the broken links

Ahh, that makes sense - I did not actually look at the code. Would they also break the build? Maybe we should add a summary reporting that is output as well for them so it is easier to see what is going on? I had to look quite closely to actually notice these in the output.

nilubava:
@janpio 
> Maybe we should add a summary reporting that is output as well

Yes, that makes sense. I will add this

content/300-guides/300-upgrade-guides/800-upgrade-from-prisma-1/01-how-to-upgrade.mdx
```diff
@@ -17,7 +17,7 @@ The upgrade documentation consists of several pages, here's an overview of how t
 
 In addition to these two pages, there are various _practical guides_ that walk you through an example scenario of the upgrade process:
 
-- [Upgrading the Prisma layer](/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-the-prisma-layer-postgres): No matter what your Prisma 1 setup looks like, you should **always start your upgrade process by following this guide**.
+- [Upgrading the Prisma layer](/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-the-prisma-layer-postgres-postgres-postgres-postgres-postgres): No matter what your Prisma 1 setup looks like, you should **always start your upgrade process by following this guide**.
```
janpio: This look broken 😆

nilubava: Ha! that was a logic glitch 😆 I have fixed those now

nilubava:
@janpio Looks like it can be merged now. I have fixed the redirects using the new script and it works as expected. I compared with Screaming Frog 308s and looks like it caught everything.

Should we make it a warning and not throw an error if links fail? This might help the doc writers to fix the links when they have time and not block further PRs (due to link errors). Once we have confidence that we have fixed all the links, we can think about making it an error again. WDYT?

PR #4620, opened 2023-03-30
change: Remove percy from repo

nilubava:
#4619


config.ts
```diff
@@ -7,7 +7,6 @@ const siteConfig = {
   },
   redirects: [
     // Added this object for redirect because gatsby filters through the redirects key in gatsby-browser.js.
-    // This object will create a "dummy" page in percy, creating a 404 page of `/foo/index`
```
janpio, line 10: How do we screenshot 404 for LP?

src/styles/layout.css
janpio, line 1: How do we hide these things for LP?

src/utils/stats.js
```diff
@@ -25,12 +25,7 @@ module.exports = {
   trackPage(page) {
     const { host } = window.location
 
-    if (
-      host.includes('netlify') ||
-      host.includes('localhost') ||
-      host.includes('percy') ||
-      host.includes('vercel')
-    ) {
+    if (host.includes('netlify') || host.includes('localhost') || host.includes('vercel')) {
```
janpio, line 28: Is LP covered via `localhost`?
Then a comment would be useful.

PR #4624, opened 2023-03-30
ci: Comment with suggestions for redirects

janpio:
This PR adds a script that leaves a comment with suggested redirects, for when you rename or delete files.


janpio:
Seems this was already merged some other way.

PR #4629, opened 2023-03-31
Adds section on viewing connection pool status with logging & metrics

### LINKED ISSUE #575: Query Engine: Size of connection pool
## Suggested action

Add note to Query Engine docs that size of connection pool can be logged with logging/debugging options (probably `prisma:engine` or just `*`)

sabinadams:
Closes #575

Adds a section named `Viewing the connection pool size` to `Prisma Client -> Working with Prisma Client -> Connection pool` that covers retrieving information about the connection pool size and state of the pool & connections using logging & metrics.


content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -57,6 +57,115 @@ datasource db {
...
+```text no-copy
+prisma:info Starting a postgresql pool with 21 connections.
+```
+
+</cmdResult>
+</CodeWithResult>
+
+When the `PrismaClient` class was instantiated, the logging notified `stdout` that a connection pool with 21 connections was started.
+
+If you need even more insights into the size of your connection pool and the amount of in-use and idle connection, you can use metrics.
```
janpio: Note that metrics is still a preview feature, so that should be mentioned here.

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
```diff
@@ -57,6 +57,115 @@ datasource db {
...
+<cmdResult>
+
+```text no-copy
+prisma:info Starting a postgresql pool with 21 connections.
+```
+
+</cmdResult>
+</CodeWithResult>
+
+When the `PrismaClient` class was instantiated, the logging notified `stdout` that a connection pool with 21 connections was started.
```
janpio, line 95: Hmmmm. Documenting this kinda makes it a public API, which the logs _content_ usually is not.
I don't want people to build tooling on that and then be unhappy when we (accidentally or intentionally) change that.
Ideas how to handle?

nikolasburk: We could add a warning that the output of `log: ['info']` can change in any release?

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
nikolasburk: how about this?

```suggestion
If you need even more insights into the size of your connection pool and the amount of in-use and idle connection, you can use the [metrics](/concepts/components/prisma-client/metrics) feature (which is currently in Preview).
```

content/200-concepts/100-components/02-prisma-client/000-working-with-prismaclient/115-connection-pool.mdx
nikolasburk:
```suggestion
When the `PrismaClient` class was instantiated, the logging notified `stdout` that a connection pool with 21 connections was started.

<Admonition type="warning">

Note that the output generated by `log: ['info']` can change in any release without notice. Be aware of this in case you are relying on the output in your application or a tool that you're building.

</Admonition>
```

nikolasburk (commented):
@janpio I think this is really useful content. I addressed your comments with two suggestions, what do you think of these? Let's commit them and then merge?

janpio (approved):

PR #4633, opened 2023-04-02
Update 100-connect-your-database.mdx

cryskram:
## Describe this PR

`relationMode` attribute is no more required

## Changes

The `relationMode` attribute in the `datasource` is not required for version 4.12.0 onwards

### DIFF
--- a/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
+++ b/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
@@ -130,16 +130,6 @@ datasource db {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


janpio:
Uhm yes it is? Only the preview feature flag got removed some versions ago. Using Prisma with a PlanetScale database still requires `relationMode=prisma`.

janpio (changes requested):
Uhm yes it is? Only the preview feature flag got removed some versions ago. 
Using Prisma with a PlanetScale database still requires relationMode=prisma.

cryskram:
But with one of my recent projects, ignoring the line `relationMode = "prisma"` in the `datasource` block, passed without any errors. I am using Prisma version `4.12.0`. 
So does it make `relationMode` optional?

janpio:
Are you using relations in that project?
`relationMode` is only required if you want to use relations and migrate those on PlanetScale.

cryskram:
My apologies for not understanding the concept adequately and opening a pull request. Yes, you are right, `relationMode` is only required when we have relations in our schema. Else it is not required.
Thanks for getting it into the light. Once again, my apologies.

janpio:
All good. Thanks for spending the time here!

PR #4647, opened 2023-04-10
Add note about using `upsert` as `findOrCreate`

### LINKED ISSUE #640: Explain how `upsert` can behave as a hypothetical `findOrCreate`
# Problem

Currently, people are wondering about having a `findOrCreate` method in the API (https://github.com/prisma/prisma-client-js/issues/85).
We also have recently introduced `connectOrCreate` as an experimental feature (#568).

It happens that `upsert` behaves like a hypothetical `findOrCreate` method when receiving an empty `update` parameter.
There has been a decision to avoid adding shortcuts in the API to already supported methods to avoid unnecessarily expand the surface to test, but having a documentation update would help people figuring that out.

# Suggested solutions

 * We update the [`upsert` documentation](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/crud#upsert) to cover the usage with an empty `update` parameter.
 * Optionally (probably a bad idea): we add a `findOrCreate` section which explains one should use `upsert` as documented above to have this behavior

TasinIshmam:
Adds a note about using `upsert` as a workaround for `findOrCreate`. 

Fixes #640

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/030-crud.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/030-crud.mdx
@@ -789,7 +789,7 @@ From version 4.6.0, Prisma carries out upserts with database native SQL commands
 
 Prisma does not have a `findOrCreate` query. You can use `upsert` as a workaround. To make `upsert` behave like a `findOrCreate` method, provide an empty `update` parameter to `upsert`. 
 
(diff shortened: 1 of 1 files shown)


nikolasburk (approved):
looks great!

PR #4648, opened 2023-04-10
Add documentation on prepared statement caching

### LINKED ISSUE #495: PostgreSQL statement caching
The upcoming 2.1.0 release will introduce prepared statement caching for PostgreSQL connector. The cache will allow us to reuse compiled statements in identical parameterized queries.

Example

```sql
SELECT * FROM user WHERE name = "John";
SELECT * FROM user WHERE name = "Brenda";
```

The two queries after parameterization will be the same, and the second query can skip the preparing step, saving database CPU and one extra roundtrip. Query after parameterization:

```sql
SELECT * FROM user WHERE name = $1
```

The connection string for PostgreSQL has now an extra parameter `statement_cache_size=N`, where one can tweak the number of items stored in the cache, N being the number of unique statements stored. After the cache gets full, adding a new statement will remove the oldest statement from the cache.

By default, we store 500 statements per connection. This means if using more than one connection per instance of Prisma, every connection will have a separate cache. Testing this with 33 connections and cache of 500, the average extra memory needed was about 14-18 megabytes. Caching made our test queries to be 12-33% faster depending on the query.

Due to the nature of pgBouncer, if the `pgbouncer` parameter is set to `true`, the statement cache is automatically disabled for that connection.

TasinIshmam:
Fixes #495


content/200-concepts/200-database-connectors/03-postgresql.mdx
```diff
@@ -212,3 +213,11 @@ model Device {
   data Unsupported("circle")
 }
 ```
+
+## Prepared statement caching
+
+A [prepared statement](https://www.postgresql.org/docs/current/sql-prepare.html) is a feature that can be used to optimize performance. A prepared statement is parsed, compiled, and optimized only once and then can be executed directly multiple times without the overhead of parsing the query again. This way, Prisma Client's [query engine](https://www.prisma.io/docs/concepts/components/prisma-engines/query-engine) does not repeatedly prepare the same statement but can reuse an existing one which reduces database CPU usage and query latency.
```
nikolasburk: Prisma's behaviour is currently not super clear from this.

> This way, Prisma Client's [query engine](https://www.prisma.io/docs/concepts/components/prisma-engines/query-engine) does not repeatedly prepare the same statement but can reuse an existing one which reduces database CPU usage and query latency.

What does "this way" refer to exactly here? The two previous sentences were about "prepared statements" but it doesnt't say _how_ Prisma utilizes prepared statements exactly. It also doesn't state whether the dveloper can in any way influence this behaviour.

nikolasburk (changes requested):
requesting a small change for clafirication here, otherwise looks great!

content/200-concepts/200-database-connectors/03-postgresql.mdx
```diff
@@ -68,23 +68,24 @@ postgresql://USER:PASSWORD@HOST:PORT/DATABASE?KEY1=VALUE&KEY2=VALUE&KEY3=VALUE
...
+| `host`                 | No       |                        | Points to a directory that contains a socket to be used for the connection                                                                                                        |
+| `socket_timeout`       | No       |                        | Maximum number of seconds to wait until a single query terminates                                                                                                                 |
+| `pgbouncer`            | No       | `false`                | Configure the Engine to [enable PgBouncer compatibility mode](/guides/performance-and-optimization/connection-management/configure-pg-bouncer)                                    |
+| `statement_cache_size` | No       | `500`                  | Since 2.1.0: Speciefies the number of [prepared statements](#prepared-statement-caching) cached per connection                                                                    |
```
janpio:
```suggestion
| `statement_cache_size` | No       | `500`                  | Since 2.1.0: Specifies the number of [prepared statements](#prepared-statement-caching) cached per connection                                                                    |
```

content/200-concepts/200-database-connectors/03-postgresql.mdx
TasinIshmam: Clarified. Please have another look @nikolasburk

nikolasburk (approved):

PR #4650, opened 2023-04-10
Clarify PrismaPromise behavior

### LINKED ISSUE #800: Clarify when Prisma queries are run (PrismaPromise behavior)
@Sytten brought up a good point in Slack that we don't really talk about when queries are executed.

In this example:

```js
import { User } from '@prisma/client'

export default async (app: Application, { id }: DeleteInput): Promise<User> => {
  const operations: Promise<any>[] = []

  // Delete onboarding
  operations.push(app.prisma.onboarding.deleteMany({ where: { userId: id } }))

  // Delete availability application
  const availabilityApp = await app.models.availabilityApp.findOne({ userId: id })

  if (availabilityApp) {
    operations.push(
      app.prisma.availabilityTemplate.deleteMany({
        where: { appId: availabilityApp.id },
      }),
      app.prisma.availabilityApp.delete({ where: { id: availabilityApp.id } })
    )
  }

  // Delete user
  const user = await app.prisma.user.findOne({ where: { id } })
  operations.push(app.prisma.user.delete({ where: { id } }))

  // Execute all operations
  await app.prisma.transaction(operations)
  return user!
}
```

You might expect the operations to run in the order that they appear, but this is not the case. They run more like this:

```js
await app.models.availabilityApp.findOne({ userId: id })
await app.prisma.user.findOne({ where: { id } })

// batched together at the end
app.prisma.onboarding.deleteMany({ where: { userId: id } })
app.prisma.availabilityTemplate.deleteMany({
  where: { appId: availabilityApp.id },
}),
app.prisma.availabilityApp.delete({ where: { id: availabilityApp.id } })
app.prisma.user.delete({ where: { id } })
```

This is because Prisma queries are then-ables, meaning they only execute when you call `await` or `.then()` or `.catch()`. This is called lazy evaluation. This is different then a regular promise which starts executing immediately. This would be called eager evaluation or immediate evaluation.

This was an understandable source of confusion for @Sytten because most libraries using promises eagerly evaluate. For example:

```js
const fs = require('fs/promises')
fs.readFile("./test.txt") // starts reading the file immediately!
```

But in the Prisma client:

```js
const deleteOperation = app.prisma.onboarding.deleteMany({ where: { userId: id } }) // doesn't do anything
// ...
deleteOperation.then() // now it executes
// or..
await deleteOperation // now it executes.
```

TasinIshmam:
Fixes #800

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/index.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/index.mdx
@@ -108,8 +108,8 @@ const users = await prisma.user.findMany()
 
 <Admonition type="info">
 
-- Note that queries will only run if used with `await` or chained with `.then()`. When using [`$transaction`](transactions)s, this makes it possible for the client to pass all the queries on to the [query engine](/concepts/components/prisma-engines/query-engine) as a transaction.
-- The client methods are ["thenable"](https://masteringjs.io/tutorials/fundamentals/thenable), they return a PrismaPromise which is implemented like a JavaScript Promise, except for the query to be executed they need to be awaited or chained with `.then()`.
+- Note that queries will only run if used with `await` or chained with `.then()` or `.catch()`. When using [`$transaction`](transactions)s, this makes it possible for the client to pass all the queries on to the [query engine](/concepts/components/prisma-engines/query-engine) as a transaction.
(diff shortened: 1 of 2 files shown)


nikolasburk (approved):
looks great!

PR #4654, opened 2023-04-11
add reference to crud page to home page

### LINKED ISSUE #4579: Increase the prevalence of the CRUD doc
Posted in slack, but was asked to make it a doc request, so here I am doing that.

> [10:52 AM](https://prisma.slack.com/archives/C01739JGFCM/p1677091944685479) isaacs
 this might be kind of a weird request, but there should be a faster path to get to this docs page: https://www.prisma.io/docs/concepts/components/prisma-client/crud  Like, it should be linked in the header of every page, it's 99% of what you need when working day to day on an app that uses Prisma, and easily a significant portion of Prisma's value to me as a developer, but tbh, I can never find it without searching my history, it's so buried.  Maybe even register a short domain name that just 301 redirects to it.  [pcrud.dev](http://pcrud.dev/) is available.
> [10:55 AM](https://prisma.slack.com/archives/C01739JGFCM/p1677092130530349) isaacs
 Most of the other topics in the docs, you need like one time, and hardly ever again, or they're concepts that you learn once and ok now you understand that, and don't need to read again.  But every day it seems, I've got TypeScript yelling at me, and I'm going "wait, is it select or include? Do I put this in data, or right in the object?" and have to go digging for the CRUD page that just shows the answer immediately.  It's a beautiful piece of documentation, I've even thought about just printing it out and mounting it on the wall by my desk.
> [10:55 AM](https://prisma.slack.com/archives/C01739JGFCM/p1677092159150829)
Not saying the rest of the docs aren't good or useful, they absolutely are, but nothing compares to the CRUD page.

nikolasburk:
Fixes #4579

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -105,6 +105,11 @@ const siteConfig = {
             url: 'concepts/components/prisma-engines',
             codeBlock: false,
           },
+          {
+            text: 'CRUD',
+            url: 'concepts/components/prisma-client/crud',
+            codeBlock: false,
+          },
         ],
       },
       {


isaacs:
Thanks!

PR #4659, opened 2023-04-13
fix: broken guide link in doc for `prototyping-schema-db-push`

thatkid02:
This commit fixes a broken link that was leading to a 404-error page. Right-click action will redirect to a proper page but on-click leads to 404-error page.

## Describe this PR

 - `prototyping-schema-db-push` guide exists under migrate page but existing link in the current page leads to 404-error page.

## Changes

- updated working link

## What issue does this fix?

- redirects to correct guide page for on-click action.

### DIFF
--- a/content/200-concepts/100-components/03-prisma-migrate/150-db-push.mdx
+++ b/content/200-concepts/100-components/03-prisma-migrate/150-db-push.mdx
@@ -44,7 +44,7 @@ See [Schema prototyping with `db push`](/guides/database/prototyping-schema-db-p
 
 ## Can I use Prisma Migrate and <inlinecode>db push</inlinecode> together?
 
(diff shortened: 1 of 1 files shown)


janpio:
@nilubava 
Why does this return an error on normal click?
Why did the link checker not find this and make us fix it earlier?

thatkid02:
@janpio Sorry for this invalid PR. Seems like all redirects are failing with on-click. I tried out with click actions on chrome, firefox and mobile chrome.
I have some issues with authentication of npm fontawesome so couldn't run locally.

janpio:
No, your PR is absolutely correct! This highlights we have a problem, and fixes it.

janpio:
I raised this internally now so we can figure out why those links don't work when staying inside the app, but only if you open them in a new tab or copy the URL.

PR #4669, opened 2023-04-17
chore (docs/reference): update JSON Array filtering example and docs 

### LINKED ISSUE #4413: Validate example of JSON Array filtering
Examples in our documentation of [array_starts_with](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#array_starts_with) and [array_ends_with](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#array_ends_with) doesn't work as expected.

There is a note that mentions that
> the value of array_starts_with must be an array and not a string, even if the array only contains a single value

But on passing an array as mentioned, no records are returned. But if only a string is passed, then correct records are returned.

We might need to validate if the note is incorrect or not.

For reproduction, here's the linked GitHub Discussion: [#17327](https://github.com/prisma/prisma/discussions/17327)

ruheni:
Closes #4413

- Fixes "PostgresSQL" -> "PostgreSQL" typos
-  Updates JSON array filtering examples and admonition


content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -323,7 +323,7 @@ const getActiveUsers = await prisma.user.findMany({
 
 ### Sort by relevance (PostgreSQL)
 
-In [3.5.0](https://github.com/prisma/prisma/releases/3.5.0) and later, when using PostgresSQL you can sort records by relevance to the query using the `_relevance` keyword. This uses the relevance ranking functions from PostgreSQL's full text search feature, which are explained further in [the PostgreSQL documentation](https://www.postgresql.org/docs/12/textsearch-controls.html).
```
janpio, line 326: Interesting why cSpell did not warn about this already... that is why we have a spell checker.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -5371,18 +5371,12 @@ const getUsers = await prisma.user.findMany({
...
     },
   },
 })
 ```
 
-<Admonition type="info">
-
-**Note**: In PostgresSQL, the value of `array_starts_with` must be an array and not a string, even if the array only contains a single value.
-
-</Admonition>
```
janpio, line 5384: Source for change?

ruheni: Internal slack conversation: https://prisma-company.slack.com/archives/C02FZENHD4N/p1680705609602039

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
ruheni: It may be likely that cSpell accepts both spellings and isn't as strict as we(Prisma) are on terminology 🤔

PR #4684, opened 2023-04-21
Update broken url

jordanstephens:
## Describe this PR
FIxing a broken link. Looks like this was overlooked in #4602.


janpio:
Thanks @jordanstephens, this got fixed in the meantime. Thanks for opening the PR anyway!

PR #4728, opened 2023-05-02
fix: add missing binary targets to the "Prisma schema reference" page

aqrln:
## Describe this PR

This PR adds missing binary targets to the "`binaryTargets` options"
section of the "Prisma schema reference" page, and fixes incorrect or
outdated information.

Most of these tables will not be necessary anymore once we unify glibc
engines on x86_64 as well, but for now we have to maintain them while
they are here.

## Changes

- Add missing `linux-musl-arm64-openssl-1.1.x` and
  `linux-musl-arm64-openssl-3.0.x` targets.
- Remove outdated comment about Alpine not being supported on ARM64.
- Adjust the "Linux ARM64" section accordingly to reflect that this
  binary target is only for glibc-based distros but not Alpine.
- Add the CPU architecture where appropriate.
- Fix incorrect entry in the CentOS section.
- Add missing entries to the rest of the sections.

## What issue does this fix?

Closes: https://github.com/prisma/client-planning/issues/372

### DIFF
--- a/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
+++ b/content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
@@ -191,16 +191,18 @@ The following tables list all supported operating systems with the name of platf
 
 ##### Linux (Alpine)
 
-Currently supported for x86_64 only.
-
-| Build OS            | Prisma engine build name     | OpenSSL |
-| :------------------ | :--------------------------- | :-----: |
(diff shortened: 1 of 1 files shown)


janpio (approved):

aqrln:
Approved by @ruheni in slack, I'll go ahead and merge it.

Failing checks are expected, Percy is not used anymore while LostPixel doesn't currently work because of problems on their side.

PR #4747, opened 2023-05-05
fix: change folder to directory

Jolg42:
This page was mixing folder and directory

I think “directory” fits better (https://stackoverflow.com/questions/5078676/what-is-the-difference-between-a-directory-and-a-folder) (edited)

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/08-views.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/08-views.mdx
@@ -289,7 +289,7 @@ When re-introspecting your database, any custom changes to your view definitions
 
 #### The `views` directory
 
-Introspection of a database with one or more existing views will also create a new `views` directory within your `prisma` directory (starting with Prisma version 4.12.0). This directory will contain a directory named after your database's schema which contains a `.sql` file for each view that was introspected in that schema. Each file will be named after an individual view and will contain the query the related view defines.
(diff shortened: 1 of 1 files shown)


ruheni (approved):

content/200-concepts/100-components/01-prisma-schema/08-views.mdx
```diff
@@ -289,7 +289,7 @@ When re-introspecting your database, any custom changes to your view definitions
 
 #### The `views` directory
 
-Introspection of a database with one or more existing views will also create a new `views` folder within your `prisma` directory (starting with Prisma version 4.12.0). This folder will contain a folder named after your database's schema which contains a `.sql` file for each view that was introspected in that schema. Each file will be named after an individual view and will contain the query the related view defines.
+Introspection of a database with one or more existing views will also create a new `views` directory within your `prisma` directory (starting with Prisma version 4.12.0). This directory will contain a directory named after your database's schema which contains a `.sql` file for each view that was introspected in that schema. Each file will be named after an individual view and will contain the query the related view defines.
```
janpio: Maybe also?
```suggestion
Introspection of a database with one or more existing views will also create a new `views` directory within your `prisma` directory (starting with Prisma version 4.12.0). This directory will contain a subdirectory named after your database's schema which contains a `.sql` file for each view that was introspected in that schema. Each file will be named after an individual view and will contain the query the related view defines.
```

janpio (approved):

PR #4748, opened 2023-05-05
fix: remove fontawesome

nilubava:
https://github.com/prisma/lens/issues/524


nilubava:
@carlagn Works locally but not on Vercel deployments

PR #4749, opened 2023-05-05
docs: add specific notes for serverless deployments when using Supabase

jharrell:
## Describe this PR

This adds a Supabase specific section on deployment to serverless envs. Notably, it calls out the need for `pgbouncer=true` to be appended to the end of the `DATABASE_URL` connection string.

### DIFF
--- a/content/300-guides/050-database/880-supabase.mdx
+++ b/content/300-guides/050-database/880-supabase.mdx
@@ -31,14 +31,14 @@ Many aspects of using Prisma with Supabase are just like using Prisma with any o
 ## Specific considerations
 
 If you'd like to use the [connection pooling feature](https://supabase.com/docs/guides/integrations/prisma#connection-pooling-with-supabase) available with Supabase, you will
-need to add `pgbouncer=true` to the end of your `DATABASE_URL` string.
+need to add `pgbouncer=true` to the end of the `DATABASE_URL` environment variable used in the `datasource.url` property:
 
 ```env file=.env
 # Connect to Supabase with PgBouncer.
 DATABASE_URL="postgres://postgres:__PASSWORD__@db.__YOUR_SUPABASE_PROJECT__.supabase.co:6543/postgres?pgbouncer=true"
 ```
 
-If you would like to use the Prisma CLI in order to perform other actions on your database (e.g. migrations) you will need to add a `DIRECT_URL` string so that the CLI can bypass PgBouncer:
+If you would like to use the Prisma CLI in order to perform other actions on your database (e.g. migrations) you will need to add a `DIRECT_URL` environment variable to use in the `datasource.directUrl` property so that the CLI can bypass PgBouncer:
 
 ```env file=.env highlight=4-5;add
 # Connect to Supabase with PgBouncer.
--- a/content/300-guides/050-database/880-supabase.mdx
+++ b/content/300-guides/050-database/880-supabase.mdx
@@ -30,19 +30,27 @@ Many aspects of using Prisma with Supabase are just like using Prisma with any o
 
 ## Specific considerations
 
-If your application is being deployed to a serverless environment, make sure that you follow our general guidance on [connection pooling](https://www.prisma.io/docs/guides/performance-and-optimization/connection-management#recommended-connection-pool-size-1). Supabase makes this easy by [providing connection pooling](https://supabase.com/docs/guides/integrations/prisma#connection-pooling-with-supabase) compatible with PgBouncer.
-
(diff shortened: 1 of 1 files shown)


content/300-guides/050-database/880-supabase.mdx
```diff
@@ -28,6 +28,28 @@ Many aspects of using Prisma with Supabase are just like using Prisma with any o
 - use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to Supabase
 - use [Prisma Client](/concepts/components/prisma-client) in your application to talk to the database server at Supabase
 
+## Specific considerations
+
+If your application is being deployed to a serverless environment, make sure that you follow our general guidance on [connection pooling](https://www.prisma.io/docs/guides/performance-and-optimization/connection-management#recommended-connection-pool-size-1). Supabase makes this easy by [providing connection pooling](https://supabase.com/docs/guides/integrations/prisma#connection-pooling-with-supabase) compatible with PgBouncer.
```
janpio: The specific link for "connection pooling" you chose is unfortunately terrible, as it is just invalid guidance we have in our docs. It makes no sense to use a connection pool size of 1 if you have a pooler before your database that prevents you running out of connections.

jharrell: Ah okay, bummer. I can change this to be more specific to Supabase and say that a connection pooler should be used. In this case, Supabase's PgBouncer implementation

content/300-guides/050-database/880-supabase.mdx
```diff
@@ -28,6 +28,28 @@ Many aspects of using Prisma with Supabase are just like using Prisma with any o
...
+
+In short, you will need to update your Prisma schema to include `url` as well as `direct_url` as discussed [here](https://www.prisma.io/docs/data-platform/data-proxy/prisma-cli-with-data-proxy#set-a-direct-database-connection-url-in-your-prisma-schema). In addition, you will need to add `pgbouncer=true` to the end of your `DATABASE_URL` string.
```
janpio: I would turn this around. Default action is to use `pgbouncer=true`. And then if you also want to use Prisma CLI to e.g. migrate your database (and you probably want to), `directUrl` makes it possible to also define an unpooled connection string with `directUrl` that is used instead of the default `url` one.

content/300-guides/050-database/880-supabase.mdx
```diff
@@ -28,6 +28,28 @@ Many aspects of using Prisma with Supabase are just like using Prisma with any o
 - use [`db push`](/concepts/components/prisma-migrate/db-push) to push changes in your schema to Supabase
 - use [Prisma Client](/concepts/components/prisma-client) in your application to talk to the database server at Supabase
 
+## Specific considerations
```
janpio, line 31: I think on our side, we could "widen" the description here a bit form the specific use case (as Supabase does it):
If you want to use the database connection with pooling enabled that Supabase offers, these are the things to do.
(Of course we can also mention the use case serverless as the reason for actually wanting to do so - but I guess many users do so in any case as the URL pretty prominent in the Supabase UI)

content/300-guides/050-database/880-supabase.mdx
```diff
@@ -28,8 +28,46 @@ Many aspects of using Prisma with Supabase are just like using Prisma with any o
...
+
+If you'd like to use the [connection pooling feature](https://supabase.com/docs/guides/integrations/prisma#connection-pooling-with-supabase) available with Supabase, you will
+need to add `pgbouncer=true` to the end of your `DATABASE_URL` string.
+
+```env file=.env
+# Connect to Supabase with PgBouncer.
+DATABASE_URL="postgres://postgres:__PASSWORD__@db.__YOUR_SUPABASE_PROJECT__.supabase.co:6543/postgres?pgbouncer=true"
+```
+
+If you would like to use the Prisma CLI in order to perform other actions on your database (e.g. migrations) you will need to add a `DIRECT_URL` string so that the CLI can bypass PgBouncer:
```
janpio: Sorry to be nitpicky here, but `DATABASE_URL` and `DIRECT_URL` are just random env var strings, they could also be `FOO` and `BAR`. The relevant bits are `url` and `directUrl` in the `datasource` block.

You could write:
```
... to the end of the `DATABASE_URL` environment variable used in the `datasource.url` property:
```
and equivalent for the direct URL.

jharrell: I appreciate the feedback! I'll clarify my language here.

PR #4757, opened 2023-05-10
chore: Getting started improvements

### LINKED ISSUE #2747: Getting Started guide improvements (all guides)
These changes apply to all the Getting Started guides. Originally documented as part of the Notion section [here](https://www.notion.so/prismaio/PlanetScale-Getting-Started-feedback-c7cc8b185473451c8cbef2c809a78bb1#f6282418abca4d67842ff8d9bee6fbc1).

Related issue: #2820 for the PlanetScale guide. Many of the points there may apply to all other guides.

In all guides, the following sections need updating:

### Add to existing databases > Querying the database

These need to be updated to match the explanations in 'Start from scratch > Querying the database'. In particular, need to explain that the first time you run `npx ts-node index.ts` it will return an empty array.

###  Add to existing databases > Introspection

- The section starting with “Sometimes though, you may want to make additional changes to the names of the columns...” is kind of out of place for a getting started tutorial. Would be better somewhere else
    - Suggestion: All this could go into a collapsible "Learn how to fix this, if that is not the case" to make the page much shorter.
- PostgreSQL: should now have referential actions in introspected SQL. This probably applies to other DBs too - need to rerun the commands and see. See https://github.com/prisma/docs/issues/2752

### Add to existing databases > Connect your database and Start from scratch > Connect your database
    
- In the 'Connect your database' section starting from: ‘The url is set from an environment variable’: Before this sentence would be a good place to add a structuring headline that lets the user know that everything that follows is now about the connection string or URL. (vs. choosing the connector in the first half of the document)

### LINKED ISSUE #3368: Get started: clarify role of schema
https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/install-prisma-client-typescript-postgres

In my opinion, we should more clearly explain in our get started guide - probably in the above file - that when a user changes their schema, this doesn't change the database - it's just the model that Prisma uses to build the client.

ruheni:
- Closes #3368 (clarify role of schema & workflow)
- Closes #2747 (connection URL format heading & clarification for query response for the first query after introspection)


content/100-getting-started/02-setup-prisma/100-start-from-scratch/120-mongodb/200-install-prisma-client.mdx
```diff
@@ -19,7 +19,7 @@ Notice that the install command automatically invokes `prisma generate` for you
 
 ![Install and generate Prisma Client](./../../../images/prisma-client-install-and-generate.png)
 
-Whenever you make changes to your Prisma schema in the future, you manually need to invoke `prisma generate` in order to accommodate the changes in your Prisma Client API.
+Whenever you update your Prisma schema, you will need to regenerate Prisma Client using the `prisma generate` command to update your Prisma Client API.
```
nikolasburk: I'm actually not sure about the recommended workflow on MongoDB but I thought you were supposed to run `prisma db push` (which would trigger `generate`), or do we always recommend to just run `prisma generate`?

nikolasburk (commented):
looks good, I had one minor question but I think this should be good to be merged

content/100-getting-started/02-setup-prisma/100-start-from-scratch/120-mongodb/200-install-prisma-client.mdx
ruheni: Seems this workflow is also not present in the getting started guide from scratch. I think `prisma db push` creates unique and id fields. 😅

`prisma generate` should work but I'll confirm it before merging the PR

nikolasburk: > `prisma generate` should work but I'll confirm it before merging the PR

Yeah, the question is what is best practice is that we want to recommend to our users and my understanding was that `prisma db push` should be used. (I think we should strive to make it as simple as possible for them and not teach them two approaches that do the same thing but teach one approach that should be default and then maybe at some point they can learn about a different way to achieve the same goal. ).

ruheni: It makes sense to have the `prisma db push` command.

content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/200-install-prisma-client.mdx
```diff
@@ -19,7 +19,7 @@ Notice that the install command automatically invokes `prisma generate` for you
 
 ![Install and generate Prisma Client](../../../images/prisma-client-install-and-generate.png)
 
-Whenever you update your Prisma schema, you will have to update your database schema using either `prisma migrate dev` or `prisma db push`. This will keep your database schema in sync with your Prisma schema. The commands will also regenerate Prisma Client to update your Prisma Client API.
+Whenever you update your Prisma schema, you will have to update your database schema using either `prisma migrate dev` or `prisma db push`. This will keep your database schema in sync with your Prisma schema. The commands will also regenerate Prisma Client.
```
janpio: The information that this updates the Prisma Client API to match the Prisma schema is relevant, I don't think we can assume people understand that intuitively.

ruheni: 🤔

I'll include that in another PR.

PR #4759, opened 2023-05-10
chore: improve wording around the `$on` method in the Reference

### LINKED ISSUE #2888: Clarify documentation for the `$on` method
The wording for the `prisma.$on` method could be improved. [This user](https://github.com/prisma/prisma/discussions/11580) was confused and tried to use it to subscribe to a custom Postgres event.

[Docs currently read like this](https://github.com/prisma/prisma/discussions/11580):

> The $on() method allows you to subscribe to events. With those you can subscribe to [logging events](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log) or the [exit hook](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-management#exit-hooks).

We could mention those are the only two instances where you can use `$on` and maybe provide the examples inline.

ruheni:
Closes #2888

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -5467,7 +5467,7 @@ $connect(): Promise<void>
 <!-- prettier-ignore -->
 ### <inlinecode>$on()</inlinecode>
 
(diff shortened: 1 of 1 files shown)


nikolasburk (approved):
lgtm

janpio (approved):

PR #4760, opened 2023-05-10
Add Example for `now()` attribute function

### LINKED ISSUE #2754: No example for `now()`
Example is missing here somehow: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#now

nurul3101:
## Changes

Added an example for now() function.

## What issue does this fix?
Closes #2754


content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -3234,6 +3234,34 @@ Set a timestamp of the time when a record is created.
...
+
+```prisma
+model User {
+  id        String   @id @default(uuid()) @map("_id")
+  createdAt DateTime @default(now())
+}
+```
+
+</tab>
+</TabbedContent>
```
ruheni: I think we can do away with the `TabbedContent` here since the behavior & syntax is the same in both relational dbs & MongoDB

janpio: As far as I know the convention in the docs is not to do that, because both example _require_ an ID and the id definition of MongoDB has to be different.

ruheni: We could keep the example schemas more succinct. I.e., keep it focused on what the reference item is about. In this case, we could omit the `@default()` attribute function for the `id` field:

```prisma
model User {
  id        String   @id
  createdAt DateTime @default(now())
}
```

Could you point me to the conventions for our Reference content @janpio? I can't find it in our docs.

janpio: I just look at all the existing content.

I don't think the example would be valid for MongoDB.

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -3234,6 +3234,34 @@ Set a timestamp of the time when a record is created.
...
+#### Examples
+
+##### Set current timestamp value when a record is created
+
+<TabbedContent tabs={[<FileWithIcon text="Relational databases" icon="database"/>, <FileWithIcon text="MongoDB" icon="database"/>]}>
+<tab>
+
+```prisma
+model User {
+  id        String   @id @default(uuid())
```
nikolasburk:
```suggestion
  id        String   @id 
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
```diff
@@ -3234,6 +3234,34 @@ Set a timestamp of the time when a record is created.
...
+  createdAt DateTime @default(now())
+}
+```
+
+</tab>
+<tab>
+
+```prisma
+model User {
+  id        String   @id @default(uuid()) @map("_id")
```
nikolasburk:
```suggestion
  id        String   @id @default(auto()) @map("_id")  @db.ObjectId
```

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
nikolasburk: I don't have a super strong opinion here but would generally tend to include an ID (so that the entire model is valid). For relational DBs I'm ok with just using `@id` but for MongoDB we should be using the full blurb that's needed because (afaiu) that's really the only way an ID for MongoDB can be defined.

nikolasburk (changes requested):

content/400-reference/200-api-reference/100-prisma-schema-reference.mdx
nikolasburk: by the way, I opened another issue to double-check whether `cuid()` and `uuid()` are even valid as default values on MongoDB: https://github.com/prisma/docs/issues/4766

nikolasburk (approved):

PR #4795, opened 2023-05-23
chore: remove deprecated `introspect` command in the home page

### LINKED ISSUE #1840: Use db pull instead of introspect as default in docs
## Suggested action

Update all references to `introspect` to `db pull` in all code samples. This includes moving everything under `introspect` to `db pull` here: https://www.prisma.io/docs/reference/api-reference/command-reference#introspect

ruheni:
Closes #1840

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -296,8 +296,8 @@ const siteConfig = {
         icon: 'CLI',
         links: [
           {
-            text: 'introspect',
-            url: 'reference/api-reference/command-reference#introspect',
+            text: 'db',
+            url: 'reference/api-reference/command-reference#db',
             codeBlock: true,
           },
           {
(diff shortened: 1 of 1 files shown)


Jolg42:
Nice, but technically there are missing too?

[version (-v)](https://www.prisma.io/docs/reference/api-reference/command-reference#version--v)
[init](https://www.prisma.io/docs/reference/api-reference/command-reference#init)
[generate](https://www.prisma.io/docs/reference/api-reference/command-reference#generate)
[validate](https://www.prisma.io/docs/reference/api-reference/command-reference#validate)
[format](https://www.prisma.io/docs/reference/api-reference/command-reference#format)

Jolg42 (approved):

ruheni:
That was something I also considered, and it was conflicting. I linked out to the commands that are "more prominent." Also, if we add more commands, we might break the consistency of the size of the cards.

This will be a quick fix for now. We'll think about the docs homepage and give it a fresh look soon. 🤞

CleanShot 2023-05-24 at 18 30 24@2x

Jolg42:
Yes though it feels arbitrary, also it doesn't convey the idea that there are more commands. (Note: generate is a "prominent" command, for example)

Issue #4778, opened 2023-05-17
Mention `directUrl` in the Data Proxy docs


nikolasburk:
Step 1 in this [section](https://www.prisma.io/docs/data-platform/data-proxy/use-data-proxy#steps-1) in the Data Proxy docs should have a hint about:

- the fact that changing `DATABASE_URL` to the proxy URL will disable Prisma Migrate
- link to `directUrl` as a solution for this

nikolasburk:
https://www.prisma.io/docs/data-platform/data-proxy/prisma-cli-with-data-proxy#set-a-direct-database-connection-url-in-your-prisma-schema

nikolasburk:
Better: tell people how to set `directUrl` directly because that'll pretty much always be always want they want

PR #4802, opened 2023-05-24
docs: explain how to set `directUrl`(#4778)

ankur-arch:
- explain why `directUrl` is needed
- demonstrate how to add `directUrl`

### DIFF
--- a/content/800-data-platform/050-data-proxy/100-use-data-proxy.mdx
+++ b/content/800-data-platform/050-data-proxy/100-use-data-proxy.mdx
@@ -119,10 +119,32 @@ You can generate and use Prisma Client for the Data Proxy in the following envir
    <Admonition>
 
    **Note**<br /><br />
-   For security reasons, we strongly recommend that you always use environment variables for your database connection strings and Data Proxy strings. Do not hard-code them into your `schema.prisma` file.<br /><br />
-   For more information, see [Using `.env` files](/guides/development-environment/environment-variables#using-env-files).
+   For security reasons, use environment variables for database connection strings and Data Proxy strings. Do not hard-code them into `schema.prisma`. For more information, see [Using `.env` files](/guides/development-environment/environment-variables#using-env-files).
 
    </Admonition>
+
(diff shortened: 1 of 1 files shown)


content/800-data-platform/050-data-proxy/100-use-data-proxy.mdx
```diff
@@ -119,11 +119,33 @@ You can generate and use Prisma Client for the Data Proxy in the following envir
...
+   ```
+
+   Then, add the direct database connection string in the `.env` file.
+
+   ```bash file=.env highlight=2;add
+   DATABASE_URL="prisma://****:****@****"
+   + DIRECT_URL="postgresql://****:****@****"
+   ```
+
+   However, in older versions, you need to overwrite the `DATABASE_URL` environment variable with the direct database connection string when running migrations. Refer to the [Prisma CLI commands with the Data Proxy](https://www.prisma.io/docs/data-platform/data-proxy/prisma-cli-with-data-proxy) documentation for more information.
```
janpio, line 147: This is now content under the headline `Generate Prisma Client for the Data Proxy`. Shouldn't this be under a different top level heading?

ankur-arch: The content was added to ensure users have a `directUrl` setup when generating the Prisma Client for Data Proxy. As step 1 replaces the database connection, step 2 indicates that migrations won't work without a direct URL and we show how to set that up. I agree it does not go well with the heading, I could move this into a new section below? 
@nikolasburk  @janpio - something like 'Using the direct database connection'

janpio: Yes, `directUrl` is only for Prisma CLI so maybe that can be used. The current headline uses "Prisma Client", so another one about using Prisma CLI in that situation would make sense.

nikolasburk: I agree that the headline isn't very accurate any more, nice catch Jan!

Maybe some of these work better now: 

- Configure project to use Data Proxy
- Configure Prisma project to use Data Proxy
- Configure local project to use Data Proxy
- Configure local Prisma project to use Data Proxy

ankur-arch: Changing the headline to  `Configure Prisma project to use Data Proxy`

PR #4806, opened 2023-05-24
Docs: Unaffected row highlighting instead of actual affected row

yasirunilan:
## Describe this PR

In the introspection documents related to Typescript/Javascript for Postgres, after manual editing, the affected row is highlighted in the given example code. But for the Postgres example, it highlights the unaffected row instead of the affected row. The issue only exists for Postgresql.

## Changes
- Updated the line number that needed to be highlighted.

The following shows the difference between the Postgres example and the MySQL example.
MySQL

mysql

Postgres

postgress


yasirunilan:
@all-contributors please add @yasirunilan for doc and example

ruheni (approved):

ruheni:
Thanks for catching this. 🙏

LGTM

PR #4808, opened 2023-05-24
chore: update community generators section

### LINKED ISSUE #1868: Document community middleware (similar to community generators)
First example: https://www.npmjs.com/package/prisma-lrucache-middleware

### LINKED ISSUE #4121: Consider removing the duplicated generators material and just linking to the main list of generators
Our list of generators is duplicated:
- Docs setion: https://www.prisma.io/docs/concepts/components/prisma-schema/generators#community-generators
- Main Prisma site page: https://www.prisma.io/ecosystem

We could remove the list and simply link to the page on the main site.

Thread about this: https://www.prisma.io/docs/concepts/components/prisma-schema/generators#community-generators

However, Jan points out in that thread, users can do PRs on the docs page to add their own generators.

So we need to decide whether to remove the material in the docs section (and add a link) or not.

### LINKED ISSUE #4221: Serverless - create a docs landing page
We currently have no landing page for serverless. We should consider adding one, to 
- explain that Prisma supports serverless
- link out to the various guides

ruheni:
## Changes

This PR removes the Community generators and directs our community members to the [ecosystem page](https://prisma.io/ecosystem) and Typeform to submit their packages.

Closes #1868
Closes #4121


vercel.json
```diff
@@ -1430,6 +1430,10 @@
     {
       "source": "/docs/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-the-prisma-layer-postgres",
       "destination": "/docs/guides/upgrade-guides/upgrade-from-prisma-1/upgrading-the-prisma-layer-postgresql"
+    },
+    {
+      "source": "/docs/concepts/components/prisma-schema/generators#community-generators",
```
janpio: Can you redirect specific anchors on a page?

janpio:
I still think the docs are the better place, and that `/ecosystem` is not very usable to discover generators.

vercel.json
ruheni: I removed this. I forgot to push the commit

ruheni:
I made a few changes: adding back the  "Community generators" heading, a short description, and a link out to the Ecosystem page.

> I still think the docs are the better place, and that `/ecosystem` is not very usable to discover generators.

While the docs are the right place, having one page where we aggregate all generators, middleware, client extensions, and other types of packages would be valuable. 

- Why do you think it is not very useful to discover generators via the `/ecosystem` page? 
- Is it because we need to talk about it more (internally & externally)? 
- How could we make it the default? 
- Is this something we could improve with SEO & talking about it on social?

nikolasburk:
> I still think the docs are the better place, and that /ecosystem is not very usable to discover generators.

Agree with this in principle! However, we want to unify the places where we list the community packages so removing them from the docs seemed like the best option. 

One idea to improve the situation could be to ask web dev to implement a _direct link_ to the **Generators** section (with the filter applied) on the ecosystem page. That way, at least people would directly navigate to the relevant packages instead of needing to scroll down and applying a filter when they arrive on the ecosystem page.

image
 

How about we ask web dev to create

ruheni:
> How about we ask web dev to create

That’s on their radar. (From a chat with Julie yesterday)

PR #4824, opened 2023-05-30
new filtering FAQ section with info on LIKE and ILIKE

### LINKED ISSUE #4821: Prisma `equals`, `contains`, `startsWith`, and `endsWith` in string filtering doesn't escape SQL characters
Stems from https://github.com/prisma/prisma/issues/19506

The current behavior can be unexpected (as the operators above don't escape wildcard characters) As a consequence, it's worth clarifying in the docs. Please read the original issue report.

jharrell:
Brief PR describing how our filtering uses `LIKE` and `ILIKE` in cases where the DB is sql-like

Resolves #4821 

Happy to link to this section in other areas of the docs. Where would that be most useful? Maybe editing the note at [filter conditions and operators](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference?query=react%20server%20components&page=1#filter-conditions-and-operators) would be sufficient?


content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +234,36 @@ const users = await prisma.user.findMany({
 
 See also: [Case sensitivity](case-sensitivity)
 
+### Filtering FAQs
+
+#### How does filtering work at the database level?
+
+For MySQL and PostgreSQL we are utilizing the [`LIKE`](https://www.w3schools.com/sql/sql_like.asp) (and [`ILIKE`](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE)) operator to search for a given pattern. These operators have built-in pattern matching using symbols unique to `LIKE`. These include `%` for zero or more characters (similar to `*` in other regex implementations) and `_` for one character (similar to `.`).
```
ruheni:
```suggestion
For MySQL and PostgreSQL, Prisma Client utilizes the [`LIKE`](https://www.w3schools.com/sql/sql_like.asp) (and [`ILIKE`](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE)) operator to search for a given pattern. The operators have built-in pattern matching using symbols unique to `LIKE`. The pattern-matching symbols include `%` for zero or more characters (similar to `*` in other regex implementations) and `_` for one character (similar to `.`)
```

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +234,36 @@ const users = await prisma.user.findMany({
...
+
+If you would like to match the literal characters `%` or `_`, make sure to escape those characters. For example:
```
ruheni:
```suggestion
To match the literal characters, `%` or `_`, make sure you escape those characters. For example:
```

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +234,36 @@ const users = await prisma.user.findMany({
...
+const users = await prisma.user.findMany({
+  where: {
+    name: {
+      startsWith: '_',
+    },
+  },
+})
+```
+
+The above query will match any name that starts with any character. That is to say, any user where name is not `NULL`. If you instead wanted to find any user whose name starts with a literal `_` you could do
```
ruheni:
```suggestion
The above query will match any name starting with any character — any user whose name is not `NULL`. If you instead wanted to find any user whose name starts with a literal `_`, you could do:
```

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +234,36 @@ const users = await prisma.user.findMany({
...
+
+```ts highlight=4
+const users = await prisma.user.findMany({
+  where: {
+    name: {
+      startsWith: '_', // note that the `_` character is escaped
```
ruheni: It would be valuable to have a more concrete example with a name

jharrell: fixed!

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +234,36 @@ const users = await prisma.user.findMany({
...
+})
+```
+
+The above query will match any user whose name starts with a character followed by `benny` such as `7benny` or `&benny`. If you instead wanted to find any user whose name starts with the literal string `_benny`, you could do:
+
+```ts highlight=4
+const users = await prisma.user.findMany({
+  where: {
+    name: {
+      startsWith: '\\_benny', // note that the `_` character is escaped
```
jharrell: For whatever reason my local machine is stripping out `\`. adding `\\` is not working. 😠 
```suggestion
      startsWith: '\_benny', // note that the `_` character is escaped
```

ruheni: Would escaping the second / with \ work? 🤔

jharrell: the `//` isn't the problem, it's `'\_benny'` which some sort of git commit hook transfers to `'_benny'`. 

Committing this via GH UI seems to work though

ruheni: Would another commit (from the same or different PR) reset this? That might be worth considering. I would also recommend internally raising this in `#website`/ `#docs` channels.

janpio: That is probably via https://github.com/prisma/docs/blob/main/.husky/pre-commit Pain in the ass forever, and I have no idea why it is there in a way that messes with your stuff before committing (vs. a linter that complains when not run manually as a CI job or similar)

jharrell: Fixed! Thank you @nilubava

janpio:
Is this really a FAQ? Wouldn't this be more something for the "Filtering" introduction at https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting#filtering, before we have the different examples?

jharrell:
@janpio I think this content is still good as a FAQ but I agree that it's good information to at least be aware of before learning how to filter. I added an admonition where you suggested.

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +240,40 @@ const users = await prisma.user.findMany({
...
+
+#### How does filtering work at the database level?
+
+For MySQL and PostgreSQL, Prisma Client utilizes the [`LIKE`](https://www.w3schools.com/sql/sql_like.asp) (and [`ILIKE`](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE)) operator to search for a given pattern. The operators have built-in pattern matching using symbols unique to `LIKE`. The pattern-matching symbols include `%` for zero or more characters (similar to `*` in other regex implementations) and `_` for one character (similar to `.`)
+
+To match the literal characters, `%` or `_`, make sure you escape those characters. For example:
+
+<!-- prettier-ignore-start -->
+```ts
+const users = await prisma.user.findMany({
```
ruheni, line 253: What is the difference between the 2 queries? They seem to have the same filter 😅

jharrell: oops!

content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
```diff
@@ -234,6 +240,40 @@ const users = await prisma.user.findMany({
...
+  where: {
+    name: {
+      startsWith: '\_benny',
```
jharrell: Thanks for the catch @ruheni 
```suggestion
      startsWith: '_benny',
```

ruheni (approved):

runticle:
Hey is the information added in this PR accurate? I'm using prisma 5.14 and the following prsima query does not escape the `_`.

```
  const results = await prisma.xyz.findMany({
    where: {
      name: {
        startsWith: `a\_`,
      },
    },
  });
```

salahmedamin:
> Hey is the information added in this PR accurate? I'm using prisma 5.14 and the following prsima query does not escape the `_`.
> 
> ```
>   const results = await prisma.xyz.findMany({
>     where: {
>       name: {
>         startsWith: `a\_`,
>       },
>     },
>   });
> ```

I don't know if this is still a thing for you, but the docs say to add "\\" twice to escape it itself !

PR #4837, opened 2023-06-02
Update 100-use-data-proxy.mdx

nikolasburk:


content/800-data-platform/050-data-proxy/100-use-data-proxy.mdx
```diff
@@ -239,7 +239,7 @@ The _default_ value for the `connection_limit` in each Data Proxy instance is `5
 Here is a sample connection string for the Data Proxy connection that uses the `connection_limit` argument:
```
janpio, line 240: Line 235 is also wrong - this is not the Data Proxy connection string, but the direct database connection string.

nikolasburk: good catch, thanks!

PR #4841, opened 2023-06-06
Use globalThis instead of global

andyjy:
Node's `global` object is deprecated - we should use `globalThis` instead, [as instructed in the NodeJS docs](https://nodejs.org/api/globals.html#global). 

This also fixes things to work as intended when testing locally under the Vercel Edge runtime (which doesn't support `global`).


ruheni (approved):

ruheni:
Thanks for catching this 🙏

PR #4860, opened 2023-06-07
fix(env): Remove section about having to use dotenv to use env vars in code

### LINKED ISSUE #19519: importing prisma client pollutes `process.env` without using dotenv
### Bug description

If you put MY_TEST_VAR in a .env file and import a prisma client, then `process.env.MY_TEST_VAR` will be defined, despite the fact that we are not using e.g. dotenv.

### How to reproduce

prisma/schema.prisma:
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL_MAIN")
}

model Locations {
  LocationID                   Int                            @id
  LocationName                 String                         @db.VarChar(45)
}
```

package.json:
```json
{
  "type": "module",
  "scripts": {
    "start": "node app.js"
  },
  "dependencies": {
    "@prisma/client": "^4.14.1",
    "prisma": "^4.14.1"
  }
}
```

.env:
```bash
MY_TEST_VAR="THIS IS THE VALUE OF MY_TEST_VAR"
```

app.js:
```javascript
console.log(process.env.MY_TEST_VAR);
```

`npx prisma generate`
`npm start`
(logs the variable, not undefined)

### Expected behavior

The expected behaviour is that undefined is logged

So what is said [here](https://www.prisma.io/docs/guides/development-environment/environment-variables#using-environment-variables-in-your-code) is not a bit misleading, because the variables will be loaded anyway

"If you want environment variables to be evaluated at runtime, you need to load them manually in your application code (for example, by using [dotenv](https://github.com/motdotla/dotenv))"

### Prisma information

(same as above)
prisma/schema.prisma:
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL_MAIN")
}

model Locations {
  LocationID                   Int                            @id
  LocationName                 String                         @db.VarChar(45)
}
```

### Environment & setup

Mac and linux

### Prisma Version

```
prisma                  : 4.15.0
@prisma/client          : 4.15.0
Current platform        : darwin-arm64
Query Engine (Node-API) : libquery-engine 8fbc245156db7124f997f4cecdd8d1219e360944 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 8fbc245156db7124f997f4cecdd8d1219e360944 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Format Wasm             : @prisma/prisma-fmt-wasm 4.15.0-28.8fbc245156db7124f997f4cecdd8d1219e360944
Default Engines Hash    : 8fbc245156db7124f997f4cecdd8d1219e360944
Studio                  : 0.484.0
```

janpio:
`.env` file is automatically loaded, so no `dotenv` needed.

closes https://github.com/prisma/prisma/issues/19519


millsp (approved):

millsp:
The removal is correct, but I am not sure this closes the mentioned issue? The person is complaining that Prisma Client loads and pollutes their env although it shouldn't (I think we have multiple reports of that?).

janpio:
Yes, as you say we have other issues about that. This user was mostly misguided by the docs that tell them otherwise - so fixing the docs solves their immediate complaint.

PR #4894, opened 2023-06-12
fix: interactive transaction version changes admonition

ruheni:
Add admonition around interactive transaction version changes


janpio (changes requested):
Seems to leak unrelated Prisma Client extensions changes.

Issue #4891, opened 2023-06-12
Formatting of lists within admonitions


meletj:
Lists in admonitions seems to be missing the "indentation": the text should align with the start of the sentence and not with the bullet itself. 
That makes them harder to read. 

Example from: https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#compare-columns-in-the-same-table

![image](https://github.com/prisma/docs/assets/70962362/213e2e5b-b798-4122-ab7e-97cc85817543)

PR #4895, opened 2023-06-12
fix: interactive transaction version changes admonition

ruheni:
Add admonition around interactive transaction version changes


janpio (approved):
👍 

But also another case of https://github.com/prisma/docs/issues/4891

PR #4915, opened 2023-06-14
temp: split client reference into subpages

janpio:


janpio:
Still an interesting idea that we might revisit. Problem are the downsides of not having everything on one page any more that you can Ctrl+F.

nikolasburk:
@janpio this is a PR that I'd simply close. I actually think we should align on whether we _want_ to split the reference into sub-pages (different people might have different opinions on that). Or otherwise leave open and defer?

PR #4916, opened 2023-06-14
feat(frontmatter): Document required vs. optional frontmatter

janpio:
See https://prisma-company.slack.com/archives/C5Z9TH6N9/p1686763363942739

### DIFF
--- a/content/600-about/200-prisma-docs/30-docs-components/03-frontmatter.mdx
+++ b/content/600-about/200-prisma-docs/30-docs-components/03-frontmatter.mdx
@@ -6,7 +6,9 @@ search: false
 
 <TopBlock></TopBlock>
 
-## <inlinecode>title</inlinecode>
+## Required
+
+### <inlinecode>title</inlinecode>
 
 The `<h1>` title of the page. This title appears at the top of the page, and in the left-hand navigation.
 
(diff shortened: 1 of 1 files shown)


content/600-about/200-prisma-docs/30-docs-components/03-frontmatter.mdx
```diff
@@ -43,29 +48,35 @@ Note:
...
+### <inlinecode>preview</inlinecode>
 
 Accepts `true` or `false` (defaults to `false`).
 
 Adds a `preview` label to a page in the left-hand navigation.
 
-## <inlinecode>toc</inlinecode>
+### <inlinecode>toc</inlinecode>
 
 Enable or disable table of contents navigation on the page (defaults to `false`). For example:
```
janpio:
```suggestion
Enable or disable table of contents navigation on the page (defaults to `true`). For example:
```

meletj (commented):
Overall LGTM. I'll let Nilu take a second look too. 
I think we could adjust this to use 'you' instead of 'we' . _If we wish, we can make the `metaTitle` more descriptive than the `title`._

janpio:
Unfortunately seems that the frontmatter page was changed too much so there is now a conflict. This will need to be re-applied to the current state @nilubava, @carlagn, @meletj.

nikolasburk:
Yeah, I tried to go through it in the web editor but wasn't able to properly resolve everything. Let's defer 🙏

janpio:
Reordering is almost impossible to get right here in the GitHub UI. I think someone will need to look at the preview of this, resort the current state of the frontmatter page, and then add the same headlines as here.

PR #4934, opened 2023-06-17
Make it clear that the code is an example of an incorrect schema

brunocrosier:
Make it clear that the code is an example of an incorrect schema

## Changes

Add comment in schema to avoid confusion


nikolasburk:
Thanks for your help @brunocrosier 💚

PR #4940, opened 2023-06-19
fix: update note in admonition

ruheni:


janpio (changes requested):
See https://prisma-company.slack.com/archives/C5Z9TH6N9/p1687203018588259?thread_ts=1687196786.567859&cid=C5Z9TH6N9

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -3227,10 +3227,9 @@ const result = await prisma.user.update({
 ```
 
 ## Filter conditions and operators
-
 <Admonition type="info">
 
-**Note:** The following examples show how to compare columns in separate tables.
+**Note:** The following examples show how to compare columns in the same table.
```
nikolasburk:
```suggestion
```

nikolasburk: Tbh, I find this sentence quite confusing overall and don't see its value. The bullets are useful, but why the note? What should it say? 

If we don't know what it should say, let's just drop the sentence and keep the useful bullets to get this merged?

janpio: I think it is supposed to qualify the bullets and the examples.

The examples all compare a field to a value `name` and `'Eleanor'` for example.
But since 4.3.0 it is also possible to compare fields to fields, so `name` to `lastname` for example.

So the `equals` examples for are really only one specific kind of example, there should possible als obe one that shows comparing the field to another field. But seems that is currently document under another headline, "Compare columns in the same table", that the first bullet links to.

(That headline should be "Compare fields in the same model" at least as the current words are wrong in Prisma context.)

Suggestion:
1. Fix the linked-to headline (and the words in the admonition here)
2. Add example for field to field comparison here as well
3. Then indeed remove the Note as it is not needed any more

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -3227,10 +3227,9 @@ const result = await prisma.user.update({
...
 
 - From version 4.3.0, you can also use these operators to [compare columns in the same table, with the `<model>.fields` property](#compare-columns-in-the-same-table).
```
nikolasburk:
```suggestion
- From version 4.3.0, you can also use these operators to [compare _fields_ in the same model](#compare-columns-in-the-same-table), with the `<model>.fields` property.
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -3227,12 +3227,11 @@ const result = await prisma.user.update({
...
 ## Filter conditions and operators
-
 <Admonition type="info">
 
-**Note:** The following examples show how to compare columns in separate tables.
+**Note:** The following examples show how to compare columns in the same table.
 
-- From version 4.3.0, you can also use these operators to [compare columns in the same table, with the `<model>.fields` property](#compare-columns-in-the-same-table).
+- From version 4.3.0, you can also use these operators to [compare _fields_ in the same model](#compare-columns-in-the-same-table), with the `<model>.fields` property.
 - In versions before 4.3.0, you can [compare columns in the same table with raw queries](/guides/other/troubleshooting-orm/help-articles/comparing-columns-through-raw-queries).
```
nikolasburk:
```suggestion
- In versions before 4.3.0, you can compare fields in the same model [with raw queries](/guides/other/troubleshooting-orm/help-articles/comparing-columns-through-raw-queries).
```

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -3227,13 +3227,12 @@ const result = await prisma.user.update({
...
 ## Filter conditions and operators
-
 <Admonition type="info">
 
-**Note:** The following examples show how to compare columns in separate tables.
+**Note:** The following examples show how to compare columns in the same table.
 
-- From version 4.3.0, you can also use these operators to [compare columns in the same table, with the `<model>.fields` property](#compare-columns-in-the-same-table).
-- In versions before 4.3.0, you can [compare columns in the same table with raw queries](/guides/other/troubleshooting-orm/help-articles/comparing-columns-through-raw-queries).
+- From version 4.3.0, you can also use these operators to [compare _fields_ in the same model](#compare-columns-in-the-same-table), with the `<model>.fields` property.
```
nikolasburk:
```suggestion
- From version 4.3.0, you can also use these operators to compare _fields_ in the same model [with the `<model>.fields` property](#compare-columns-in-the-same-table).
```

janpio (approved):

PR #4926, opened 2023-06-15
docs: reorder content on shared extensions

ruheni:
reorder content on the shared extensions page

Closes #4899

PR #4942, opened 2023-06-20
docs: add top level `$allOperations` method docs

### LINKED ISSUE #4870: Document top-level `$allOperations` method

ruheni:
It will also be documented in the shared extensions docs in this PR: #4926

Closes #4870


content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -148,6 +154,19 @@ const xprisma = prisma.$extends({
 })
 ```
 
+Alternatively, you can also use the `$allOperations` method available directly from the `query` extension component to extend all operations on all models:
```
SevInf: I'd expand this sentence and explain the difference between new method and `$allModels.$allOperations`  that existed before.
- `$allModels.$allOperations` is called for all model operations, i.e. the ones where you have to specify model when calling it (`prisma.user.findMany`). It is not called for raw queries which are not bound to a specific model.
- top-level `$allOperations` is called on both model operations and raw queries. In case it is called on a raw query, `model` property passed to callback will be `null`.

ruheni: I've moved it to be in its own heading/ section

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -30,11 +30,18 @@ Use the `$extends` [client-level method](/reference/api-reference/prisma-client-
...
+  - [Modify a specific operation in a specific model](#modify-a-specific-operation-in-a-specific-model)
+  - [Modify a specific operation in all models of your schema](#modify-a-specific-operation-in-all-models-of-your-schema)
+  - [Modify all operations in a specific model](#modify-all-operations-in-a-specific-model)
+  - [Modify all Prisma Client operations](#modify-all-prisma-client-operations)
+  - [Modify all operations in all models of your schema](#modify-all-operations-in-all-models-of-your-schema)
+  - [Modify a top-level raw query operation](#modify-a-top-level-raw-query-operation)
+  - [Mutate the result of a query](#mutate-the-result-of-a-query)
+- [Wrap a query into a batch transaction](#wrap-a-query-into-a-batch-transaction)
+- [Query extensions versus middlewares](#query-extensions-versus-middlewares)
+  - [If you use the query extension component and middlewares](#if-you-use-the-query-extension-component-and-middlewares)
```
janpio: This looks auto generated and completely wrong. It should only list the ways how you can modify a client.
The wording of the links is also customized to be _not_ the same as the headlines.

janpio (dismissed):
TOC instead of curated list of links, see https://github.com/prisma/docs/pull/4942/files#r1235089030

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -140,7 +166,7 @@ const xprisma = prisma.$extends({
   query: {
     $allModels: {
       $allOperations({ model, operation, args, query }) {
-        // handle all operations
+        // modify all operations on all models
```
janpio: The example does not modify at all, it just calls the original operation

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -129,6 +136,25 @@ const xprisma = prisma.$extends({
...
+### Modify all Prisma Client operations
+
+Use the `$allOperations` method to modify all query methods present in Prisma Client. The `$allOperations` can be used on both model operations and raw queries.
+
+You can modify all methods as follows:
+
+```ts
+const xprisma = prisma.$extends({
+  query: {
+    $allOperations({ model, operation, args, query }) {
```
janpio, line 137: OT: @SevInf @millsp Why the mismatch `query` and `operations` here btw? Do we call the same thing two different names on different levels?

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -129,6 +136,25 @@ const xprisma = prisma.$extends({
...
+      // modify all operations
+      return query(args)
+    },
+  },
+})
+```
+
+In the event a raw query is invoked, the `model` argument passed to the callback will be `null`.
```
janpio: What will be `operation` in this case?

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -129,6 +136,25 @@ const xprisma = prisma.$extends({
 })
 ```
 
+### Modify all Prisma Client operations
+
+Use the `$allOperations` method to modify all query methods present in Prisma Client. The `$allOperations` can be used on both model operations and raw queries.
+
+You can modify all methods as follows:
+
+```ts
```
janpio, line 134: Should we have an example here what `model`, `operations`, `args` and `query` actually contain when you call it? Maybe an "example example" below?

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
ruheni: Thanks for catching this. It was auto-generated indeed by one of my VS Code extensions. I'll resolve it

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
ruheni: Ideally, yes.

I didn't have the time to play with it yet, but I'll do my best to come up with a concrete example.

We could take this a step further, review all examples for Prisma Client extensions docs, and update them to have more concrete example use cases. The current Prisma Client extensions docs read like Reference docs. 😔

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
ruheni: `operation` will be the method that is invoked by Prisma Client, e.g. `$queryRaw`, or `findMany` etc

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
ruheni: I've added a query logging example let me know if that works

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
SevInf: `query` and `operation` are two different things:
- `operation` - name of the operation performed, i.e.  `findFirst`
- `query` is a function for performing the actual query, (`query(args)`)

And they exist in every query callback.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
millsp: `model` would be `undefined`, not `null`.

millsp: `operation` would be the name of the operation called eg `$queryRaw` (string)

millsp: `query` is the callback to execute it

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -116,14 +117,55 @@ const xprisma = prisma.$extends({
...
+  query: {
+    $allOperations({ model, operation, args, query }) {
+      /* your custom logic for modifying all Prisma Client operations here */
+      return query(args)
+    },
+  },
+})
+```
+
+In the event a raw query is invoked, the `model` argument passed to the callback will be `undefined`.
```
jharrell: should we link to our raw query docs?

jharrell (approved):
One small comment but not really a blocker.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
ruheni:
```suggestion
In the event a [raw query](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access) is invoked, the `model` argument passed to the callback will be `undefined`.
```

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
janpio: Isn't the `query` a `findFirst` in this example?

janpio: By the way, I did not mean the `query` parameter (which is a function) - I meant line 147 which defined the "query" extension type. Aren't we extending the operation?

PR #4948, opened 2023-06-21
chore: link to Prisma Migrate API reference from Prisma Migrate (Concepts) page

ruheni:


janpio (approved):
Not well integrated, but does the job.

PR #4978, opened 2023-06-25
MSSQL: use `database` in connection string instead of `initialCatalog`

manuel3108:
## Describe this PR

For Microsoft SQL Server the documentation is inconsistent for using `database` or `initialCatalog`. I gues this was changed recently, but only the first occurrence gut corrected.

Currently for me, using `initialCatalog` does not seem to work. At least for an hosted Azure SQL Database. After many hours of debugging and enabling debug logs, i could see that it was not correctly changing the database. I always got different errros, mostly that my database was [empty](https://www.prisma.io/docs/reference/api-reference/error-reference#p4000).

After a lot of reasearch i pinpointed, that I apparently copied the connection string with the sample data, which currently uses `initialConnection`

## Changes

Use `database` in all provided MSSQL connection strings

## What issue does this fix?

Documentation inconsistency

--


ruheni (approved):
Thank you for catching that @manuel3108. 🙌

Here's a link to SQL Server's docs for more context: https://learn.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver16#:~:text=databaseName%2C%0Adatabase — for anyone else who is curious why

PR #4983, opened 2023-06-28
[docs] adds missing attribute to data object

redpandatronicsuk:
## Describe this PR

The Sequelize example has a `name` attribute, which is missing from the Prisma example

## Changes

It adds the `name` attribute

## What issue does this fix?

That the data written to the database for both Sequelize and Prisma are the same.

### DIFF
--- a/content/200-concepts/300-more/400-comparisons/02-prisma-and-sequelize.mdx
+++ b/content/200-concepts/300-more/400-comparisons/02-prisma-and-sequelize.mdx
@@ -236,6 +236,7 @@ const posts = await Post.findAll({
 ```ts
 const user = await prisma.user.create({
   data: {
+    name: 'Alice',
     email: 'alice@prisma.io',
   },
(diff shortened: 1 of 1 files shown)


ruheni (commented):
👋 

Thank you for your contribution. 

In this particular example, there is no missing field. We use the following data model for most of our examples in the docs:

```prisma
model User {
  id    Int     @id @default(autoincrement())
  email String  @unique
  name  String?
  posts Post[]
}

model Post {
  id        Int      @id @default(autoincrement())
  title     String
  content   String?
  published Boolean  @default(false)

  author    User?    @relation(fields: [authorId], references: [id])
  authorId  Int?
}
```

The `name` field is nullable and can therefore be omitted.

jharrell:
Thank you for your pull request @redpandatronicsuk but as @ruheni mentioned, this is intended. Closing this 🙏

PR #4987, opened 2023-07-03
make `jsonProtocol` GA

### LINKED ISSUE #4854: Prisma Client Docs: make `jsonProtocol` GA
Part of https://github.com/prisma/prisma/issues/19310

Currently, only mentioned in https://github.com/prisma/docs/blob/main/content/200-concepts/100-components/250-preview-features/050-client-preview-features.mdx?plain=1

So removing it would be the only thing to do?

jharrell:
Fixes #4854

## Describe this PR

jsonProtocol is no longer a preview feature, so removing it from this list! There were no pages specific to the feature, so no links have been changed.


janpio (changes requested):
Any preview feature removed here needs to be listed in the table below.

janpio (approved):

janpio:
Additionally:
Technically we _should_ have had a page already talking about what this preview feature does. Then we could turn that into the "history" version now. Probably to late for that now, which means that the docs will not explain to the users of Prisam 4.11.0 to 5.0.0 how to improve their performance.

PR #4990, opened 2023-07-03
fix: TOC Highlighting for top elements

### LINKED ISSUE #4905: TOC highlights first item in the case of large TopBlock paragraphs where headings are not displayed yet
Context: https://prisma-company.slack.com/archives/C5Z9TH6N9/p1686595863282829

carlagn:
## Describe this PR

The right-side TOC for docs does not highlight the correct top section, and after scrolling keeps the first element highlighted even if it's not in view.

To quote the issue report: 

> This page is a good example why our current "table of contents" (right hand sidebar) can be confusing:
https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields
All I see on my screen is under the primary "Working with Json fields" headline.
But the TOC already underlines "Use cases for Json fields".
When I scroll down, I actually have no way to go back to the initial content that introduces the concept of Json fields.

## Changes

- Reset activeId on intersection observer when no heading is visible. (WIP: this can only happen when window is at top!)
- Add "Back to Top" button on the right corner.

## What issue does this fix?

Fixes [#4905](https://github.com/prisma/docs/issues/4905)

N/a


meletj (approved):

nilubava:
@carlagn We have decided not to implement the "Back to top" button now. So is this PR still relevant?

PR #5011, opened 2023-07-06
update: lens pkg

carlagn:
With the current leanest version of lens


nilubava:
@carlagn Please dont delete the local fonts in docs. It is there until we figure out a better way to serve fonts

src/styles/layout.css
```diff
@@ -1,83 +1,3 @@
-/** Inter Regular */
```
nilubava: Please do not remove the fonts here

src/styles/all.css
```diff
@@ -0,0 +1,944 @@
+.fa {
```
nilubava: Are these icons used in docs homepage from the pro package? Then, I think we should not expose these icons in this public repo. We can load these icons from Lens

src/styles/layout.css
carlagn: they come from lens though? this is a repeated import

src/styles/all.css
carlagn: I tried it without the icons from lens and it didn't load them :/

src/styles/layout.css
nilubava: We had to keep the fonts local to avoid the loading issue.

PR #5013, opened 2023-07-06
fix: inline code boldness on headings

### LINKED ISSUE #5014: Font boldness on heading inline code

nilubava:


janpio:
I think in bigger headlines the `code` font-size is now a bit much:

- image
- image
- image

nilubava:
@janpio Can you review again?

janpio:
Now it is the same size as before, but bold - that is definitely better in that it does not add another design weirdness. 

But the font size is probably too small really for a headline. 

Examples:
- image
- image
- image

nilubava:
@janpio If it is the same font, the it would look like the image in your first feedback.
image

janpio:
I know, the code font is "visually bigger" - so if we want to tweak this, we'll probably have to play with the font-size a bit to adapt to all these use cases.

nilubava:
A page where this occurs: https://www.prisma.io/docs/about/prisma-docs/about-the-docs#the-user-and-post-data-model

janpio (approved):
I think the bolding is already good.

Optimally we would also bump up the font size to match the rest of the text in the headlines, but that can be a separate PR.

PR #5035, opened 2023-07-11
chore: add admonition mandatory `cockroachdb` provider

### LINKED ISSUE #5000: Update cockroachdb provider to specific string instead of postgres
Should now read `cockroachdb` instead of `postgresql`

I think this is already the case in our docs, but we should go through and make sure we are consistent.

ruheni:
Closes #5000


content/200-concepts/200-database-connectors/08-cockroachdb.mdx
```diff
@@ -32,6 +32,12 @@ The fields passed to the `datasource` block are:
 - `provider`: Specifies the `cockroachdb` data source connector.
 - `url`: Specifies the [connection URL](#connection-details) for the CockroachDB database server. In this case, an [environment variable is used](/concepts/components/prisma-schema#accessing-environment-variables-from-the-schema) to provide the connection URL.
 
+<Admonition type="info">
+
+While `cockroachdb` and `postgresql` connectors are similar, it is mandatory to use the `cockroachdb` connector instead of `postgresql` from version 5.0.0.
```
Jolg42:
```suggestion
While `cockroachdb` and `postgresql` connectors are similar, it is mandatory to use the `cockroachdb` connector instead of `postgresql` when connecting to a CockroachDB database from version 5.0.0.
```
to be explicit that it's about the database here

Jolg42 (approved):

janpio (approved):

PR #5038, opened 2023-07-11
Update 250-querying-the-database.mdx

ecancino:
## Describe this PR

Updating `Write your first query with Prisma Client` example to use finally

## Changes

Removing `.then` in example and use `.finally`

## What issue does this fix?

It uses the promise API correctly


jharrell (approved):

jharrell:
@ecancino thank you for your contribution!

PR #5047, opened 2023-07-14
Fix: Update 600-upgrading-to-prisma-5.mdx

MiccWan:
Add `await` to make try-catch works

## Describe this PR

An example in the guide demonstrates the usage of `findUniqueOrThrow` along with try-catch without "await" it.
The promise will be reject without triggering the try-catch in this case.

## Changes

Add `await` before the call

## What issue does this fix?

None

None


jharrell (approved):

jharrell:
@MiccWan thank you for your contribution!

PR #5048, opened 2023-07-14
Update 050-prisma-client-reference.mdx

rgmvisser:
## Describe this PR

I ran into this where the upsert query will be a SELECT + INSERT if the update field is an empty object. This is a bit unintuitive, maybe even a bug?

## Changes

Add another condition to when an upsert is used

## What issue does this fix?

Unexpected SELECT + INSERT instead of a UPDATE ON CONFLICT

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -1036,6 +1036,7 @@ Prisma Client uses a database upsert for an `upsert` query when the query meets 
 - The query modifies only one model
 - There is only one unique field in the `upsert`'s `where` option
 - The unique field in the `where` option and the unique field in the `create` option have the same value
+- The update field is not an empty object
 
 If your query does not meet these criteria, then Prisma Client handles the upsert itself.
 


janpio:
Do I understand correctly that in your opinion an empty object should be treated like no object, where it then would use the native upsert?

rgmvisser:
It's more that if you have something like:
```
prisma.object.upsert({
      where: {
          hash
      },
      create: {
        hash,
        tenantId: tenant.id,
      },
      update: {},
    });
````

It will produce a select and than an insert. However, when dealing with potential race conditions (calling this twice at the same time for instance), the order might be: SELECT 1-> no result, SELECT  2-> no result,  INSERT 1 -> new row, INSERT 2 -> Error of duplicated unique field "hash"

If it would be a native UPDATE ON CONFLICT, this would be handled in the case of a race condition by the DB. 

At the moment I solve this by calling the function like this:
```
prisma.object.upsert({
      where: {
          hash
      },
      create: {
        hash,
        tenantId: tenant.id,
      },
      update: {
            hash
      },
    });
````

This forces Prisma to do a native upsert (even though the hash never changes/updates). 

Let me know if this explanation makes more sense?

janpio:
Would you expect Prisam to make the assumption that it is valid to use the unchanged `hash` as the `update` value if case tehre is no explicit, non-empty `update` object provided?

rgmvisser:
> Would you expect Prisam to make the assumption that it is valid to use the unchanged hash as the update value if case tehre is no explicit, non-empty update object provided?

No definitely not as that also might not be expected. Probably more something like: `INSERT ... ON CONFLICT DO NOTHING` (Only DO NOTHING if the update field is an empty object)

janpio:
Ah ok, that makes more sense and is a valid suggestion. That sounded a tiny bit too scary.

Can you open a feature request in prisma/prisma repository? I think I would prefer to treat this as a missing feature or even bug instead of documenting this as intentional.

rgmvisser:
Sure, submitted here: https://github.com/prisma/prisma/issues/20229

janpio:
Thanks! Shared with the team that is working on that part of Prisma Client.

Should we close this PR here then, or do you think we need to merge it in the meantime?

rgmvisser:
You're the PM, so up to you 😃

jharrell:
I'll make a call for @janpio here and say that we'll close this in favor of work coming out of https://github.com/prisma/prisma/issues/20229

Thank you for your report and explanation @rgmvisser !

PR #5063, opened 2023-07-17
Document how to escape SQL Server connection string parts

### LINKED ISSUE #2750: Document how to escape SQL Server connection string parts
See https://github.com/prisma/prisma/issues/11276#issuecomment-1017989820

jharrell:
Fixes #2750

Documents how to escape special characters in SQL Server connection strings.


janpio:
Do we document the need to escape for other database in a similar location? If not, that would be a good follow up PR - we have similar questions for MySQL and PostgreSQL. 

(I think the original intent of the issue was to _add_ it to where we already talk about escaping credentials, which only lists some other databases but not SQL Server)

content/200-concepts/200-database-connectors/08-sql-server/index.mdx
```diff
@@ -37,6 +37,22 @@ The following example uses SQL authentication (username and password) with an en
...
+
+```terminal
+:\=;/[]{}  # these are characters that will need to be escaped
+```
+
+To escape these characters, use curly braces `{}` around each special character. As an example:
+
+```terminal
+sqlserver://HOST:PORT;database=DATABASE;user=MyServer{/}MyUser;password=ThisIsA{:}SecurePassword{;};encrypt=true
+```
```
ruheni: 🤔

Would adding curly braces around a property's value also work when escaping special characters? Would that be a more convenient way of escaping special characters? e.g. 

```
sqlserver://HOST:PORT;database=DATABASE;user={MyServer/};MyUser;password={ThisIsA:SecurePassword;};encrypt=true
```

jharrell:
> Do we document the need to escape for other database in a similar location? If not, that would be a good follow up PR - we have similar questions for MySQL and PostgreSQL.
> 
> (I think the original intent of the issue was to _add_ it to where we already talk about escaping credentials, which only lists some other databases but not SQL Server)

@janpio we do have info on escaping characters in other DBs. I've added a sentence for SQL Server in those locations.

content/200-concepts/200-database-connectors/08-sql-server/index.mdx
jharrell: referencing this documentation: https://learn.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url?view=sql-server-ver16#escaping-values-in-the-connection-url

It looks like this is possible! The example would look something like this 
```
sqlserver://HOST:PORT;database=DATABASE;user={MyServer/MyUser};password={ThisIsA:SecurePassword;};encrypt=true
```

janpio:
Should we also add specific inforamtion to each of the other database subpages equal to the SQL Server one that was added?

jharrell:
@janpio created a follow-up task #5067 and assigned myself

PR #5068, opened 2023-07-18
Mention `_relevance` in fulltext search docs

### LINKED ISSUE #5045: Mention `_relevance` in fulltext search docs
## Problem

Information about ranking search results is not surfaced in the docs for fulltext search here https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search

Instead it is fragmented and only documented here, not very discoverable.
(https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#sort-post-by-relevance-of-the-title)

## Suggested solution
The fulltext search docs should at least give a high level overview of all features related to fulltext search, even if it just links to other parts of the docs that better document each of the features. It is easy to falsely conclude that Prisma is not capable of basic search ranking (which makes fulltext search support in Prisma appear pretty useless to be frank, when in fact it is more useful and robust that the docs would seem to suggest)

jharrell:
Fixes #5045

## Describe this PR

Added a section to full-text search explaining `_relevance` and linking to existing `orderBy` docs. 

Also, the docs referenced PostgreSQL only... but I was able to generate a MySQL client with `_relevance` so maybe we fixed that at some point.


content/200-concepts/100-components/02-prisma-client/060-full-text-search.mdx
```diff
@@ -166,6 +166,27 @@ MySQL also has `>`, `<` and `~` operators for altering the ranking order of sear
...
 
+## Sorting results by <inlinecode>\_relevance</inlinecode>
+
+<Admonition type="warning">
+
+Sorting by relevance is only available for PostgreSQL and MySQL.
+
+</Admonition>
+
+In addition to [Prisma's default orderBy behavior](/content/reference/api-reference/prisma-client-reference#orderBy), full-text search also adds sorting by relevance to a given string or strings. As an example, if you wanted to order posts by their relevance to the term `'database'` in their title, you could use the following:
```
janpio:
```suggestion
In addition to [Prisma's default `orderBy` behavior](/content/reference/api-reference/prisma-client-reference#orderBy), full-text search also adds sorting by relevance to a given string or strings. As an example, if you wanted to order posts by their relevance to the term `'database'` in their title, you could use the following:
```

janpio (approved):

PR #5094, opened 2023-07-24
fix(supported-databases): PlanetScale is fully supported

janpio:


content/400-reference/300-database-reference/03-supported-databases.mdx
```diff
@@ -21,6 +21,7 @@ Prisma currently supports the following databases:
 | MySQL                        | 5.7     |
 | MySQL                        | 8       |
 | MariaDB                      | 10      |
+| PlanetScale                  | \*      |
```
ruheni, line 24: Isn't PlanetScale a provider rather than a database? 🤔

janpio: Yes, but it was listed on this page here before. I just fixed it being marked as preview.

Users will potentially look for Planetscale here indeed, especially as it does not really communicate the MySQL version and so on - so this is a "incorrectness" I am ok with.

Jolg42 (approved):

content/400-reference/300-database-reference/03-supported-databases.mdx
Jolg42: I'm ok with this too, it's the best place to put it for now and if we have another idea we can always change this later

PR #5104, opened 2023-07-25
fix(one-to-many-relations): Forward relation first, back relation second

janpio:
Before started with the back relation, which is weird in my head.


content/200-concepts/100-components/01-prisma-schema/06-relations/200-one-to-many-relations.mdx
```diff
@@ -205,42 +205,42 @@ In MongoDB, the only difference between a 1-1 and a 1-n is the number of documen
...
-  posts Post[]
-}
-
 model Post {
   id       Int   @id @default(autoincrement())
   author   User? @relation(fields: [authorId], references: [id])
   authorId Int?
 }
+
+model User {
```
jharrell: I'm not sure if I agree with this shuffling. We're describing 1-n relationships, but if you read this top to bottom this is n-1. I can understand why the description above was changed but IMO these should stay as they were. Open to being wrong.

janpio: Fair.

I assume July Jan's focus was to put the model with `@relation` first (forward relation), and then the back relation.

nikolasburk: I don't have a strong opinion here but I don't think changing it makes it clearer necessarily. In (almost) all other code snippets I can think of we write `User` first and then `Post`, so I would keep the order here for consistency.

nikolasburk:
I don't have a strong opinion on this one (because it just makes some slight cosmetic changes). @janpio how do you want to proceed here? Personally, I'd just close the PR but if you still think this should be clarified maybe update the PR to keep the order in the code snippets with `User` first and then `Post` and then merge it?!

janpio:
The core of the changes in this PR was the changed description above.
I undid the change in the schemas.

PR #5106, opened 2023-07-27
Adds note about the inability to use filter conditions to filter a JS…

### LINKED ISSUE #5098: Improve docs for `findUnique`, mention that filters, like `equals` cannot be used
Docs page: https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#findunique

Related to https://github.com/prisma/prisma/discussions/20367

We have proper type/runtime validation errors when the field in not Json at the moment, so it's obvious that using filters doesn't work.
But for a Json column all queries will silently return `null`

Example that will not behave as expected with a Json column
```ts
await prisma.address.findUnique({where: {jsonColumn: {equals: {hello: "world"}}})
```
This will try to match a database entry where the value is `{equals: {hello: "world"}}`

Example that will behave 
```ts
await prisma.address.findUnique({where: {jsonColumn: {hello: "world"}}})
```
This will try to match a database entry where the value is `{hello: "world"}`

sabinadams:
Adds note to the client API reference for `findUnique` letting readers know the cannot use filter conditions when filtering a JSON field.

Closes #5098

### DIFF
--- a/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
+++ b/content/400-reference/200-api-reference/050-prisma-client-reference.mdx
@@ -461,6 +461,7 @@ Use model queries to perform CRUD operations on your models. See also: [CRUD](/c
 
 - Prisma's dataloader [automatically batches `findUnique` queries](/guides/performance-and-optimization/query-optimization-performance#solving-n1-in-graphql-with-findunique-and-prismas-dataloader) with the same `select` and `where` parameters.
 - If you want the query to throw an error if the record is not found, then consider using [`findUniqueOrThrow`](#finduniqueorthrow) instead.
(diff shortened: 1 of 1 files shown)


jharrell (approved):

janpio:
Isn't that true for all Prisma types, but for most it is caught by the types and runtime, but only for Json we can not do that? Currently it sounds this is like a Json only limitation. But Json is just the only one where we can not differentiate and catch it.

(cc @Jolg42)

sabinadams:
@janpio Not sure I understand. Are you saying filter conditions just don't work at all?

The issue referenced in this PR is that if you do something like:

```ts
await prisma.address.findUnique({where: {jsonColumn: {equals: {hello: "world"}}})
```

It searches the DB for a record that literally equals:

```json
{ 
   "equals": {
      "hello": "world"
   }
}
```

janpio:
> Are you saying filter conditions just don't work at all?

That is how I understood the following sentence from the issue:
> We have proper type/runtime validation errors when the field in not Json at the moment, so it's obvious that using filters doesn't work.

But of course that also does not make too much sense.

PR #5107, opened 2023-07-27
docs: Add Remix to list of web frameworks

kentcdodds:
## Changes

In the REST documentation, there are links to web frameworks. I added Remix to the list.


nikolasburk:
Thanks Kent 💚  I took the liberty to put Remix at the end of the list and just merged another PR with that change: https://github.com/prisma/docs/pull/5460

(Just because we think that the other frameworks are the more "typical" ones that people would use to build their REST APIs where Remix is probably more often used for [fullstack](https://www.prisma.io/docs/concepts/overview/prisma-in-your-stack/fullstack) apps).

Thanks for taking the time to help us improve our docs 🙏

PR #5122, opened 2023-08-02
fix(prisma-client-transactions-guide): Make $transaction([]) explicit

janpio:
Right now the document refers to the batch transaction API only as "`$transaction` API". But interactive transactions of course also use `$transaction()`, just with a different parameter (functions instead of array).

Calling it the "`$transaction([])` API" at least lets us differentiate.

This needs to be cleaned up on a higher level though as well: https://github.com/prisma/docs/issues/5123


content/300-guides/100-performance-and-optimization/100-prisma-client-transactions-guide.mdx
```diff
@@ -79,7 +79,7 @@ Consider using nested writes if:
 
 <Tip>
 
-If you [pre-compute your IDs, you can choose between a nested write or using the `$transaction` API](#scenario-pre-computed-ids-and-the-transaction-api).
+If you [pre-compute your IDs, you can choose between a nested write or using the `$transaction([])` API](#scenario-pre-computed-ids-and-the-transaction-api).
```
ruheni, line 82: Q: Would referring to `$transaction([])` as sequential transaction operations make sense? If so, I would be happy to take care of the updates 🤔

Example usage: 
```suggestion
If you [pre-compute your IDs, you can choose between a nested write or using sequential transaction operations](#scenario-pre-computed-ids-and-the-transaction-api).
```

janpio: A further rewrite of these two pages is definitely justified, potentially including your suggestion. I created https://github.com/prisma/docs/issues/5123 for the follow up work here.

ruheni: I'll add this (rewriting the transactions content page) to the agenda for the Docs grooming session later this week with Jon & Niko.

janpio: Could we still approve and merge this in the meantime?

ruheni (approved):

content/300-guides/100-performance-and-optimization/100-prisma-client-transactions-guide.mdx
ruheni: I've created a new GH issue for further updates to our transaction guides.

PR #5126, opened 2023-08-02
chore: update `every` & `some` examples in the API reference

### LINKED ISSUE #5116: No `every` example (shows `some` instead, plus typo in wrong example)
Our every example is actually for some and has a typo: 
https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#get-all-user-records-where-all-posts-are-published-and-a-least-one-related-post-mentions-prisma

ruheni:
Closes #5116


jharrell (approved):
I think it would be nice to combine these in the future to show off what is possible, but not in our client reference 🙂

janpio (approved):

PR #5164, opened 2023-08-15
fix(admonition-component): Prevent overlapping text

### LINKED ISSUE #5163: overlapping text when code blocks are in admonitions
Visible here: https://www.prisma.io/docs/concepts/database-connectors/sql-server#connection-details

image

ThomasCharuel:
## Describe this PR

Elements within the Admonition component may overlap with adjacent elements.
Here is an example:
https://www.prisma.io/docs/concepts/database-connectors/sql-server#connection-details

image

Switching the component's `display: flex` to `display: block` resolves this bug without altering the component's intended behavior:

image

## Changes

- Change display property from `flex` to `block`.
- Remove duplicate style property `padding-bottom: 8px`

## What issue does this fix?

Fixes #5163

### DIFF
--- a/src/components/shortcodes/admonition.tsx
+++ b/src/components/shortcodes/admonition.tsx
@@ -60,7 +60,7 @@ const AdmonitionWrapper = styled.span<{ type?: string }>`
   padding-bottom: 8px;
   margin: 2rem 0px;
   position: relative;
-  display: flex;
+  display: block;
   max-width: 100%;
   pre {
     font-weight: normal;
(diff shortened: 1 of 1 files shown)


ThomasCharuel:
Hello, is it possible to re-run the checks?
I don't see how their failures are related to the changes in this PR. If they are related please tell me what should be done to make it work.
Thanks :)

janpio:
Nope, you can fully ignore those - we have no good way to run those check for external contributions. Someone will just have to look at your code directly.

nilubava:
Thanks for catching this error and the PR @ThomasCharuel 👍

nilubava (approved):

PR #5169, opened 2023-08-15
Small improvements to Netlify guide

### LINKED ISSUE #5165: Small improvements to Netlify guide
### Explain binary target

I really like this section! The only info that I think we should add is what does `"rhel-openssl-1.0.x"` mean exactly?  A short inline explanation of what this value is, what it means and _why_ we need to put it there would be helpful I think.

image

### Fix terminal output

The code block is broken, the `DATABASE_URL ` and `postgresql://postgres:__PASSWORD__@__HOST__:__PORT__/__DB_NAME__ |` should probably be next to each other but instead are rendered with a line break.

image

jharrell:
Fixes #5165


content/300-guides/200-deployment/201-serverless/500-deploy-to-netlify.mdx
```diff
@@ -18,14 +18,14 @@ Before you can follow this guide, you will need to set up your application to be
 
 ## Binary targets in <inlinecode>schema.prisma</inlinecode>
 
-The Prisma schema should contain the following in the `generator` block:
+Since your code is being bundled and deployed to Netlify's environment, which isn't necessarily the same as your local environment, you will need to set [`binaryTargets`](/reference/api-reference/prisma-schema-reference#binarytargets-options) in order to bundle the query engine that is compatible with Netlify. If you do not set this option, your deployed code will have an incorrect query engine deployed with it and will not function.
```
Jolg42:
```suggestion
Since your code is being deployed to Netlify's environment, which isn't necessarily the same as your local environment, you will need to set [`binaryTargets`](/reference/api-reference/prisma-schema-reference#binarytargets-options) in order to download the query engine that is compatible with Netlify runtime during the build step. If you do not set this option, your deployed code will have an incorrect query engine deployed with it and will not function.
```

Jolg42: I think we can avoid saying `bundle` here, the engine is not bundled actually, but installed in their build runner.

jharrell: Makes total sense. Thank you @Jolg42 !

Jolg42 (approved):

content/300-guides/200-deployment/201-serverless/500-deploy-to-netlify.mdx
```diff
@@ -18,14 +18,14 @@ Before you can follow this guide, you will need to set up your application to be
 
 ## Binary targets in <inlinecode>schema.prisma</inlinecode>
 
-The Prisma schema should contain the following in the `generator` block:
+Since your code is being deployed to Netlify's environment, which isn't necessarily the same as your development environment, you will need to set [`binaryTargets`](/reference/api-reference/prisma-schema-reference#binarytargets-options) in order to download the query engine that is compatible with the Netlify runtime during your build step. If you do not set this option, your deployed code will have an incorrect query engine deployed with it and will not function.
```
Jolg42, line 21: I just realized again the difference for Netlify, is If I remember correctly that the build env is different than the deployment env actually 🙈

Jolg42: This sentence is still correct, I think

PR #5176, opened 2023-08-16
docs: improve `$allModels` documentation

### LINKED ISSUE #4840: Documentation about $allModels on model extensions is not clear
The documentation on model Prisma extensions (https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model) is not very clear regarding how the `$allModels` resource should be used.

It's clear that it adds the same methods to all models in your Prisma scheme, but the documentation should have better examples on how these methods should be implemented, because there are issues that aren't addressed:

* https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model#add-a-custom-method-to-all-models-in-your-schema : inside the sample `validate(data: unknown)` method, how can we know which model is the method being called on? Getting the model's name from the context (https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model#get-the-current-model-name-at-runtime) is not a valid solution, because the name returned is a `string` and doing something like `prisma[name].findMany(...)` will raise TypeScript errors.

* https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model#advanced-type-safety-improve-the-type-safety-and-developer-experience-of-your-custom-model-methods : why are there so many Prisma tools for type safety (`Exact`, `Args`, etc.) if inside the function we need to override type safety (`return {} as any`)? One can imagine the reasoning behind this, but it's not clear how the logic of a custom method should be implemented. A function that only contains `return {} as any` is far from a good example.

* Why does the `validate` function in https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model#example-1 has only one argument while the `customCall` has two? Why the first argument in `customCall` appears to be a reference to the model it's being called on, and why does `validate` apparently doesn't receive this information? Again, having real code samples instead of just `/* Generic validation logic */` or `return {} as any` would be very helpful.

* https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model#advanced-type-safety-improve-the-type-safety-and-developer-experience-of-your-custom-model-methods : the `customCall` example doesn't explain what the `x: Prisma.Exact<...` argument is, nor how it should be used. It also fails to show that trying to access `x.customProperty` _doesn't_ work and causes a TS error: `Property 'customProperty' does not exist on type 'string | number | bigint | boolean | [] | { [x: string]: string extends keyof A ? Exact<A[keyof A & string], any> : never; }'.`. A working code sample that shows how this `x` parameter should be used could also be very helpful.

I'd provide working samples for all the situations above, but I don't believe the workarounds I did to use `$allModels` on my project are the most appropriate ones.

### LINKED ISSUE #5167: Client extensions: Use `prisma` as variable name and instantiate example code snippets using `new PrismaClient()`
https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions

For example, this code snippet:

```ts
const xprisma = prisma.$extends({
  name: `signUp`, // Optional: name appears in error logs
  model: {        // This is a `model` component
    user: { ... } // The extension logic for the `user` model goes inside the curly braces
  },
})
```

should be rewritten to:

```ts
const prisma = new PrismaClient().$extends({
  name: `signUp`, // Optional: name appears in error logs
  model: {        // This is a `model` component
    user: { ... } // The extension logic for the `user` model goes inside the curly braces
  },
})
```

or this one: 

```ts
// First of all, store your original Prisma Client in a variable as usual

// Declare an extended client that has an extensionA and extensionB
const prismaAB = prisma.$extends(extensionA).$extends(extensionB)
```

to:

```ts
// Declare an extended client that has an extensionA and extensionB
const prisma = new PrismaClient
  .$extends(extensionA)
  .$extends(extensionB)
```

ruheni:
closes #4840
closes #5167

This PR adds more concrete examples for the `$allModels` model client extension to help our users build a better mental model of how or when to use them.

This PR also cleans up the usage of `xprisma`.

Provide more concrete examples based on user feedback linked to the issue.


content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
```diff
@@ -23,48 +23,39 @@ The following type utilities are available in Prisma Client:
...
+      ): Promise<boolean> {
+        // Retrieve the current model at runtime
+        const context = Prisma.getExtensionContext(this)
+
+        // Prisma Client query that retrieves data based
+        const result = await (context as any).findFirst({ where })
+        return result !== null
       },
     },
   },
```
ruheni, line 49: I'm curious whether we should keep this example because it provides extensive usage on how to use TS to build an extension.

If anyone has a better example than the one I've added, please share it, and I'll include it in the docs.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
```diff
@@ -68,22 +68,22 @@ const xprisma = prisma.$extends({
...
+const user = await prisma.user.findFirst()
 
 // return the user's full name, such as "John Doe"
 console.log(user.fullName)
 ```
 
 In above example, the input `user` of `compute` is automatically typed according to the object defined in `needs`. `firstName` and `lastName` are of type `string`, because they are specified in `needs`. If they are not specified in `needs`, then they cannot be accessed.
 
-When you call a method in an extension, use the constant name from your `$extends` statement, not `prisma`. In the above example, `xprisma.user.findFirst` computes the virtual field, but `prisma.user.findFirst` would not.
+When you call a method in an extension, use the constant name from your `$extends` statement, not `prisma`. In the above example, `prisma.user.findFirst` computes the virtual field, but `prisma.user.findFirst` would not.
```
jharrell: This comment doesn't make sense anymore since we've renamed `xprisma` to `prisma`

content/200-concepts/100-components/02-prisma-client/053-client-extensions/140-shared-extensions.mdx
```diff
@@ -155,10 +155,10 @@ To import the `find-or-create` extension from the example above, and wrap your c
...
-const user = await xprisma.user.findOrCreate({ ... })
+const prisma = new PrismaClient().$extends(findOrCreate)
+const user = await prisma.user.findOrCreate({ ... })
 
-xprisma.user.findOrCreate()
+await prisma.user.findOrCreate()
 ```
 
-When you call a method in an extension, use the constant name from your `$extends` statement, not `prisma`. In the above example,`xprisma.user.findOrCreate` works, but `prisma.user.findOrCreate` does not, because the original `prisma` is not modified.
+When you call a method in an extension, use the constant name from your `$extends` statement, not `prisma`. In the above example,`prisma.user.findOrCreate` works, but `prisma.user.findOrCreate` does not, because the original `prisma` is not modified.
```
jharrell: ditto above comment

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
jharrell: I'm not sure I follow your concern. I think in-depth examples can be very good and show how our utilities work in a "real world" example.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
janpio: If this was the only place where this was highlighted, we now might need to surface this in another place as i is a common pattern how to use extensions.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/100-model.mdx
```diff
@@ -138,13 +147,18 @@ You can get the name of the current model at runtime with `Prisma.getExtensionCo
...
+// `context` refers to the current model
+const context = Prisma.getExtensionContext(this)
+
+// `context.name` returns the name of the current model
+console.log(context.name)
 
-// `ctx.name` returns the name of the current model
-console.log(ctx.name)
+// Usage
+await(context as any).findFirst({ args })
```
janpio, line 157: this looks wrong to me

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
```diff
@@ -23,48 +23,39 @@ The following type utilities are available in Prisma Client:
...
+      // Define a new `exists` operation on all models
+      // T is a generic type that corresponds to the current model
+      async exists<T>(
+        // `this` refers to the current type, e.g. `prisma.user` at runtime
         this: T,
-        x: Prisma.Exact<
-          A,
-          // For `customCall`, use the arguments from model `T` and the
-          // operation `findFirst`. Add `customProperty` to the operation.
-          Prisma.Args<T, 'findFirst'> & { customProperty: boolean }
```
janpio, line 47: How to add a custom property does not seem to be part of the next example.

janpio (approved):

millsp:
@ruheni would this comment be worth tackling via this PR as well?
https://prisma-company.slack.com/archives/C043X6JBND6/p1691704235991279

content/200-concepts/100-components/02-prisma-client/053-client-extensions/100-model.mdx
ruheni: Are you referring to the type-casting or whether the example snippet would work?

If it's related to type-casting, Serhii created an issue a few days ago: https://github.com/prisma/prisma/issues/20786

The snippet is functional and we use it in the [`exists-fn`](https://github.com/prisma/prisma-client-extensions/blob/4b7fe4549121a8d1eb73499230f84dd1b91733ba/exists-fn/script.ts#L11) example extension.

janpio: Just looked wrong on a first skim, so 🤷 If it works, resolve and ignore.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -35,7 +35,9 @@ Use the `query` extension component to modify queries. You can modify a custom q
 To create a custom query, use the following structure:
 
 ```ts
-const xprisma = prisma.$extends({
+const prisma = new PrismaClient()
+
+const extendedPrisma = prisma.$extends({
```
jharrell: Do we want to do this? After a brief glance I can't tell why we need both `prisma` and `extendedPrisma`
```suggestion
const prisma = new PrismaClient().$extends({
```

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -53,7 +55,9 @@ The properties are as follows:
 The `query` object can contain functions that map to the names of the [Prisma Client operations](/reference/api-reference/prisma-client-reference#model-queries), such as `findUnique`, `findFirst`, `findMany`, `count`, and `create`. The following example modifies `user.findMany` to a use a customized query that finds only users who are older than 18 years:
 
 ```ts
-const xprisma = prisma.$extends({
+const prisma = new PrismaClient()
+
+const extendedPrisma = prisma.$extends({
```
jharrell: ditto above

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -66,10 +70,10 @@ const xprisma = prisma.$extends({
   },
 })
 
-await xprisma.user.findMany() // returns users whose age is greater than 18
+await extendedPrisma.user.findMany() // returns users whose age is greater than 18
```
jharrell: ditto
```suggestion
await prisma.user.findMany() // returns users whose age is greater than 18
```

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
```diff
@@ -66,10 +70,10 @@ const xprisma = prisma.$extends({
...
 ```
 
-In the above example, a call to `xprisma.user.findMany` triggers `query.user.findMany`. Each callback receives a type-safe `{ model, operation, args, query }` object that describes the query. This object has the following properties:
+In the above example, a call to `extendedPrisma.user.findMany` triggers `query.user.findMany`. Each callback receives a type-safe `{ model, operation, args, query }` object that describes the query. This object has the following properties:
```
jharrell: etc

jharrell (approved):
LGTM. Comment about `extendedPrisma` in one file but just looking for clarification really.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/120-query.mdx
ruheni: In this example, the `query` extensions modify the methods available in Prisma Client. This might lead to undesired behavior if applied to the primary Prisma Client instance and one forgets about it. I'm not sure if we mention this anywhere but I think we generally recommend doing that on a separate Prisma Client instance

jharrell: Cool! That makes sense. Let's just make an admonition describing that.

ruheni: I've made the update. Let me know if the wording could be better 😄

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
```diff
@@ -26,7 +26,7 @@ Use the `result` extension component to add custom fields and methods to query r
 To add a custom [field](#add-a-custom-field-to-query-results) or [method](#add-a-custom-method-to-the-result-object) to query results, use the following structure. In this example, we add the custom field `myComputedField` to the result of a `user` model query.
```
jharrell, line 26: @ruheni I think we should add a similar admonition to this file

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
```diff
@@ -68,22 +70,22 @@ const xprisma = prisma.$extends({
   },
 })
 
-const user = await xprisma.user.findFirst()
+const user = await prisma.user.findFirst()
```
jharrell, line 71: If we want to use `extendedPrisma` in this file as well, then this should be changed.
```suggestion
const user = await extendedPrisma.user.findFirst()
```

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
```diff
@@ -68,22 +70,22 @@ const xprisma = prisma.$extends({
...
-When you call a method in an extension, use the constant name from your `$extends` statement, not `prisma`. In the above example, `xprisma.user.findFirst` computes the virtual field, but `prisma.user.findFirst` would not.
+When you call a method in an extension, use the constant name from your `$extends` statement, not `prisma`. In the above example, `extendedPrisma.user.findFirst` computes the virtual field, but `prisma.user.findFirst` would not.
 
 ### Re-use a computed field in another computed field
 
 The following example computes a user's title and full name in a type-safe way. `titleFullName` is a computed field that reuses the `fullName` computed field.
 
 ```ts
-const xprisma = prisma
+const prisma = new PrismaClient()
```
jharrell, line 84: Are we using `extendedPrisma` in this file? If so:
```suggestion
const prisma = new PrismaClient()
const extendedPrisma = prisma
```

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
```diff
@@ -140,9 +144,9 @@ const xprisma = prisma.$extends({
   },
 })
 
-const user = await xprisma.user.findUniqueOrThrow({ where: { id: someId } })
+const user = await prisma.user.findUniqueOrThrow({ where: { id: someId } })
```
jharrell, line 143:
```suggestion
const user = await extendedPrisma.user.findUniqueOrThrow({ where: { id: someId } })
```

content/200-concepts/100-components/02-prisma-client/053-client-extensions/140-shared-extensions.mdx
```diff
@@ -155,10 +155,11 @@ To import the `find-or-create` extension from the example above, and wrap your c
...
 import findOrCreate from 'prisma-extension-find-or-create'
 
-const xprisma = prisma.$extends(findOrCreate)
-const user = await xprisma.user.findOrCreate({ ... })
+const prisma = new PrismaClient()
+const extendedPrisma = prisma.$extends(findOrCreate)
+const user = await prisma.user.findOrCreate({ ... })
 
-xprisma.user.findOrCreate()
+await prisma.user.findOrCreate()
```
jharrell:
```suggestion
await extendedPrisma.user.findOrCreate()
```

jharrell (commented):
Found some instances where `prisma` vs `extendedPrisma` is inconsistent.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/130-result.mdx
ruheni: We could do without the admonition here because the extension modifies the results of a query and but not the query sent to the database 😄

ruheni: Got some feedback from Serhii ([Internal conversation](https://prisma-company.slack.com/archives/C043X6JBND6/p1693828548570949?thread_ts=1693825660.132309&cid=C043X6JBND6))

>My opinion, pretty much for all kinds of extensions:
you need to store primary instance in a separate variable if you want to use it later (without the extension) or you want to use independent instances of extensions with different settings within your app (for example, RLS extension would need session information and can not be applied globally)
you are fine to use inline version if you won’t need primary instance for anything else: most result, client and model extensions would fall into this category but there could be exceptions.

TODO: 

- [x] Update the `query` extension docs, removing instances of `extendedPrisma` in favor of extending the primary instance.

- [x] I'll create a separate issue for us to document the difference between the two — primary vs "variant" (https://github.com/prisma/docs/issues/5270)

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
```diff
@@ -23,50 +23,81 @@ The type utilities available are:
...
+        this: T,
+
+        // The `exists` function will use the `where` arguments from the current model, `T`, and the `findFirst` operation
+        where: Prisma.Args<T, 'findFirst'>['where']
+      ): Promise<boolean> {
+        // Retrieve the current model at runtime
+        const context = Prisma.getExtensionContext(this)
 
-const xprisma = prisma.$extends({
+        // Prisma Client query that retrieves data based
```
millsp, line 44: Should we place these same comments in the other code example?

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
```diff
@@ -23,50 +23,81 @@ The type utilities available are:
...
+          // and intersect it with `CacheStrategy` as part of `findMany` arguments
+          Prisma.Args<T, 'findMany'> & CacheStrategy
         >
-
-        // Get the correct result types for the model of type `T`,
-        // and the arguments of type `A` for `findFirst`.
-        // `Prisma.Result` computes the result for a given operation
-        // such as `select {id: true}` in function `main` below.
-        //
-      ): Prisma.Result<T, A, 'findFirst'> {
```
millsp, line 55: Does `Prisma.Result` have a spot where it is showcased?

ruheni: I've added it to the caching example.

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
ruheni: I've removed the duplicate comments from the second example. :)

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
```diff
@@ -23,50 +23,79 @@ The type utilities available are:
...
+  await prisma.post.findMany({
+    cacheStrategy: {
+      ttl: 360,
+      swr: 60,
     },
   })
 }
 ```
+
+The example here is only conceptual. For the actual caching to work, you will have to implement the logic.
```
Jolg42: add a link to Accelerate?
to suggest that they can use accelerate if they want a caching solution

jharrell (approved):

content/200-concepts/100-components/02-prisma-client/053-client-extensions/150-type-utilities.mdx
```diff
@@ -23,50 +23,79 @@ The type utilities available are:
...
+  await prisma.post.findMany({
+    cacheStrategy: {
+      ttl: 360,
+      swr: 60,
     },
   })
 }
 ```
+
+The example here is only conceptual. For the actual caching to work, you will have to implement the logic. If you're interested in a caching extension/ service, we recommend taking a look at [Prisma Accelerate](https://www.prisma.io/data-platform/accelerate).
```
janpio, line 101: additional space after `extension/`

PR #5192, opened 2023-08-18
fix(accelerate): add a mention for edge runtimes

millsp:
Quickly mention that Accelerate can be run in Cloudflare and other edge runtimes. In the ORM, when you run `prisma generate`, we will suggest in the successful generation message to use Accelerate for `@prisma/client/edge`. So I think my addition is required here at least. Ideally we could also point to guides once we revamped them.

### DIFF
--- a/content/800-data-platform/100-accelerate/200-getting-started.mdx
+++ b/content/800-data-platform/100-accelerate/200-getting-started.mdx
@@ -124,6 +124,15 @@ import { withAccelerate } from '@prisma/extension-accelerate'
 const prisma = new PrismaClient().$extends(withAccelerate())
 ```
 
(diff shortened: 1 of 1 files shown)


content/800-data-platform/100-accelerate/200-getting-started.mdx
```diff
@@ -124,6 +124,15 @@ import { withAccelerate } from '@prisma/extension-accelerate'
 const prisma = new PrismaClient().$extends(withAccelerate())
 ```
 
+If you are going to deploy to an edge runtime (like Cloudflare Workers, or Vercel Edge Functions, Deno Deploy, or Netlify Edge Functions), use our edge client instead:
```
jharrell:
```suggestion
If you are going to deploy to an edge runtime (like Cloudflare Workers, Vercel Edge Functions, Deno Deploy, or Netlify Edge Functions), use our edge client instead:
```

jharrell (approved):

PR #5196, opened 2023-08-20
- added max width constraint to avoid overflow of text

### LINKED ISSUE #5114: Code block overflows when used in an admonition component
Example page: https://www.prisma.io/docs/concepts/database-connectors/sql-server

CleanShot 2023-07-31 at 15 15 14@2x

kuluruvineeth:
## Describe this PR

Code block overflows when used in an admonition component 

## Changes

Added max-width: inherit;

## What issue does this fix?
Screenshot 2023-08-20 at 7 18 28 PM

Fixes #5114


jharrell:
Hi there! We recently updated our docs site to Docusaurus. We have a few remaining issues to resolve with code blocks and admonitions, but I think this is now more consistent. Please feel free to open another PR if it's not resolved!

PR #5201, opened 2023-08-21
docs: add `datasourceUrl` to Prisma Client Reference

ruheni:


jharrell (approved):

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -101,6 +101,26 @@ datasource db {
 }
 ```
 
+### <inlinecode>datasourceUrl</inlinecode>
+
+Shorthand for programmatically overriding the [`datasource`](#datasources) property.
```
janpio: Not really a shorthand, but just an additional way of doing it.

content/400-reference/200-api-reference/050-prisma-client-reference.mdx
```diff
@@ -101,6 +101,26 @@ datasource db {
...
+| -------------------------- | -------------------- | ----------------------------------------------------------------------------- |
+| Database connection string | `'file:./dev_qa.db'` | The database [connection URL](/reference/database-reference/connection-urls). |
+
+#### Examples
+
+```ts
+import { PrismaClient } from '@prisma/client'
+
+const prisma = new PrismaClient({
+  datasourceUrl: 'file:./dev_qa.db',
```
janpio: 99% of our users want a connection string there, not a SQLite file path. Let's use that as the example.

ruheni: The file path is still a connection string, only that it's specific to SQLite, correct?

janpio: Yes, but for 99% of our users it will look unexpected nad not how what they are trying to do and "scan" for.

SevInf:
Should we also mention new option in `datasources` section? At the moment, there is nothing verbose option can do that short can not, so maybe it worth pointing out that since 5.2 there is shorter way of doing it?

SevInf (approved):

PR #5208, opened 2023-08-22
ci(lost-pixel): Extract list of changed pages and output to observe + improve

janpio:
Basis for being able to give LP the list of actually changes pages so they can minimize the pages they screenshot, leading to lower cost and duration.


janpio:
I think this should be merged @nilubava @carlagn even though we currently do not submit it to LP - but this is the basis we can use for that.

jharrell:
I don't think that this is going to work. What if we change an mdx component that appears on every page, but no content pages? 

Additionally, we have the ability to use [partial pages](https://docusaurus.io/docs/markdown-features/react#importing-markdown) now, which this would also miss. 

I agree that the time spent during the lost-pixel check is long, but at this point I don't think it's worth changing. Do you agree?

janpio:
For PRs that only change some .mdx files (which is 99%) this would cut down the check time (and also cost for LP) significantly. Of course for changes of mdx components, or styles, or any non content, the full docs would need to be tested.

PR #5215, opened 2023-08-23
feat: new mdx component for version info

nilubava:
Empty component that does nothing, besides making it possible to mark all the version information in all the files.
When we have done that, we can easily find it and decide how to visually deal with it.


janpio:
Awesome! Now looking at the rendered output, I realized that no one will be able to differentiate a normal paragraph from one that is using this component 🤦. Adding a `class="versioninfo"` should be free and invisible, right? That way one can inspect code, or use a custom stylesheet to style those things a bit to make it easier to detect.

PR #5234, opened 2023-08-28
Update index.mdx

austin-mccollom:
Prisma Migrate is declarative. You define the desired state of your schema in Prisma Schema Language, and the tool figures out the changes needed to go from the current state to that desired state. https://en.wikipedia.org/wiki/Declarative_programming


ruheni (approved):
Thanks for catching this!

PR #5235, opened 2023-08-29
fix: suggested change for flaky tests by Dima

nilubava:
Context: 

Fix for flaky tests by Dima

### DIFF
--- a/lostpixel.config.js
+++ b/lostpixel.config.js
@@ -28,7 +28,13 @@ module.exports = {
         /* hide cookie banner */
         #onetrust-consent-sdk {
           display: none;
-        }`,
+        }
+        
+        /* reset menu item alignment */
+        #sidebar-holder li a {
+          vertical-align: baseline;
+        }
+        `,
     })
   },
 }


janpio:
@d-ivashchuk Unrelated: This problem still exists for me:
https://app.lost-pixel.com/app/repos/7/clb5ek3mm1772001qqg7yban38?build=cllw4u9px2kiund0ett3o9wc0&diff=cllw4yd9006rfa20exnnsil4r
image

d-ivashchuk:
> @d-ivashchuk Unrelated: This problem still exists for me: https://app.lost-pixel.com/app/repos/7/clb5ek3mm1772001qqg7yban38?build=cllw4u9px2kiund0ett3o9wc0&diff=cllw4yd9006rfa20exnnsil4r image

hmm, what exactly you feel is not right here? the line height changed a bit, so one image is smaller vertically than the other one 🤔 on the side-by-side comparison they should be pretty aligned if we are talking about alignment there. 

but please elaborate more, would be happy to fix it :D

janpio:
The "left" image is ~50 pixels lower then the "right" one, with "transparent" space on top. This makes the slider view completely useless in actually figuring out the differences by sliding around.

d-ivashchuk:
> The "left" image is ~50 pixels lower, with "transparent" space on top. This makes the slider view completely useless.

so basically the improvement here would be to align them as they are aligned on _**side-by-side**_ view 👍🏼

chriskalmar:
This is the new view for the slider:
image

PR #5246, opened 2023-08-31
docs: extension limitations for nested operations

### LINKED ISSUE #5217: docs: make it obvious that extensions don't work with nested reads/writes
I found this caveat buried in one of the [examples](https://github.com/prisma/prisma-client-extensions/blob/8661dbcf22456b603204246ca5dac65cd8264a36/input-validation/README.md#L9):

> **NOTE**: Query extensions do not currently work for nested operations. In this example, validations are only run on the top level `data` object passed to methods such as `prisma.product.create()`. Validations implemented this way do not automatically run for [nested writes](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#nested-writes).

This seems like a pretty important caveat that should be in a more-obvious place (perhaps the extension docs site).

ruheni:
Closes #5217

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/053-client-extensions/index.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/053-client-extensions/index.mdx
@@ -196,3 +196,7 @@ if (xPrisma.$connect) {
   xPrisma.$connect()
 }
 ```
+
+### Usage with nested operations
+
+Prisma Client extensions currently do not support nested read and write operations. If you're interested in this feature, let us know by opening a [feature request](https://github.com/prisma/prisma/issues/new?assignees=&labels=&template=feature_request.md&title=) with your use case. 


jharrell (approved):

content/200-concepts/100-components/02-prisma-client/053-client-extensions/index.mdx
```diff
@@ -196,3 +196,7 @@ if (xPrisma.$connect) {
   xPrisma.$connect()
 }
 ```
+
+### Usage with nested operations
+
+Prisma Client extensions currently do not support nested read and write operations. If you're interested in this feature, let us know by opening a [feature request](https://github.com/prisma/prisma/issues/new?assignees=&labels=&template=feature_request.md&title=) with your use case.
```
janpio, line 202: We don't really want separate feature requests for this.
We should have one already from @SevInf probably, and ask for comment with use case on that.

PR #5247, opened 2023-08-31
Update note about `migrate deploy` warning for edited migrations

### LINKED ISSUE #5244: Migration history conflicts
In the 'Migration troubleshooting in production' docs (https://www.prisma.io/docs/guides/migrate/production-troubleshooting#migration-history-conflicts) it's said that `prisma migrate deploy` issues a warning if an already applied migration has been edited. However, the warning doesn't seem to be shown anywhere. 
Did I misunderstand something, or maybe should enable some special log level?

ruheni:
Closes #5244

### DIFF
--- a/content/300-guides/025-migrate/800-production-troubleshooting.mdx
+++ b/content/300-guides/025-migrate/800-production-troubleshooting.mdx
@@ -202,6 +202,13 @@ Your local migration history now yields the same result as the state your produc
 
 `prisma migrate deploy` issues a warning if an already applied migration has been edited - however, it does not stop the migration process. To remove the warnings, restore the original migration from source control.
 
+<Admonition type="warning">
+
+From version [3.12.0](https://github.com/prisma/prisma/releases/tag/3.12.0), `prisma migrate deploy` does not surface a warning if a migration is edited after it is edited. The warning will be surfaced by `prisma migrate dev`.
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
Let's turn this around - remove the default text that does not apply since 3.12.0 any more, and instead add an info admonition that until 3.12.0 this was happening.

And the `migrate dev` warning probably should be put somewhere else as normal information - that was happening before and is now.

Jolg42 (approved):

ruheni:
I've updated the text in the admonition.

@janpio We also document what happens if a migration is edited here: https://www.prisma.io/docs/concepts/components/prisma-migrate/migration-histories#:~:text=Run%20prisma%20migrate%20dev%20%2D%20Prisma%20Migrate%20detects%20that%20a%20migration%20has%20changed%2C%20and%20asks%20to%20reset%20the%20database%3A. `prisma migrate dev` will prompt for a reset if a migration is edited after it's applied.

janpio (approved):
The default text of the docs, when you ignore the admonition, is still not correct since 3.12.0.
Optimally our docs should always represent the _current_ state, and use admonitions and asides to highlight old behavior or when something changed to the current state.

But ok, not super important here.

PR #5250, opened 2023-08-31
fix(metrics): 2 "Global labels" examples should be enough

janpio:
Right now the full diff is super distracting and does not add much.


jharrell (approved):

ruheni:
I would, personally, recommend keeping the top-level properties, i.e., counters, gauges & histograms in the example snippet.

ruheni (approved):

PR #5228, opened 2023-08-26
added default export for global prisma client

pundang:
## Describe this PR

Missing default export in prisma client for instantiating the client in Next.js

## Changes

- Updated `400-nextjs-prisma-client-dev-practices.mdx`

## What issue does this fix?

There's no issue

## Any other relevant information

None

PR #5253, opened 2023-08-31
update import example for global prisma client

pundang:
## Describe this PR

I forgot to update the import example in #5228

## Changes

- Updated import example for global client

## What issue does this fix?

Fixes #5228

No :trollface:


ruheni:
👋 

Thanks for catching that. I merged #5263 before seeing your PR. In case you catch any other inconsistencies, don't hesitate to create a PR/ issue :)

PR #5256, opened 2023-09-01
fix: remove distributed from accelerate definition

ankur-arch:
- Accelerate is not a distributed cache; it's globally available.
- It's connection pooling features also benefit edge applications

### DIFF
--- a/content/800-data-platform/100-accelerate/100-what-is-accelerate.mdx
+++ b/content/800-data-platform/100-accelerate/100-what-is-accelerate.mdx
@@ -14,7 +14,7 @@ Its main features are:
 
 - a global cache
 - scalable connection pool for serverless and edge applications
-- usage of Prisma Client at the Edge (e.g. in Cloudflare Workers or Vercel Edge Functions)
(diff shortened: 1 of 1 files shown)


nikolasburk (commented):
I find it a bit confusing the that bullet 2 lowercases "edge" but bullet 3 uppercases it as "Edge". maybe we should lower case "at the edge" as well if we generally don't uppercase it?

content/800-data-platform/100-accelerate/100-what-is-accelerate.mdx
```diff
@@ -12,8 +12,8 @@ toc: true
 
 Its main features are:
 
-- a globally distributed cache
-- scalable connection pool for serverless applications
+- a global cache
+- scalable connection pool for serverless and edge applications
 - usage of Prisma Client at the Edge (e.g. in Cloudflare Workers or Vercel Edge Functions)
```
ankur-arch:
```suggestion
- usage of Prisma Client at the edge (e.g. in Cloudflare Workers or Vercel Edge Functions)
```

PR #5261, opened 2023-09-02
chore(docs): update prisma import

joshxfi:
## Describe this PR

On the article [Best practice for instantiating PrismaClient with Next.js](https://www.prisma.io/docs/guides/other/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices):

`prisma` is exported via `export default` & was imported as `import { prisma } from './db'`. 
It should be imported as `import prisma from './db'`.
Otherwise, it should be `export const prisma = ...` instead of using default.

## Changes
\- `import { prisma } from './db'`
\+`import prisma from './db'`


ruheni:
Thanks for catching this, @joshxfi. I merged #5263 before getting to this PR. Don't hesitate to create a PR if you catch any other inconsistencies in our docs. 🙂

For now, I'll go ahead and close this PR.

PR #5275, opened 2023-09-04

sably:

### DIFF
--- a/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
+++ b/content/200-concepts/100-components/01-prisma-schema/06-relations/500-troubleshooting-relations.mdx
@@ -24,7 +24,7 @@ model Animal {
 }
 ```
 
-The resulting relation table in SQL looks as follows, where `A` represents prey (`a_eats`) and `B` represents predators (`a_eatenBy`):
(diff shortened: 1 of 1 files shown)


janpio:
Aside: Why do we even use the `a_` and `b_` prefixes here for the relation field names?

sably:
> Aside: Why do we even use the `a_` and `b_` prefixes here for the relation field names?

 they affect how you reference and query the data.

janpio:
Yes, but `eats` and `eatenBy` would make much more sense here. I don'T know that the `a_` and `b_` mean here.

(This was more a question to whoever wrote this page and will review your edit)

sably:
> 

whenever I write this page again I take a new conclusion? why?

sably:
> Yes, but `eats` and `eatenBy` would make much more sense here. I don'T know that the `a_` and `b_` mean here.
> 
> (This was more a question to whoever wrote this page and will review your edit)

whenever I read this page again I take a new conclusion. why?

nikolasburk:
Thanks for the PR @sably 💚 

Also: @janpio I agree with you, this example is confusing. I created an issue to review and improve it: https://github.com/prisma/docs/issues/5488

Nonetheless, the change from @sably is definitely correct since it catches a typo and makes the example (at least technically) correct.

PR #5289, opened 2023-09-07
Use `terminal` instead of `sh` language annotation on terminal commands

### LINKED ISSUE #5287: Use `terminal` instead of `sh` language annotation on terminal commands
For example on this page: https://www.prisma.io/docs/guides/deployment/serverless/deploy-to-aws-lambda

But while on it, maybe do a global search in the docs for other occurrences?

jharrell:
Fixes #5287

## Describe this PR

Updates most usage of `sh` to `terminal` for code snippets.

### DIFF
--- a/content/300-guides/200-deployment/201-serverless/400-deploy-to-aws-lambda.mdx
+++ b/content/300-guides/200-deployment/201-serverless/400-deploy-to-aws-lambda.mdx
@@ -43,7 +43,7 @@ Your functions will need the `DATABASE_URL` environment variable to access the d
 
 First, make sure that the plugin is installed:
 
-```sh
+```terminal
 npm install -D serverless-dotenv-plugin
 ```
 
@@ -59,14 +59,14 @@ The environment variables in your `.env` file will now be automatically loaded o
 <CodeWithResult expanded="{true}">
 <cmd>
 
-```sh
+```terminal
 serverless package
 ```
 
 </cmd>
 <cmdResult>
 
-```sh no-copy
+```terminal no-copy
 Running "serverless" from node_modules
 DOTENV: Loading environment variables from .env:
          - DATABASE_URL
@@ -122,7 +122,7 @@ If you use `serverless-webpack`, you will need additional configuration so that 
 
 First, ensure the following webpack dependencies are installed:
 
-```sh
+```terminal
 npm install --save-dev webpack webpack-node-externals copy-webpack-plugin serverless-webpack
 ```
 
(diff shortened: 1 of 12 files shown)


content/300-guides/200-deployment/201-serverless/400-deploy-to-aws-lambda.mdx
```diff
@@ -195,7 +195,7 @@ custom:
         -- find . -name "libquery_engine-*" -not -name "libquery_engine-rhel-openssl-*" | xargs rm
 ```
 
-If you are deploying to [Lambda functions with ARM64 architecture](#lambda-functions-with-arm64-architectures) you should update the find command to the following:
+If you are deploying to [Lambda functions with ARM64 architecture](#lambda-functions-with-arm64-architectures) you should update the `find` command to the following:
 
 ```sh
```
nikolasburk:
```suggestion
```terminal
```

nikolasburk (commented):
I believe I found one instance of `sh` that slipped, or was that left intentionally? 

Also, this is personal preference but I would prefer if the CLI results on this page were hidden by default.

image

The results are very noisy and distract from the actual important bit (the command that needs to be run).

nikolasburk (commented):
I believe I found one instance of `sh` that slipped, or was that left intentionally? 

Also, this is personal preference but I would prefer if the CLI results on this page were hidden by default.

image

The results are very noisy and distract from the actual important bit (the command that needs to be run).

content/300-guides/200-deployment/201-serverless/400-deploy-to-aws-lambda.mdx
jharrell: Keeping this was intentional since it's not a command to be run in the terminal, it's suggesting how you change the `find` command on line 195

Maybe this will work? 
````suggestion
```
````

Comment (nikolasburk) [reply content/300-guides/200-deployment/201-serverless/400-deploy-to-aws-lambda.mdx]: Ah, sorry 🤦 

In that case, I actually think it's most clear if we repeat the entire file. I'll make a separate suggestion.

File: content/300-guides/200-deployment/201-serverless/400-deploy-to-aws-lambda.mdx [hunk 2]
```diff
@@ -195,7 +195,7 @@ custom:
...
 find . -name "libquery_engine-*" -not -name "libquery_engine-arm64-openssl-*" | xargs rm
```
nikolasburk:
```suggestion
```yaml file=serverless.yml highlight=6;add
custom:
  webpack:
    packagerOptions:
      scripts:
        - prisma generate
        -- find . -name "libquery_engine-*" -not -name "libquery_engine-arm64-openssl-*" | xargs rm
```

nikolasburk (approved):
looks good now, thanks for the quick action on this Jon!

PR #5297, opened 2023-09-11
Fix case on `pgbouncer=true` param

NicolasMahe:
## Describe this PR

Fix the case of the param `pgbouncer` in page "Connection management".
It must be lowercased, this page currently has it with a uppercased `B`.
All other pages have already the correct lowercased version.

## Changes

- Fix case on param `pgbouncer` in page "Connection management"


ruheni (approved):

ruheni:
Thanks for catching this! 🙏

PR #5304, opened 2023-09-12
Update P2002 error reference

grukillerkams:
Added how to access the actual fields causing violation


janpio:
Thanks, but the `constraint` here is a placeholder that is fille in the code side - and might be different from what is actually in `meta.target`. But you are correct, we are lacking documention on this in some other form.

PR #5319, opened 2023-09-15
fix: table dimensions + alignments

carlagn:
Fixes [#505](https://github.com/prisma/website/issues/505)

Please refer to this comment: https://github.com/prisma/website/issues/505#issuecomment-1700941316

Main updates to https://docs-git-fix-table-dimensions-alignments-prisma.vercel.app//reference/database-reference/database-features/#relational-database-features


meletj (changes requested):
- [x] Microsoft SQL server column under 'Constraints' table should also be center aligned

I checked the comment on the other PR, but I couldn't find the reasoning behind changing the width back to standard. Any reason we cannot leave it as 'wide: true'?

carlagn:
@meletj 
> * [ ]  Microsoft SQL server column under 'Constraints' table should also be center aligned
> 
> I checked the comment on the other PR, but I couldn't find the reasoning behind changing the width back to standard. Any reason we cannot leave it as 'wide: true'?

@Techboidesign left a comment [here](https://github.com/prisma/website/issues/505#issuecomment-1719330375) on that for you. Asking the same actually: wether or not we keep the `wide` frontmatter.

meletj:
I'd say we keep it to 'wide: true' for this page - this is a one-off / special case.

meletj (approved):

PR #5324, opened 2023-09-18
change: page title for tech switcher paths to have `:` instead of `|`

nilubava:
Page title for tech switcher paths to have `:` instead of `|` since the title suffix is `| Prisma Docs`

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -35,7 +35,7 @@ const getTitle = (frontmatter: any, lang?: any, db?: any) => {
     const titleEntry = frontmatter.techMetaTitles
       ? frontmatter.techMetaTitles.find((item: any) => item.name === queryParam)
       : null
-    pageSeoTitle = titleEntry ? titleEntry.value : `${pageSeoTitle} | ${queryParam}`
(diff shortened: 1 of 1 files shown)


gatsby-node.ts
```diff
@@ -35,7 +35,7 @@ const getTitle = (frontmatter: any, lang?: any, db?: any) => {
     const titleEntry = frontmatter.techMetaTitles
       ? frontmatter.techMetaTitles.find((item: any) => item.name === queryParam)
       : null
-    pageSeoTitle = titleEntry ? titleEntry.value : `${pageSeoTitle} | ${queryParam}`
+    pageSeoTitle = titleEntry ? titleEntry.value : `${pageSeoTitle} : ${queryParam}`
```
ruheni, line 38: I'm not sure if this is quite correct 😄 
The current title contains `typescript-postgresql` from `query-param` I think:

```
Start from scratch with relational databases (15 min) : typescript-postgresql
```

ruheni (dismissed):

gatsby-node.ts
ruheni: This seems to be the current behaviour. You can go ahead and ignore the comment 😅

PR #5330, opened 2023-09-19
excluding-fields.mdx: add obfuscate fields to Going Further list

vsilman:
## Describe this PR

We wanted to exclude certain fields from our models. Information to obfuscate fields was only located within GitHub. Documentation should show this as an alternative to exclusion. 

## Changes

- Add obfuscate fields to Going Further list in Exclude Fields


ruheni (approved):
Thanks for adding this to our docs! Much appreciated.

PR #5339, opened 2023-09-22
docs: fix incorrect link to bark client extension

jthomaschewski:
## Changes

Link to bark was incorrectly pointing to a completely unrelated pagination extension.
I fixed the link.

Maybe it makes sense to add the pagination in a new row.


janpio:
This seems to have gone wrong already when @ruheni added the list initially.

ruheni (approved):

ruheni:
Thanks for catching this 🙏

PR #5343, opened 2023-09-24
Update some mentions of Data Browser

petradonka:
I think we'll also need to adjust the Cloudflare Worker and the Deno deployment guides to use Accelerate, @ankur-arch, is the best way to do that to create an issue in this repo?


ankur-arch:
@petradonka additional issues would be better.

content/200-concepts/100-components/06-prisma-studio.mdx
```diff
@@ -10,13 +10,6 @@ Prisma Studio is a visual editor for the data in your database. Note that Prisma
 
 Run `npx prisma studio` in your terminal.
 
-<Admonition>
-
-**Note**<br /><br />
-The features of Prisma Studio are similar to those in the Data Browser in the Prisma Data Platform. See the [Data Browser](/data-platform/classic-projects/platform/data-browser) documentation.
-
```
janpio: Related: We'll want to move the existing docs from Data Browser to Studio.

nikolasburk (approved):
Was this change made to make it explicit that Data Browser lives on `cloud.prisma.io` as opposed to `console.prisma.io`? If so, I guess it looks good to me 👍

PR #5345, opened 2023-09-25
Update 160-baselining.mdx

GuilleAmutio:
Specify the initial migration that the document is working with

## Describe this PR

This PR aims to make more visible to the user that the name of the migration is the folder that was created in the previous steps.

## Changes

Instead of ```20210426141759_initial-migration-for-db``` using ```0_init``` shows the user that the name of the folder is indeed the name of the migration


janpio (approved):

ruheni:
Thanks for catching this and updating the docs @GuilleAmutio! 🙏

PR #5348, opened 2023-09-26
add quick admonition for prisma during vercel deployment

jharrell:


content/300-guides/200-deployment/201-serverless/300-deploy-to-vercel.mdx
```diff
@@ -32,6 +32,12 @@ Vercel will automatically cache dependencies on deployment. For most application
 
 This will re-generate Prisma Client at build time so that your deployment always has an up-to-date client.
 
+<Admonition>
+
+If you see `prisma: command not found` errors during your deployment to Vercel, you are missing `prisma` in your dependencies. By default, `prisma` is a dev dependency and may need to be moved to be a standard dependency.
```
janpio, line 37: The second sentence is not correct I think and it should not matter if it is a dev or normal dependency - the problem should only happen if it is not a dependency at all.

PR #5349, opened 2023-09-28
Fix yarn add --dev command

hyperflux:
Yarn uses `--dev` to add devDependencies, not `--save-dev`

## Describe this PR

This yarn command was incorrectly using npm's `--save-dev` flag. Just fixing that little issue.

## Changes

Easy fix!

## What issue does this fix?

Now anybody who copies this command and runs it will add prisma cli to their devDependencies.


ruheni:
Thanks for catching this @hyperflux! 🙏

PR #5370, opened 2023-10-12
Fixed update event filter example

benjs04:
Changed `create` to `update` in the update filter event example.

## Describe this PR

The update event filter example was using `create` instead of `update`, which I corrected.

## Changes

Changed `create` to `update` in the example showing how to use the filter when subscribing to update events.


ruheni (approved):

ruheni:
Thanks for catching this! I appreciate the contribution.

PR #5382, opened 2023-10-16
perf: improve newsletter form

iiio2:
## Describe this PR

The newsletter form in the footer was working even if no input is provided. It was showing `Thank you` to the client after the click event even if no input data. It is fixed now.


iiio2:
@janpio still not reviewed.

janpio:
Yes, we will get to review this when we have time. This does not sound super urgent, so we will take care of it when it fits.

PR #5383, opened 2023-10-17
Update 063-excluding-fields.mdx

mubaidr:
## Describe this PR

Update exclude function to make it short and compatible with typescript version 4.4+

## Changes

Just updated example `exclude` function definition

## What issue does this fix?

Exclude function does not support latest typescript version, gives warnings. 

N/A

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
@@ -22,13 +22,12 @@ The following is a type-safe `exclude` function returns a user without the `pass
 
 ```tsx
 // Exclude keys from user
-function exclude<User, Key extends keyof User>(
-  user: User,
-  keys: Key[]
-): Omit<User, Key> {
-  return Object.fromEntries(
-    Object.entries(user).filter(([key]) => !keys.includes(key))
-  )
+function exclude<T extends object, K extends keyof T>(
+  obj: T,
+  key: K,
+): Omit<T, K> {
+  const { [key]: omitted, ...rest } = obj
+  return rest
 }
 
 function main() {


content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
```diff
@@ -22,13 +22,12 @@ The following is a type-safe `exclude` function returns a user without the `pass
 
 ```tsx
 // Exclude keys from user
```
nikolasburk, line 24:
```suggestion
// Exclude keys from user (requires TypeScript > 4.4)
```

content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
```diff
@@ -22,13 +22,12 @@ The following is a type-safe `exclude` function returns a user without the `pass
...
-function exclude<User, Key extends keyof User>(
-  user: User,
-  keys: Key[]
-): Omit<User, Key> {
-  return Object.fromEntries(
-    Object.entries(user).filter(([key]) => !keys.includes(key))
-  )
+function exclude<T extends object, K extends keyof T>(
+  obj: T,
```
nikolasburk, line 26: can we make this optional?

```suggestion
  obj: T | null,
```

content/200-concepts/100-components/02-prisma-client/063-excluding-fields.mdx
```diff
@@ -22,13 +22,12 @@ The following is a type-safe `exclude` function returns a user without the `pass
...
+  key: K,
```
nikolasburk, line 27: is it possible to keep this as `K[]`?

nikolasburk (changes requested):
Hey @mubaidr, thanks a lot for this PR 🙏  it looks great 💚 

I tested the code and it worked on my machine 🎉 

The main concern I have is that it changes the signature of the `exclude` function and now only allows to exclude a single key, i.e. `exclude(user, 'password')` instead of multiple keys (passed as an array), e.g. `exclude(user, ['id', 'password'])` as it's intended.

- Would you be able to update this and make the `exclude` function accept an _array_ of keys?
- Also, can you please make the input `obj` optional?
- And finally (since you're on it): This didn't come from your PR but I noticed that the example in `main` actually doesn't run 🤦 

It currently has the following query:

```ts
const user = await prisma.user.findUnique({ where: 1 })
```

but it should be:

```ts
const user = await prisma.user.findUnique({
  where: {
    id: 1
  }
})
```

Thanks for caring about our docs and taking the time to improve them, that's much appreciated!!

PR #5401, opened 2023-10-26
feat: add testing section for Accelerate

ankur-arch:
Add a section for testing Accelerate


luanvdw (approved):

janpio:
Should this section be called "Evaluating" instead? With "Testing" one usually means to test for correctness of the software one has written, while this here describes a process to evaluate the characteristics of Accelerate. For actual "Testing" this would be completely inadequate. @luanvdw

PR #5405, opened 2023-10-29
fix(pagination): Explain what happens when cursor does not exist

### LINKED ISSUE #14283: non-existent cursor user experience
## Problem

Right now, If we provide a `cursor` that does not exist in the table (and the table contains records) Prisma returns `null` even when though there are elements after and before the cursor and a `take` was specified.

```
this.prismaService.brand.findMany({
    cursor: {
        id: body.cursor,
    },
    take: body.take ? body.take : DEFAULT_TAKE,
    include: {
        pdf: true,
        variants: true,
    },
    orderBy: {
        id: 'asc',
    },
})
```
## Suggested solution

Update the docs, specifically the [cursor-based pagination guide](https://www.prisma.io/docs/concepts/components/prisma-client/pagination#cursor-based-pagination) to indicate that supplying a non-existent `cursor` will yield `null`.

## Alternatives

Another possible behavior is to return the closest next elements to the cursor. here is an example implementation using `where`.

```
this.prismaService.brand.findMany({
    where: {
        id: {
            ...(body.take < 0 && {
                lt: body.cursor,
            }),
            ...(body.take > 0 && {
                gt: body.cursor,
            }),
        },
    },
    take: body.take ? body.take : DEFAULT_TAKE,
    include: {
        pdf: true,
        variants: true,
    },
}),
```

## Additional context

None.

janpio:
closes https://github.com/prisma/prisma/issues/14283


content/200-concepts/100-components/02-prisma-client/055-pagination.mdx
```diff
@@ -158,6 +158,10 @@ If you guess the value of the next cursor, you will page to an unknown location
 
 No, cursor pagination does not use cursors in the underlying database ([e.g. PostgreSQL](https://www.postgresql.org/docs/9.2/plpgsql-cursors.html)).
 
+#### What happens if the cursor value does not exist?
+
+Using a nonexistent cursor returns `null`. Prisma does not locate adjacent values.
```
janpio:
```suggestion
Using a nonexistent cursor returns `null`. Prisma does not try to locate adjacent values.
```

jharrell (approved):

nikolasburk:
Merging since it's a helpful addition to the docs and @jharrell  has approved it ✅  thanks Jan!

PR #5409, opened 2023-10-31
Correcting Escaped Code Example

shery:
In js, according to the normal string escape rules, need to add `\` before `\`

### DIFF
--- a/content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
+++ b/content/200-concepts/100-components/02-prisma-client/050-filtering-and-sorting.mdx
@@ -267,7 +267,7 @@ The above query will match any user whose name starts with a character followed 
 const users = await prisma.user.findMany({
   where: {
     name: {
-      startsWith: '\_benny', // note that the `_` character is escaped
+      startsWith: '\\_benny', // note that the `_` character is escaped, preceding `\` with `\` when included in a string
     },
   },
 })


jharrell:
@shery Hey! Thanks for your PR. I think you might be mistaken. We're not trying to escape `\` we're just trying to escape `_` so `\` should not be escaped. 

To put it another way, the example shows that the underlying database will understand `_benny` and `\_benny` differently as the latter will treat `_` as a literal character instead of a wildcard.

shery:
Hi @jharrell, If executing SQL statements through the mysql Client, then `\_` will work as expected (matching the `_` character). However, Prisma Client runs in a JavaScript environment, and due to JavaScript's string escape rules (escaping special characters with `\`), `\_` will be interpreted as escaping `_` in JavaScript string, so the actual in-memory character is `_`, causing the finally generated SQL query statement to not contain `\`. This results in `_` not being escaped.

Here is a simple example: 

| id | name | age | city |
|-|-|-|-|  
| 1 | John Doe | 35 | new_york |
| 2 | Jane Smith | 28 | newxyork |  

```js
const arr1 = await prisma.user.findMany({
  where: {
    city: {
      contains: 'new\_york',
    },
  },
})

console.log(arr1)
// [{"id":1,"name":"John Doe","age":35,"city":"new_york"},{"id":2,"name":"Jane Smith","age":28,"city":"newxyork"}]

const arr2 = await prisma.user.findMany({
  where: {
    city: {
      contains: 'new\\_york',
    },
  },
})

console.log(arr2)
// [{"id":1,"name":"John Doe","age":35,"city":"new_york"}]
```

jharrell:
@shery Huh. How about that. You are correct! Thank you for correcting and apologies for my mistake.

jharrell (approved):

PR #5427, opened 2023-11-06
feat: PgBouncer => external connection pooler

janpio:
Generalize in prep for upcoming PgBouncer changes.


content/300-guides/100-performance-and-optimization/150-connection-management/200-configure-for-external-connection-pooler.mdx
```diff
@@ -1,38 +1,44 @@
 ---
-title: Configure Prisma Client with PgBouncer
+title: Configure Prisma Client with an external connection pooler (like PgBouncer)
```
ruheni, line 2:
```suggestion
title: Configure Prisma Client with an external connection pooler
```

I don't think we need it in the title but we could include it as part of the metaTitle and metaDescription.

content/300-guides/100-performance-and-optimization/150-connection-management/200-configure-for-external-connection-pooler.mdx
```diff
@@ -1,38 +1,44 @@
...
 
-To use Prisma Client with PgBouncer from a serverless function, add the `?pgbouncer=true` flag to the PostgreSQL connection URL:
+### Set PgBouncer to transaction mode
 
-```
-postgresql://USER:PASSWORD@HOST:PORT/DATABASE?pgbouncer=true
-```
+For Prisma Client to work reliably, PgBouncer must run in [**Transaction mode**](https://www.pgbouncer.org/features.html).
+
+_Transaction mode_ offers a connection for every transaction – a requirement for the Prisma Client to work with PgBouncer.
```
ruheni:
```suggestion
Transaction mode offers a connection for every transaction – a requirement for the Prisma Client to work with PgBouncer.
```

We don't need this here as the first mention of transaction mode is/ was bolded

ruheni (approved):
LGTM. 

I've left minor comments. We can merge it once they're resolved ~and the merge conflict is also resolved~

content/300-guides/100-performance-and-optimization/150-connection-management/200-configure-for-external-connection-pooler.mdx
janpio: Let's keep it so Google does not get weird ideas to rank other pages from ours for `PgBouncer`. Title has the biggest effect on that.

PR #5432, opened 2023-11-07
add snippet installing ws types

ruheni:


Jolg42 (approved):
Looks good, I think it was the only place where this was mentioned.

ruheni:
>  I think it was the only place where this was mentioned.

Correct. The "Database drivers" page links to this page. 😄

PR #5445, opened 2023-11-13
fix(accelerate): Rename "Testing" to "Evaluating"

janpio:
The content of this page has little to do with _testing_ as it is usually understood in software and Prisma context, which is about making sure that the software does the correct thing. 

Instead it is about _evaluating_ the performance and functionality of Prisma Accelerate.

Hence I renamed the page and replaced the term in the content.


luanvdw (approved):
Thanks @janpio

PR #5451, opened 2023-11-16
fix: remove legacy billing for data platform

ankur-arch:
The billing keeps showing up in Kapa search results, hence we're removing this.


nikolasburk (approved):
lgtm 👍

nikolasburk (approved):
lgtm 👍

PR #5456, opened 2023-11-16
Update Prisma docs to clarify 'migrate dev' behavior for schema changes

PutziSan:
## Describe this PR

I propose this documentation update because I encountered confusion about how to integrate migrations from my colleagues locally without generating a new migration file. Initially, I didn't find clear documentation on this and ended up using `migrate deploy`. However, when I accidentally ran `migrate dev` without making any changes to the schema, I realized how it actually works. This update aims to clarify this aspect for others who might face the same confusion.

## Changes

- Updated the "Create and apply migrations" section in the Prisma documentation to clarify that the `migrate dev` command only generates a new migration file if it detects changes in the `schema.prisma` file. This change explicitly informs users that new migrations are created only when necessary, based on modifications to the schema.

## What issue does this fix?

This update doesn't correspond to an existing issue in the repository. It's a proactive clarification based on personal experience and confusion encountered while using Prisma Migrate in a development environment.

This change should make it easier for developers to understand the behavior of the `migrate dev` command, particularly in scenarios involving the integration of migrations from team members, without the unnecessary creation of new migration files. It's a clarification based on practical usage and understanding of the Prisma Migrate process.


content/200-concepts/100-components/03-prisma-migrate/080-migrate-development-production.mdx
```diff
@@ -32,6 +32,7 @@ This command:
 1. Reruns the existing migration history in the [shadow database](shadow-database) in order to detect schema drift (edited or deleted migration file, or a manual changes to the database schema)
 1. Applies pending migrations to the shadow database (for example, new migrations created by colleagues)
 1. Generates a new migration from any changes you made to the Prisma schema before running `migrate dev`
+1. If it detects changes to the Prisma schema, it generates a new migration from these changes
```
nikolasburk: Isn't the point you added equivalent to the previous point? The way I read them, they express exactly the same (i.e. "a new migration is generated if there were changes to the Prisma schema").

So, with your current suggestions, I think 3. and 4. would be redundant.

However, I do find your version a bit clearer so my suggestion would be to drop the existing line and keep yours.

```suggestion
1. If it detects changes to the Prisma schema, it generates a new migration from these changes
```

image

nikolasburk (changes requested):
Hey @PutziSan 👋  thanks a lot for this PR and taking the time to improve our docs 💚 

I've left a comment with a suggestion, please let me know what you think!

janpio (approved):

nikolasburk (approved):

nikolasburk:
Thanks for helping us out, @PutziSan 🎉 💚

PR #5468, opened 2023-11-17
docs: update landing page for getting started

nikolasburk:
- remove wall of text & make product descriptions more concise
- introduce separate sections for getting started with ORM and PDP
- add CTAs for Pulse and Accelerate

### DIFF
--- a/content/100-getting-started/index.mdx
+++ b/content/100-getting-started/index.mdx
@@ -10,9 +10,9 @@ Welcome to the Prisma getting started section! 👋
 
 [**Prisma ORM**](/concepts/overview/what-is-prisma) is an [open source](https://github.com/prisma/prisma) next-generation database toolkit that helps with data modeling, data access, schema migrations and more.
 
-In addition to Prisma ORM, you can use the products of the [Prisma Data Platform](https://console.prisma.io) for different use cases:
+In addition to Prisma ORM, you can use the products of the [Prisma Console](https://console.prisma.io) for different use cases:
 
-- [**Prisma Accelerate**](/data-platform/accelerate/what-is-accelerate): A scalable connection pool and (optional) global database cache.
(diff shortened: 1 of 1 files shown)


petradonka:
@nikolasburk, this is a good start! I'm excited to have a central page to get started with any of our products.

Could you make some adjustments to better represent that our products are the Prisma ORM, Prisma Accelerate and Prisma Pulse? The Prisma Data Platform is just a loose container around some of these and not a product itself. 
For example, "Get started with Prisma Data Platform" shouldn't be a thing on it's own, but rather "Get started with Prisma Accelerate", "Get started with Prisma Pulse".
We should also use product blurbs that are better aligned with other parts of the website. For example the "(optional) global database cache" - while I get that you'd like to express that it's something configurable by the user and they can just use a connection pooler, it might be confusing or just inconsistent with the rest of our messaging.
These blurbs are pretty easy to adjust, so I'm more interested in seeing an iteration on not mentioning PDP much, and better representing the Accelerate and Pulse are the products :)

nikolasburk:
Thanks for the feedback, @petradonka, that's super helpful!

> Could you make some adjustments to better represent that our products are the Prisma ORM, Prisma Accelerate and Prisma Pulse? The Prisma Data Platform is just a loose container around some of these and not a product itself.
For example, "Get started with Prisma Data Platform" shouldn't be a thing on it's own, but rather "Get started with Prisma Accelerate", "Get started with Prisma Pulse".

Fair point! I've adjusted it to look as follows:

image

> We should also use product blurbs that are better aligned with other parts of the website. For example the "(optional) global database cache" - while I get that you'd like to express that it's something configurable by the user and they can just use a connection pooler, it might be confusing or just inconsistent with the rest of our messaging.

I removed the "optional" for the caching part of Accelerate. Were there other things you found that didn't align with the messaging we use for the products otherwise?

petradonka (approved):
Looks good, let's get this out there! 🙌

PR #5473, opened 2023-11-18
Offer clear field @map example

olivierlacan:
There's no clear field-mapping example in the [Prisma schema reference documentation](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#model-fields).

Users have to find their way to the [Using custom model and field names](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/use-custom-model-and-field-names#using-map-and-map-to-rename-fields-and-models-in-the-prisma-client-api) to find a clear example. 

I added a direct link to this section in the note itself, to make it more discoverable.


nikolasburk:
That's a helpful change, thanks a lot for taking the time to improve our documentation @olivierlacan 💚

PR #5476, opened 2023-11-20
[TEST] Update config.ts

nikolasburk:
just to test what happens to the sidenav when the top-level nav item is removed


nilubava:
@nikolasburk check out this PR: https://github.com/prisma/docs/pull/5484

PR #5479, opened 2023-11-21
Update 03-supported-databases.mdx

nikolasburk:

### DIFF
--- a/content/400-reference/300-database-reference/03-supported-databases.mdx
+++ b/content/400-reference/300-database-reference/03-supported-databases.mdx
@@ -6,9 +6,13 @@ metaDescription: 'This page lists all the databases and their versions that are 
 
 <TopBlock>
 
-Prisma currently supports the following databases.
+Prisma currently supports the following databases. 
+
+> See also: [System requirements](/reference/system-requirements).
 
 </TopBlock>
+
+An asterisk (\*) indicates that the version number is not relevant; either all versions are supported, there is not a public version number, etc.
 
 ## Self-hosted databases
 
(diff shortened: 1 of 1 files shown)


content/400-reference/300-database-reference/03-supported-databases.mdx
```diff
@@ -43,11 +45,11 @@ Prisma currently supports the following databases.
...
 
-Note that a fixed version of SQLite is shipped with every Prisma release.
-
-An asterisk (\*) indicates that the version number is not relevant; either all versions are supported, there is not a public version number, etc.
 
 &sup1; This does not include support for [Data API for Aurora Serverless](https://github.com/prisma/prisma/issues/1964).
 
+---
+
+An asterisk (\*) indicates that the version number is not relevant; either all versions are supported, there is not a public version number, etc.
```
janpio: Let's move this up to line 12, so it just comes before both headlines.

content/400-reference/300-database-reference/03-supported-databases.mdx
```diff
@@ -43,11 +45,11 @@ Prisma currently supports the following databases.
 | CockroachDB-as-a-Service     | \*      |
 | MongoDB Atlas                | \*      |
 
-Note that a fixed version of SQLite is shipped with every Prisma release.
-
-An asterisk (\*) indicates that the version number is not relevant; either all versions are supported, there is not a public version number, etc.
 
 &sup1; This does not include support for [Data API for Aurora Serverless](https://github.com/prisma/prisma/issues/1964).
 
+---
```
janpio: Seems this actually escapes out of the content area on rendering - if we ever want to use `---` we should get that fixed.

PR #5484, opened 2023-11-22
update: hidden prop on nav item

### LINKED ISSUE #5477: Sidenav breaks when top-level item gets hidden
See this PR here: https://github.com/prisma/docs/pull/5476

nilubava:
#5477 

Add prop `hidden: true` to hide any top nav item

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -45,7 +45,7 @@ const siteConfig = {
         type: 'bucket',
         bucketName: '/800-data-platform',
       },
-      { text: 'About', to: '/about', type: 'bucket', bucketName: '/600-about' },
+      { text: 'About', to: '/about', type: 'bucket', bucketName: '/600-about', hidden: true },
     ],
     search: {
       indexName: process.env.GATSBY_ALGOLIA_INDEX_NAME,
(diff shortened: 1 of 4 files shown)


config.ts
```diff
@@ -45,7 +45,7 @@ const siteConfig = {
         type: 'bucket',
         bucketName: '/800-data-platform',
       },
-      { text: 'About', to: '/about', type: 'bucket', bucketName: '/600-about' },
+      { text: 'About', to: '/about', type: 'bucket', bucketName: '/600-about', hidden: true },
```
nikolasburk, line 48: this was only for testing, right? for now I think we should keep the About tab.

```suggestion
```

nikolasburk (commented):
@nilubava this looks great, thanks a lot! 

I have one suggestion to remove the `hidden` flag from the `About` page because we want to keep it in the top-nav (for now).

config.ts
nikolasburk:
```suggestion
      { text: 'About', to: '/about', type: 'bucket', bucketName: '/600-about' },
```

nikolasburk (approved):

PR #5500, opened 2023-11-24
Document P6010 Accelerate error code

marcoieni:
We are in the process of adding this error code to accelerate. Let's merge this when we ship this change 👍 
Also, let me know if you like the wording :)


marcoieni:
This error is now returned by Accelerate in production 👍

jharrell:
@MarcoIeni looks like this PR was never merged. Should error code `p6010` still be documented? I can take this on if so.

marcoieni:
yes, please

PR #5529, opened 2023-12-08
remove concepts sidenav item from accelerate

nikolasburk:

### DIFF
--- a/content/300-accelerate/250-connection-pooling.mdx
+++ b/content/300-accelerate/250-connection-pooling.mdx
@@ -1,9 +1,7 @@
 ---
 title: 'Connection Pooling'
 metaTitle: 'Accelerate: Connection Pooling'
-metaDescription: 'Learn about everything you need to know to use Accelerate's connection pooling.'
-tocDepth: 3
-toc: true
+metaDescription: "Learn about everything you need to know to use Accelerate's connection pooling."
 ---
 
 ## Connection pooling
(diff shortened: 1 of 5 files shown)


content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -0,0 +1,63 @@
...
+title: 'Connection Pooling'
+metaTitle: 'Accelerate: Connection Pooling'
+metaDescription: "Learn about everything you need to know to use Accelerate's connection pooling."
+---
+
+## Connection pooling
+
+A [connection pool](https://en.wikipedia.org/wiki/Connection_pool#:~:text=In%20software%20engineering%2C%20a%20connection,executing%20commands%20on%20a%20database.) is a storage of database connections that can be reused for future requests to the database. When a new connection is requested, it is retrieved from the pool if one is available. Once the connection is no longer needed, it is returned to the pool for reuse.
+
+Connection pooling is important as it allows you to reuse existing connections instead of creating new ones, which can be expensive. It also allows you to limit the number of connections to the database, which can help prevent the database from being overloaded.
```
ankur-arch:
```suggestion
[Connection pooling](https://www.prisma.io/dataguide/database-tools/connection-pooling) is important as it allows you to reuse existing connections instead of creating new ones, which can be an expensive operation. 

The efficient management of database connections allows the database to process more queries without exhausting the available database connections, making your application more scalable. 
```

Limiting the number of database connections is natively built on to relational DBMS's, so I think it would be nicer to keep the focus on the benefits of connection pooling. Plus added a small link about connection pooling on our data-guide.

nikolasburk: Nice, thanks! (FWIW, I didn't change any of the actual text, just split the existing content into separate files 🙏 )

ankur-arch (approved):

PR #5532, opened 2023-12-11
Sort tables alphabetically and add Neon to managed list

michrome:
For ease of reading, sorted tables alphabetically. Added Neon to managed list.


jharrell (approved):
Thank you for the contribution! Confirming this passes our preview deployment tests and then will merge.

PR #5549, opened 2023-12-15
(upgrade guides) yarn upgrade alias -> yarn up

dearlordylord:
## Describe this PR

`yarn up` will work with both yarn 1 and yarn 2.

## Changes

yarn upgrade -> yarn up

## What issue does this fix?

guides doesn't work for yarn 2

##

opened instead of 3-week older PR against your older codebase that went out of sync


jharrell:
@Firfi just confirming: 

`yarn upgrade` does not work w/ Yarn 2 (berry) but `yarn up` works for both Yarn 1 and 2?

dearlordylord:
> @Firfi just confirming:
> 
> `yarn upgrade` does not work w/ Yarn 2 (berry) but `yarn up` works for both Yarn 1 and 2?

@jharrell correct

jharrell (approved):

jharrell:
Thank you very much for your contribution, @Firfi !

PR #5561, opened 2024-01-04
correct information about default connection pool timeout

### LINKED ISSUE #5535: Documentation about connection pool timeout misleading
> The default connection pool timeout is 10 seconds. If the query engine does not process a particular query within that time, it throws an exception and moves on to the next query in the queue.

This is incorrect. The connection pool timeout includes fetching a connection from the pool, not the processing of the query.

Here is a more correct statement:

> The default connection pool timeout is 10 seconds. If the query engine does not get a connection from the connection pool within that time, it throws an exception and moves on to the next query in the queue.

jharrell:


content/200-orm/200-prisma-client/000-setup-and-configuration/050-databases-connections/115-connection-pool.mdx
```diff
@@ -176,7 +176,7 @@ For more details on what is available in the metrics output, see the [About metr
 
 #### Default pool timeout
 
-The default connection pool timeout is 10 seconds. If the query engine does not process a particular query within that time, it throws an exception and moves on to the next query in the queue.
+The default connection pool timeout is 10 seconds. If the query engine does not get a connection from the connection pool within that time, it throws an exception and moves on to the next query in the queue.
```
ruheni:
```suggestion
The default connection pool timeout is 10 seconds. If the Query Engine does not get a connection from the database connection pool within that time, it throws an exception and moves on to the next query in the queue.
```
I presume it's from the database's connection pool, correct? If not, feel free to ignore my suggestion. 😄

ruheni (approved):

mhessdev (approved):

PR #5563, opened 2024-01-05
add versioninfo when relevant

jharrell:
This is for our effort to understand the locations where we have version specific info.


content/200-orm/050-overview/500-databases/850-planetscale.mdx
```diff
@@ -65,8 +65,12 @@ If you try to push to a production branch, you will get the [error message](/orm
 
 ## How to emulate relations in Prisma Client
 
+<versioninfo>
+
 PlanetScale does not allow foreign keys in its database schema. By default, Prisma uses foreign keys in the underlying database to enforce relations between fields in your Prisma schema. In Prisma versions 3.1.1 and later, you can [emulate relations in Prisma Client with the `prisma` relation mode](/orm/prisma-schema/data-model/relations/relation-mode#emulate-relations-in-prisma-with-the-prisma-relation-mode), which avoids the need for foreign keys in the database.
 
+</versioninfo>
```
janpio:
```suggestion
PlanetScale does not allow foreign keys in its database schema. By default, Prisma uses foreign keys in the underlying database to enforce relations between fields in your Prisma schema. <versioninfo>In Prisma versions 3.1.1 and later, you can [emulate relations in Prisma Client with the `prisma` relation mode](/orm/prisma-schema/data-model/relations/relation-mode#emulate-relations-in-prisma-with-the-prisma-relation-mode), which avoids the need for foreign keys in the database.</versioninfo>
```
Would something like this work with the element as well?

content/200-orm/050-overview/500-databases/850-planetscale.mdx
```diff
@@ -79,12 +83,20 @@ datasource db {
...
+
 The ability to set the relation mode was introduced as part of the `referentialIntegrity` preview feature in Prisma version 3.1.1, and is generally available in Prisma versions 4.8.0 and later.<br /><br />The `relationMode` field was renamed in Prisma version 4.5.0, and was previously named `referentialIntegrity`.
 
+</versioninfo>
+
 </Admonition>
 
+<versioninfo>
+
 If you use relations in your Prisma schema with the default `"foreignKeys"` option for the `referentialIntegrity` field, PlanetScale will error when Prisma tries to create foreign keys. In versions 2.27.0 and later, Prisma will output the [P3021 error message](/orm/reference/error-reference#p3021).
```
janpio: OT: Is that actually correct with `refentialIntegrity` or should it be `relationMode` in this sentence?

content/200-orm/100-prisma-schema/20-data-model/10-models.mdx
```diff
@@ -425,7 +425,11 @@ Refer to the [relations documentation](relations) for more examples and informat
 
 ### Native types mapping
 
-Version [2.17.0](https://github.com/prisma/prisma/releases/tag/2.17.0) and later support **native database type attributes** (type attributes) that describe the underlying database type:
+<versioninfo>
+  Version [2.17.0](https://github.com/prisma/prisma/releases/tag/2.17.0) and
+  later support **native database type attributes** (type attributes) that
+  describe the underlying database type:
+</versioninfo>
```
janpio, line 432: This is a good example where I think the docs would benefit from rewriting this in a way without mentioning the version information, and than add that information in parens after the sentence: "(This is available since version 2.17.0)".

janpio:
Seems usage of the new element breaks some Markdown:

image
via https://app.lost-pixel.com/app/repos/7/clb5ek3mm1772001qqg7yban38?build=clr16ymc5apx1mh0e6fkx85j5&diff=clr1723f802j6cx0eak397eo0

image
via https://app.lost-pixel.com/app/repos/7/clb5ek3mm1772001qqg7yban38?build=clr16ymc5apx1mh0e6fkx85j5&diff=clr1723f802ixcx0eb2m8zfyh

It also seems to mess with line height in some places:
image
https://app.lost-pixel.com/app/repos/7/clb5ek3mm1772001qqg7yban38?build=clr16ymc5apx1mh0e6fkx85j5&diff=clr1723fa02k3cx0eyq1ankor

content/200-orm/050-overview/500-databases/850-planetscale.mdx
jharrell: Yup, you're right. This should be `relationMode`.

jharrell:
@janpio yup, it looks like back ticks don't work, but `<inlinecode></inlinecode>` works. I'll start updating.

janpio:
@jharrell Is that maybe fixable? Because that means a _lot_ of changes that make the content less portable between inside `<versioninfo>` and outside `<versioninfo>` :/ Should we wait until this is fixed maybe?

Issue #5507, opened 2023-11-27
Share a sidenav block across several top-level items


nikolasburk:
For the [docs restructuring](https://github.com/prisma/docs/pull/5501), we'll have the following top-level nav items:

- **Get Started**
- **ORM**
- **Accelerate**
- **Pulse**
- **Platform** (hidden)
- **About** (hidden)

We want to share the sidenav content from the **Platform** and duplicate it under **Accelerate** and **Pulse**. So for example, if someone clicks on **Accelerate**, they should first see the "native" sidenav content from the top-level **Accelerate** item. And below that, we want to duplicate the sidenav content from **Platform**.

PR #5569, opened 2024-01-10
share sidenav between buckets

nilubava:
#5507 

Screenshot 2024-01-10 at 21 08 34

You can use `includeSidenav: '/sidenav-bucket-name'` to include another section as sidenav


janpio (commented):
This is not really frontmatter, but independent sidebar navigation?

nikolasburk:
I had a strange issue where the **Local development** page under [**Accelerate**](https://docs-git-feat-share-sidenav-prisma.vercel.app/accelerate) was shown as a heading in the sidenav as well. This happened only on first load though and a refresh fixed it.

image

I wasn't able to reproduce it afterwards so I _guess_ it's fine?

nikolasburk:
Otherwise this generally looks good although the heading structure is a bit weird. This might be better after the [**LEGACY PLATFORM** docs have been deleted](https://prisma-company.slack.com/archives/C5Z9TH6N9/p1704963510000019). So let's wait until the deletion is confirmed and executed before merging this.

PR #5576, opened 2024-01-16
Update 400-api-reference.mdx

nikolasburk:

### DIFF
--- a/content/300-accelerate/400-api-reference.mdx
+++ b/content/300-accelerate/400-api-reference.mdx
@@ -116,6 +116,6 @@ interface AccelerateInfo {
 
 ## Errors
 
-Accelerate-related errors start with `P6xxx`.
+Prisma Accelerate-related errors start with `P6xxx`.
 
 You can find the full error code reference for Prisma Accelerate [here](/orm/reference/error-reference#accelerate).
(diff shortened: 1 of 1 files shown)


content/300-accelerate/400-api-reference.mdx
```diff
@@ -113,3 +113,9 @@ interface AccelerateInfo {
...
 | `signature`    | `String`                                                     | The unique signature of the Prisma operation.                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
+
+## Errors
+
+Accelerate-related errors start with `P6xxx`.
```
nikolasburk: this line is quoted from the [actual error reference](https://www.prisma.io/docs/orm/reference/error-reference#accelerate):

image

ankur-arch: LGTM 👍 , maybe both places should read `Prisma Accelerate` rather than Accelerate.

ankur-arch:
```suggestion
Prisma Accelerate-related errors start with `P6xxx`.
```

nikolasburk: I don't like how this looks but ChatGPT says it's the way to go 😂 

image

nikolasburk: changed "Accelerate" to "Prisma Accelerate" on the other page as well: https://github.com/prisma/docs/pull/5577

PR #5587, opened 2024-01-19
patch: clear mentions of cloud projects

ankur-arch:


content/300-accelerate/600-faq.mdx
```diff
@@ -64,7 +64,7 @@ Under the hood, Accelerate uses Cloudflare, which uses [Anycast](https://www.clo
 
 ## How can I invalidate a cache on Accelerate?
 
-You can invalidate your cache on a project level up to five times a day. This can be done via the Cloud Project's Accelerate configuration page.
+You can invalidate your cache on a project level up to five times a day. This can be done via the Platform Console's Accelerate configuration page.
```
nikolasburk: I wonder if we even need "Platform Console's" here:

```suggestion
You can invalidate your cache on a project level up to five times a day. This can be done via the Accelerate configuration page.
```

content/400-pulse/200-getting-started.mdx
```diff
@@ -135,13 +135,13 @@ In a project using [Prisma Client](/orm/prisma-client), run the following comman
...
-### Store your Cloud Projects API key in your .env file
+### Store your Pulse API key in your .env file
 
-The Pulse extension requires you to use a Cloud Projects API key.
+The Pulse extension requires you to use an API key.
 
 <Admonition type="info">
 
-You should have received an API key from the Cloud Projects dashboard when setting up your project.
+You should have received an API key from the Console Platform dashboard when setting up your project.
```
nikolasburk: would this be more accurate maybe?

```suggestion
You should have received an API key when you added Prisma Pulse to your project in the Platform Console.
```

also, I think we should add a note in step 2. that an API key will be created at the end.

nikolasburk:
generally looks good to me! I left two small suggestions that you can consider 🙏

content/300-accelerate/600-faq.mdx
ankur-arch: I like the change on this one 🙌 !

content/400-pulse/200-getting-started.mdx
ankur-arch: @nikolasburk, I added a note for that in the latest commit.

nikolasburk (approved):
thanks for incorporation my suggestions, this looks good to me know 👍

mhessdev:
I would suggest updating everything to match the branding. So instead of `Accelerate` and `Pulse` it's Prisma Accelerate and `Prisma Pulse.` Otherwise, it looks good!

mhessdev (changes requested):
Use proper branding, Prisma Accelerate and Prisma Pulse

ankur-arch:
@mhessdev, thanks for the review 🙌 !

For your suggestion, I think it makes sense for me to handle that in a separate PR.

> So instead of Accelerate and Pulse it's Prisma Accelerate and Prisma Pulse. Otherwise, it looks good!

There are plenty of mentions of Pulse and Accelerate. Doing a global search in Vscode I can see that there are:
- Approx 163 mentions of Pulse in 19 files
- Approx 370 mentions of Accelerate in 30 files.

Additionally, this PR is just intended to clear out the mentions of Cloud Projects, so I wanted to keep the changes specific to the issue title.  
WDYT?

mhessdev (approved):

PR #5612, opened 2024-02-02
docs: Make DIRECT_URL naming consistent with the previous pattern

kof:
## Describe this PR

Current naming makes a mess in any env, because DIRECT_URL doesn't tell what that is, at the very least it has to stay consistent with the previous pattern, in ideal case both will be renamed because they are semantically wrong.

## Changes

More consistent env variable naming

## What issue does this fix?

Now people are copying bad naming and getting "DIRECT_URL" as a global env variable which says nothing about its usage and isn't consistent with Prisma's other url


jharrell:
Hi there! Thank you very much for the contribution 🙏 

Unfortunately, I don't think we'll be able to include this. While you're correct, `DATABASE_DIRECT_URL` is a more clear variable, we have several years of content pointing to the current setup of `DATABASE_URL` and `DIRECT_URL`. 

Again, thank you very much, and I'll also bring notes back to our team to consider this naming change in the future.

PR #5613, opened 2024-02-05
Custom output with `driverAdapters`

### LINKED ISSUE #5609: Add 5.9.0+ specific instructions for using custom output with `driverAdapters`
In 5.9.0, we have a few changes under the hood that cause issues when using the driverAdapters preview feature with a client with a custom output location. 

In these cases, it's important to follow the following:
1. _don't_ use a relative import (like `../prisma/client` or `./client`)
2. _do_ use a linked dependency using your dependency manager of choice

```
Once you know where your client is located, add your own client as a dependency.
  - For npm, run `npm add db@./path/to/your/client`
  - For pnpm, run `pnpm add db@link:./path/to/your/client`
  - For yarn, run `yarn add db@link:./path/to/your/client`
```

3. If using typescript, make sure to set [module](https://www.typescriptlang.org/tsconfig#module) and [moduleResolution](https://www.typescriptlang.org/tsconfig#moduleResolution) to `nodenext` and `nodenext` or `bundler`, respectively.

4. Now import from the custom dependency above! `import { PrismaClient } from "db"`

jharrell:
Fixes #5609

## Description

Added a few changes for a specific 5.9.0+ interaction between custom output paths and the driver adapters Preview feature.

### DIFF
--- a/content/200-orm/050-overview/500-databases/200-database-drivers.mdx
+++ b/content/200-orm/050-overview/500-databases/200-database-drivers.mdx
@@ -50,7 +50,7 @@ You can also build your own driver adapter for the database you're using. The fo
 
 To use this feature:
 
-1. Update the `previewFeatures` block in your schema to include the the `driverAdapters` preview feature:
+1. Update the `previewFeatures` block in your schema to include the the `driverAdapters` Preview feature:
 
    ```prisma
    generator client {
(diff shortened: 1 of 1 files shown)


content/200-orm/050-overview/500-databases/200-database-drivers.mdx
```diff
@@ -70,3 +70,63 @@ To use this feature:
...
+```
+
+Instead, you will need to use a linked dependency.
+
+<TabbedContent tabs={[<FileWithIcon text="npm" icon="file"/>, <FileWithIcon text="pnpm" icon="file"/>, <FileWithIcon text="yarn" icon="file"/>]}>
+
+<tab>
+
+```terminal
+npm add db@./src/generated/client
```
janpio, line 103: What does this really do? Does this add this path as an alias `db`?

jharrell: Yes, this adds the generated client as a dependency with the alias `db`

janpio: I think that would be good to add to the description. I assume I am not alone in not having used that before, hence not being super clear what it does.

PR #5615, opened 2024-02-06
Cross-reference nested upsert and upsert()

### LINKED ISSUE #5614: Clarify and cross-reference upsert documentation
Jan's message:
"Searching for upsert in our docs has this as #1 result for me right now: https://www.prisma.io/docs/orm/reference/prisma-client-reference#upsert-1
This is actually only the "nested" upsert that can be used inside of update().

The real upsert() is documented here: https://www.prisma.io/docs/orm/reference/prisma-client-reference#upsert
I assume I am not the only one that finds that a bit confusing? Can we add some clarifications and cross links to make all this more discoverable?

(There is also different kind of upsert documentation here that is #1 on Google for prisma upsert: https://www.prisma.io/docs/orm/prisma-client/queries/crud#update-or-create-records. It links to the reference, but is missing the typical "API" bubble we usually have for those kinds of links I think.)"

grpatel:
Fixes #5614 

- [ ] Cross reference two upsert sections

### DIFF
--- a/content/200-orm/500-reference/050-prisma-client-reference.mdx
+++ b/content/200-orm/500-reference/050-prisma-client-reference.mdx
@@ -896,7 +896,7 @@ const user = await prisma.user.update({
 ```
 
 ### <inlinecode>upsert()</inlinecode>
-
+*This section covers the usage of the general upsert() operation. To learn about using nested upsert queries within update(), reference the documentation [here](#upsert-1).*
 `upsert` does the following:
 
 - If an existing database record satisfies the `where` condition, it updates that record
@@ -3149,7 +3149,7 @@ const user = await prisma.user.update({
 ### <inlinecode>upsert</inlinecode>
 
 <AlgoliaTerm name="apiReference" value="upsert" />
-
+*This section covers the usage of nested upsert within update(). To learn about the [`upsert()`](#upsert) operation, reference the linked documentation.*
 A nested `upsert` query updates a related record if it exists, or creates a new related record.
 
 #### Examples
--- a/content/200-orm/500-reference/050-prisma-client-reference.mdx
+++ b/content/200-orm/500-reference/050-prisma-client-reference.mdx
@@ -902,6 +902,7 @@ const user = await prisma.user.update({
 This section covers the usage of the `upsert()` operation. To learn about using [nested upsert queries](#upsert-1) within `update()`, reference the linked documentation.
 
 </Admonition>
+
 `upsert` does the following:
 
 - If an existing database record satisfies the `where` condition, it updates that record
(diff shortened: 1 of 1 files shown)


content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -896,7 +896,7 @@ const user = await prisma.user.update({
 ```
 
 ### <inlinecode>upsert()</inlinecode>
-
+*This section covers the usage of the general upsert() operation. To learn about using nested upsert queries within update(), reference the documentation [here](#upsert-1).*
```
```suggestion

*This section covers the usage of the general `upsert()` operation. To learn about using nested upsert queries within `update()`, reference the documentation [here](#upsert-1).*

```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -3149,7 +3149,7 @@ const user = await prisma.user.update({
 ### <inlinecode>upsert</inlinecode>
 
 <AlgoliaTerm name="apiReference" value="upsert" />
-
+*This section covers the usage of nested upsert within update(). To learn about the [`upsert()`](#upsert) operation, reference the linked documentation.*
```
```suggestion

*This section covers the usage of nested upsert within `update()`. To learn about the [`upsert()`](#upsert) operation, reference the linked documentation.*

```

content/200-orm/500-reference/050-prisma-client-reference.mdx
janpio: Can we find a better link text than "here"? I think the docs have a styleguide that includes not using "here" and similar (or had in the past).

grpatel:
```suggestion

*This section covers the usage of the general `upsert()` operation. To learn about using [nested upsert queries](#upsert-1) within `update()`, reference the linked documentation.*

```

Ya, I changed it to follow the same format as the other. Does that work?

janpio: Definitely better for me, your colleagues hopefully know the _actual_ style guide and if that is ok with it.

janpio (dismissed):
(to prevent accidental merge)

We need to check the preview again after suggestions ahve been merged, right now it looks like this:
image

content/200-orm/500-reference/050-prisma-client-reference.mdx
grpatel:
```suggestion

<Admonition type="info">

This section covers the usage of the `upsert()` operation. To learn about using [nested upsert queries](#upsert-1) within `update()`, reference the linked documentation.

</Admonition>
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -3150,6 +3155,8 @@ const user = await prisma.user.update({
 
 <AlgoliaTerm name="apiReference" value="upsert" />
 
+*This section covers the usage of nested upsert within `update()`. To learn about the [`upsert()`](#upsert) operation, reference the linked documentation.*
```
jharrell: Let's wrap this in an `Admonition` as well

jharrell (approved):

PR #5658, opened 2024-02-20
add new errors

### LINKED ISSUE #5639: Some error codes are not documented
Some of the recently added error codes are missing from https://www.prisma.io/docs/orm/reference/error-reference#prisma-client-query-engine

`P2029` - Query parameter limit exceeded error: {message}.
`P2035` - Assertion violation on the database: `{database_error}`
`P2036` - Error in external connector  (id {id}) **Note**: this one should never technically be seen by users, but code exists anyway.

In 5.10, there will also be a new addition:
`P2037` - Too many database connections opened.

I've heard there is a script for getting those out of the codebase automatically.

nikolasburk:
closes https://github.com/prisma/docs/issues/5639

### DIFF
--- a/content/200-orm/500-reference/250-error-reference.mdx
+++ b/content/200-orm/500-reference/250-error-reference.mdx
@@ -300,6 +300,10 @@ Possible P1012 error messages:
 
 "Transaction API error: {error}"
 
+#### <inlinecode>P2029</inlinecode>
+
+"Query parameter limit exceeded error: {message}"
+
 #### <inlinecode>P2030</inlinecode>
 
 "Cannot find a fulltext index to use for the search, try adding a @@fulltext([Fields...]) to your schema"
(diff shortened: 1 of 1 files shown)


content/200-orm/500-reference/250-error-reference.mdx
```diff
@@ -316,6 +320,18 @@ Possible P1012 error messages:
...
+
+"Assertion violation on the database: {database_error}"
+
+#### <inlinecode>P2036</inlinecode>
+
+"Error in external connector (id {id})"
+
+#### <inlinecode>P2037</inlinecode>
+
+"Too many database connections opened"
```
janpio:
```suggestion
"Too many database connections opened: {message}"
```
A mistake like this should not be possible with the script. Are we not using this any more?

nikolasburk: I just copied over the descriptions from here: https://github.com/prisma/docs/issues/5639#issue-2136235549

About the script: I think I've heard that we have smth like this but I don't know where it is / how to use it. Do you have any details about it by any chance? (If not no worries, I'll do some research myself as well.)

Happy to update the PR with the actual output of the script once I figured out how to use it.

PR #5682, opened 2024-02-27
update mobile layout for navbars

carlagn:


src/css/custom.css
```diff
@@ -21,32 +21,68 @@
...
+    color: rgb(113, 128, 150) !important;
+  }
 }
 .menu__link--sublist {
   margin-bottom: 0;
 }
-.menu__link:hover {
-  color: rgb(26, 32, 44) !important;
+.menu__link:not(.navbar-login-btn):hover {
+  /* color: rgb(26, 32, 44) !important; */
```
jharrell: Is this meant to be commented?

(Same question for lines 83-85)

jharrell (approved):
Quick question on some commented out code.

Also, saw the SVGs were resized. Just wanted to confirm that was intentional.

src/css/custom.css
carlagn: Will delete the commented lines altogether.
The SVGs were updated per @Techboidesign's input

jharrell (approved):

PR #5683, opened 2024-02-27
Add disclaimer regarding usage of direct foreign key values

igorrocha:
## Describe this PR

The [examples section](https://www.prisma.io/docs/orm/reference/prisma-client-reference#examples-27) of the `connect` nested query page currently indicates that you can set foreign keys directly. 

However, when you try to use both the direct way of setting a foreign key and the `connect` query, you get a type error, due to the fact that it is unpractical to generate types for every combination of inputs using direct assignment and  `connect`, as described in [this issue comment](https://github.com/prisma/prisma/issues/4322#issuecomment-737976117).

Since this behavior is intentional, and the error you get when trying to combine both approaches isn't very clear, I realized a note on the documentation could prevent some people stumbling on this problem.

## Changes

Added a short disclaimer explaining that you can only use one of the two approaches in the same query (the sentence after the code example).

image

## What issue does this fix?

Not applicable.

I decided to link the main repository's relevant issue to the docs rather than explaining the reasoning behind the limitation, because I figured the disclaimer would get too long otherwise. Please feel free to improve on the disclaimer in case it isn't optimal as is.


content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -2875,6 +2875,8 @@ const user = await prisma.profile.create({
 })
 ```
 
+However, you can't use both the direct approach and the `connect` approach in the same query. See [this issue comment](https://github.com/prisma/prisma/issues/4322#issuecomment-737976117) for details.
```
janpio, line 3065: Internal note to my colleagues: Can we move the actual information into this sentence to avoid linking to an old issue? The statement from Harshit is still true as far as I know - but might be wort double checking to expand on it.

jharrell (approved):

PR #5685, opened 2024-02-28
Fixed inconsistency in documentation about .$queryRaw methods (updated 050-raw-queries.mdx)

nikelborm:
## Describe this PR

Fixed little inconsistency in documentation

## Changes

- added that you can't run `select 1; select 2` (any amount of queries not equal 1) with every raw query method (`$queryRaw`, `$executeRaw`, `$queryRawUnsafe`, `$executeRawUnsafe`) instead of only 2 (`$executeRaw`,`$executeRawUnsafe`)


jharrell:
Nice catch! Thank you for the contribution

janpio:
We should probably repeat this remark multiple times for each function instead of moving it up to the list in the introduction. Many more users will miss it completely this way, don't you think @jharrell?

PR #5688, opened 2024-02-29
feat: docusaurus homepage

carlagn:
## Describe this PR

Docusaurus homepage updated to the current live layout. 

## Changes

Update theming and the index page.

## What issue does this fix?

https://linear.app/prisma-company/issue/DA-65/docs-homepage

- [x] responsive
- [x] dark/light mode


config.ts
```diff
@@ -0,0 +1,180 @@
+const siteConfigData = {
```
jharrell, line 1: Is there a way we can do away with this file? I know it's a bit of extra work but doing something [like this](https://github.com/prisma/docs/blob/docusaurus/src/components/HomepageFeatures/index.tsx) feels a bit better to me.

config.ts
```diff
@@ -0,0 +1,180 @@
...
+    ],
+    CommunityLinksData: [
+      // {
+      //   id: 'slack',
+      //   title: 'Slack',
+      //   description: 'Engage with a vibrant community of developers and companies adopting Prisma',
+      //   icon: 'fa-brands fa-slack',
+      //   link: 'https://slack.prisma.io/',
+      //   linkText: 'Join our channel',
+      // },
```
jharrell, line 107:
```suggestion
```

src/components/icons/technologies/CockroachDB.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react'
```
jharrell, line 1: Is there a reason why we need these as TSX components rather than just svgs?

jharrell (approved):
If it works, it works, but I'd like to see a few changes. Happy to take those on myself to free you up.

config.ts
carlagn: To have content + logic on the same file? sure!

PR #5698, opened 2024-03-05
Prevent typescript not recommending actual prisma variable

Miigaarino:
### Best practice for instantiating Prisma Client with Next.js

By declaring global variable named prisma, typescript thinks it is an global variable and should not be imported from somewhere.

image

## Changes

Renaming global variable that contains the prisma object

## What issue does this fix?

Typescript now recommends actual prisma object separate to prismaGlobal variable we declared that is only needed to not instantiate prisma client again
image


jharrell:
Thank you for the fix and contribution!

PR #5702, opened 2024-03-06
fix(models): post field was not lowercase

ledenis:
## Describe this PR

`post` field was not lowercase. Confusing because `Post` is the model name.


jharrell:
Thank you very much for the contribution and fix 🙏

PR #5706, opened 2024-03-10
Add documentation for MariaDB users

ottok:
## Describe this PR

I have never used Prisma before, and it took me quite a lot of extra effort to figure out if Prisma supports MariaDB at all because the documentation was only mentioning MySQL. After some digging, I found that indeed Prisma uses both https://www.npmjs.com/package/mariadb for the connector and https://hub.docker.com/_/mariadb as part of the CI, and MariaDB support is actually excellent.

[Sequalize has very clear MariaDB documentation](https://sequelize.org/docs/v7/other-topics/dialect-specific-things/#mariadb), and if you merge this PR then Prisma will be on-par.

## Changes

Each commit is [atomic](https://optimizedbyotto.com/post/good-git-commit/#1-commits-should-be-atomic) and carefully documented - please review this PR by looking commit-by-commit looking at the message and the change at the same time. 

Looking at [git log in 500-databases](https://github.com/prisma/docs/commits/main/content/200-orm/050-overview/500-databases) and e.g. 4358a4169ef3ff042d70b825a9cf3e16bf71e48e seems you prefer to squash when merging. Please merge this PR as-is **without** squashing to avoid loosing the per-commit documentation.

I am also happy to iterate on the contents of this PR based on your feedback. I am also happy to rebase frequently.

### DIFF
--- a/content/200-orm/050-overview/500-databases/306-mariadb.mdx
+++ b/content/200-orm/050-overview/500-databases/306-mariadb.mdx
@@ -1,0 +1,206 @@
+---
+title: 'MySQL'
+metaTitle: 'MySQL database connector'
+metaDescription: 'This page explains how Prisma can connect to a MySQL database using the MySQL database connector.'
+tocDepth: 3
+---
+
+<TopBlock>
+
+The MySQL data source connector connects Prisma ORM to a [MySQL](https://www.mysql.com/) database server.
+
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
I am not sure this is the best way to mention MariaDB in our docs, so I am using "Request changes" as a way to prevent an accidental merge.

We should find a more lightweight way to make sure our users can know that for us MariaDB and MySQL are the same.

ottok:
Hi!

Thanks for reply. If you don't want a new page, can you share if you can
accept the other commits in this PR so I can prepare them for merge?

ottok:
Closing this as #5713 was merged.

PR #5707, opened 2024-03-11
Remove unnecessary spacing

oscarmorrison:
## Describe this PR

Ref: https://hackmd.io/@jenc/H1bgodhlo#Horizontal-Alignment

## Changes

This PR proposes a minor but essential change in the codebase. 

## What issue does this fix?

This PR proposes a minor but essential change in the codebase. 

The primary motivation behind this change is to improve the readability and maintainability of our code. While aligning the equals signs (=) might seem more visually appealing, it is not a recommended practice. 

Here's why:

1. **Inconsistency**: When new lines are added, all existing lines need to be realigned, which can lead to inconsistencies if not done carefully. This change will help avoid such situations.

2. **Difficult to manage**: This kind of formatting becomes difficult to manage, especially when we have lines of different lengths. It also becomes troublesome when we have to add or remove a variable with a longer name, as we need to adjust all other lines accordingly.

3. **Poor readability**: While it may seem that aligning the equals sign enhances readability, it can actually make the code harder to read. Our eyes are naturally good at scanning vertically, not horizontally. Hence, it's easier to read code when it's formatted with a simple key-value pair on each line.

4. **Version Control Difficulties**: In a version control system like Git, this kind of formatting can lead to unnecessary line changes, making it harder to track actual changes in the code.

By implementing this change, we aim to make our codebase more consistent and easier to manage, leading to more efficient development in the long run.

### DIFF
--- a/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
+++ b/content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/100-connect-your-database.mdx
@@ -16,7 +16,7 @@ To connect your database, you need to set the `url` field of the `datasource` bl
 ```prisma file=prisma/schema.prisma
 datasource db {
   provider = "postgresql"
-  url      = env("DATABASE_URL")
+  url = env("DATABASE_URL")
 }
 ```
 
@@ -70,7 +70,7 @@ DATABASE_URL="postgresql://janedoe:janedoe@localhost:5432/janedoe?schema=hello-p
 ```prisma file=prisma/schema.prisma
 datasource db {
   provider = "postgresql"
-  url      = env("DATABASE_URL")
+  url = env("DATABASE_URL")
 }
 ```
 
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
(Using "Request changes" as "blocking merge")

Unfortunately that is not how Prisma works, using `prisma format` will format the files with horizontal spacing.

oscarmorrison:
> (Using "Request changes" as "blocking merge")
> 
> Unfortunately that is not how Prisma works, using `prisma format` will format the files with horizontal spacing.

any reason why? really a bit of a formatting anti-pattern here. Really enjoy prisma, but seeing this throughout the docs is a bit frustrating

janpio:
The docs reflect how the tool works.
And the tools works like this because 5+ years ago someone made the decision that Prisma's formatting rules should use horizontal spacing. Changing that now would be a major breaking change that would cause hundreds of thousands of Prisma schemas to be modified after upgrading to a newer Prisma version, so that is not a very attractive option.

I don't think today we would make the same decision as back in the day, but looking at the big picture this is fortunately not that big of a problem and our users can usually find a way to live with it - or just reformat their Prisma schema files manually.

jharrell:
Thank you for the PR! Unfortunately, as @janpio said, this is baked into our schema linting and formatting. This would be better served as a feature request in https://github.com/prisma/prisma

PR #5729, opened 2024-03-16
Update 250-should-you-use-prisma.mdx

aqiank:
Update link for Make the Right Thing The Easy Thing slides

## Describe this PR

It seems like the domain is no longer available for the Make the Right Thing The Easy Thing slide. This updates it to one that is still available.

## Changes

- Update Make the Right Thing The Easy Thing slides link

## What issue does this fix?

It fixes an invalid link on the page.

I used the link put inside the slides but there's also this link https://jlengstorf.github.io/presentations/make-the-right-thing-the-easy-thing/


jharrell (approved):
bummer about that link. Thank you for finding a working one!

PR #5731, opened 2024-03-19
feat: add troubleshooting guide for accelerate

ankur-arch:
Adds troubleshooting guide for P6004, P6008, and P6009 errors. 
Adds references to the troubleshooting guide from the limitations and the error page.


content/300-accelerate/650-troubleshoot.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+### Possible causes for <inlinecode>[P6009](/orm/reference/error-reference#p6009-responsesizelimitexceeded)</inlinecode>
+
+#### Transmitting images/files in response
+
+This error may arise if images or files stored in base64 format within your table are being fetched, resulting in a large response size. Storing assets in base64 format is generally discouraged because it significantly impacts database performance and scalability. In addition to performance, it makes database backups slow and significantly increases the cost of storing routine backups.
+
+**Suggested solution:** Store the image/file in a BLOB store like Cloudflare [R2](https://developers.cloudflare.com/r2/), [AWS S3](https://aws.amazon.com/pm/serv-s3/), [Cloudinary](https://cloudinary.com/), etc. These services would allow you to store the assets optimally and return a URL through which you should be able to access the asset. You can store this URL in your table instead of storing the asset in a base64 format. This would substantially reduce the size of your response.
```
mavilein: I would remove the mention of base64 encoding. Files can either be stored as raw bytes or something like base64. This error would also be returned if the file is stored as raw bytes. If you remove the mention of base64 and only talk about storing files in the database this will read better imo.

content/300-accelerate/650-troubleshoot.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+#### Incorrect/Missing <inlinecode>[where](/orm/reference/prisma-client-reference#where)</inlinecode> or <inlinecode>[select](/orm/reference/prisma-client-reference#select)</inlinecode> clause
```
mavilein: Everything that is said makes sense. It might be more readable to break out the where and select into separate paragraphs. So one for for `missing select` and one for `incorrect where`. It reads a bit confusing since this block talks about two issues at once. If you want to group those together I would add a headline a la `Overfetching of data`.

content/300-accelerate/650-troubleshoot.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+**Suggested solution:** Double-check if the `where` clause is filtering the data as you expect. Consider using [pagination](/orm/prisma-client/queries/pagination) to prevent a large number of records from being fetched. Using the [select](/orm/reference/prisma-client-reference#select) clause to return only the fields that are being consumed can also be helpful in reducing the response size.
+
+#### Fetching a large volume of data
+
+In many data processing workflows, especially those involving ETL (Extract-Transform-Load) processes or scheduled CRON jobs, there's a need to extract large amounts of data from data sources (like databases, APIs, or file systems) for analysis, reporting, or further processing. If you are running an ETL/CRON workload that fetches a huge chunk of data for analytical processing then you might run into this limit.
+
+**Suggested solution:** Consider splitting your query to fetch data in batches to prevent the 5MB limit from being reached. By splitting your query to retrieve data in batches, you ensure that each operation fetches only a portion of the total data volume, thereby not exceeding the size limit for a single fetch operation.
```
mavilein, line 37: This is the same point as the previous one related to pagination, right?

content/300-accelerate/650-troubleshoot.mdx
```diff
@@ -0,0 +1,124 @@
...
+This limit should align with your database's maximum number of connections.
+
+By default, the connection limit is set to 10 unless a
+different `connection_limit` is specified in your database connection string.
+
+#### Long-running queries
+
+Queries may be slow to respond, hitting the 10-second timeout even when connections are available. This could happen if a very large amount of data is being fetched in a single query or if appropriate indexes are missing from the table.
+
+**Suggested solution**: Identify the slow-running queries and fetch only the required data. If you only need to use specific fields from the table, then you can use the select clause in your queries to prevent unnecessary data from being fetched. Additionally, you could consider adding appropriate indexes based on the query to ensure that data is fetched efficiently.
```
mavilein, line 63: You could also suggest to isolate long running queries into their own Environment. This way long running queries can't affect the transactional queries.

content/300-accelerate/650-troubleshoot.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+**Suggestion solution**: If normally quick queries are intermittently slow or timing out without any modifications to them, it's probable that competing queries are exerting pressure on the same database tables. To diagnose this, adopt monitoring tools or leverage your database's inherent capabilities to observe reads, writes, and wait times. Such monitoring will unveil activity patterns or spikes that align with the observed performance dips.
+
+Moreover, it's crucial to periodically scrutinize and refine essential queries and verify that tables are properly indexed. This proactive approach minimizes the vulnerability of these queries to slowdowns caused by competing workloads.
+
+### Considerations for <inlinecode>[P6009](/orm/reference/error-reference#p6009-responsesizelimitexceeded)</inlinecode> and <inlinecode>[P6004](/orm/reference/error-reference#p6004-querytimeout)</inlinecode> errors
+
+For runtimes that support Prisma ORM natively, you could consider creating two `PrismaClient` Instances. One with the Accelerate connection string (prefixed with `prisma://`) and the other one with the direct database connection string (prefixed with `postgres://`, `mysql://` etc)
```
mavilein: I think it would be worth to highlight the main idea of this approach: Bypass Accelerate for certain queries. For the reader it might not be clear that this is the idea.

content/300-accelerate/650-troubleshoot.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+However, please note that the available connections would be split between both of your `PrismaClient` Instances. It's crucial to understand the implications of managing multiple instances, particularly in regards to direct database connections. Utilizing a `PrismaClient` instance with a direct database connection string means that this connection will interact directly with your database.
+
+This approach requires careful consideration because the direct connections and those managed by Accelerate share the same underlying database connection pool. This can lead to competition for resources, potentially affecting the performance and availability of your database services.
+
+Additionally, direct connections could have a significant impact on your database's performance and availability. Operations that consume a considerable amount of resources could potentially degrade the service for other users or processes that rely on the same database.
```
mavilein, line 84: I think this could emphasize that users really need to know what they are doing. This approach bypasses the safeguards provided by Accelerate. One really has to know what they are doing in order to warrant this approach. And most often those that are aware of this will find a better solution anyway.

PR #5741, opened 2024-03-21
feat: remove "new" tag from platform environments

ankur-arch:


content/500-platform/20-concepts/30-environments.mdx
```diff
@@ -4,7 +4,6 @@ metaTitle: 'Platform Environments'
 metaDescription: 'Learn about Environments in Platform Console.'
 tocDepth: 3
 toc: true
```
ankur-arch, line 6:
```suggestion
toc: true
highlight: false
```

PR #5751, opened 2024-03-26
remove feedback component

meletj:
removing the visual feedback component (was this page helpful?)
This PR only removes the visual component taking into account the full widget will be removed in Docusaurus migration.

### DIFF
--- a/src/components/pageBottom.tsx
+++ b/src/components/pageBottom.tsx
@@ -131,58 +131,6 @@ const PageBottom = ({ editDocsPath }: any) => {
 
   return (
     <PageBottomWrapper>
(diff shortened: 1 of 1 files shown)


src/components/pageBottom.tsx
```diff
@@ -131,58 +131,6 @@ const PageBottom = ({ editDocsPath }: any) => {
 
   return (
     <PageBottomWrapper>
-      <Feedback>
```
mhessdev, line 134: I would have just commented it. with <!-- but thats ok

mhessdev (approved):

PR #5763, opened 2024-04-01
Docusaurus: Fix links on getting started page

jharrell:
Resolves DA-135


petradonka:
@jharrell, will this work fine when the URL is prisma.io/docs? Wouldn't the /docs be needed then? I might be asking stupid questions. :)

jharrell:
Should work fine. This is saying "navigate to location relative to the base" where the base (right now) is a `pages.dev` url when in the future it will be `prisma.io/docs`

PR #5850, opened 2024-04-10
Docusaurus swizzle search

mhessdev:
## Describe this PR

Made a swizzle component for saerch so we can customize things, also added some prettier files


.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+	"trailingComma": "es5",
+	"tabWidth": 4,
+	"semi": false,
+	"singleQuote": true
+}
```
jharrell: adapted from the live site:
```suggestion
{
  "printWidth": 100,
  "jsxBracketSameLine": false,
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "es5",
  "semi": false,
  "plugins": ["prettier-plugin-prisma"],
}
```

PR #5869, opened 2024-04-10
re-added old prettier configs

mhessdev:
Saving files was causing massive diffs. So i reverted prettier configs to match how they were during all of development

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,17 @@
 {
+    "printWidth": 100,
+    "jsxBracketSameLine": false,
+    "singleQuote": true,
+    "tabWidth": 2,
     "trailingComma": "es5",
-    "tabWidth": 4,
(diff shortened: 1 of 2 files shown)


petradonka (approved):

.prettierrc
```diff
@@ -1,6 +1,17 @@
...
+    "printWidth": 100,
+    "jsxBracketSameLine": false,
+    "singleQuote": true,
+    "tabWidth": 2,
     "trailingComma": "es5",
-    "tabWidth": 4,
     "semi": false,
-    "singleQuote": true
+    "plugins": ["prettier-plugin-prisma"],
+    "overrides": [
```
jharrell, line 9: You can remove this block since we’re excluding mdx

jharrell (approved):

PR #5871, opened 2024-04-10
Fix canonical

mhessdev:
had to add Head to the index js to override the canonical being set by config. This seemed the only way to do this.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -4,14 +4,5 @@
     "singleQuote": true,
     "tabWidth": 2,
     "trailingComma": "es5",
-    "semi": false,
-    "plugins": ["prettier-plugin-prisma"],
-    "overrides": [
-        {
-            "files": "*.mdx",
-            "options": {
-                "printWidth": 80
-            }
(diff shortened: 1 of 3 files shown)


.prettierrc
```diff
@@ -1,17 +1,8 @@
 {
-    "printWidth": 100,
-    "jsxBracketSameLine": false,
-    "singleQuote": true,
-    "tabWidth": 2,
-    "trailingComma": "es5",
-    "semi": false,
-    "plugins": ["prettier-plugin-prisma"],
```
petradonka, line 8: plugins is removed also, intentional?

petradonka:
Changes seem good to me, but I'd like someone else to double-check and approve

.prettierrc
mhessdev: Yes, pretty sure we don't need it. Correct @jharrell ?

jharrell: correct

jharrell (approved):

PR #5886, opened 2024-04-11
move files and make sure assets are loaded from the correct path

jharrell:


docusaurus.config.ts
```diff
@@ -170,10 +170,10 @@ const config: Config = {
...
       indexName: 'prisma',
       contextualSearch: false,
-      replaceSearchResultPathname: {
-        from: '/docs/',
-        to: DOCUSAURUS_BASE_URL,
-      },
+      // replaceSearchResultPathname: {
+      //   from: '/docs/',
+      //   to: DOCUSAURUS_BASE_URL,
+      // },
```
jharrell:
```suggestion
      replaceSearchResultPathname: {
        from: '/docs/',
        to: DOCUSAURUS_BASE_URL,
      },
```

PR #5897, opened 2024-04-11
document `omit` query option

grpatel:


content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -527,6 +527,7 @@ Use model queries to perform CRUD operations on your models. See also: [CRUD](/o
...
 | `select`                        | `XOR<UserSelect, null>`  | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned object.                                                                                                                                                                                    |
 | `include`                       | `XOR<UserInclude, null>` | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned object.                                                                                                                                                                    |
+| `exclude`                       | `XOR<UserExclude, null>` | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```
jharrell:
```suggestion
| `omit`                       | `XOR<UserOmit, null>` | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -757,6 +758,7 @@ We introduced `findFirstOrThrow` in v4.0.0. It replaces the [`rejectOnNotFound`]
...
 | `select`               | `XOR<PostSelect, null>`                                       | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned object.                                                                                                                                                                                    |
 | `include`              | `XOR<PostInclude, null>`                                      | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned object.                                                                                                                                                                    |
+| `exclude`              | `XOR<PostExclude, null>`                                      | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```
jharrell:
```suggestion
| `omit`              | `XOR<PostOmit, null>`                                      | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -796,6 +798,7 @@ const user = await prisma.user.findMany({
...
 | [`select`](#select)    | `XOR<UserSelect, null>`                                  | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned object.                                                                                                                                                                                    |
 | [`include`](#include)  | `XOR<UserInclude, null>`                                 | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned object.                                                                                                                                                                    |
+| [`exclude`](#exclude)  | `XOR<UserExclude, null>`                                 | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```
jharrell:
```suggestion
| [`omit`](#omit)  | `XOR<UserOmit, null>`                                 | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -903,6 +906,7 @@ prisma:query COMMIT
...
 | [`select`](#select)    | `XOR<UserSelect, null>`                                | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned object.                                                                                                                                                                                    |
 | [`include`](#include)  | `XOR<UserInclude, null>`                               | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned object.                                                                                                                                                                    |
+| [`exclude`](#exclude)  | `XOR<UserExclude, null>`                               | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```
jharrell:
```suggestion
| [`omit`](#omit)  | `XOR<UserOmit, null>`                               | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -951,6 +955,7 @@ This section covers the usage of the `upsert()` operation. To learn about using
...
 | [`select`](#select)    | `XOR<UserSelect, null>`                                 | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned object.                                                                                                                                                                                    |
 | [`include`](#include)  | `XOR<UserInclude, null>`                                | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned object.                                                                                                                                                                    |
+| [`exclude`](#exclude)  | `XOR<UserExclude, null>`                                | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```
jharrell:
```suggestion
| [`omit`](#omit)  | `XOR<UserOmit, null>`                                | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1156,6 +1161,7 @@ To delete records that match a certain criteria, use [`deleteMany`](#deletemany)
...
 | [`select`](#select)    | `XOR<UserSelect, null>`  | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned object.                                                                                                                                                                                    |
 | [`include`](#include)  | `XOR<UserInclude, null>` | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned object.                                                                                                                                                                    |
+| [`exclude`](#exclude)  | `XOR<UserExclude, null>` | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```
jharrell:
```suggestion
| [`omit`](#omit)  | `XOR<UserOmit, null>` | No       | Specifies which properties to omit on the returned object.                                                                                                                                                                    |
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
 </cmdResult>
 </CodeWithResult>
 
+### `exclude` (Preview)
```
jharrell:
```suggestion
### `omit` (Preview)
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+`exclude` defines which fields are omitted in the object that Prisma Client returns.
```
jharrell:
```suggestion
`omit` defines which fields are omitted in the object that Prisma Client returns.
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+#### Remarks
+
+- You cannot combine `exclude` and `select` since they serve opposite purposes
```
jharrell:
```suggestion
- You cannot combine `omit` and `select` since they serve opposite purposes
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+#### Examples
+
+##### Exclude the `password` field from all User records
```
jharrell:
```suggestion
##### Omit the `password` field from all User records
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+<CodeWithResult>
+<cmd>
+
+```ts
+const result = await prisma.user.findMany({
+  exclude: {
```
jharrell:
```suggestion
  omit: {
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+    email: 'rose@prisma.io',
+    name: 'Rose',
+  },
+]
+```
+
+</cmdResult>
+</CodeWithResult>
+
+##### Exclude `title` of all User posts with `include`
```
jharrell:
```suggestion
##### Omit `title` of all User posts with `include`
```

jharrell: This could probably use a bit of a better description too.

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+<CodeWithResult>
+<cmd>
+
+```ts
+const results = await prisma.user.findMany({
+  exclude: {
```
jharrell:
```suggestion
  omit: {
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+    password: true,
+  },
+  include: {
+    posts: {
+      exclude: {
```
jharrell:
```suggestion
      omit: {
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+      },
+    ],
+  },
+]
+```
+
+</cmdResult>
+</CodeWithResult>
+
+#### Generated types for `exclude`
```
jharrell:
```suggestion
#### Generated types for `omit`
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+The following example demonstrates how to use the [`validator`](/orm/prisma-client/type-safety/prisma-validator) with `exclude`:
```
jharrell:
```suggestion
The following example demonstrates how to use the [`validator`](/orm/prisma-client/type-safety/prisma-validator) with `omit`:
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1867,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+```ts
+const excludePassword = Prisma.validator<Prisma.UserExclude>()({
```
jharrell:
```suggestion
const omitPassword = Prisma.validator<Prisma.UserOmit>()({
```

jharrell (dismissed):
Totally understand that the API is in flux, but wanted to block w/ request changes so the wrong API doesn't enter the docs. Just ping me/yell at me and I'll remove the hold 🙏

grpatel:
Ah was gonna rename today, thanks for beating me to it 😛 🙏 @jharrell

SevInf (changes requested):
Overall looks good to me so far, but could we also document the behavior of `omit` together with `result` extensions? I mentioned it [here](https://prisma-company.slack.com/archives/C058VM009HT/p1712741786528389)

grpatel:
> Overall looks good to me so far, but could we also document the behavior of `omit` together with `result` extensions? I mentioned it [here](https://prisma-company.slack.com/archives/C058VM009HT/p1712741786528389)

@SevInf I've followed up with you on the thread to clarify

content/200-orm/500-reference/050-prisma-client-reference.mdx
grpatel: changing with to inside maybe?
```suggestion
##### Omit `title` from User posts inside `include`
```

grpatel: Maybe something like this?
```suggestion
##### Use `include` to `omit` `title` from User posts
```

jharrell:
I would also (personally) expect this to be documented next to `select` instead of next to `include`. Is there a reason we have it there?

grpatel:
Thanks for fixing table format @jharrell. Also, we had a discussion around ordering and decided on ordering by predicted frequency of usage. Quoting @Weakky (Flavian)'s message here:

> My 2cts - I see two ways to organize/sort this:
By “similarity”: select, exclude, include
By frequency of usage: select, include, exclude (this is just a hunch, I don’t know yet how much ppl are gonna use exclude)
I think I’d go with frequency of usage.
select, include, exclude may also be more aligned with the progression of needs of a user. What I mean is, you just installed the PrismaClient, you wanna know how to select data. Then you think about how to fetch relations. Then, maybe, you start wondering how to exclude data-sensitive columns.
I’m not opinionated on this though, choose whatever you think is best based on these 2cts. The sorting by similarity is just as fine. (edited)

jharrell:
@grpatel @Weakky 

totally makes sense. What about a mix of the two? Similar "groups" organized by usage. 

In that way, I would expect Select/Omit to be grouped, etc etc

I'm not going to block based on that though. It was just my first thought as I was looking at the list.

content/200-orm/500-reference/050-prisma-client-reference.mdx
Weakky: What's important in the use-case shown, IMO, is that one can omit fields from nested relations. Not that we're using `include` specifically. eg:

```
- Omit the `title` fields from all User posts
- Omit the `title` fields from all Post relations
```

grpatel: I like this one
```suggestion
##### Omit the `title` fields from all Post relations
```

content/200-orm/200-prisma-client/300-client-extensions/130-result.mdx
janpio, line 1: Unrelated changes in this file!?

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1777,23 +1784,9 @@ const result = await prisma.user.findMany({
 
 #### Generated types for `select`
 
-The following examples demonstrate how to use the [`validator`](/orm/prisma-client/type-safety/prisma-validator) with `select`:
-
-- `SelectAndInclude`
-- `UserSelect`
+The following example demonstrates how to use the [`validator`](/orm/prisma-client/type-safety/prisma-validator) with `select`:
```
janpio, line 1787: Intentional changes here?

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1868,123 @@ const usersWithCount = await prisma.user.findMany({
...
+
+`omit` defines which fields are excluded in the object that Prisma Client returns.
+
+#### Remarks
+
+- You cannot combine `omit` and `select` since they serve opposite purposes
+
+#### Examples
+
+##### Omit the `password` field from all User records
```
```suggestion
##### Omit the `password` field from all `User` records
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1889,6 +1868,123 @@ const usersWithCount = await prisma.user.findMany({
...
+    email: 'rose@prisma.io',
+    name: 'Rose',
+  },
+]
+```
+
+</cmdResult>
+</CodeWithResult>
+
+##### Omit the `title` fields from all Post relations
```
```suggestion
##### Omit the `title` fields from all `Post` relations
```

content/200-orm/200-prisma-client/300-client-extensions/130-result.mdx
grpatel: This is after my discussion with Niko. We thought the main heading was redundant and then wanted to adjust the other heading sizes accordingly before adding the new info about omit. Is there anything that you don't agree with?

content/200-orm/500-reference/050-prisma-client-reference.mdx
janpio: Is calling this "Posts relations" even correct though?
It is the `Post` model, but aren't the relations named "posts"?

content/200-orm/500-reference/050-prisma-client-reference.mdx
grpatel: Yes, after my discussion with @SevInf. Wanted to keep only `UserSelect`

content/200-orm/200-prisma-client/300-client-extensions/130-result.mdx
janpio: Yes, it is not related to `omit` so should not be part of this PR. It makes it much harder to reason about this PR by itself - which now means I'll have to figure out if the headline indeed is redundant before approving.

content/200-orm/500-reference/050-prisma-client-reference.mdx
SevInf: Yes, `SelectAndInclude` example does not really make sense - it *requires* `select` and `include` to be used together which in turn is not allowed by either type- or runtime- validation in the client. It's an internal type which used for validation (basically, only to say "If user's input looks like that then complain")

content/200-orm/500-reference/050-prisma-client-reference.mdx
grpatel: One of the docs examples says "Include the `posts` relation..." So maybe we can change it to: 
In order to make the relation part explicit
```suggestion
##### Omit the `title` fields from all `User`'s `posts` relation
```

cc: @Weakky

content/200-orm/200-prisma-client/300-client-extensions/130-result.mdx
janpio: Ok, after looking at the other pages I think the previous state followed the same pattern as with the other extension types. 

But no strong feelings, so 🤷

grpatel: Ya my bad for not creating a separate PR for this. I'd like to keep it this way since I've talked with Niko about this. In the future, we can update the other pages to keep the format aligned

SevInf (approved):

jharrell (approved):

PR #5901, opened 2024-04-12
fix title backticks in extension pages

jharrell:
Docusaurus doesn't like backticks (`) in titles.

Before:
image


content/200-orm/200-prisma-client/300-client-extensions/100-model.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: '`model`: Add custom methods to your models'
+title: 'Add custom methods to Prisma models'
```
petradonka, line 2: Don't these changes take away from the meaning of the title, as they previously better expressed what kind of client extension they're about?

jharrell: I think the meaning is slightly different, but (IMO) the new and old titles focus on what you want to do. In this case, adding custom methods. 

How you do that (again in this case, `model` extensions) is described in the page.

PR #5946, opened 2024-04-25
Fix code snippet won't compile issue

onlyice:
The original code won't compile with TS compiler, due to the missing field `content` in the data. Add the field to eliminate compiler error.

### DIFF
--- a/content/100-getting-started/01-quickstart.mdx
+++ b/content/100-getting-started/01-quickstart.mdx
@@ -276,6 +276,7 @@ async function main() {
       posts: {
         create: {
           title: 'Hello World',
+          content: 'This is a test post',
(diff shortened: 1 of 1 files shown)


jharrell:
Hi @onlyice ,

That shouldn't be the case. The `Post` model's `content` field is marked as optional:

```prisma
model Post {
  title     String
  content   String?
  published Boolean @default(false)
  author    User    @relation(fields: [authorId], references: [id])
  authorId  Int
}
```

Could you confirm that your model is set up correctly?

onlyice:
Hi @jharrell, I can confirm that the problem is caused by the reason you stated. Thanks for your reply.

PR #5970, opened 2024-05-02
add PDP concepts overview

nikolasburk:
This PR introduces a number of changes to the PDP docs:
- refactors "concepts" and move **Workspaces**, **Projects** and **Environments** out of the sidebar and puts all relevant information into **About** page
- elaborates on **Workspaces**, **Projects** and **Environments** with simple language and gives examples of "typical" use/workflows
- gives more context on billing with concrete examples
- adds concrete instructions to configure the Accelerate connection pool size


content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -36,17 +32,17 @@ This means that Accelerate will create a maximum of 5 connections to the databas
 
 ### Configuring the connection pool size
 
-The connection pool size can be configured to a value other than the default via the _database connection string_.
+The connection pool size can be configured to a value other than the default via the _database connection string_ in your PDP project (_not_ the connection string in the Prisma schema) via the `connection_limit` parameter. For example, to set a connection pool size of `10` like this: `postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=10`.
 
-![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
+To change the size of the connection pool:
+1. Open the [Platform Console](https://console.prisma.io/).
+1. Navigate to the _project_ in which you're using Accelerate.
```
nurul3101: Suggestion: Navigate to the _project_ in the appropriate workspace in which you're using Accelerate.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,117 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
 displayed_sidebar: platformSidebar
 ---
 
-<TopBlock>
+## Overview
 
-Platform Console enables you to integrate the latest Prisma Data Platform products into your application
+Platform Console enables you to integrate the latest Prisma Data Platform (PDP) products into your application
 
 - [Accelerate](/accelerate/what-is-accelerate): Global database cache with scalable connection pooling
```
nurul3101, line 12: Should it be `scalable` or `scaleable`?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,117 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+In each workspace, you can:
+
+- view and manage all projects (and their environments) in that workspace.
+- manage billing, i.e. select a [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site), configure payment methods, or view the invoice history.
+- view the usage of your enabled PDP products across all projects in that workspace.
+- manage team members and change their roles / permissions in the workspace.
+
+### Project
+
+In each workspace, you can create several projects. A project typically represents an application (a product or service) your organization is offering. You typically have one [Prisma schema](/orm/prisma-schema/) per project.
```
nurul3101: Maybe we can add that the number of projects you can create in a workspace depends on the plan you are on.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,117 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+| **Environments** | 2 / project | 6 / project | 12 / project | Custom         |
+
+## Billing
+
+### Per-workspace billing
+
+Billing in PDP is done on a per-workspace basis:
+
+- A [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) is selected per workspace. That means, your PDP account can have multiple workspaces where each workspace uses a different pricing plan.
+- A payment method is select per workspace. That means, your PDP account can have multiple workspaces where each workspace has a different payment method.
```
nurul3101:
```suggestion
- A payment method is selected per workspace. That means, your PDP account can have multiple workspaces where each workspace has a different payment method.
```

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -36,17 +32,17 @@ This means that Accelerate will create a maximum of 5 connections to the databas
...
+1. Select the _environment_ where you want to configure Accelerate's connection pool size.
+1. Update the connection string with by appending the `connection_limit` argument.
```
nurul3101:
```suggestion
1. Update the connection string by appending the `connection_limit` argument. 
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,117 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- if you want to try out a particular plan for only 2 days, you can do so and you'll only be charged for the respective fraction of the plan's price (2 days out of the entire month).
+- if you downgrade your pricing plan (e.g. from **Business** to **Pro**) after 10 days of a 30-day month, you'll be charged 1/3 (i.e. 10 days) for the **Business** plan and 2/3 (i.e. 20 days) for the **Pro** plan.
+
+Visit our [pricing page](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) for more details.
+
+### Downgrading a pricing plan
+
+If you downgrade a pricing plan, you may need to delete some of your projects and/or their environments in order to adhere to the [limits](#environment) of the newly selected plan.
+
+For example, if your workspace is on a **Business** plan and current has 14 (out of 15) projects, you will need to delete at least 4 projects to adhere to the project limit of the **Pro** plan. Additionally, you need to make sure that the remaining projects don't have more than 6 environments per project to adhere to the environment limit of the **Pro** plan.
```
nurul3101:
```suggestion
For example, if your workspace is on a **Business** plan and currently has 14 (out of 15) projects, you will need to delete at least 4 projects to adhere to the project limit of the **Pro** plan. Additionally, you need to make sure that the remaining projects don't have more than 6 environments per project to adhere to the environment limit of the **Pro** plan.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,117 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+## Programmatic access via the Platform CLI
+
+In addition to the GUI of the Platform Console, thge Prisma CLI provides another way to interact with your PDP account and manage PDP products in your projects.
```
nurul3101:
```suggestion
In addition to the GUI of the Platform Console, the Prisma CLI provides another way to interact with your PDP account and manage PDP products in your projects.
```

content/500-platform/50-support.mdx
```diff
@@ -7,26 +7,32 @@ toc: true
...
+### Dedicated support
 
 Dedicated contact person.
+
+## Deleting your PDP account
+
+
+If you want to delete your PDP account, **email us at <a href="mailto:support@prisma.io">support@prisma.io</a>**. 
+
+To ensure that you're not accidentally disabling any infrastructure powering onbe of your applications, we require that you **disable Accelerate and Pulse in _all_ environments of _all_ your projects** that live in the account to be deleted,
```
nurul3101:
```suggestion
To ensure that you're not accidentally disabling any infrastructure powering one of your applications, we require that you **disable Accelerate and Pulse in _all_ environments of _all_ your projects** that live in the account to be deleted.
```

content/300-accelerate/250-connection-pooling.mdx
nikolasburk:
```suggestion
1. Select the _workspace_ of the _project_ in which you're using Accelerate and navigate into that _project_.
```

content/500-platform/10-about.mdx
nikolasburk: not sure 🤔 looks like "scaleable" is UK spelling maybe? I think "scalable" is fine (looks like its the more common option after a quick Google search).

content/500-platform/10-about.mdx
nikolasburk: great idea!! I actually had done this but hadn't pushed the latest version with that change 🤦 just pushed these changes though!

content/500-platform/10-about.mdx
nurul3101: Let's go with scalable then 👍

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -36,17 +32,17 @@ This means that Accelerate will create a maximum of 5 connections to the databas
 
 ### Configuring the connection pool size
 
-The connection pool size can be configured to a value other than the default via the _database connection string_.
+The connection pool size can be configured to a value other than the default via the _database connection string_ in your PDP project (_not_ the connection string in the Prisma schema) via the `connection_limit` parameter. For example, to set a connection pool size of `10` like this: `postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=10`.
```
petradonka:
```suggestion
The connection pool size can be configured via the _database connection string_ in your Platform project (_not_ the Accelerate connection string in your Prisma schema) via the `connection_limit` parameter. For example, set a connection pool size of `10` like this: `postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=10`.

If nothing is configured, a default value will be used.
```

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -36,17 +32,17 @@ This means that Accelerate will create a maximum of 5 connections to the databas
...
 
-![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
+To change the size of the connection pool:
+1. Open the [Platform Console](https://console.prisma.io/).
+1. Select the _workspace_ of the _project_ in which you're using Accelerate and navigate into that _project_.
```
petradonka:
```suggestion
1. Select the _project_ in which you're using Accelerate. You might need to select a different _workspace_ if you can't see your project.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
 displayed_sidebar: platformSidebar
 ---
 
-<TopBlock>
+## Overview
 
-Platform Console enables you to integrate the latest Prisma Data Platform products into your application
+Platform Console enables you to integrate the latest Prisma Data Platform (PDP) products into your application
```
petradonka:
```suggestion
The [Platform Console](http://console.prisma.io) enables you to manage and configure your projects that use Prisma Data Platform (PDP) products, and helps you integrate them into your application:
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
 
 - [Accelerate](/accelerate/what-is-accelerate): Global database cache with scalable connection pooling
 - [Pulse](/pulse/what-is-pulse): A managed Change Data Capture (CDC) service that enables real-time database subscriptions
 
-[Go to Platform Console](https://console.prisma.io)
+Open the [Platform Console](https://console.prisma.io).
```
petradonka:
```suggestion
```

Let's remove this and rather link "Platform Console" above.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
 
-</TopBlock>
+## Concepts
+
+The Platform Console workflows are based on four main concepts:
+
+- [**Account**](#account): In order to use PDP products, you need to have a PDP account. An _organization_ will typically create one account to manage all their workspaces, projects and environments.
```
petradonka: > An _organization_ will typically create one account to manage all their workspaces, projects and environments.

I don't understand this part, it kinda sounds like an org/company would create a shared account that everyone uses, which is not right. Am I missing something?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- [**Workspaces**](#workspace): An account can have multiple workspaces. A workspace typically represents a _team_ within an _organization_. **Billing happens on a per-workspace basis**, i.e. an _organization_ receives an invoice at the end of the month that captures all cost associated with the projects in a given workspace.
```
petradonka:
```suggestion
- [**Workspaces**](#workspace): A user can belong to multiple workspaces. A workspace typically represents a _team_ of individuals working together on one or more projects. **Billing is on a workspace level**, i.e. the invoice for a workspace at the end of the month captures all costs associated with the projects in a given workspace.
```

I tried simplifying a little. I believe account might be better named "user", and I'd avoid bringing in the term/concept "organization".

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- [**Project**](#project): A workspace can have multiple projects. A project typically represents the _application_ of a _product_ or _service_ a team is working on.
```
petradonka:
```suggestion
- [**Project**](#project): A project belongs to a workspace. It typically represents the _application_ or _service_ a team is working on.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- [**Environments**](#environment): A project can have multiple environments. An environment typically maps to a _development stage_, like `dev`, `staging` or `prod`. **API keys are provisioned on the environment-level**.
```
petradonka:
```suggestion
- [**Environments**](#environment): An environment belongs to a project. It typically maps to a _development stage_, like `dev`, `staging` or `prod`. **API keys are provisioned on the environment level**, and products are configured per environment as well (e.g. the database connection string used for Accelerate).
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+Here is a visual illustration of how these concepts relate to each other:
+
+![](/img/platform/pdp-concepts.png)
```
petradonka, line 27: Could we rather use ACME as the workspace, where there might be a different projects (apps, services, etc)? I think it'd be a bit simpler, and would map better to most of our users.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+### Account
```
petradonka:
```suggestion
### User Account
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+Your account is the basis for all interactions with PDP products. You can use it to manage your workspaces (and their projects).
```
petradonka: I find this a little confusing, because the env API key is also a way to interact with products. What do you mean here exactly?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+If you need to delete your account, go [here](/platform/support#deleting-your-pdp-account).
+
+### Workspace
+
+In your PDP account, you can create several workspaces. A workspace is an isolated space to host projects in which you use PDP products.
```
petradonka: Hmh, interesting, I'd rather say that accounts belong to workspaces, which would mean that "in your account, you can create workspaces" is not exactly correct. How about this?

```suggestion
You can create several workspaces. A workspace is an isolated space to host projects.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+In each workspace, you can:
+
+- view and manage all projects (and their environments) in that workspace.
+- manage billing, i.e. select a [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site), configure payment methods, or view the invoice history.
+- view the usage of your enabled PDP products across all projects in that workspace.
+- manage team members and change their roles / permissions in the workspace.
+
+### Project
+
+In each workspace, you can create several projects. A project typically represents an application (a product or service) your organization is offering. You typically have one [Prisma schema](/orm/prisma-schema/) per project.
```
petradonka:
```suggestion
In each workspace, you can create several projects. A project typically represents an application (a product or service). You typically have one [Prisma schema](/orm/prisma-schema/) per project.
```

I'd avoid using "organization" here, as well as "offering", which implies a commercial product. Great callout on the Prisma schema!

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+In each project, you can:
+
+- view and manage all environments in that project.
+
+The number of project you can create in a workspace depends on the [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) configured in that workspace:
+
+Here's an overview of the number of projects per workspace on each plan:
```
petradonka: I see the conversation above :) I'd actually remove this, and just let people look on the pricing page to minimize the places where we have a source of truth. If we ever decide to change this, it'll be a pain to find all places we mention explicit numbers.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+### Project
+
+In each workspace, you can create several projects. A project typically represents an application (a product or service) your organization is offering. You typically have one [Prisma schema](/orm/prisma-schema/) per project.
+
+In each project, you can:
+
+- view and manage all environments in that project.
+
+The number of project you can create in a workspace depends on the [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) configured in that workspace:
```
petradonka:
```suggestion
The number of project you can create in a workspace depends on the [pricing plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) configured in that workspace:
```

I'd rather use a direct prisma.io link, is there a good reason I might be missing to use a pris.ly link?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+| ------------ | ----------- | ------- | ------------ | -------------- |
+| **Projects** | 5           | 10      | 15           | Custom         |
+
+### Environment
+
+An environment is an isolated space to provision PDP products for a given project. Environments typically map to development stages, like `dev`, `staging` or `prod`.
+
+In each environment, you can:
+
+- enable, disable and manage PDP products (Pulse, Accelerate, ...).
```
petradonka:
```suggestion
- enable, disable and configure PDP products (Pulse, Accelerate, ...).
```

How about configure instead of manage?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- for **Accelerate**:
+  - set your database connection string.
+  - generate API keys.
```
petradonka: Hmh, would you say that you generate API keys per product or per environment? You could use the same API key for Accelerate and Pulse on the same env, right?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+  - change the connection pool size.
+  - enable static IPs.
```
petradonka:
```suggestion
  - enable static IP.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- for **Pulse**:
+  - set your database connection string (of a [Pulse-compatible database](/pulse/database-setup)).
+  - configure the _region_ where Pulse is running.
```
petradonka, line 71: Should we mention region for Accelerate also?

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+  - configure database replication.
+  - generate API keys.
+
+The number of environments you can create in a project depends on the [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) configured in your workspace:
```
petradonka: Same comment here about the prisma.io vs. pris.ly link & mentioning the specific numbers of envs allowed per plan

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+Here's an overview of the number of environments per project on each plan:
+
+|                  | **Starter** | **Pro**     | **Business** | **Enterprise** |
+| ---------------- | ----------- | ----------- | ------------ | -------------- |
+| **Environments** | 2 / project | 6 / project | 12 / project | Custom         |
+
+## Billing
+
+The [pricing plan](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) you select in your workspace determins how many projects and environments you can create in that workspace:
```
petradonka: I'd remove this section before "per-workspace billing". It's a very limited subset of what you get with plans, and I also wouldn't want to duplicate the information.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+|                  | **Starter** | **Pro**     | **Business** | **Enterprise** |
+| ---------------- | :---------- | ----------- | ------------ | -------------- |
+| **Projects**     | 5           | 10          | 15           | Custom         |
+| **Environments** | 2 / project | 6 / project | 12 / project | Custom         |
+
+### Per-workspace billing
+
+Billing in PDP is done on a per-workspace basis:
```
petradonka:
```suggestion
Billing is set up on a per-workspace basis:
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+- A pricing plan is selected per workspace. That means, your PDP account can have multiple workspaces where each workspace uses a different pricing plan.
```
petradonka:
```suggestion
- A subscription plan is selected per workspace. That means, your user can belong to multiple workspaces where each workspace uses a different plan.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- A payment method is selected per workspace. That means, your PDP account can have multiple workspaces where each workspace has a different payment method.
```
petradonka:
```suggestion
- A payment method is selected per workspace. That means, your user can belong to multiple workspaces where each workspace has a different payment method.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+At the end of the billing period, **your selected payment method will be charged with the usage costs of the PDP products across _all_ the projects (and their environments) in that workspace**.
```
petradonka:
```suggestion
At the end of a billing period, your selected payment method will be charged with the incurred costs of products across _all_ projects (and their environments) in that workspace.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+You can configure all billing details in the **Billing** section of your project.
```
petradonka:
```suggestion
You can configure all billing details in the **Billing** section of your workspace.
```

See? :) It needs clarity & explanation, even we mix it up sometimes.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+### Prorated billing
+
+All our paid plans are prorated, which means **you're only billed for the duration you use a specific plan**.
```
petradonka:
```suggestion
All base plan prices are prorated, which means you're only billed for the duration of your subscription to a specific plan. In addition, you're also billed for any usage costs you've incurred during your subscription.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+For example:
+
+- if you subscribe to our **Pro** plan for only 15 days in a 30-day month, you'll only be charged for half of the plan's price.
+- if you want to try out a particular plan for only 2 days, you can do so and you'll only be charged for the respective fraction of the plan's price (2 days out of the entire month).
```
petradonka: I'd  remove this, as we might introduce a minimum charge in the future and I don't want to encourage doing this. Some people have upgraded to Pro for just an hour to get direct support, which is not great for us, as they didn't _really_ pay the price for it.

```suggestion
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+You can configure all billing details in the **Billing** section of your project.
+
+### Prorated billing
+
+All our paid plans are prorated, which means **you're only billed for the duration you use a specific plan**.
+
+For example:
+
+- if you subscribe to our **Pro** plan for only 15 days in a 30-day month, you'll only be charged for half of the plan's price.
```
petradonka: How about taking this bullet point as subscribing in the middle of the month, while the last one is about cancelling/downgrading mid-month?

```suggestion
- if you subscribe to our **Pro** plan on the 15th day of a month, you'll only be charged the base plan price for the days left in that month.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+- if you want to try out a particular plan for only 2 days, you can do so and you'll only be charged for the respective fraction of the plan's price (2 days out of the entire month).
+- if you downgrade your pricing plan (e.g. from **Business** to **Pro**) after 10 days of a 30-day month, you'll be charged 1/3 (i.e. 10 days) for the **Business** plan and 2/3 (i.e. 20 days) for the **Pro** plan.
```
petradonka:
```suggestion
- if you downgrade your subscription plan (e.g. from **Business** to **Pro**) after 10 days of a 30-day month, you'll be charged for 10 days of the base price of the **Business** plan and 20 days for the base price of the **Pro** plan.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+Visit our [pricing page](https://pris.ly/pricing?utm_source=prisma-docs&utm_medium=docs-site) for more details.
```
petradonka:
```suggestion
Visit our [pricing page](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) for more details.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+### Downgrading a pricing plan
```
petradonka:
```suggestion
### Downgrading a subscription plan
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+If you downgrade a pricing plan, you may need to delete some of your projects and/or their environments in order to adhere to the [limits](#environment) of the newly selected plan.
```
petradonka:
```suggestion
If you downgrade a subscription plan, you may need to delete some of your projects and/or their environments in order to adhere to the [limits](http://prisma.io/pricing) of the newly selected plan.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+For example, if your workspace is on a **Business** plan and current has 14 (out of 15) projects, you will need to delete at least 4 projects to adhere to the project limit of the **Pro** plan. Additionally, you need to make sure that the remaining projects don't have more than 6 environments per project to adhere to the environment limit of the **Pro** plan.
+
+## Programmatic access via the Platform CLI
+
+In addition to the GUI of the Platform Console, thge Prisma CLI provides another way to interact with your PDP account and manage PDP products in your projects.
```
petradonka:
```suggestion
In addition to the web interface of the Platform Console, the Prisma CLI provides another way to interact with your PDP account and manage PDP products.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+This is if you need programmatic access to your account, e.g. for integrating it into CI workflows.
```
petradonka:
```suggestion
This can be useful if you need programmatic access, e.g. for integrating it into CI workflows.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+Read more about the Prisma CLI [here](/platform/platform-cli).
```
petradonka:
```suggestion
Read more about the [Prisma CLI](/platform/platform-cli).
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,133 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+## API keys
+
+An API key is required to authenticate requests from your Prisma Client to products such as Prisma Accelerate and Prisma Pulse. You may generate multiple API keys per environment and manage those via the **API Keys** section in a project.
```
petradonka:
```suggestion
An API key is required to authenticate requests from your Prisma Client to products such as Prisma Accelerate and Prisma Pulse. You may generate multiple API keys per environment and manage those via the **API Keys** section in an environment.
```

Env or project? 🤔

content/500-platform/50-support.mdx
```diff
@@ -7,26 +7,32 @@ toc: true
...
 
-### Dedicated Support
+### Dedicated support
 
 Dedicated contact person.
+
+## Deleting your PDP account
+
+
+If you want to delete your PDP account, **email us at <a href="mailto:support@prisma.io">support@prisma.io</a>**.
```
petradonka, line 36: I'd remove this, so we don't have to update it when we implement this. We can link to the Console, where these instructions also show up.

petradonka (changes requested):

content/500-platform/10-about.mdx
nikolasburk: I'm actually not super familiar with the workflows so I'm very open to feedback.

Just to state my understanding: 

> it kinda sounds like an org/company would create a shared account that everyone uses

I thought this would have been the case but it might well be wrong! 

Do we expect a single company to create multiple PDP accounts? My understanding was the the PDP concepts map to "organizations" as follows:
- Account -> Organization (company, agency, individual, ...)
- Workspace -> Team
- Project -> Product / Service / Application
- Environment -> Dev stage

Is that not correct?

nikolasburk: Ah, never mind! Of course every member of the org will have their own account and there isn't a _shared_ account for everyone across the company. I see how my phrasing is confusing here (I've even confused myself 😅 ), will try to make it clearer!

nikolasburk: How about this:

- [**Account**](#account): In order to use PDP products, you need to have a PDP account. A _user_ will typically create one account to manage all their workspaces, projects and environments. The _user_ can also be invited to join other workspaces to collaborate on the projects in that workspace.

content/500-platform/10-about.mdx
nikolasburk: Agree, will update the illustration.

nikolasburk: how about this?

image

there are still two ACME-related workspaces, do you think one of them should be renamed/removed or is it ok like this?

content/500-platform/10-about.mdx
nikolasburk: I basically meant it as "Without a user account you can't do anything in the PDP, so the account is the basis/foundation/prerequisite for any interactions with PDP product".

Does that make sense? But if it's not clear, I can try to phrase this better!

nikolasburk: Maybe just rephrasing "basis" to "prerequisite" makes that intention clearer?

"A user account is the prerequisite for any interactions with PDP products. You can use it to manage your workspaces (and their projects)."

nikolasburk: Expanding even more:

"A user account is the prerequisite for any interactions with PDP products. You can use it to manage your workspaces (and their projects). A user account can be invited to collaborate on workspaces created by other users as well."

content/500-platform/10-about.mdx
nikolasburk: tbh I copied over the link that Ankur used previously and I don't really have an opinion on this.

I guess when using pris.ly there's an element of flexibility where if the target URL changes, we only need to update the pris.ly link but not the places where we use it. on the other hand, the `prisma.io/pricing` url likely won't change so that's maybe not a concern.

nikolasburk: updated all pris.ly/pricing links to prisma.io/pricing 👍

content/500-platform/10-about.mdx
nikolasburk: I thought it's useful because IMO it helps to strengthen the understanding of workspaces, projects and environments and how they relate to each other (and also because Nurul brought it up as something people are confused about and that we may want to document).

> If we ever decide to change this, it'll be a pain to find all places we mention explicit numbers.

I'd take the tradeoff of better/clearer explanations and needing to go through the docs to update pricing information in favor over not showing this info.

Do you feel strongly about removing this information? If so, happy to do it of course!

content/500-platform/10-about.mdx
nikolasburk: Hmm, good point! I think there are actually _two_ pieces of information here:

- _where_ is an API key generated (as in: _where in the console UI_?) -> on the _environment_ level (that's what I was referring to)
- _what_ does an API key apply to (as in: _where can it be used_)? -> I'm actually not sure.... "You could use the same API key for Accelerate and Pulse on the same env" so it's one API key per env that can be used for _both_ products? that would seem unexpected to me but might very well be the case!

content/500-platform/10-about.mdx
nikolasburk: added: 

  - configure the _region_ where Accelerate's connection pool is running.

content/500-platform/10-about.mdx
nikolasburk: I can commit this but personally find "pricing plan" clearer than "subscription plan" because:
- "pricing" is also part of the target url and leaves no room for ambiguity misinterpration
-  "subscription" is generally a bit more ambiguous IMO (also in the context of Pulse where we have "subscriptions")
- I _believe_ I/we also used "pricing plan" everywhere else on this page

nikolasburk: I updated the bullet with your other suggestions though:

- A pricing plan is selected per workspace. That means, a user account can belong to multiple workspaces where each workspace uses a different plan.

content/500-platform/10-about.mdx
nikolasburk: updated to: 

- A payment method is selected per workspace. That means, a user account can belong to multiple workspaces where each workspace has a different payment method.

content/500-platform/10-about.mdx
nikolasburk: makes sense!

content/500-platform/10-about.mdx
nikolasburk: > I can commit this but personally find "pricing plan" clearer than "subscription plan"

I saw you made the same suggestion elsewhere. Just confirm that we want to go with "subscription plan" and I'll do a find/replace on the page 👍

content/500-platform/50-support.mdx
nikolasburk: > I'd remove this,

What exactly does "this" refer to here? is it the entire section **Deleting your PDP account** or just this sentence "If you want to delete your PDP account, **email us at <a href="mailto:support@prisma.io">support@prisma.io</a>**."?

FWIW, this was a request from Nurul because he needs to explain this to people over and over.

petradonka: > this sentence "If you want to delete your PDP account, email us at [support@prisma.io](mailto:support@prisma.io)."?

This sentence :) 

@nurul3101, what do you think? Do you get a lot of questions about this, even considering that it's on the Console in your settings?

content/500-platform/10-about.mdx
petradonka: Yep, you found the confusing parts 😄 I'd just call this a user, vs. an account. I think it makes things more clear. You also have a user ID, and you have user settings, etc.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,132 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+### Workspace
+
+You can create several workspaces. A workspace is an isolated space to host projects. A workspace can have multiple user accounts associated with it so that multiple users can collaborate on the the projects in the workspace.
+
+In each workspace, you can:
+
+- view and manage all projects (and their environments) in that workspace.
+- manage billing, i.e. select a [pricing plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site), configure payment methods, or view the invoice history.
+- view the usage of your enabled PDP products across all projects in that workspace.
+- manage team members and change their roles / permissions in the workspace.
```
petradonka: Do we actually have roles, permissions right now?

content/500-platform/10-about.mdx
petradonka: If you guys think it's valuable, let's leave it. I can see your reasoning to strengthen understanding. It's likely not something we'll change a lot.

content/500-platform/10-about.mdx
petradonka: >"You could use the same API key for Accelerate and Pulse on the same env" so it's one API key per env that can be used for both products?

Yes, that's my understanding. If we're not sure, let's please ask and check :)

content/500-platform/10-about.mdx
petradonka: That's how I usually put it :) I have never used pricing plan anywhere. We usually talk about subscriptions to talk about what you have active that is on a specific plan, so I think subscription plan and subscription base price is the right terminology.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,132 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+- enable, disable and configure PDP products (Pulse, Accelerate, ...).
+- for **Accelerate**:
+  - set your database connection string.
+  - configure the _region_ where Accelerate's connection pool is running.
+  - generate API keys.
+  - change the connection pool size.
+  - enable static IP.
+- for **Pulse**:
+  - set your database connection string (of a [Pulse-compatible database](/pulse/database-setup)).
```
petradonka, line 70:
```suggestion
  - set your database connection string.
```

Reading this again, I think this is implied. You can also only use a compatible connection string for Accelerate, but we don't call it out.

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,132 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
+
+|                  | **Starter** | **Pro**     | **Business** | **Enterprise** |
+| ---------------- | ----------- | ----------- | ------------ | -------------- |
+| **Environments** | 2 / project | 6 / project | 12 / project | Custom         |
+
+## Billing
+
+The [pricing plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) you select in your workspace determins how many projects and environments you can create in that workspace:
+
+|                  | **Starter** | **Pro**     | **Business** | **Enterprise** |
```
petradonka, line 80: Even if we leave other mentions of these limits, I'd remove it from this section. It's literally right under the previous table, seems very duplicated.

content/500-platform/10-about.mdx
petradonka: I think this is better! I like the differentiation between personal and work workspaces

content/500-platform/10-about.mdx
nikolasburk: I like the extra callout because it indicates that a "normal" DB isn't enough for Pulse and link people to a place where they can learn more. To me, that's valuable extra context that I like to add in places where Pulse is being mentioned.

> You can also only use a compatible connection string for Accelerate, but we don't call it out.

IMO, that's because _any_ publicly accessible DB is compatible with Accelerate.  But with Pulse, you need to go through additional DB setup.

I would prefer leaving it in but can drop it if you think it doesn't add value.

content/500-platform/10-about.mdx
nikolasburk: Actually no, good catch! 

image

Do we want to still mention this with a "coming soon" maybe? Or drop altogether?


```suggestion
- (coming soon) manage team members and change their roles / permissions in the workspace.
```

content/500-platform/10-about.mdx
nikolasburk: I've updated the bullet list like this:

- enable, disable and configure PDP products (Pulse, Accelerate, ...).
- generate API keys.
- for **Accelerate**:
  - set your database connection string.
  - configure the _region_ where Accelerate's connection pool is running.
  - change the connection pool size.
  - enable static IP.
- for **Pulse**:
  - set your database connection string (of a [Pulse-compatible database](/pulse/database-setup)).
  - configure the _region_ where Pulse is running.
  - configure database replication.

content/500-platform/10-about.mdx
nikolasburk: done with find/replace ✅

mhessdev:
Screenshot 2024-05-06 at 1 42 32 PM

I would possibly note what environment is default for a project. Does it have a name?

mhessdev:
Maybe I missed it, but I'm not sure. I see something stating that you need a single API key for a single environment, even if that environment has both Prisma Pulse and Prisma Accelerate enabled. (although I might be wrong on that working)

content/500-platform/10-about.mdx
petradonka: Nah, probably drop. "Manager team members" can mean that also, but we don't know when we'd add this, when it's coming

content/500-platform/10-about.mdx
petradonka: I'm happy with that reasoning

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,119 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
...
-[Go to Platform Console](https://console.prisma.io)
 
-</TopBlock>
+## Concepts
+
+The Platform Console workflows are based on four main concepts:
+
+- [**User account**](#user-account): In order to use PDP products, you need to have a PDP user account. A _user_ will typically create one user account to manage all their workspaces, projects and environments. The _user_ can also be invited to join other workspaces to collaborate on the projects in that workspace.
+- [**Workspaces**](#workspace): A user account can belong to multiple workspaces. A workspace typically represents a _team_ of individuals working together on one or more projects. **Billing is on a workspace level**, i.e. the invoice for a workspace at the end of the month captures all costs associated with the projects in a given workspace.
+- [**Project**](#project): A project belongs to a workspace. It typically represents the _application_ or _service_ a team is working on.
```
petradonka:
```suggestion
- [**Projects**](#project): A project belongs to a workspace. It typically represents the _application_ or _service_ a team is working on.
```

content/500-platform/10-about.mdx
```diff
@@ -5,13 +5,119 @@ metaDescription: 'Learn about the Platform Console to integrate the Prisma Data
 displayed_sidebar: platformSidebar
 ---
 
-<TopBlock>
+## Overview
 
-Platform Console enables you to integrate the latest Prisma Data Platform products into your application
+The [Platform Console](http://console.prisma.io) enables you to manage and configure your projects that use Prisma Data Platform (PDP) products, and helps you integrate them into your application:
```
petradonka: We might not want to resolve this in this PR, but I'll mention it and we can discuss:

Do we want to call the Console:
- Console
- Platform Console
?

I'd imagine that eventually we'll move towards Console and the Prisma CLI (including what we call Platform CLI now) to be top level entities in our docs. The Platform CLI is not a different CLI than the Prisma CLI, so the current differentiation might become confusing long-term.
That'd mean that Console and the CLI are interfaces to manage, configure and work with our products, ORM, Accelerate, Pulse, etc.

We might want to already start calling the Console "Console", to ease that simplification. It also aligns well with the URL being console.prisma.io.

Curious what you think, it's somewhat out of scope of this specific PR.

nikolasburk: Hmm, good question! I don't think I have a strong opinion and would be happy with either approach. Just "Console" is shorter and simpler, which when in doubt, I think is the preferred option. 

> That'd mean that Console and the CLI are interfaces to manage, configure and work with our products, ORM, Accelerate, Pulse, etc.

That makes sense!

> We might want to already start calling the Console "Console", to ease that simplification. It also aligns well with the URL being console.prisma.io.

I can do a find/replace to ensure we're only mentioning "Console" in this PR.

content/500-platform/10-about.mdx
nikolasburk: Sounds good! But we should mention that other users can be invited to collaborate, right?

```suggestion
- invite other users to collaborate in the workspace.
```

content/500-platform/10-about.mdx
nikolasburk: I do sometimes find it helpful to be a bit more elaborative (is that a word? 😄 ) when it helps with context. For example, we have this in the frontmatter of the About page:

```
title: 'About'
metaTitle: 'Platform Console: About'
```

In these cases, should we:
- keep "Platform Console"?
- abbreviate to "Console" (as everywhere else)?
- allow ourselves to be pragmatic and maybe disambiguate this with "PDP Console"?

nikolasburk: More occurrences in the frontmatter:

image

petradonka: Let's do Console then.

PDP Console wouldn't be something we ever say. Platform Console would be acceptable, but I think it makes sense to go with Console now.

content/500-platform/50-support.mdx
nikolasburk: I also should have been more clear: 

The thing that Nurul told me he needs to explain frequently is that people need to  explicitly disable Accelerate and Pulse in _all_ environments of _all_ projects, so that's what we should be documenting.

However, that hint feels off to me if we don't also mention here how an account is being deleted. 

My suggestion would be:
- keep everything as is
- delete the entire `## Deleting your PDP account` section

Any preferences @petradonka @nurul3101?

petradonka: It's fine, lets leave it as-is then. I can see how that'd be useful. I do doubt that people will come here to read up on this, but it doesn't hurt to have it.

petradonka (approved):

petradonka:
> ## Redirect check
> This PR probably requires the following redirects to be added to static/_redirects:
> 
> ```
> /platform/concepts/workspaces /docs/##( TODO: Path of page that replaces deleted page )##
> /platform/concepts/projects /docs/##( TODO: Path of page that replaces deleted page )##
> /platform/concepts/environments /docs/##( TODO: Path of page that replaces deleted page )##
> /platform/concepts /docs/##( TODO: Path of page that replaces deleted page )##
> ```

@nikolasburk, did we add these redirects?

content/500-platform/10-about.mdx
```diff
@@ -1,17 +1,129 @@
...
+If you need to delete your user account, go [here](/platform/support#deleting-your-pdp-account).
+
+### Workspace
+
+You can create several workspaces. A workspace is an isolated space to host projects. A workspace can have multiple user accounts associated with it so that multiple users can collaborate on the the projects in the workspace.
+
+In each workspace, you can:
+
+- view and manage all projects (and their environments) in that workspace.
+- manage billing, i.e. select a [subscription plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site), configure payment methods, or view the invoice history.
```
meletj, line 42: @nikolasburk when you fix the redirects, can you also update the UTM here to have source=docs so we're consistent and medium=platform-docs? Thank you

content/500-platform/10-about.mdx
```diff
@@ -1,17 +1,129 @@
...
+
+### Project
+
+In each workspace, you can create several projects. A project typically represents an application (a product or service). You typically have one [Prisma schema](/orm/prisma-schema/) per project.
+
+In each project, you can:
+
+- view and manage all environments in that project.
+
+The number of project you can create in a workspace depends on the [subscription plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) configured in that workspace.
```
meletj, line 54: same as above re: UTM

content/500-platform/10-about.mdx
```diff
@@ -1,17 +1,129 @@
...
+- for **Pulse**:
+  - set your database connection string (of a [Pulse-compatible database](/pulse/database-setup)).
+  - configure the _region_ where Pulse is running.
+  - configure database replication.
+
+The number of environments you can create in a project depends on the [subscription plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) configured in your workspace.
+
+## Billing
+
+The [subscription plan](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) you select in your workspace determines how many projects and environments you can create in that workspace:
```
meletj, line 78: Same as above re UTM

content/500-platform/10-about.mdx
```diff
@@ -1,17 +1,129 @@
...
+### Prorated billing
+
+All base plan prices are prorated, which means you're only billed for the duration of your subscription to a specific plan. In addition, you're also billed for any usage costs you've incurred during your subscription.
+
+For example:
+
+- if you subscribe to our **Pro** plan on the 15th day of a month, you'll only be charged the base plan price for the days left in that month.
+- if you downgrade your subscription plan (e.g. from **Business** to **Pro**) after 10 days of a 30-day month, you'll be charged for 10 days of the base price of the **Business** plan and 20 days for the base price of the **Pro** plan.
+
+Visit our [pricing page](https://prisma.io/pricing?utm_source=prisma-docs&utm_medium=docs-site) for more details.
```
meletj, line 105: same re: UTM

nikolasburk:
Thanks for the ping @meletj and @petradonka! 

I added redirects and update utm in this PR: https://github.com/prisma/docs/pull/5999

PR #5980, opened 2024-05-04
Update 055-pagination.mdx

vikaswakde:
in the given example, the number to be skipped should be `20` and not `200` to get the remaining statement correct.


jharrell (changes requested):
Hi there,

Thank you for the PR! I’m not sure your change is correct, though.

We can definitely improve the wording, but what is meant here is that `skip` is always equal to the number of pages you want to skip times the number of results per page (`take`)

Since we want to get all results on page 21, we skip `20 * 10` and then take `10`.

vikaswakde:
oh so sorry I misunderstood while reading!
i am really sorry.

PR #5985, opened 2024-05-08
fix: fix quickstart schema about the nested create many

Fim98:


nikolasburk (changes requested):
Hey @liyunfu1998, thanks a lot for this PR. I actually like the change to use `[]` because it illustrates that you can provide _multiple_ objects to the `create` option in this nested write query.

However, I would like to make this point even more clear by including a second object then:

```ts
  data: {
    name: 'Bob',
    email: 'bob@prisma.io',
    posts: {
      create: [{
        title: 'Hello World',
        published: true
      }, {
        title: 'My second post',
        content: 'This is still a draft.'
      }],
    },
  },
})
console.log(user)
}
```

Would you be open to making these additional changes?

If we change this, we also must make two other changes on that page:
- update the line higlighting annotation on the code block accordingly (`highlight=6-17;add`)
- update the terminal output in the "4.3. Explore relation queries with Prisma Client" section to include the data from the new post

Could you add these changes to the PR?

Fim98:
> Hey @liyunfu1998, thanks a lot for this PR. I actually like the change to use `[]` because it illustrates that you can provide _multiple_ objects to the `create` option in this nested write query.
> 
> However, I would like to make this point even more clear by including a second object then:
> 
> ```ts
> const user = await prisma.user.create({
>   data: {
>     name: 'Bob',
>     email: 'bob@prisma.io',
>     posts: {
>       create: [{
>         title: 'Hello World',
>         published: true
>       }, {
>         title: 'My second post',
>         content: 'This is still a draft.'
>       }],
>     },
>   },
> })
> console.log(user)
> }
> ```
> 
> Would you be open to making these additional changes?
> 
> If we change this, we also must make two other changes on that page:
> 
> * update the line higlighting annotation on the code block accordingly (`highlight=6-17;add`)
> * update the terminal output in the "4.3. Explore relation queries with Prisma Client" section to include the data from the new post
> 
> Could you add these changes to the PR?

thank you, i am updated pr, added highlight and findMany output

nikolasburk (approved):
this looks great now, thanks so much @liyunfu1998 🎉

PR #5988, opened 2024-05-08
add `createManyAndReturn()` documentation

jharrell:


content/200-orm/200-prisma-client/100-queries/030-crud.mdx
```diff
@@ -294,6 +294,58 @@ The following video demonstrates how to use `createMany()` and [faker.js](https:
...
+  coinflips: []
+}]
+```
+
+</cmdResult>
+</CodeWithResult>
+
+:::warning
+
+`createManyAndReturn()` is not available with `relationLoadStrategy: join`.
```
janpio: Why is that @Weakky?

content/200-orm/200-prisma-client/100-queries/058-transactions.mdx
```diff
@@ -95,6 +95,7 @@ The following bulk operations run as transactions:
 - `deleteMany`
 - `updateMany`
 - `createMany`
+- `createManyAndReturn`
```
janpio: opportunistic:
```suggestion
- `deleteMany()`
- `updateMany()`
- `createMany()`
- `createManyAndReturn()`
```

content/200-orm/200-prisma-client/100-queries/058-transactions.mdx
```diff
@@ -746,6 +747,7 @@ Bulk writes allow you to write multiple records of the same type in a single tra
 - `updateMany`
 - `deleteMany`
 - `createMany`
+- `createManyAndReturn`
```
janpio: opportunistic:
```suggestion
- `updateMany()`
- `deleteMany()`
- `createMany()`
- `createManyAndReturn()`
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -825,7 +825,7 @@ const user = await prisma.user.create({
 
 ##### Create multiple new records
 
-In most cases, you can carry out batch inserts with the [`createMany`](#createmany) query. However, [there are scenarios where `create` is the best option to insert multiple records](#remarks-10).
+In most cases, you can carry out batch inserts with the [`createMany()`](#createmany) or [`createManyAndReturn()`](#createmanyandreturn) queries. However, [there are scenarios where `create` is the best option to insert multiple records](#remarks-10).
```
janpio: opportunistic:
```suggestion
In most cases, you can carry out batch inserts with the [`createMany()`](#createmany) or [`createManyAndReturn()`](#createmanyandreturn) queries. However, [there are scenarios where `create()` is the best option to insert multiple records](#remarks-10).
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1222,20 +1222,20 @@ const deleteUser = await prisma.user.delete({
 
 `createMany` creates multiple records in a transaction.
 
-#### Options
+**Options**
```
janpio: This is problematic, as it will change the anchor of all the other anchors of the "Options" headlines.

Why should this not be a headline any more?

(Also applies for "Return Type", Remarks and so on below)

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1255,6 +1255,54 @@ const users = await prisma.user.createMany({
...
+`createManyAndReturn` creates multiple records and returns the resulting objects.
+
+:::info
+
+`createManyAndReturn()` is only available in Prisma ORM version 5.14.0 and up.
+`createManyAndReturn()` is only available for PostgreSQL, CockroachDB, and SQLite.
+
+:::
+
+**Options**
```
```suggestion
#### Options
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1255,6 +1255,54 @@ const users = await prisma.user.createMany({
...
+| [`select`](#select)            | `XOR<UserSelect, null>`  | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned objects.                                                                                 |
+| [`omit`](#omit-preview)        | `XOR<UserOmit, null>`    | No       | Specifies which properties to exclude on the returned objects. In [Preview](/orm/more/releases#preview) since 5.13.0. Mutually exclusive with `select`.                                            |
+| [`include`](#include)          | `XOR<UserInclude, null>` | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned objects.                                                                                 |
+| `skipDuplicates?` | `boolean`                         | No       | Do not insert records with unique fields or ID fields that already exist. Only supported by databases that support [`ON CONFLICT DO NOTHING`](https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT). This excludes MongoDB and SQLServer |
+
+**Remarks**
```
```suggestion
#### Remarks
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1255,6 +1255,54 @@ const users = await prisma.user.createMany({
...
+
+- The `skipDuplicates` option is not supported by SQLite.
+- You **cannot** create or connect relations by using nested `create`, `createMany`, `connect`, `connectOrCreate` queries inside a top-level `createManyAndReturn()` query.
+- When relations are included via `include`, a separate query is generated per relation.
+- `relationLoadStrategy: join` is not supported.
+
+**Return type**
```
```suggestion
#### Return type
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
janpio: This suggested change will change the numbering of all the anchors for the "Options" headlines below this one and will require manual updates and fixes across the whole docs unfortunately.

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1255,6 +1255,54 @@ const users = await prisma.user.createMany({
...
+**Create and return several new users**
+
+```ts
+const users = await prisma.user.createManyAndReturn({
+  data: [
+    { name: 'Sonali', email: 'sonali@prisma.io' },
+    { name: 'Alex', email: 'alex@prisma.io' },
+  ],
+})
+```
```
janpio, line 1308: Should this have an example for what is being returned?

jharrell:
@mhessdev

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1244,7 +1244,7 @@ const deleteUser = await prisma.user.delete({
 
 #### Examples
 
-##### Create several new users
+**Create several new users**
```
```suggestion
##### Create several new users
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1255,6 +1255,54 @@ const users = await prisma.user.createMany({
...
+`createManyAndReturn` creates multiple records and returns the resulting objects.
+
+:::info
+
+`createManyAndReturn()` is only available in Prisma ORM version 5.14.0 and up.
+`createManyAndReturn()` is only available for PostgreSQL, CockroachDB, and SQLite.
+
+:::
+
+#### Options
```
janpio, line 1269: This addition changes the anchor "name" of all the "Options" headlines below this (as those are just incrementally numbered). Are there no links to these headlines anywhere in our docs that need to be adapted (plus 1 in the anchor name)? Otherwise these links will not point to the _wrong_ headline.

content/200-orm/200-prisma-client/100-queries/030-crud.mdx
janpio: Isn't it the other way around anyway, that `relationLoadStrategy: join` is not available with `createManyAndReturn()`?
If so, why is this worth mentioning here? I could not find a similar mention for `createMany()` where I guess the same is also true?

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -1255,6 +1255,54 @@ const users = await prisma.user.createMany({
...
+| [`select`](#select)            | `XOR<UserSelect, null>`  | No       | [Specifies which properties to include](/orm/prisma-client/queries/select-fields) on the returned objects.                                                                                 |
+| [`omit`](#omit-preview)        | `XOR<UserOmit, null>`    | No       | Specifies which properties to exclude on the returned objects. In [Preview](/orm/more/releases#preview) since 5.13.0. Mutually exclusive with `select`.                                            |
+| [`include`](#include)          | `XOR<UserInclude, null>` | No       | [Specifies which relations should be eagerly loaded](/orm/prisma-client/queries/relation-queries) on the returned objects.                                                                                 |
+| `skipDuplicates?` | `boolean`                         | No       | Do not insert records with unique fields or ID fields that already exist. Only supported by databases that support [`ON CONFLICT DO NOTHING`](https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT). This excludes MongoDB and SQLServer |
```
Jolg42, line 1277: I'm not sure if we have a test for `skipDuplicates` with `createManyAndReturn` @Weakky do you know?

content/200-orm/200-prisma-client/100-queries/030-crud.mdx
Weakky: > Why is that @Weakky?

Because it's impossible to join relations on a `RETURNING` clause, so we have to send separate queries for every relation. But we don't have in-memory joining implemented for `relationLoadStrategy: join`.


> Isn't it the other way around anyway, that relationLoadStrategy: join is not available with createManyAndReturn()?

I'm not sure I understand the nuance between relationLoadStrategy not being available with createManyAndReturn and the other way around?

> I could not find a similar mention for createMany() where I guess the same is also true?

It is irrelevant for `createMany` because you cannot fetch relations anyway.

All in all, I don't mind removing this if you think it's confusing.

Weakky: It is missing an important point: only relations that have their foreign key present in the model that is being created can be fetched.

So `User.post` can be fetched in this case:

```prisma
model User {
  id Int @id
  
  postId Int? @unique
  post Post? @relation(fields: [postId], references: [id])
}

model Post {
  id Int @id

  author User?
}
```

But cannot in all these cases:

1) 1-1 not inlined in `User`

```prisma
model User {
  id Int @id
  
  post Post?
}

model Post {
  id Int @id

  authorId Int? @unique
  author User? @relation(fields: [authorId], references: [id])
}
```

2) 1-m

```prisma
model User {
  id Int @id
  
  posts Post[]
}

model Post {
  id Int @id

  authorId Int? @unique
  author User? @relation(fields: [authorId], references: [id])
}
```

3) m2m

```prisma
model User {
  id Int @id
  
  posts Post[]
}

model Post {
  id Int @id

  authors User[]
}
```

janpio: > I'm not sure I understand the nuance between relationLoadStrategy not being available with createManyAndReturn and the other way around?

Let me expand:

> a) `createManyAndReturn()` is not available with `relationLoadStrategy: join`.

This implies that you can top level define `relationLoadStrategy: join` and then `createManyAndReturn()` will not be present in the API.

On the other hand:

> b) `relationLoadStrategy: join` is not available with `createManyAndReturn()`

This says that  `relationLoadStrategy: join` is not a valid option for the configuration inside a `createManyAndReturn()`.

If I understood correctly, you want to communicate b).

content/200-orm/500-reference/050-prisma-client-reference.mdx
jharrell: Double checked. We only link to `#options-7` in the docs, which happens before this.

content/200-orm/200-prisma-client/100-queries/030-crud.mdx
jharrell: > It is missing an important point: only relations that have their foreign key present in the model that is being created can be fetched.

@Weakky could I get some clarity here? Is this the case because it's not possible to create relations where the foreign key is not present in `createMany()` so that limitation is passed to `createManyAndReturn()`?

Weakky: Yes. It's not a `createMany` limitation per se, it's a database limitation.

content/200-orm/500-reference/050-prisma-client-reference.mdx
Weakky: We do have a test

Weakky: https://github.com/prisma/prisma-engines/blob/54465718776ea99c622bd1c060a5fa445952886d/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many_and_return.rs#L214-L228

PR #6002, opened 2024-05-14
fix(sql-server): Reorder and fix connection string information

janpio:
- Reorder so that arguments are close to example string
- Fix headline words and structure

### DIFF
--- a/content/200-orm/050-overview/500-databases/800-sql-server/index.mdx
+++ b/content/200-orm/050-overview/500-databases/800-sql-server/index.mdx
@@ -35,7 +35,7 @@ The connection URL used to connect to an Microsoft SQL Server database follows t
 The following example uses SQL authentication (username and password) with an enabled TLS encrypted connection:
 
 ```
-sqlserver://HOST:PORT;database=DATABASE;user=USER;password=PASSWORD;encrypt=true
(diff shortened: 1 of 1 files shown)


content/200-orm/050-overview/500-databases/800-sql-server/index.mdx
```diff
@@ -35,7 +35,7 @@ The connection URL used to connect to an Microsoft SQL Server database follows t
 The following example uses SQL authentication (username and password) with an enabled TLS encrypted connection:
 
 ```
-sqlserver://HOST:PORT;database=DATABASE;user=USER;password=PASSWORD;encrypt=true
+sqlserver://HOST[:PORT];database=DATABASE;user=USER;password=PASSWORD;encrypt=true
```
nikolasburk, line 38: just for my own understanding: the square brackets are _actually_ part of the connection string, right? i.e. if you don't have square brackets, the connection string is invalid?

nikolasburk (approved):
had one small question but nothing that should block this, thanks Jan!

content/200-orm/050-overview/500-databases/800-sql-server/index.mdx
janpio: Ah no, this means that the port is optional. The named instances just have a "hostname" and the port is figured out under the hood (see the example below), so it can just be left out.

PR #6030, opened 2024-05-22
fix(raw-queries): Rewrite "PostgreSQL typecasting fixes" to "Raw query typecasting behavior"

janpio:
This was a super weird section that did not follow our usual style at all.

### DIFF
--- a/content/200-orm/200-prisma-client/100-queries/090-raw-database-access/050-raw-queries.mdx
+++ b/content/200-orm/200-prisma-client/100-queries/090-raw-database-access/050-raw-queries.mdx
@@ -414,9 +414,9 @@ The solution in this case is to explicitly cast `42` to the `text` type:
 await prisma.$queryRaw`SELECT LENGTH(${42}::text);`
 ```
 
-<Admonition type="info">
+:::info
 
-**Feature availability:** This funtionality is [Generally Available](/orm/more/releases#generally-available-ga) since version 4.0.0. In v3.14.x and v3.15.x, it was were available with the preview feature `improvedQueryRaw`.
+**Feature availability:** This funtionality is [Generally Available](/orm/more/releases#generally-available-ga) since version 4.0.0. In v3.14.x and v3.15.x, it was available with the preview feature `improvedQueryRaw`.
 
 For the example above before version 4.0.0, Prisma ORM silently coerces `42` to `text` and does not require the explicit cast. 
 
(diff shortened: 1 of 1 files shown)


content/200-orm/200-prisma-client/100-queries/090-raw-database-access/050-raw-queries.mdx
```diff
@@ -391,44 +391,46 @@ The following table shows the conversion between types used in the database and
 
 Note that the exact name for each database type will vary between databases – for example, the boolean type is known as `boolean` in PostgreSQL and `STRING` in CockroachDB. See the [Scalar types reference](/orm/reference/prisma-schema-reference#model-field-scalar-types) for full details of type names for each database.
 
-### PostgreSQL typecasting fixes
+### Raw query typecasting behavior
 
-Prisma ORM resolves a number of issues with typecasting in PostgreSQL.
+Raw queries with Prisma Client might require parameters to be in the expected types of the SQL function or query and does not do subtle implicit casts.
```
jharrell:
```suggestion
Raw queries with Prisma Client might require parameters to be in the expected types of the SQL function or query. Prisma Client does not do subtle, implicit casts.
```

content/200-orm/200-prisma-client/100-queries/090-raw-database-access/050-raw-queries.mdx
```diff
@@ -391,44 +391,46 @@ The following table shows the conversion between types used in the database and
...
+On the other hand the following raw query now works correctly, returning an integer result, and failed before:
+
+```ts
+await prisma.$queryRaw`SELECT ${1.5}::int as int`
+
+// Now: [{ int: 2 }]
+// Before: db error: ERROR: incorrect binary data format in bind parameter 1
+```
+
+</Admonition>
```
jharrell:
````suggestion
:::info

**Feature availability:** This funtionality is [Generally Available](/orm/more/releases#generally-available-ga) since version 4.0.0. In v3.14.x and v3.15.x, it was available with the preview feature `improvedQueryRaw`.

For the example above before version 4.0.0, Prisma ORM silently coerces `42` to `text` and does not require the explicit cast. 

On the other hand the following raw query now works correctly, returning an integer result, and failed before:

```ts
await prisma.$queryRaw`SELECT ${1.5}::int as int`

// Now: [{ int: 2 }]
// Before: db error: ERROR: incorrect binary data format in bind parameter 1
```

:::
````

jharrell (approved):

PR #6034, opened 2024-05-23
feat: WEB-381 Run prettier on docs repo

carlagn:
Fixes [#WEB-381](https://linear.app/prisma-company/issue/WEB-381/run-prettier-config-on-all-docs-files)

⚠️ I didn't **include** `*.mdx` because we've had issues before of layouts breaking due to certain off-placed formatting.


jharrell:
How much would I have to pay for us to consider single quotes 😆

PR #6043, opened 2024-05-28
Update 250-opentelemetry-tracing.mdx: use `SEMRESATTRS_XXXXX` constants and remove deprecated imports

Jolg42:
Removed deprecated import and replace.

Random find of the day.

See 
```
/**
 * Definition of available values for SemanticResourceAttributes
 * This type is used for backward compatibility, you should use the individual exported
 * constants SemanticResourceAttributes_XXXXX rather than the exported constant map. As any single reference
 * to a constant map value will result in all strings being included into your bundle.
 * @deprecated Use the SEMRESATTRS_XXXXX constants rather than the SemanticResourceAttributes.XXXXX for bundle minification.
 */
export declare type SemanticResourceAttributes = {
```
....


Jolg42:
Note, I lost rights to create a branch on the docs repo at some point 🤷🏼 
So this is from a fork.

jharrell (approved):

PR #6048, opened 2024-05-29
feat(docs): add drizzle migration guide

jharrell:


nikolasburk (approved):
Generally looks good to me but I noticed that we're missing a step to _generate_ Prisma Client right with `npx prisma generate`, right? I also checked the other guides and ti seems like this step is missing there two. 

A fix could be as easy as updating step 3:
- update the section heading to: `## Step 3. Install and generate Prisma Client`
- add another short paragraph the code snippet with `npx prisma generate`

PR #6049, opened 2024-05-30
Link deleteMany in CRUD reference

olliechick:
Consistent with other links in sections of that page


olliechick:
I'm not sure why the checks aren't successful; I've used the same format as other links on that page from what I can tell

janpio:
I think those just do not work with PRs from forks - the changes look technically fine to me.

jharrell (approved):

jharrell:
Thank you @olliechick ! @janpio is correct, our checks currently don't work with forks correctly.

PR #6069, opened 2024-06-05
fix(full-text-search): fix _relevance headline

janpio:
No idea why it said incorrect `\_relevance` besides it being funny: `\_o_/`


jharrell (approved):
@janpio the character needed to be escaped with Gatsby. Thank you for the catch 😄

PR #6081, opened 2024-06-08
fix(deploy-to-cloudflare): Use --datasource-provider sqlite in prisma init

janpio:
…


content/200-orm/200-prisma-client/500-deployment/301-edge/450-deploy-to-cloudflare.mdx
```diff
@@ -131,7 +131,7 @@ If you don't have that yet, you can run these commands:
 npm create cloudflare@latest prisma-cloudflare-worker-example -- --type hello-world
 cd prisma-cloudflare-worker-example
 npm install prisma --save-dev
-npx prisma init
+npx prisma init --datasource-provider sqlite
```
jharrell, line 134: I'm not sure if this is correct. The next section is `PostgreSQL (traditional)` which would leave me to believe that if you followed these instructions you would run into issues.

janpio: Good point, I was coming from the D1 docs which has 0 instructions on how to set it up but just links here. I guess D1 users have to figure that out themselves.

PR #6087, opened 2024-06-10
Clarify stance on computation functions

jharrell:
Resolves DA-305


janpio (commented):
Shouldn't we just _remove_ the old approach? Or hide it at least by default behind a `<details>`?

jharrell:
@janpio my logic for not removing the section entirely was because we still have documentation for versions of Prisma pre-4.16.0

hiding behind details is a good idea. I'll do that.

janpio (approved):

PR #6097, opened 2024-06-12
improve api reference docs for pulse return types

nikolasburk:


janpio (commented):
Maybe I am just not smart enough, but here is my feedback that this might be to complicated for me: https://prisma-company.slack.com/archives/C053YT1U9M0/p1718183855095329?thread_ts=1718127951.708439&cid=C053YT1U9M0

ankur-arch (approved):

nikolasburk:
Merged as is @janpio since this seems to be the most natural way to display TS typings. Let's see if we get further feedback for this 🙏

PR #6103, opened 2024-06-13
tweak display of code in headings

jharrell:
Resolves DA-349


janpio:
Definitely improves it for the "bigger" headlines, but for smaller ones the padding is now "missing" that makes the grey background detectable visually. For example on https://fix-inline-code.docs-51g.pages.dev/about/docs-components/frontmatter

janpio:
Another example where it is now a bit too little: 
image
(left old, right new)

janpio (approved):

PR #6112, opened 2024-06-18
Add local and global omit documentation

jharrell:
Resolves DA-327

Relates to DA-315


content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
-Prisma Client doesn't have a native way of excluding fields yet, but it's easy to create a function that you can use to exclude certain fields in a type-safe way.
+:::info
+
+As of Prisma ORM 5.16.0, excluding fields globally and locally is supported via the `omitApi` Preview feature.
+
+:::
+
+## Excluding a field globally using `omit`
+
+The following is a type-safe way to exclude a field globally using the `omitApi` Preview feature.
```
ankur-arch:
```suggestion
The following is a type-safe way to exclude a field globally using the [`omitApi` Preview feature](/orm/reference/preview-features):
```

ankur-arch: Maybe linking to what preview features are might be a good idea?
https://www.prisma.io/docs/orm/reference/preview-features

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
+
+</TabItem>
 
-</TopBlock>
+</TabbedContent>
+
+## Excluding a field locally using `omit`
 
-## Excluding the password field
+The following is a type-safe way to exclude a field locally using the `omitApi` Preview feature.
```
ankur-arch:
```suggestion
The following is a type-safe way to exclude a field locally using the `omitApi` Preview feature:
```

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
+})
+```
+
+## When to use `omit` globally or locally
+
+When beginning to use `omit`, it's important to understand why a field is being omitted in the first place. 
+
+- If you are omitting a field due to security reasons, it's best to omit it globally. 
+- If you are omitting a field because it's not needed in a query, it's best to omit it locally.
+
```
ankur-arch:
```suggestion
It's important to understand when to omit a field globally or locally:

- If you are omitting a field due to security reasons, it's best to omit it globally. For example: Globally omitting the password field from a User model so that sensitive information doesn't accidentally get exposed.
- If you are omitting a field because it's not needed in a query, it's best to omit it locally.
```

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
+## Excluding the password field without using `omit`
+
+:::note
+
+The `omitApi` Preview feature, released in Prisma ORM 5.13.0, is the preferred way of omitting fields.
```
ankur-arch:
```suggestion
The `omitApi` Preview feature, released in Prisma ORM 5.13.0, is the preferred way of omitting fields from a query result.
```

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
+</TabbedContent>
+
+## How to select a previously omitted field
+
+If you [omit a field globally](#excluding-a-field-globally-using-omit), you can select it by setting `omit` to `false` in any query.
+
+```tsx
+const user = await prisma.user.findUnique({
+  omit: {
+    password: false // The password field is now selected.
```
mhessdev, line 158: Should we also add an example of doing this with multiple fields?

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
+  },
+  where: {
+    id: 1
+  }
+})
+```
+
+## When to use `omit` globally or locally
```
mhessdev, line 170: maybe also use this section to describe more of the differences? For instance global omit is used in the definition of the client and the local omit is used on the query level.

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,137 @@ metaTitle: 'Excluding fields'
...
+    id: 1 
+  } 
+})
+```
+
+</TabItem>
+
+</TabbedContent>
+
+## How to select a previously omitted field
```
jharrell, line 130:
````suggestion
## How to omit multiple fields

Omitting multiple fields works the same as selecting multiple fields: add multiple key-value pairs to the omit option.
Using the same schema as before, you could omit password and email with the following:

```tsx

// password and email are excluded
const user = await prisma.user.findUnique({
  omit: {
    email: true,
    password: true,
  },
  where: {
    id: 1,
  },
})
```

Multiple fields can be omitted locally and globally.

## How to select a previously omitted field
````

File: content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx (hunk 5 above)
Comment (jharrell) [reply content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx:158]: added above

File: content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx [hunk 8]
```diff
@@ -4,15 +4,158 @@ metaTitle: 'Excluding fields'
...
+})
+```
 
-## Excluding the password field
+## When to use `omit` globally or locally
+
+It's important to understand when to omit a field globally or locally:
+
+- If you are omitting a field due to security reasons, it's best to omit it globally. For example: Globally omitting the password field from a User model so that sensitive information doesn't accidentally get exposed.
+- If you are omitting a field because it's not needed in a query, it's best to omit it locally.
```
jharrell:
```suggestion
It's important to understand when to omit a field globally or locally:

- If you are omitting a field due to security reasons, it's best to omit it globally. For example: Globally omitting the password field from a User model so that sensitive information doesn't accidentally get exposed.
- If you are omitting a field because it's not needed in a query, it's best to omit it locally.

Locally omitted fields are omitted at the query level. This means that what fields are omitted can differ per-query, but there is no "default" omitted fields. On the other hand, globally omitted fields are defined at when a Prisma Client is instantiated and are then automatically omitted from every subsequent query on that Client. 

To omit different fields or select previous omitted fields, you will need to instantiate a new Prisma Client or [override a global omit with a local select](#how-to-select-a-previously-omitted-field).

```

ankur-arch (dismissed):

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,163 @@ metaTitle: 'Excluding fields'
...
+    id: 1,
+  },
+})
+```
+
+Multiple fields can be omitted locally and globally.
+
+## How to select a previously omitted field
+
+If you [omit a field globally](#excluding-a-field-globally-using-omit), you can select it by setting `omit` to `false` in any query.
```
SevInf: Global `omit` also does not apply to explicit `select`.
So, `{ select: { password: true } }` will also include the password.

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,163 @@ metaTitle: 'Excluding fields'
...
+
+Locally omitted fields are omitted at the query level. This means that what fields are omitted can differ per-query, but there is no "default" omitted fields. On the other hand, globally omitted fields are defined at when a Prisma Client is instantiated and are then automatically omitted from every subsequent query on that Client. 
+
+To omit different fields or select previous omitted fields, you will need to instantiate a new Prisma Client or [override a global omit with a local select](#how-to-select-a-previously-omitted-field).
+
+## Excluding the password field without using `omit`
+
+:::note
+
+The `omitApi` Preview feature, released in Prisma ORM 5.13.0, is the preferred way of omitting fields from a query result.
```
SevInf: Should we mention that global omit is/was not available until 5.16.0?

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,163 @@ metaTitle: 'Excluding fields'
...
+    id: 1
+  }
+})
+```
+
+## When to use `omit` globally or locally
+
+It's important to understand when to omit a field globally or locally:
+
+- If you are omitting a field due to security reasons, it's best to omit it globally. For example: Globally omitting the password field from a User model so that sensitive information doesn't accidentally get exposed.
```
SevInf: Not sure I like `security reasons` here. Sounds like it is too much for what this is - yes, it can protect you from accidental data leak, but it is very easy to access omitted file if you need it - "security feature" might imply stronger guarantees. Maybe expand on what we mean here or be even more specific that we mean only accidental data leaks.

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,15 +4,163 @@ metaTitle: 'Excluding fields'
...
+- If you are omitting a field because it's not needed in a query, it's best to omit it locally.
+
+Locally omitted fields are omitted at the query level. This means that what fields are omitted can differ per-query, but there is no "default" omitted fields. On the other hand, globally omitted fields are defined at when a Prisma Client is instantiated and are then automatically omitted from every subsequent query on that Client.
```
SevInf: Might be my evening-before-vacation brain, but I for some reason had really hard time parsing this sentence. Maybe there is an easier way to say "local omit affects only query it specified in, global affects every query without explicit `select`"?

mhessdev (approved):
Please rebase before merging

PR #6113, opened 2024-06-18
docs: install nuxt module using nuxi cli

RihanArfan:
Changes install docs to install using `nuxi module add`. This is the new standard way of installing modules across the Nuxt ecosystem, as it automatically adds the module to `nuxt.config.ts`.


jharrell (approved):
Thanks for catching this!

PR #6125, opened 2024-06-25
fix(deployment): use absolute path for prisma directory

luisjunco:
In the section "Deploying database changes using GitHub Actions", the current version of the docs uses a relative path to point to the prisma directory and, as a result, the workflow will never be triggered. This PR solves that issue, by using an [absolute path](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths).

Current: `./prisma/migrations/**`
After PR: `prisma/migrations/**`


jharrell (approved):
Thank you for your contribution!

PR #6130, opened 2024-06-26
fix: WEB-497 Revise external link arrow

carlagn:
For a non-breaking space, adding the external link icon with `background-image`.

Fixes #WEB-497

### DIFF
--- a/src/theme/MDXComponents.tsx
+++ b/src/theme/MDXComponents.tsx
@@ -13,6 +13,7 @@ import CollapseBox from "@site/src/components/collapsible";
 import TopSection from "@site/src/components/topSection";
 import { useLocation } from "@docusaurus/router";
 import styles from "./styles.module.scss";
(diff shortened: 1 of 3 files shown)


mhessdev (approved):

src/theme/MDXComponents.tsx
```diff
@@ -119,20 +120,8 @@ const StyledLink: React.FC<React.PropsWithChildren<ComponentProps<"a">>> = ({
     return <Link {...props}>{children}</Link>;
   else
     return (
-      <a {...props} target="_blank" rel="openeer noreferrer">
+      <a {...props} target="_blank" rel="openeer noreferrer" className={clsx(props.className, styles.externalLink)}>
```
jharrell, line 123: Would it be possible for us to use a `<Link>` component here instead of `a`?

PR #6131, opened 2024-06-27
fix: remove connection pooling calculation

ankur-arch:


jharrell (approved):

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -20,15 +20,7 @@ For more information about connection pooling in Prisma, see the documentation [
 
 ### Default connection pool size
 
-By default, Accelerate calculates a default connection pool size using the formula `num_physical_cpus * 2 + 1`.
```
mhessdev, line 23: We don't need any of this information? Seems odd to remove all of this and replace it with a single line

mhessdev (approved):

content/300-accelerate/250-connection-pooling.mdx
jharrell: It sounds like this information was false. There is no default connection pool size calculation, it's just 10.

(at least that's what I gather from this PR)

ankur-arch: @mhessdev, @jharrell the default is 10, hence the simple line 😄 . Here's a [comment](https://prisma-company.slack.com/archives/C04F60M3G06/p1718715175978299?thread_ts=1718704568.817239&cid=C04F60M3G06) from Marcus.

PR #6134, opened 2024-06-28
enhance: make reconfiguring section more obvious

ankur-arch:
User's are missing the section which is important, hence convert it into a tip so that it's more obvious.

### DIFF
--- a/content/400-pulse/200-getting-started.mdx
+++ b/content/400-pulse/200-getting-started.mdx
@@ -89,11 +89,15 @@ const prisma = new PrismaClient().$extends(
 )
 ```
 
+:::tip
+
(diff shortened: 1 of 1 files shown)


content/400-pulse/200-getting-started.mdx
```diff
@@ -89,12 +89,16 @@ const prisma = new PrismaClient().$extends(
 )
 ```
 
+:::tip
```
nikolasburk: I think a warning may be more appropriate than a tip maybe?

ankur-arch: I agree that would be better 👍 !

nikolasburk (dismissed):

nikolasburk (approved):

PR #6135, opened 2024-06-29

rawatdev:
Add the correct mongodb code in it instead of sql.


jharrell (approved):
Thank you for the catch and the fix!

PR #6138, opened 2024-07-02
fix(docs): correct relation fields documentation

### LINKED ISSUE #4354: Review the use of "relation fields" in Prisma Client and Prisma schema docs
Feedback from @janpio in https://prisma-company.slack.com/archives/C5Z9TH6N9/p1672819200234839:

I think this usage of "relation fields" is slightly misleading:
https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#select-specific-relation-fields
vs.
https://www.prisma.io/docs/concepts/components/prisma-schema/relations#relation-fields
The first one are the fields of a relation that should be returned via select, but not really "relation fields" as we usually define them (as the fields that define a relation with @relation)

Some more feedback for the "Relation Fields" docs:
https://www.prisma.io/docs/concepts/components/prisma-schema/relations#relation-fields

1. It would probably be great to just highlight the relation fields in the Prisma schema examples we have there instead of relying on the table below.
2. The Prisma schema and the tables below have a different order of fields (author vs. authorId). That makes things harder to follow.
3. "annotated relation field" could already link to the headline below - not everyone will naturally make that connection.
4. The "relation scalar field" is sometimes just called "relation scalar" (without "field")
5. Is the [note about "read only"](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#relation-scalar-fields) correct?

jharrell:
resolves DA-671
resolves #4354

### DIFF
--- a/content/200-orm/100-prisma-schema/20-data-model/20-relations/index.mdx
+++ b/content/200-orm/100-prisma-schema/20-data-model/20-relations/index.mdx
@@ -295,7 +295,7 @@ model Category {
 
 <Admonition type="info">
 
-This schema is the same as the [example data model](/orm/prisma-schema/data-model/models) but has all [scalar fields](/orm/prisma-schema/data-model/models#scalar-fields) removed (except for the required [relation scalars](/orm/prisma-schema/data-model/relations#relation-scalar-fields)) so you can focus on the [relation fields](#relation-fields).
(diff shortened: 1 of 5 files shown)


mhessdev (dismissed):

mhessdev:
Fix build before merging please :D

nikolasburk (approved):
lgtm 👍

PR #6140, opened 2024-07-03
feat: add faq for accelerate transaction billing #DA-935

ankur-arch:


content/300-accelerate/600-faq.mdx
```diff
@@ -125,3 +125,27 @@ In the rare event of a service disruption, falling back to a direct connection w
 If there is a service disruption, it's recommended to verify on the [status page](https://pris.ly/data-platform-status). You can reach out to one of Prisma's [support channels](/platform/support) for assistance.
 
 > **Note:** Additionally, it's worth noting that some edge function runtime environments may not support direct connections with Prisma ORM. For further details, refer to our [Edge functions documentation](/orm/prisma-client/deployment/edge/overview).
+
+Sure! Here's the revised FAQ item:
```
mhessdev: Why is this here? Forget to remove the chatgpt part?

mhessdev (changes requested):

content/300-accelerate/600-faq.mdx
```diff
@@ -125,3 +125,27 @@ In the rare event of a service disruption, falling back to a direct connection w
...
+However, when using the [`$transaction` API for sequential client operations](/orm/prisma-client/queries/transactions#sequential-prisma-client-operations), regardless of the number of queries within the array, it counts as only one billable query. For example:
+
+```ts
+await prisma.$transaction([
+  prisma.user.deleteMany({ where: { name: 'John Doe' } }),
+  prisma.user.createMany({ data }),
+]);
+```
+
+If you don't need the additional functionality of [interactive transactions](/orm/prisma-client/queries/transactions#interactive-transactions), you can save costs by relying on [array-based transactions](/orm/prisma-client/queries/transactions#sequential-prisma-client-operations) instead.
```
rtbenfield: The array format also performs better on Accelerate as it can execute as one round-trip to the database, whereas interactive transactions necessitate a round-trip to for start, commit, and each individual operation on the transaction. This may be worth calling out as another incentive.

content/300-accelerate/600-faq.mdx
ankur-arch: My bad.

ankur-arch:
```suggestion
```

content/300-accelerate/600-faq.mdx
ankur-arch: Thanks @rtbenfield , will add that in too.

mhessdev (dismissed):

content/300-accelerate/600-faq.mdx
```diff
@@ -125,3 +125,25 @@ In the rare event of a service disruption, falling back to a direct connection w
...
+However, when using the [`$transaction` API for sequential client operations](/orm/prisma-client/queries/transactions#sequential-prisma-client-operations), regardless of the number of queries within the array, it counts as only one billable query. For example:
+
+```ts
+await prisma.$transaction([
+  prisma.user.deleteMany({ where: { name: 'John Doe' } }),
+  prisma.user.createMany({ data }),
+]);
+```
+
+If you don't need [interactive transactions](/orm/prisma-client/queries/transactions#interactive-transactions), you can save costs and improve performance by using [array-based transactions](/orm/prisma-client/queries/transactions#sequential-prisma-client-operations). Array-based transactions perform better on Accelerate because they execute in one round-trip to the database, while interactive transactions require separate round-trips for start, commit, and each individual operation on the transaction.
```
ankur-arch: @rtbenfield added the perf incentive here aswell. Let me know if it looks good to you.

rtbenfield: This looks great 👍🏻  I think the "sequential operations transaction API" name you had previously is what the ORM docs use as well as ["$transaction([]) API"](https://www.prisma.io/docs/orm/prisma-client/queries/transactions#transaction-api). I just call it the array format because that is usually more recognizable what I'm referring to. I'm not sure which is more correct in this case.

ankur-arch:
```suggestion
If you don't need [interactive transactions](/orm/prisma-client/queries/transactions#interactive-transactions), you can save costs and improve performance by using [sequential operations transactions](/orm/prisma-client/queries/transactions#sequential-prisma-client-operations). Sequential operations transactions perform better on Accelerate because they execute in one round-trip to the database, while interactive transactions require separate round-trips for start, commit, and each individual operation on the transaction.
```

rtbenfield (approved):

PR #6141, opened 2024-07-03
fix: document proper pool timeout behaviour #DA-919

ankur-arch:


content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -40,7 +40,7 @@ To change the size of the connection pool:
 
 ### Configuring the connection pool timeout
 
-The connection pool timeout is the duration, measured in seconds, during which the query engine must process a specific query; failing to do so within this time frame results in an exception being thrown, and the system proceeds to the next query in the queue.
+The connection pool timeout is the maximum number of seconds to wait for a new connection to be opened. If the connection is not opened within this time frame, an exception is thrown. A value of 0 means there is no timeout.
```
jharrell: > The connection pool timeout is the maximum number of seconds to wait for a new connection to be opened.

A connection from where, to where? I think that's the critical information missing here.

ankur-arch:
```suggestion
The connection pool timeout is the maximum number of seconds to wait for a new connection to be opened. If the connection is not opened from the pool to the database within this time frame, an exception is thrown. A value of 0 means there is no timeout.
```

ankur-arch: @jharrell is this better?

jharrell: Yeah that's in the right direction. I think what I'm trying to get at is that a user should understand where the problem is going wrong. Is it between Prisma Client and Accelerate? Accelerate and the DB? 

From your update it seems we mean the latter. So I'd tweak your copy to something like this 
```suggestion
The connection pool timeout is the maximum number of seconds that Accelerate's internal connection pool waits for a new connection to your database to succeed and open. If the connection is not opened within this time frame, an exception is thrown. A value of 0 means there is no timeout.
```

Something like that, verbatim not needed just more clear to an end user.

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -40,7 +40,7 @@ To change the size of the connection pool:
 
 ### Configuring the connection pool timeout
 
-The connection pool timeout is the duration, measured in seconds, during which the query engine must process a specific query; failing to do so within this time frame results in an exception being thrown, and the system proceeds to the next query in the queue.
+The connection pool timeout is the maximum number of seconds that Accelerate's internal connection pool waits for a new connection to your database to succeed and open. If the connection is not opened within this time frame, an exception is thrown. A value of 0 means there is no timeout.
```
rtbenfield: Slightly more nuance here 🙂 This isn't the time to establish the connection, but rather the time the query will wait in the queue for a free connection to be available.

```suggestion
The connection pool timeout is the maximum number of seconds that a query will block while waiting for a connection from Accelerate's internal connection pool. This occurs if the number of concurrent requests exceeds the connection limit, resulting in queueing of additional requests until a free connection becomes available. An exception is thrown if a free connection does not become available within the pool timeout. The connection pool timeout can be disabled by setting the value to 0.
```

rtbenfield (approved):

PR #6142, opened 2024-07-03
fix(docs): better docs for case-insensitive filtering with SQLite. 

### LINKED ISSUE #1858: Improve case sensitivity section for SQLite provider
image

This [section](https://www.prisma.io/docs/concepts/components/prisma-client/case-sensitivity#sqlite-provider)'s wording is a bit confusing on whether or not SQLite supports case-insensitive filtering or not.
It should be clear that the default filtering mode is case-sensitive and that SQLite supports case insensitive only if your data is pure ASCII characters.

It's also not clear how I would actually use this workaround:

```SQL
CREATE TABLE mytable (
    sample TEXT COLLATE NOCASE   /* collating sequence NOCASE */
);
```

So might be a good idea to include a resource that clarifies where to include this snippet

jharrell:
Resolves DA-460
Resolves #1858

### DIFF
--- a/content/200-orm/200-prisma-client/100-queries/070-case-sensitivity.mdx
+++ b/content/200-orm/200-prisma-client/100-queries/070-case-sensitivity.mdx
@@ -170,13 +170,35 @@ The MongoDB uses a RegEx rule for case-insensitive filtering.
 
 ### SQLite provider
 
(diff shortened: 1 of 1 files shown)


mhessdev (dismissed):

ankur-arch:
@jharrell should we add a case for when user's want to apply case-insensitive filtering to an existing field (already migrated).

jharrell:
@ankur-arch Thanks for the suggestion. I updated the docs with that and it's good for review again.

mhessdev (approved):

PR #6143, opened 2024-07-03
fix(docs): document behavior of `not` with `NULL` values

### LINKED ISSUE #3812: Improve documentation around `IS NULL` and `not` filter condition
We should add something about database `NULL` handling to the `not` filter condition to make this more discoverable. Eeither a remark, or a link to a page that explains the whole topic better - it is a recurring problem for our users. (Might also apply to other filter conditions)

Context: https://prisma-company.slack.com/archives/C03V1UE750D/p1663774582942709

Related issue for filter conditions: https://github.com/prisma/docs/issues/3813

jharrell:
Resolves DA-623
Resolves #3812

### DIFF
--- a/content/200-orm/500-reference/050-prisma-client-reference.mdx
+++ b/content/200-orm/500-reference/050-prisma-client-reference.mdx
@@ -3560,13 +3560,34 @@ const result = await prisma.user.findMany({
 })
 ```
 
+:::warning
+
+`not` will return all items that do not match a given value. However, if the column is nullable, `NULL` values will not be returned. If you require null values to be returned, use an [`OR`](#or) to include `NULL` values.
+
+:::
+
+##### Return all users where `name` does **not** equal `"Eleanor"` **including** users where `name` is `NULL`
+
+```ts
+await prisma.user.findMany({
+  where: {
+    OR: [
+      { name: { not: 'Eleanor' } },
+      { name: null }
+    ]
+  }
+})
+```
+
 ### `in`
 
 Value `n` exists in list.
 
-#### Remarks
+:::note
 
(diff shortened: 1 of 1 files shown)


mhessdev (dismissed):

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -3560,13 +3560,34 @@ const result = await prisma.user.findMany({
...
+
 ### `in`
 
 Value `n` exists in list.
 
-#### Remarks
+:::note
 
-- `null` values are not returned. For example, if you combine `in` and `NOT` to return user whose name is _not_ in the list, users with `null` value names are not returned.
+`null` values are not returned. For example, if you combine `in` and `NOT` to return user whose name is _not_ in the list, users with `null` value names are not returned.
```
ankur-arch:
```suggestion
`null` values are not returned. For example, if you combine `in` and `NOT` to return a user whose name is _not_ in the list, users with `null` value names are not returned.
```

content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -3560,13 +3560,34 @@ const result = await prisma.user.findMany({
 })
 ```
 
+:::warning
+
+`not` will return all items that do not match a given value. However, if the column is nullable, `NULL` values will not be returned. If you require null values to be returned, use an [`OR`](#or) to include `NULL` values.
```
ankur-arch:
```suggestion
`not` will return all items that do not match a given value. However, if the column is nullable, `NULL` values will not be returned. If you require null values to be returned, use an [`OR`](#or) operator to include `NULL` values.
```

ankur-arch (approved):

PR #6155, opened 2024-07-21
feat(docs): add --check flag to format docs

baileywickham:
Let me know if there is anywhere else this info should be added


baileywickham:
Hello, is there any update on this? @Dev-Connections

jharrell (approved):
Sorry for the delay. Thank you for the contribution!

PR #6158, opened 2024-07-23
accelerate autoscaling

jharrell:
Resolves DA-1085


content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,45 @@ metaTitle: 'Accelerate: Connection Pooling'
...
 
-Accelerate has a default `connection_limit` set to `10`.
+By default, Accelerate has a default connection limit of `10`. In most cases, this connection limit is sufficient. However, in high traffic or compute environments, you may need to [increase this value](#configuring-the-connection-pool-size). For the [Pro plan]() there is a maximum connection limit of `100`. For [Business plan]() users the maximum is `1000`.
 
 ### Configuring the connection pool size
 
-The connection pool size can be configured via the _database connection string_ in your Platform project (_not_ the Accelerate connection string in your Prisma schema) via the `connection_limit` parameter. For example, set a connection pool size of `10` like this: `postgresql://USER:PASSWORD@HOST:PORT/DATABASE?connection_limit=10`.
+The connection pool size can be configured in your Platform project via the `Connection limit` setting. For example, set a connection pool size of `10` like this:
+
+![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
```
jharrell: @luanvdw this is a filler image, we'll need this.

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,45 @@ metaTitle: 'Accelerate: Connection Pooling'
...
+If nothing is configured, the default value of `10` will be used.
 
 To change the size of the connection pool:
 1. Open the [Console](https://console.prisma.io/login).
 1. Select the _project_ in which you're using Accelerate. You might need to select a different _workspace_ if you can't see your project.
 1. Select the _environment_ where you want to configure Accelerate's connection pool size.
-1. Update the connection string by appending the `connection_limit` argument. 
+1. Update the connection limit via the `Connection limit` setting.
 
 ![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
```
jharrell: @luanvdw this will need to be updated as well to match the new instructions.

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,45 @@ metaTitle: 'Accelerate: Connection Pooling'
...
-</Admonition>
+Accelerate provides built-in connection pooling by default. By using Accelerate, you get the benefits of connection pooling without having to configure anything. The efficient management of database connections allows the database to process more queries without exhausting the available database connections, making your application more scalable.
+
+In some cases, however, it may be beneficial to further configure connection pooling in order to optimize the performance of your application.
 
-### Default connection pool size
+### Connection Pooling in Accelerate
 
-Accelerate has a default `connection_limit` set to `10`.
+By default, Accelerate has a default connection limit of `10`. In most cases, this connection limit is sufficient. However, in high traffic or compute environments, you may need to [increase this value](#configuring-the-connection-pool-size). For the [Pro plan]() there is a maximum connection limit of `100`. For [Business plan]() users the maximum is `1000`.
```
jharrell: @luanvdw I intended for `pro plan` and `business plan` to be links here but I can't find good spots to link to. Let me know if I'm missing something.

mhessdev: I think for now, just the pricing page. However, I believe it would be beneficial to have the Accelerate header include a link, and I think there's an issue for that. We can try to implement this before the launch if we think it's a good approach.

mhessdev: I think I would rather see this as a bulleted list than a paragraph. It would make it easier to see the differences between each plan. Let me know if that idea doesn't make sense.

content/300-accelerate/250-connection-pooling.mdx
mhessdev: Can we not put it in the URL anymore?

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,45 @@ metaTitle: 'Accelerate: Connection Pooling'
...
 1. Select the _project_ in which you're using Accelerate. You might need to select a different _workspace_ if you can't see your project.
 1. Select the _environment_ where you want to configure Accelerate's connection pool size.
-1. Update the connection string by appending the `connection_limit` argument. 
+1. Update the connection limit via the `Connection limit` setting.
 
 ![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
 
+#### Autoscaling
 
+With the Pro and Business plans of the Prisma Data Platform, you can increase your connection limit above the default of `10`. This allows Accelerate to allocate resources to scale up or down based on the traffic to your application. Specifically, Accelerate will begin allocating new resources when you have consistent usage greater than 90% of your currently alloted connection limit. If traffic continues to increase, Accelerate will scale up further. If traffic decreases, Accelerate will scale down.
```
mhessdev: > With the Pro and Business plans of the Prisma Data Platform, you can increase your connection limit above the default of `10`

That line reads weird

How about 

"The starter plan limits your connection limit to 10. You can increase that limit by upgrading to the Pro of Buissness plans" 

Possibly also include the bulleted list idea from above here? 

I also feel like this should follow the main description of what Autoscaling is

"Autoscalling allows Accelerate to allocate resources to scale up or down based on the traffic to your application. Specifically, Accelerate will begin allocating new resources when you have consistent usage greater than 90% of your currently alloted connection limit. If traffic continues to increase, Accelerate will scale up further. If traffic decreases, Accelerate will scale down.

The following connection limits are set on a per plan basis (don't like this wording) 
- Starter: 10
- Buissness: w.e number
- Pro: w.e number

"

content/300-accelerate/250-connection-pooling.mdx
jharrell: bulleted list or a table? What do you think?

content/300-accelerate/250-connection-pooling.mdx
jharrell: Correct. @rtbenfield let me know if I'm wrong here.

rtbenfield: No, we disregard the value from the URL when the new explicit setting is active. This was needed for backwards compatibility with existing users that may have a higher value in the URL than their plan allows. Accelerate will continue to work for them as it does today until they update their connection string and are asked to choose from the new options.

content/300-accelerate/250-connection-pooling.mdx
mhessdev: If you can make a table look nice then yea lets do that. I know we have had issues with tables in the past though.

content/300-accelerate/250-connection-pooling.mdx
mhessdev: hmmm, we should probably note this? It does seem odd though. So essentially I wont be able to set the connection limit in the URL if I'm a new users. But if I'm old and I have it set it will still work but be limited to my plan regardless of what I put in the string?

content/300-accelerate/250-connection-pooling.mdx
rtbenfield: > Specifically, Accelerate will begin allocating new resources when you have consistent usage greater than 90% of your currently alloted connection limit.

I'd prefer if we left out any specific details about scaling thresholds and focus more on how a higher connection limit enables Accelerate to process more requests concurrently. Specific details might introduce friction if we want to refine these values later. I also don't think it provides much values to the user as they just want to understand what options they have for supporting more queries and the tangible option is the connection limit. It's our job to make sure the resources we allocate can support the connections they've given us.

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,56 @@ metaTitle: 'Accelerate: Connection Pooling'
...
 To change the size of the connection pool:
 1. Open the [Console](https://console.prisma.io/login).
 1. Select the _project_ in which you're using Accelerate. You might need to select a different _workspace_ if you can't see your project.
 1. Select the _environment_ where you want to configure Accelerate's connection pool size.
-1. Update the connection string by appending the `connection_limit` argument. 
+1. Update the connection limit via the `Connection limit` setting.
 
 ![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
 
+#### Autoscaling
```
jharrell, line 33: @mhessdev updated this section.

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,56 @@ metaTitle: 'Accelerate: Connection Pooling'
...
+
+Autoscaling allows Accelerate to allocate resources to scale up or down based on the traffic to your application. Specifically, Accelerate will begin allocating new resources when you have consistent usage greater than 90% of your currently alloted connection limit. If traffic continues to increase, Accelerate will scale up further. If traffic decreases, Accelerate will scale down.
+
+Autoscaling is enabled when the Accelerate connection limit is set higher than `10`. Autoscaling is not available on the [Starter plan](https://prisma.io/pricing).
```
mhessdev: This is confusing. Below in the table it says starter is 10. But here it says its not available. But overall this section reads much better is is laid out nicely

content/300-accelerate/250-connection-pooling.mdx
rtbenfield: > But if I'm old and I have it set it will still work but be limited to my plan regardless of what I put in the string?

The plan limits are also not enforced for existing users until they update their Accelerate config. That's also intended to preserve the behavior that exists today. We don't want any users to have a sudden impact from the change and rather want to give them the opportunity to adopt the new behavior on their own terms.

content/300-accelerate/250-connection-pooling.mdx
```diff
@@ -4,39 +4,56 @@ metaTitle: 'Accelerate: Connection Pooling'
...
 1. Select the _project_ in which you're using Accelerate. You might need to select a different _workspace_ if you can't see your project.
 1. Select the _environment_ where you want to configure Accelerate's connection pool size.
-1. Update the connection string by appending the `connection_limit` argument. 
+1. Update the connection limit via the `Connection limit` setting.
 
 ![Update database connection string in Accelerate](/img/accelerate/accelerate-update-database-connection-string.png)
 
+#### Autoscaling
+
+Autoscaling allows Accelerate to allocate resources to scale up or down based on the traffic to your application. Specifically, Accelerate will begin allocating new resources when you have consistent usage greater than 90% of your currently alloted connection limit. If traffic continues to increase, Accelerate will scale up further. If traffic decreases, Accelerate will scale down.
```
rtbenfield: Reposting this as I think it got lost in one of the changesets.

> Specifically, Accelerate will begin allocating new resources when you have consistent usage greater than 90% of your currently alloted connection limit.

I'd prefer if we left out any specific details about scaling thresholds and focus more on how a higher connection limit enables Accelerate to process more requests concurrently. Specific details might introduce friction if we want to refine these values later. I also don't think it provides much values to the user as they just want to understand what options they have for supporting more queries and the tangible option is the connection limit. It's our job to make sure the resources we allocate can support the connections they've given us.

content/300-accelerate/250-connection-pooling.mdx
mhessdev: > The plan limits are also not enforced for existing users until they update their Accelerate config.

What does this mean? Is this simply toggling on the setting in the workspace UI?

rtbenfield: Right. @luanvdw maybe you have some thoughts on how that flow is in the Console? Any designs we can share here?

mhessdev (approved):

PR #6173, opened 2024-08-03
Fix Pagination Navigation in "Get Started" Section

columk1:
### Description
This PR addresses the pagination issues in the Getting Started section by updating the front matter of each MDX file to include the `pagination_prev` field.

### Issue Details
The "Previous" navigation buttons were self-referential so it was not possible to go back on most pages. In most cases clicking Previous reloaded the Typescript/Postgres version of the current page no matter what language/database configuration the user was in. The buttons worked on a few of the title pages but not in nested routes. This PR allows navigation back and forth through the docs and maintains the user's choices where possible. (There are a couple of cases where there is a missing Planetscale page so the Postgresql version was chosen in those cases.)


jharrell:
Hi there, 

Thank you so much for your contribution, but we actually just resolved this. 

Really appreciate your time and if you see any issues, please let us know!

PR #6211, opened 2024-08-14
fix: WEB-632 Update sidebar darkmode text color

carlagn:
Fixes #WEB-632


meletj (changes requested):
for some reason my reply in linear wasn't showing in the issue but the color should be 'text tertiary'  #A0AEC0

Techboidesign (approved):

meletj (approved):

PR #6221, opened 2024-08-23
fix typo in `050-raw-queries.mdx`

DevlinRocha:
- fixed typo in `050-raw-queries.mdx`
    - "be aware that ~~your~~ these functions"
    - "be aware that these functions"


jharrell (approved):

DevlinRocha:
why are these checks failing? all i did was remove 1 word

jharrell:
@DevlinRocha we have an issue where our checks cannot be run against forks. 

But also, we've change the structure of these pages recently and it looks like this typo was caught. Thank you very much for your contribution and if you see a similar issue, please do continue to open pull requests 🙂

PR #6224, opened 2024-08-27
Update 03-generators.mdx

dbgso:
I made generator to generating er diagrams for PlantUML.
Please let us know if there are any tasks that need to be completed to be reflected in the document.

https://github.com/dbgso/prisma-generator-plantuml-erd/tree/main/packages/generator


jharrell (approved):

PR #6236, opened 2024-09-05
Add filenames to TypeScript files in TypedSQL docs

karlhorky:
First of all, thanks for Prisma and TypedSQL! Really great projects and tools.

[The TypedSQL docs page](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/typedsql) contains some ambiguous code blocks.

Eg. this code block looks like you can write TypeScript directly in the SQL file:

Screenshot 2024-09-05 at 14 41 32

Adding filenames for the TypeScript code blocks should help.


jharrell:
Wow, thank you for noticing that! It looks like we have a visual issue where if two codeblocks are right next to one another, they can appear as one code block. This is a good fix, but we should also resolve the visual issue.

Thank you very much for your contribution!

jharrell (approved):

karlhorky:
@jharrell no problem, glad to help! Thanks for the review and merge!

And yeah, agreed that the underlying problem should probably be resolved as well.

If zooming in, there is a slight rounded corner and (very subtle) shadow on the edge of the code block - but agreed this is probably not enough of a visual signal that there is a new separate code block starting:

Screenshot 2024-09-13 at 17 58 52

PR #6245, opened 2024-09-12
Add Accelerate comparison page

sneub:


content/300-accelerate/900-compare.mdx
```diff
@@ -0,0 +1,97 @@
+---
+title: 'Compare Accelerate'
+metaTitle: 'Comparing Prisma Accelerate to other connection pooling options'
+metaDescription: 'Learn how Prisma Accelerate compares to other connection poolers like pgbouncer.'
+hide_table_of_contents: true
+---
+
+<TopBlock>
```
mhessdev, line 8: I might be miss-remembering Is TopBlock correct here, @jharrell ? I thought we may have a diff element for that.

content/300-accelerate/900-compare.mdx
```diff
@@ -0,0 +1,97 @@
...
+    
+    ```jsx
+    await prisma.user.findMany({
+      cacheStrategy: {
+        swr: 60,
+      },
+    });
+    ```
+    
+- Query level cache policies are critical for serious applications, so that you can control which queries are cached, and the characteristics of the policy. You may want certain data in your app to be cached for several days, other data to be cached for a just a few minutes, and other data to be not cached at all. This is only possible with Prisma Accelerate.
```
mhessdev, line 55: I honestly think this is such a huge point that people over look. I am glad you are highlighting this. I almost want it sooner in the page but I'm not sure where I would put it.

mhessdev (commented):
So, the overall goal of this page is to compare Accelerate to other options, correct? However, it feels like the content is mostly reiterating what already exists in our docs and elsewhere (not a knock just a "hey you might not know this"). While there are tables comparing Accelerate to other options, maybe diving deeper into the pain points of the other solutions—ones that Accelerate addresses—would be more beneficial, as we don’t have that content yet.

content/300-accelerate/900-compare.mdx
sneub: I copied from another page -- happy to change it though 👍

content/300-accelerate/900-compare.mdx
sneub: I've added a mention to it in the introductory section now -- I agree it's a biggy.

content/300-accelerate/900-compare.mdx
```diff
@@ -0,0 +1,106 @@
...
+| **Query-level cache policies** | ✅ | ❌ | ❌ |
+| **Authenticate with API key** | ✅ | ❌ | ❌ |
+| **Postgres compatible** | ✅ | ✅ | ❌ |
+| **MySQL compatible** | ✅ | ❌ | ✅ |
+| **MongoDB compatible** | ✅ | ❌ | ❌ |
+| **Automatic cache updates** | ❌ | ❌ | ✅ |
+
+**Why are these important?**
+
+- Integration with the Prisma client is important for existing Prisma ORM users — it means you can control caching policies directly from your existing codebase with just an extra line of code. Integration is seamless. Here is an example using [the stale-while-revalidating caching strategy](/accelerate/caching#stale-while-revalidate-swr):
```
jharrell: Provided a suggestion here because I was confused the first time I read this bullet point. I understand now how it's related to the table above, but it confused me on first read.
```suggestion
- Since Accelerate extends the Prisma client, you can control caching policies directly from your codebase with just an extra line of code. Integration is seamless. Here is an example using [the stale-while-revalidating caching strategy](/accelerate/caching#stale-while-revalidate-swr):
```

content/300-accelerate/900-compare.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+Explore how Prisma Accelerate compares to other global cache and connection pool solutions on the market, and discover what sets it apart.
+
+</TopBlock>
+
+## What makes Accelerate unique?
+
+Prisma Accelerate is chosen and loved by many for a number of key reasons which make Accelerate unique:
+
+- [**Query-Level Policies**](/accelerate/compare#accelerate-global-cache): Accelerate is the only solution that offers query-level cache policies, allowing you to control the cache strategy for each query specifically. It is common to have some values that need to be cached for a long time, others that need caching for a short time, and some that should not be cached at all. With Accelerate you can do this, and even set different cache strategies per query.
```
ankur-arch:
```suggestion
- [**Query-Level policies**](/accelerate/compare#accelerate-global-cache): Accelerate is the only solution that offers query-level cache policies, allowing you to control the cache strategy for each query specifically. It is common to have some values that need to be cached for a long time, others that need caching for a short time, and some that should not be cached at all. With Accelerate you can do this, and even set different cache strategies per query.
```
```suggestion
- [**Query-Level Policies**](/accelerate/compare#accelerate-global-cache): Accelerate is the only solution that offers query-level cache policies, allowing you to control the cache strategy for each query specifically. It is common to have some values that need to be cached for a long time, others that need caching for a short time, and some that should not be cached at all. With Accelerate you can do this, and even set different cache strategies per query.
```

content/300-accelerate/900-compare.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+## Accelerate connection pool
+
+Prisma Accelerate includes a globally hosted connection pooler, which allows you to handle peak loads without any problem. Using a connection pool is important especially for serverless infrastructure, which by nature is not able to control connection volume to the database on its own. Prisma Accelerate offers a fully managed, globally colocated option, which auto scales to support any workload.
+
+### Management
+
+|  | Accelerate | pgbouncer | pgcat | Digital Ocean (pgbouncer) | Neon (pgbouncer) | Supavisor | Hyperdrive |
+| --- | --- | --- | --- | --- | --- | --- | --- |
+| **Fully Managed** | ✅ | ❌ | ❌ | 🟠 | ❌ | ❌ | ✅ |
```
ankur-arch:
```suggestion
| **Fully managed** | ✅ | ❌ | ❌ | 🟠 | ❌ | ❌ | ✅ |
```
```suggestion
| **Fully Managed** | ✅ | ❌ | ❌ | 🟠 | ❌ | ❌ | ✅ |
```

petradonka: @sneub, how come the Neon connection pool is marked as not fully managed? 🤔 I'd think it is, but maybe missing something

content/300-accelerate/900-compare.mdx
```diff
@@ -0,0 +1,106 @@
...
+### Database Support
+
+|  | Accelerate | pgbouncer | pgcat | Digital Ocean (pgbouncer) | Neon (pgbouncer) | Supavisor | Hyperdrive |
+| --- | --- | --- | --- | --- | --- | --- | --- |
+| **PostgreSQL** | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| **MySQL** | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
+| **Planetscale** | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
+| **CockroachDB** | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
+| **MongoDB** | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
+| **MariaDB** | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
```
petradonka: I'd remove MariaDB, unless you think that people really really care about it coming soon

Are there no connection pool solutions for the other ones?

ankur-arch: @petradonka I suggested that we add MariaDB for transparency. 

> unless you think that people really really care about it coming soon

I recall some users attempting to try out Accelerate with it, @nurul3101 can elaborate regarding the demand for MariaDB.

sneub: There are for sure connection pool solutions for other ones, but this page doesn't aim to be a full representation of everything, just the commonly perceived alternatives to Accelerate. 

I'd also say that we can add to this page as we learn more -- I'd err on the side of concise and pointy (easy & useful for readers) rather than cover everything on the market (overwhelming)

petradonka: Yep, I agree. So just remove MariaDB then, if you agree :)

content/300-accelerate/900-compare.mdx
sneub: Fixed 👍

petradonka (approved):

PR #6246, opened 2024-09-13
Add an alternative way to implement libSQL replica syncing

aslauris:


content/200-orm/050-overview/500-databases/900-turso.mdx
```diff
@@ -222,3 +222,23 @@ app.use(async (req, res, next) => {
 
 app.listen(3000, () => console.log(`Server ready at http://localhost:3000`))
 ```
+It could be also implemented as a [Prisma Client extension](https://www.prisma.io/docs/orm/prisma-client/client-extensions). The below example shows auto-syncing after create, update or delete operation is performed.
```
jharrell:
```suggestion
It could be also implemented as a [Prisma Client extension](/orm/prisma-client/client-extensions). The below example shows auto-syncing after create, update or delete operation is performed.
```

jharrell (approved):

PR #6300, opened 2024-09-28
feat: respect color theme preference

twobiers:
This PR enables the `respectPrefersColorScheme` from Docusaurus. It respects the preferred color theme by the browser
The default from Docusaurus is `false` because its not assuming multiple available themes. Since the prisma docs support light and dark themes, it should be enabled.


xoyseau (approved):

jharrell (approved):

jharrell:

PR #6382, opened 2024-11-01
elaborate on pricing

nikolasburk:


ankur-arch (dismissed):

jharrell (dismissed):

mhessdev (dismissed):
Overall its good, maybe update it so it leaves wiggle room for us to change our stand in the future, the line about "Unlike traditional databases prisma postgres...... mental shift" this kinda sounds like we are saying 'we think we are right and this is the right way to do it"

content/200-orm/050-overview/500-databases/250-prisma-postgres.mdx
```diff
@@ -42,7 +42,15 @@ Learn more about the architecture in our [announcement article](https://pris.ly/
 
 ## Pricing
 
-Prisma Postgres has a pay-as-you-go pricing model for storage and queries and is billed through Prisma Accelerate and Pulse. More details on our [pricing page](https://www.prisma.io/pricing).
+Prisma Postgres has a pay-as-you-go pricing model for storage and queries.
+
+Unlike traditional database providers, Prisma Postgres charges for _query volume_, not compute. This is a mental shift compared to other products, but we believe that this ultimately presents a more intuitive model for developers to think about database cost. 
+
+Our goal is that developers need to only think about queries — we'll take care of everything else to make sure those queries can run. Developers shouldn't need to worry about compute, scaling, downtime, etc.
```
nikolasburk:
```suggestion
In contrast to traditional database providers, Prisma Postgres charges for _query volume_, not compute. 

Our goal is that developers need to only think about queries — we'll take care of everything else to make sure those queries can run. Developers shouldn't need to worry about compute, scaling, downtime, etc.
```

jharrell (approved):

mhessdev (commented):
This might change now that ppg is free?

nikolasburk:
I'm closing this one since we don't seem to want to make these strong statements about our pricing model any more and also it's free during EA. 

I'll open a separate PR to note that PPG is completely free during EA though 👍

PR #6386, opened 2024-11-04
Update 200-prisma-cli-reference.mdx

### LINKED ISSUE #6181: update doc about creating a migration (an empty migration to be more precise)
Hey guys,

I was wandering around the doc to find out how can I create an empty migration and then I ended up [herje](https://www.prisma.io/docs/orm/reference/prisma-cli-reference#migrate-dev) under options section I read that it says:

> Creates a new migration based on the changes in the schema but does not apply that migration. Run `migrate dev` to apply migration.

For `--create-only` flag. So I though I cannot use it for creating an empty migration file. Then I read [this Stackoverflow Q&A](https://stackoverflow.com/a/75439776/8784518) and tried it. Surprisingly they were right about it; you can create an empty migration without the need to have anything changed in `schema.prisma`.

Thus I think it is time to revise it to be more accurate. Maybe something like this would do the trick: "Creates a new migration based on the changes in the schema, or just creating an empty one when your schema has not change. Note that this does not apply that migration automatically. Run `migrate dev` to apply migration."

Any comment/idea/feedback?

Best of luck.

nikolasburk:
fixes #6181

### DIFF
--- a/content/200-orm/500-reference/200-prisma-cli-reference.mdx
+++ b/content/200-orm/500-reference/200-prisma-cli-reference.mdx
@@ -1043,13 +1043,14 @@ See also:
 
 #### Options
 
-| Option            | Required | Description                                                                                                                         | Default |
-| :---------------- | :------- | :---------------------------------------------------------------------------------------------------------------------------------- | :------ |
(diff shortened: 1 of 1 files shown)


mhessdev (approved):

kasir-barati (approved):
I loved this @nikolasburk. Thanks a million for taking care of this.

PR #6405, opened 2024-11-06
Add prisma-extension-casl to community extensions

dennemark:
Hi,

Adding another client extension to the docs.

I have created a [prisma client extension](https://github.com/dennemark/prisma-extension-casl) for [CASL](https://casl.js.org/v6/en/). I am using it in a redwood project with graphql where I was annoyed that I could not properly check all prisma queries i.e. in graphql relation queries. To avoid footguns I decided to implement the authorization logic in the prisma client. It can actually create pretty large queries for complicated authorization logic, but so far it works, even for ACL.

Here is an example:
```ts
function builderFactory() {
  const builder = abilityBuilder();
  const { can } = builder;
  can("read", "Post", {
    thread: {
      creatorId: 0,
    },
  });
  can("read", "Thread", "id");
  return builder;
}

const caslClient = prismaClient.$extends(
  useCaslAbilities(builderFactory, "casl")
);
const result = await caslClient.post.findMany({
  include: {
    thread: true,
  },
});
/**
 * creates a query under the hood with assistance of @casl/prisma
 *
 *{
 *   where: {
 *       AND: [{
 *           OR: [{
 *               thread: {
 *                   creatorId: 0
 *                   }
 *                }]
 *            }]
 *        }
 *   include: {
 *       thread: true
 *    }
 *
 * and result will be filtered and should look like
 * { id: 0, threadId: 0, thread: { id: 0 }, casl: ['read'] }
 */
```

### DIFF
--- a/cSpell.json
+++ b/cSpell.json
@@ -82,7 +82,8 @@
     "Solidstart",
     "Astro",
     "unikernels",
-    "Postico"
+    "Postico",
+    "CASL"
   ],
   "ignoreWords": [
     "Ania",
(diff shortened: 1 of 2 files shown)


jharrell (approved):
Thank you very much for your contribution!

PR #6419, opened 2024-11-13
fix: make difference clearer

### LINKED ISSUE #787: Test updatedAt and now() in different regions
## Context 

(See Jan's comment) Find out what date/time gets written to the database if your db is in a different region.

## Suggested action

Update reference docs for that attribute and function.

ankur-arch:
closes #787


nikolasburk (approved):
good callout 👌

PR #6429, opened 2024-11-14
Update 250-prisma-postgres.mdx

nikolasburk:
I'm actually quite unsure about what we want to write here. Just mention that PPG is free in EA (as I suggest in this PR), already give an outlook on what the pricing model will be (pay-as-you-go?) or maybe entirely drop the **Pricing** section from the page?

### DIFF
--- a/content/200-orm/050-overview/500-databases/250-prisma-postgres.mdx
+++ b/content/200-orm/050-overview/500-databases/250-prisma-postgres.mdx
@@ -73,7 +73,7 @@ const prisma = new PrismaClient()
 
 ## Pricing
 
-Prisma Postgres is completely free during [Early Access](/platform/maturity-levels#early-access). More details on our [pricing page](https://www.prisma.io/pricing).
+Prisma Postgres is free during [Early Access](/platform/maturity-levels#early-access). More details on our [pricing page](https://www.prisma.io/pricing).
 
 ## Temporary limitations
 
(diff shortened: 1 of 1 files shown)


content/200-orm/050-overview/500-databases/250-prisma-postgres.mdx
```diff
@@ -73,7 +73,7 @@ const prisma = new PrismaClient()
 
 ## Pricing
 
-Prisma Postgres has a pay-as-you-go pricing model for storage and queries and is billed through Prisma Accelerate and Pulse. More details on our [pricing page](https://www.prisma.io/pricing).
+Prisma Postgres is completely free during [Early Access](/platform/maturity-levels#early-access). More details on our [pricing page](https://www.prisma.io/pricing).
```
ankur-arch: suggestion from Marc's DAGPT to keep it simple:

```suggestion
Prisma Postgres is free during [Early Access](/platform/maturity-levels#early-access). More details on our [pricing page](https://www.prisma.io/pricing).
```

ankur-arch (approved):

ankur-arch:
> I'm actually quite unsure about what we want to write here. Just mention that PPG is free in EA (as I suggest in this PR), already give an outlook on what the pricing model will be (pay-as-you-go?) or maybe entirely drop the Pricing section from the page?

@nikolasburk, *maybe* we could encourage users to reach out and discuss their ideal usage and pricing with PPG. This could create a valuable feedback loop.

mhessdev (approved):

PR #6437, opened 2024-11-15
Clarify docs for Next.js Edge runtime users

rnbrady:
This might save others some time. Basically:

- Using `ws` in Vercel Edge functions or Cloudflare Workers will fail. 
- These runtimes provide a native WebSocket implementation which should be used.
- Dotenv isn't needed in some frameworks.

### DIFF
--- a/content/200-orm/050-overview/500-databases/890-neon.mdx
+++ b/content/200-orm/050-overview/500-databases/890-neon.mdx
@@ -196,3 +196,21 @@ const adapter = new PrismaNeon(pool, {
   schema: 'myPostgresSchema'
 })
 ```
+
+#### Next.js Edge runtime
+
+If your runtime provides a WebSocket class, you don't need to install `ws` and `@types/ws` and you might get errors if you do. If your framework provides support for `.env` files, then you don't need the `dotenv` package. For example if using Next.js with Edge runtime, the following code will work:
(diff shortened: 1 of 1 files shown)


jharrell:
Hey there @rnbrady and thank you for the PR!

We generally try and separate as many concerns as possible in our docs. What I mean by that is that this is a DB page, not a runtime or framework page. 

I think your information is still valuable. The fact that `ws` doesn't work on Vercel Edge Functions is important! But I think that information should be in other pages. 

We have a [Vercel Edge page](https://github.com/prisma/docs/blob/main/content/200-orm/200-prisma-client/500-deployment/301-edge/485-deploy-to-vercel.mdx?plain=1) and a [Next.JS considerations page](https://github.com/prisma/docs/blob/main/content/200-orm/800-more/600-help-and-troubleshooting/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx?plain=1) it might be worth linking to and/or updating instead of this page. 

Does that make sense?

rnbrady:
Oh dang, how did I miss that Vercel page. Thanks for the link.

Vercel aside, this page talks about edge environments and then gives a code example with a Nodejs dependency.

Best solution is probably to remove `ws` and `dotenv` from the code example.

PR #6459, opened 2024-11-27
fix: add `findUniqueOrThrow` example

### LINKED ISSUE #5017: `*OrThrow` queries documentation is missing an illustrating example
https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#finduniqueorthrow

image

Note that the code example here is only about this weird usage with transactions, but not the actual main use case which is much more important.

ankur-arch:
closes #5017

### DIFF
--- a/content/200-orm/500-reference/050-prisma-client-reference.mdx
+++ b/content/200-orm/500-reference/050-prisma-client-reference.mdx
@@ -632,18 +632,11 @@ We introduced `findUniqueOrThrow` in v4.0.0. It replaces the [`rejectOnNotFound`
 Here’s an example of how it works:
 
 ```ts
-await prisma.model.findUniqueOrThrow({
+await prisma.user.findUniqueOrThrow({
   where: {
     id: 1,
   },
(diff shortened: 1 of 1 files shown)


content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -621,13 +621,29 @@ const result = await prisma.user.findUnique({
...
+    id: 1,
+  },
+});
+```
+
+:::note
+
+Replace `model` with the name of your Prisma model, e.g., `prisma.user.findUniqueOrThrow` or `prisma.post.findUniqueOrThrow`.
+
+:::
```
nikolasburk: we're using concrete model names across the entire page, so no need to use a placeholder for this query:

```suggestion
Here’s an example of how it works:

```ts
await prisma.user.findUniqueOrThrow({
  where: {
    id: 1,
  },
});
```
```

nikolasburk (approved):

PR #6461, opened 2024-11-27
fix typo for isNot definition

ludralph:
A PR to fix the typo for isNot definition


nikolasburk (approved):
nice catch, thanks a lot for fixing this right away @ludralph 🎉

PR #6466, opened 2024-11-29
Added information about directUrl not needed while using Prisma Postgres

nurul3101:


content/300-accelerate/200-getting-started.mdx
```diff
@@ -72,6 +72,8 @@ Migrations and introspections will use the `directUrl` connection string rather
 
 > `directUrl` is useful for you to carry out migrations and introspections. However, you don't need `directUrl` to use Accelerate in your application.
 
+> If you are using Prisma Postgres then there is no need for `directUrl` as Prisma Migrate and Introspection work with the `prisma+postgres://` connection string.
```
ankur-arch:
```suggestion
:::note
If you are using Prisma with PostgreSQL, there is no need for `directUrl`, as Prisma Migrate and Introspection work with the `prisma+postgres://` connection string.
:::
```

ankur-arch (approved):

content/300-accelerate/200-getting-started.mdx
```diff
@@ -72,6 +72,10 @@ Migrations and introspections will use the `directUrl` connection string rather
 
 > `directUrl` is useful for you to carry out migrations and introspections. However, you don't need `directUrl` to use Accelerate in your application.
 
+:::note
+If you are using Prisma with PostgreSQL, there is no need for `directUrl`, as Prisma Migrate and Introspection work with the `prisma+postgres://` connection string.
+:::
```
nikolasburk, line 77: I understand the impulse of mentioning this here but I think it's out-of-context in the Accelerate "getting started" guide.

```suggestion
```

nurul3101: I added it here because the customers were referring to Accelerate guide while using Prisma Postgres. Do you think I should add it in Prisma Postgres docs section instead?

Issue #4845, opened 2023-06-06
Document how order of operations affects a query


nurul3101:
Based on this GitHub Discussion:
- https://github.com/prisma/prisma/discussions/19491

the execution order of `set` and `connect` changes the end result of the query. It would be a good idea to document this behaviour.

PR #6495, opened 2024-12-05
Added note about order of `set` and `connect` operations

nurul3101:
This resolves:
- https://github.com/prisma/docs/issues/4845

### DIFF
--- a/content/200-orm/500-reference/050-prisma-client-reference.mdx
+++ b/content/200-orm/500-reference/050-prisma-client-reference.mdx
@@ -3011,7 +3011,7 @@ A nested `connect` query connects a record to an existing related record by spec
   ```
   The required connected records were not found. Expected 1 records to be connected, found 0.
   ```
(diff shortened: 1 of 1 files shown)


content/200-orm/500-reference/050-prisma-client-reference.mdx
```diff
@@ -3011,6 +3011,7 @@ A nested `connect` query connects a record to an existing related record by spec
   ```
   The required connected records were not found. Expected 1 records to be connected, found 0.
   ```
+- When using `set` and `connect` together, their order significantly impacts the result. If `set` is used before `connect`, the connected records will reflect the final state after the connect operation, as set clears all previous connections and connect establishes new ones. However, if `connect` is used before `set`, the set operation will override the connect action by clearing all connected records.
```
ankur-arch:
```suggestion
- When using `set` and `connect` together, the order in which they are applied significantly impacts the result. If `set` is used before `connect`, the connected records will only reflect the final state established by the `connect` operation, as `set` clears all existing connections before `connect` establishes new ones. Conversely, if `connect` is applied before `set`, the `set` operation will override the `connect` action by clearing all connected records and replacing them with its own specified state.
```

ankur-arch (approved):

PR #6501, opened 2024-12-06
fix: #5602 remove type from keyword renders

### LINKED ISSUE #5602: Wrong syntax highlighting in Prisma schemas when `type` is used a field name
This code snippet:

```prisma
model Asset {
  id   Int       @id @default(autoincrement())
  url  String
  type AssetType

  video   Video?
  article Article?

  owner   User @relation(fields: [ownerId], references: [id])
  ownerId Int
}
```

Renders the `type` field as a keyword in our docs:

image

carlagn:
Fixes #5602

### DIFF
--- a/src/theme/prism-prisma.js
+++ b/src/theme/prism-prisma.js
@@ -1,6 +1,6 @@
 Prism.languages.prisma = Prism.languages.extend("clike", {
-  keyword: /\b(?:datasource|enum|generator|model|view)\b/,
-  "type-class-name": /(\s+)[A-Z]\w+/, ///(\b)(\s+)[A-Z]\w+/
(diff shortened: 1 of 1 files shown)


ankur-arch (dismissed):

ankur-arch (commented):
@carlagn the issue here is, sometimes *`type`* is used as a keyword. In those cases, it should be unchanged. 


For example, [here](https://www.prisma.io/docs/orm/prisma-schema/data-model/models#defining-composite-types) it should remain unchanged and highlighted.
image


But it [removes](https://4a20efd8.docs-51g.pages.dev/orm/prisma-schema/data-model/models#defining-composite-types) the highlight as we remove the 'type' keyword as a keyword.

image

mhessdev (approved):

ankur-arch (approved):

PR #6520, opened 2024-12-12
Update 380-connection-urls.mdx

nurul3101:
Updates reference to use free Postgres database from Prisma Postgres instead of Supabase.


ankur-arch (approved):

nikolasburk (approved):
nice, thanks Nurul!

PR #6523, opened 2024-12-13
Updating tracing docs for GA

jharrell:
Resolves DA-1908


content/200-orm/200-prisma-client/600-observability-and-logging/250-opentelemetry-tracing.mdx
```diff
@@ -36,7 +34,7 @@ For each trace, Prisma Client outputs a series of spans. The number and type of
 - `prisma:client:operation`: Represents the entire Prisma Client operation, from Prisma Client to the database and back. It contains details such as the model and method called by Prisma Client. Depending on the Prisma operation, it contains one or more of the following spans:
   - `prisma:client:connect`: Represents how long it takes for Prisma Client to connect to the database.
   - `prisma:client:serialize`: Represents how long it takes to validate and transform a Prisma Client operation into a query for the [query engine](/orm/more/under-the-hood/engines).
-  - `prisma:engine`: Represents how long a query takes in the query engine.
+  - `prisma:engine:query`: Represents how long a query takes in the query engine.
     - `prisma:engine:connection`: Represents how long it takes for Prisma Client to get a database connection.
     - `prisma:engine:db_query`: Represents the database query that was executed against the database. It includes the query in the tags, and how long the query took to run.
     - `prisma:engine:serialize`: Represents how long it takes to transform a database query result into a Prisma Client result.
```
aqrln: We also have separate `prisma:engine:response_json_serialization` since Prisma 5.5.0 (for library engine only, not Accelerate) because this span was confusing to the users since it doesn't include serialization of data between the engine and the client in a sense they expect, it refers to massaging the raw data returned from the database into the shape and types expected by the client.

content/200-orm/200-prisma-client/600-observability-and-logging/250-opentelemetry-tracing.mdx
```diff
@@ -71,26 +69,28 @@ If your application sends a large number of spans to a [collector](https://opent
...
 ### Step 1. Install up-to-date Prisma ORM dependencies
 
-Use version `4.2.0` or later of the `prisma`, `@prisma/client`, and `@prisma/instrumentation` npm packages.
+Use version `6.1.0` or later of the `prisma`, `@prisma/client`, and `@prisma/instrumentation` npm packages. You will also need to install the `@opentelemetry/api` package as it's a peer dependency.
 
 ```terminal
 npm install prisma@latest --save-dev
 npm install @prisma/client@latest --save
 npm install @prisma/instrumentation@latest --save
+npm install @opentelemetry/api@^1.8.0 --save
```
aqrln:
```suggestion
npm install @opentelemetry/api@latest --save
```

aqrln: 1.8.0 is the minimum supported but the users shouldn't think about it, the OTel ecosystem is on 1.9.0 now mostly

content/200-orm/200-prisma-client/600-observability-and-logging/250-opentelemetry-tracing.mdx
```diff
@@ -99,15 +99,17 @@ generator client {
...
+Now install the appropriate OpenTelemetry packages, as follows:
 
 ```console
 npm install @opentelemetry/semantic-conventions @opentelemetry/exporter-trace-otlp-http @opentelemetry/instrumentation @opentelemetry/sdk-trace-base @opentelemetry/sdk-trace-node @opentelemetry/resources
 ```
 
-### Register tracing in your application
+### Step 3: Register tracing in your application
 
 The following code provides two examples of configuring OpenTelemetry tracing in Prisma:
```
aqrln, line 116: The example below uses deprecated names of constants. The new names are
```ts
import { ATTR_SERVICE_NAME, ATTR_SERVICE_VERSION } from '@opentelemetry/semantic-conventions'
```

content/700-optimize/200-getting-started.mdx
```diff
@@ -27,9 +27,20 @@ Prisma Optimize is intended for use in local environments. Learn more in the [FA
...
 ## 2. Add Optimize to your application
 
-### 2.1. Update Your `schema.prisma` file
+### 2.1. Install the Optimize Prisma Client extension
 
-In the `generator` block of your Prisma schema, add the `tracing` preview feature:
+Install Prisma Client, the Optimize extension, and OpenTelemetry API:
+
+```bash
+npm install @prisma/client@latest @prisma/extension-optimize @opentelemetry/api@^1.8.0
```
aqrln: I don't think installing `@opentelemetry/api` should be necessary here since `@prisma/extension-optimize` depends on it already. cc @millsp @igalklebanov to confirm

Otherwise, if we need this, could we please remove `@^1.8.0`? There's no need to specify an older version that would be installed without it.

content/700-optimize/200-getting-started.mdx
```diff
@@ -27,9 +27,20 @@ Prisma Optimize is intended for use in local environments. Learn more in the [FA
...
+```
+
+We recommend using the latest versions of Prisma Client and the Optimize extension. Prisma Client supports `@opentelemetry/api` version `1.8.0` as a peer dependency.
```
aqrln: It doesn't support only 1.8.0, and latest (i.e. 1.9.x right now) is recommended.

content/200-orm/200-prisma-client/600-observability-and-logging/250-opentelemetry-tracing.mdx
jharrell: Thanks @aqrln! How does `prisma:engine:response_json_serialization` differ from `prisma:engine:serialize`? If `response_json_serialization` is serialization between engine and client, then is `prisma:engine:serialize` serialization that occurs on the client side?

aqrln: No, `prisma:engine:serialize` happens before `prisma:engine:response_json_serialization` in the engine. What happens there is the engine goes through the raw data returned by the database connector and converts the types according to the schema, fixes up the shapes of the objects (among other things, to compensate for different internal and user-facing representation of virtual selections, like relation count) and so on.

nikolasburk (dismissed):
lgtm 👍

content/200-orm/200-prisma-client/600-observability-and-logging/250-opentelemetry-tracing.mdx
```diff
@@ -99,15 +99,17 @@ generator client {
...
 
-### Step 3: Install OpenTelemetry packages
+</details>
 
-Finally, install the appropriate OpenTelemetry packages, as follows:
+### Step 2: Install OpenTelemetry packages
+
+Now install the appropriate OpenTelemetry packages, as follows:
 
 ```console
```
aqrln, line 110: FWIW I added this thing today: https://github.com/prisma/prisma/pull/25878, so now users don't have to install the `@opentelemetry/instrumentation` package (unless they need it for whatever reason separately) and can do

```ts
import { PrismaInstrumentation, registerInstrumentations } from '@prisma/instrumentation'
```

instead of

```ts
import { registerInstrumentations } from '@opentelemetry/instrumentation'
import { PrismaInstrumentation } from '@prisma/instrumentation'
```

For us, compatibility with different `@opentelemetry/instrumentation` versions is less of an issue if most users use our re-export, so perhaps it's worth to use it in the docs.

aqrln (approved):
LGTM!

PR #6530, opened 2024-12-16
Add Aiven as a managed database option supporting PostgreSQL and MySQL

ludralph:
This PR updates the Managed Databases section in the Prisma documentation to include Aiven as a supported provider. Aiven offers fully managed PostgreSQL and MySQL services, making it a relevant addition to the list of managed database providers compatible with Prisma.


jharrell (changes requested):
Holding on this until we can get confirmation from the ORM team on how we want to present this data. 

Cc @wmadden

wmadden:
Seems fine to me @jharrell but I'm not sure what value it provides unless there's specific functionality that we provide to support a particular provider.

Be aware of the list of what we support, don't support and do/don't care about listed here: https://github.com/prisma/prisma/issues/25843

jharrell (approved):

PR #6560, opened 2024-12-31
Enable Docusaurus Faster

slorber:
Your site is already on Docusaurus v3.6 but not using Docusaurus Faster yet: https://github.com/facebook/docusaurus/issues/10556

Turning this new option on can make your site build 4x faster (10s instead of 40s for me)


jharrell:
Thank you @slorber ! I'm going to create a new branch based off of this one so that our CI works 😊

jharrell:
Closed in favor of #6576

PR #6567, opened 2025-01-03
Update documentation for ORM 6.2

### LINKED ISSUE #6565: Update existing documentation for ORM 6.2
- JSON type support for SQLite
- Enum type support for SQLite
- omitApi preview feature goes GA

jharrell:
Fixes #6565
Resolves DA-2285

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/prisma/docs/pull/6567?shareId=41183709-0979-493b-80a7-861f98d39638).


content/200-orm/050-overview/500-databases/500-sqlite.mdx
```diff
@@ -46,9 +46,12 @@ The SQLite connector maps the [scalar types](/orm/prisma-schema/data-model/model
 | `Float`    | `REAL`        |
 | `Decimal`  | `DECIMAL`     |
 | `DateTime` | `NUMERIC`     |
-| `Json`     | Not supported |
+| `Json`     | \*            |
+| `Enum`     | \*            |
 | `Bytes`    | `BLOB`        |
 
+> \* As SQLite does not have native `JSON` or `ENUM` types, these are implemented via shims and consistency is maintained by Prisma ORM.
```
jacek-prisma: SQLIte does have native `JSON` types and we're using them, they're `JSONB` and `JSON` and the one we use in our implementation is `JSONB`.

The enum bit is correct, but it might be worth adding that they're backed by TEXT. Another thing worth adding might be that there is no database level or migration level enforcement for correctness of enums in SQLite as opposed to MySQL and Postgres. If someone bypasses Prisma and stores an invalid enum entry in the database, Prisma queries will fail at runtime when reading that entry. Also, in SQLite it's possible to end up with incorrect data after schema changes similarly to MongoDB (since the enums aren't checked by the database).

nikolasburk: Thanks a lot for this additional context, Jacek! That's super helpful 🙏 

I updated this section accordingly:

image

nikolasburk (approved):
I incorporated some feedback from @jacek-prisma and made some other small edits to the PR. It should be good to go for tomorrow's release 🙌

content/200-orm/200-prisma-client/100-queries/063-excluding-fields.mdx
```diff
@@ -4,44 +4,23 @@ metaTitle: 'Excluding fields'
...
-
-model User {
-  id        Int      @id @default(autoincrement())
-  firstName String
-  lastName  String
-  email     String   @unique
-  password  String
-  createdAt DateTime @default(now())
-  updatedAt DateTime @updatedAt
-}
```
jkomyno, line 37: Why did we get rid of the `User` model here?

nikolasburk: we didn't! I just switched the order of the tabs so that the **Code** is first and the **Schema** second.

image

previously I think it made sense to have the schema first to show that you needed to enable the preview feature. but now that it's not necessary any more, I think leading with the code is the more sensible thing to do.

PR #6619, opened 2025-01-24
chore: Add FAQ entry about using local databases with Prisma Pulse

ludralph:
This update adds a new FAQ entry addressing whether Prisma Pulse can work with a local database during development. It explains that Pulse requires a network-accessible database and does not support databases running on localhost. The entry also provides a workaround using tools like ngrok for testing purposes, while emphasizing the recommendation to use hosted databases for production environments.

This is based on this [discord conversation](https://discord.com/channels/937751382725886062/1331750113382760559/1331750113382760559)


nikolasburk (approved):
Looks great! My only thought is that I'm not sure how much sense it makes to mention ngrok without providing concrete instructions for how to actually use a local DB with Pulse, I'd also be ok with just the first of the three paragraphs here. Not a blocker though and this is good to get merged from my side 🙌

PR #6620, opened 2025-01-24
tweak: ⚙️ remove password k/v pair

colocated:
Removes the K/V Pair from the docs, because the sentence underneath says "notice how it DOESNT show up", whilst showing a JSON object containing what we're showing we're "omitting".

{9BD4E6D3-CF94-4F56-A76E-6158E8F31350}


jharrell (approved):

jharrell:
Thank you for the fix!

PR #6633, opened 2025-01-30
fix: remove ppg from db switcher from existing project section

ankur-arch:


nikolasburk (approved):
thank you for cleaning this up Ankur!

PR #6639, opened 2025-01-31
chore: Add documentation on counting large MongoDB collections

### LINKED ISSUE #6638: Counting Large Collections in MongoDB
[Internal Slack Conversation](https://prisma-company.slack.com/archives/C032AU58HT7/p1738312245598449)

ludralph:
Documentation regarding performance considerations for large MongoDB collections, specifically focusing on counting operations. The update highlights potential issues with `count()` on large datasets, suggests alternatives like `estimatedDocumentCount()`, and introduces the concept of counter caches for frequently accessed counts to optimize performance.

Closes #6638


jharrell (changes requested):
Thanks for this info, @ludralph ! I agree we should add this to our docs. 

Would you mind rewriting this section to not use a list and instead use a “problem” and “solution” format? Similar to our help and troubleshooting docs. 

The “problem” would be the description about count being slow in Mongo and the “solution” would be the use of estimated count or a cache.

ludralph:
> Thanks for this info, @ludralph ! I agree we should add this to our docs.
> 
> Would you mind rewriting this section to not use a list and instead use a “problem” and “solution” format? Similar to our help and troubleshooting docs.
> 
> The “problem” would be the description about count being slow in Mongo and the “solution” would be the use of estimated count or a cache.

I'll update the PR.

jharrell (approved):

PR #6685, opened 2025-02-20
docs: clarify non-postgres pnpm workspace setup

aamir0:
- Added some clarifications for modifications that need to be made
  when not using Postgres, based on my experience using MySQL
- Move note about `create next-app` command to follow the command
- Fixed a typo (their/there)


ankur-arch (approved):

ankur-arch:
Thanks for the contribution @aamir0 🙏 .  I'll see to it that it's merged.

PR #6711, opened 2025-03-05
Update 090-nextjs.mdx

Levix0501:


ankur-arch:
Hey @Levix0501 👋 , thanks for opening the PR. 

Could you add a description to specify why this change is necessary?

Levix0501:
> Hey @Levix0501 👋 , thanks for opening the PR.
> 
> Could you add a description to specify why this change is necessary?

Thanks for your review!

The original code assigns the created PrismaClient instance to globalForPrisma.prisma in development mode but never actually reads from it. This means that on every import, a new PrismaClient instance is still being created, which can lead to multiple database connections in Next.js development mode.

ankur-arch (approved):

PR #6762, opened 2025-03-21
Added missing instructions for @prisma/client for quick start

cghiurea:
Add missing @prisma/client install step to Quickstart docs

Ensures users install `@prisma/client` before using it in code examples, fixing an omitted step.

They are official instructions from setup-prisma/start-from-scratch/relational-databases/install-prisma-client-typescript-postgresql which i also included in the quick start


jharrell (approved):

jharrell:

PR #6764, opened 2025-03-23
Correct a small typo

bakert:
I came across this because a starter project I cloned was doing exactly this (and my _prisma_migrations table kept being "inexplicably" deleted!) I think the new wording is correct?


nikolasburk (approved):

nikolasburk:
Thanks a lot for catching & fixing this @bakert 🎉

PR #6776, opened 2025-03-27
Reworked Guide Titles

aidankmcalister:
Titles sorted and renamed

### DIFF
--- a/sidebars.ts
+++ b/sidebars.ts
@@ -420,7 +420,7 @@ const sidebars: SidebarsConfig = {
             "guides/tanstack-start",
             "guides/react-router-7",
             "guides/solid-start",
-          ],
+          ].sort(),
         },
         {
           type: "category",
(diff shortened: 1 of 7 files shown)


mhessdev (approved):

sidebars.ts
```diff
@@ -420,7 +420,7 @@ const sidebars: SidebarsConfig = {
             "guides/tanstack-start",
             "guides/react-router-7",
             "guides/solid-start",
-          ],
+          ].sort(),
```
jharrell, line 423: I was sitting here thinking about how annoying it was going to be to keep these organized and sorted. Wow. This is such a simple and perfect fix.

jharrell (approved):

sidebars.ts
aidankmcalister: I was about to make an object, sort through it, and then map through it. I'm glad I tried this first lol

PR #6784, opened 2025-03-30
Generated prisma client code doesn't declare types, that doesn't make sense

adriancuadrado:
https://stackoverflow.com/a/59552002/

> Your specific example is different since you are declaring a type, not a variable, types already do not compile into any JavaScript. I do not know if there is any reason to declare a type.


nikolasburk (approved):

nikolasburk:
Thanks for catching & fixing this @adriancuadrado 🎉

PR #6805, opened 2025-04-06
Fix a mistake in code example

### LINKED ISSUE #6804: Mistake in example code
## Description

The following part of the documentation specifies how to create a prisma client that attaches to the global object:

https://github.com/prisma/docs/blob/389c943d485989fccdb573ab5872849367fe498a/content/800-guides/090-nextjs.mdx?plain=1#L283-L298

This code results in a prisma client with a type of any, as well as raising the following errors:
1. 'prisma' is referenced directly or indirectly in its own type annotation. ts(2502)
2. 'prisma' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer .ts(7022)

A fix should be made in line 289:

https://github.com/prisma/docs/blob/389c943d485989fccdb573ab5872849367fe498a/content/800-guides/090-nextjs.mdx?plain=1#L289

Where `{ prisma: typeof prisma }` should become `{ prisma: typeof PrismaClient }`.

This fix removes the errors as well as makes the exported client type safe.

Mosaab-Emam:
Fixes #6804


Mosaab-Emam:
Looks like this was fixed in c0e94b601eda6a7e21ffce2ac8f2b8567b3c2ef0

PR #6808, opened 2025-04-07
Resolves Import and Type Issues in Next.js Guide

nurul3101:
In the Next.js Guide we were generating PrismaClient at a custom location but were importing it from the default location of `node_modules`. This caused issues in `seed.ts` and `prisma.ts` file.

This PR fixes both the issues.


nikolasburk (approved):

nikolasburk:
Thanks Nurul! 💚

PR #6827, opened 2025-04-13
DA-3191 Turborepo Guide Updates

aidankmcalister:


content/800-guides/080-turborepo.mdx
```diff
@@ -24,43 +24,43 @@ This guide will show you how to set up Prisma as a standalone package in a Turbo
 
 :::note
 
-This guide was tested using Turborepo version `2.3.3` and Prisma ORM version `6.1.0`.
+This guide was tested using Turborepo version `2.5.0` and Prisma ORM version `6.6.0`.
```
mhessdev: Is "tested" the right word? Yes its tested but maybe "last updated"? I also am a tad concerned abotu saying this as people may see it in the future and assume its wrong based on versions. What do you think about that?

content/800-guides/080-turborepo.mdx
```diff
@@ -122,23 +121,23 @@ Inside the `database` directory, initialize prisma by running:
...
+  npx prisma init --db --output ../generated/prisma
   ```
 
   </TabItem>
 
   <TabItem value="yarn">
 
   ```terminal
-  yarn prisma init
+  yarn prisma init --db --output ../generated/prisma
```
mhessdev, line 126: Dont we have a mdx component for this? If not we should make one so we dont have to update this everythwere

content/800-guides/080-turborepo.mdx
```diff
@@ -147,73 +146,101 @@ Inside the `database` directory, initialize prisma by running:
...
-
-Make sure to replace the `DATABASE_URL` inside `packages/database/.env` with a valid database url. 
-
-:::
-
-Add a model to your Prisma schema in `database/prisma/schema.prisma`:
+Add a model to your Prisma schema in `packages/database/prisma/schema.prisma`:
 
-```prisma
+```prisma file=packages/database/prisma/schema.prisma showLineNumbers
```
mhessdev, line 149: SHouldnt there be a new line here?

mhessdev (changes requested):

content/800-guides/080-turborepo.mdx
aidankmcalister: I think it would be helpful just in case to say the versions that it was last updated with. But I agree "tested" -> "updated". Maybe for every guide we can add something like this,

### Last Updated With
- Turborepo `2.5.0`
- Prisma `6.6.0`

content/800-guides/080-turborepo.mdx
aidankmcalister: I don't think so but I could be wrong. Not seeing one in the codebase.

content/800-guides/080-turborepo.mdx
aidankmcalister: I'm not sure what you mean.

Do you mean this?
```
//add-next-line
```prisma file=packages/database/prisma/schema.prisma showLineNumbers
```
If so, the `file=...` is just to give the codeblock a header with the file path. Not actual code

content/800-guides/080-turborepo.mdx
aidankmcalister: Not having a version prolongs its life and credibility. The only issue is if someone keeps trying to brute force it after an update but doesn't know it's outdated. Maybe put it at the bottom instead?

What about this?

### Version Compatibility

This guide was last updated to work with:
- [Turborepo `2.5`](https://turbo.build/blog/turbo-2-5)
- [Prisma `6.6.0`](https://www.prisma.io/blog/prisma-orm-6-6-0-esm-support-d1-migrations-and-prisma-mcp-server)

mhessdev (dismissed):

mhessdev (approved):

PR #6829, opened 2025-04-13

adriancuadrado:
`_b` should be `b_`


nikolasburk (approved):

nikolasburk:
Thanks so much for catching these @adriancuadrado 🎉

PR #6867, opened 2025-04-28
AWS added support ?

wilau2:
Wondering if this part of the documentation is still relevant. Is there a benchmark to compare the usage of RDS proxy to the other proxy using prisma ? 

https://aws.amazon.com/fr/blogs/database/amazon-rds-proxy-multiplexing-support-for-postgresql-extended-query-protocol/


jharrell:
Hi @wilau2 as far as I’m aware, this hasn’t changed. We have an issue here tracking this: https://github.com/prisma/prisma/issues/5866

If you have a repro where this _is_ working, please comment on that issue and we can investigate!

PR #6903, opened 2025-05-13
add supabase & neon w/ accelerate guides

nikolasburk:


content/800-guides/300-supabase-accelerate.mdx
```diff
@@ -0,0 +1,137 @@
+---
+title: 'Set up Supabase with Accelerate Connection Pool'
+metaTitle: "Set up PostgreSQL on Supabase with Prisma Accelerate's Connection Pool"
+description: "Learn how to set up PostgreSQL on Supabase with Prisma Accelerate's Connection Pool"
+sidebar_label: 'Supabase with Accelerate '
```
jharrell:
```suggestion
sidebar_label: 'Supabase with Accelerate'
```

mhessdev (dismissed):

content/800-guides/300-supabase-accelerate.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+## Prerequisites
+
+To successfully complete this guide, you need **a connection string for a PostgreSQL instance hosted on Supabase**. It typically looks similar to this:
+
+```bash no-copy
+postgresql://postgres:[YOUR-PASSWORD]@db.nzpppscrldfwlzfalbrf.supabase.co:5432/postgres
+```
+
+If you already have a project using Prisma ORM, you can skip the first two steps and jump ahead to [Step 3. Install the Accelerate extension](#3-install-the-accelerate-extension).
```
jharrell, line 25: I think I would personally make this an admonition of some kind. Additionally, should the note here say that if you’re using Prisma ORM _and_ have an Accelerate connection string?

content/800-guides/300-supabase-accelerate.mdx
```diff
@@ -0,0 +1,137 @@
...
+```
+
+The `--no-engine` option is used to omit the [query engine](/orm/more/under-the-hood/engines) in the generated Prisma Client library. The query engine manages Prisma ORM's internal connection pool and is not needed when using Prisma Accelerate.
+
+## 6. Send queries through the connection pool
+
+In your application code, you now need to apply the Accelerate extension to your Prisma Client instance:
+
+```ts
+import { PrismaClient } from "./generated/prisma"
```
jharrell, line 131: I’m worried about this as depending on a user’s project this could be any relative path. IMO we shouldn’t use this and I think the latest version of the ORM is no longer setting the path by default.

content/800-guides/310-neon-accelerate.mdx
```diff
@@ -0,0 +1,137 @@
+---
+title: 'Set up Neon with Accelerate Connection Pool'
+metaTitle: "Set up PostgreSQL on Neon with Prisma Accelerate's Connection Pool"
+description: "Learn how to set up PostgreSQL on Neon with Prisma Accelerate's Connection Pool"
+sidebar_label: 'Neon with Accelerate '
```
jharrell:
```suggestion
sidebar_label: 'Neon with Accelerate'
```

jharrell (commented):
Left some comments.

ankur-arch (approved):

PR #6911, opened 2025-05-16
add ppg serverless driver docs

nikolasburk:


content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,100 @@
...
+  connectionString: "prisma+postgres://accelerate.prisma-data.net/?api_key=...",
+});
+
+const posts = await client.query('SELECT * FROM "Post" WHERE "userId" = $1', [1]);
+```
+
+## Limitations
+
+- No transaction support.
+- No DDL support.
```
aqrln: This is not true anymore

nikolasburk:
```suggestion
```

nikolasburk: things are changing fast 😄

content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,99 @@
...
+| Name               | Type          | Required | Description                                                                                                                                                                                   |
+| :----------------- | :------------ | :------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `connectionString` | `string`      | **Yes**  | The connection string for your Prisma Postgres instance.                                                                                                                                      |
+| `deserialize`      | `Deserialize` | No       | Custom deserializer function which takes a column type OID and a raw value and returns a mapped value. It's type is defined as `type Deserialize = (value: unknown, oid: unknown) => unknown` |
+
+#### Usage
+
+```ts
+import { ppg } from "@prisma/ppg";
+import type { User, Post } from "@prisma/client";
```
sorenbs: Let's avoid using generated types in this docs page, as it gives the impression that you need to use the CLI to generate code.

sorenbs: CleanShot 2025-05-16 at 16 40 56@2x

This is an example of what V0 actually does when using Neon

nikolasburk: ah, that's a good point! my goal was to show the compatibility of the Prisma types with the returned values but I see how  that may give a wrong impression. I still think there's value in showing the compabiltiy but don't want to overemphasize it, how about all default examples don't use that and then we do a dedicated one talking about "compatibility with generated Prisma types" (or smth like that) to demonstrate that. 

or do you think this is too misleading and even with that context there wouldn't be much value in showing that?

aqrln: It's not currently compatible with Prisma types though (except in trivial cases)

aqrln: e.g. dates and decimals would definitely be incompatible

content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+const sql = ppg("prisma+postgres://accelerate.prisma-data.net/?api_key=...");
+
+const posts: Post[] = (await sql`SELECT * FROM "Post"`) as Post[];
```
aqrln: One of these `Post[]` annotations is superfluous, I'd drop the `as`

content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+const userId = 42;
+const user: User = (await sql`SELECT * FROM "User" WHERE "id" = ${userId}`) as User;
```
aqrln: ditto.

Also, the type is wrong, the query returns `User[]`.

content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,108 @@
...
+```
+
+### `Client`
+
+The `Client` class provides more low-level control and should be used with more care. It aims to be a drop-in replacement for the [`pg`](https://node-postgres.com/) driver:
```
aqrln: > It aims to be a drop-in replacement for the [`pg`](https://node-postgres.com/) driver:

It's very far from that at the moment (most importantly, it doesn't provide a persistent connection), I'd drop this.

content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,108 @@
...
+});
+
+const posts = await client.query('SELECT * FROM "Post" WHERE "userId" = $1', [1]);
+```
+
+## Limitations
+
+- No transaction support.
+- No [local Prisma Postgres](/postgres/database/local-development) support.
+- Requires ESM.
```
aqrln: Not anymore

content/250-postgres/300-database/750-serverless-driver.mdx
nikolasburk: ah, then I'll remove that entirely

content/250-postgres/300-database/750-serverless-driver.mdx
```diff
@@ -0,0 +1,177 @@
...
+This query returns an object of this structure:
+
+```ts
+{
+  columns: [
+    { name: 'id', oid: 23, type: 23 },
+    { name: 'title', oid: 25, type: 25 },
+    { name: 'content', oid: 25, type: 25 },
+    { name: 'published', oid: 16, type: 16 },
+    { name: 'authorId', oid: 23, type: 23 }
```
aqrln: The `type` field is not defined in the types and users shouldn't rely on it, only on `oid`. We should remove this field.

aqrln (approved):

PR #6913, opened 2025-05-19
feat: add more limitations

ankur-arch:


content/250-postgres/300-database/550-local-development.mdx
```diff
@@ -74,7 +74,9 @@ npx prisma migrate dev
 
 :::note
 
-Make sure the local Prisma Postgres server is running before running the `prisma migrate dev` command.
+Make sure the local Prisma Postgres server is running before running the `prisma migrate dev` command. 
+
+If you must use a different port, [append `--port <number>`](/orm/reference/prisma-cli-reference#dev) (for example, `npx prisma migrate dev --port 5422`) and update your `DATABASE_URL` (or other connection settings) to match, but be aware that custom ports can introduce unexpected behavior, so use them only when necessary.
```
nikolasburk:
```suggestion
If you must use a different port, append [`--port <number>`](/orm/reference/prisma-cli-reference#dev) (for example, `npx prisma migrate dev --port 5422`) and update your `DATABASE_URL` (or other connection settings) to match. Be aware that custom ports can introduce unexpected behavior, so use them only when necessary.
```

content/250-postgres/300-database/550-local-development.mdx
```diff
@@ -74,7 +74,9 @@ npx prisma migrate dev
 
 :::note
 
-Make sure the local Prisma Postgres server is running before running the `prisma migrate dev` command.
+Make sure the local Prisma Postgres server is running before running the `prisma migrate dev` command. 
+
+If you must use a different port, append [`--port <number>`](/orm/reference/prisma-cli-reference#dev) (for example, `npx prisma migrate dev --port 5422`) and update your `DATABASE_URL` (or other connection settings) to match. Be aware that custom ports can introduce unexpected behavior, so use them only when necessary.
```
nikolasburk: do we know what kind "unexpected behavior" could happen here? not a blocker necessarily but this seems a bit vague to me.

ankur-arch: I kept it a bit vague intentionally, as changing ports might be a bit [unstable](https://prisma-company.slack.com/archives/C08QF717GR3/p1747317679566909).

igalklebanov: Was speaking about the connection string stability - as in, it doesn't change as long as you're using the same ports everytime. There is no unexpected behavior in choosing ports different than the default ones.

ankur-arch:
```suggestion
If you must use a different port, append [`--port <number>`](/orm/reference/prisma-cli-reference#dev) (for example, `npx prisma migrate dev --port 5422`) and update your `DATABASE_URL` (or other connection settings) to match. 
```

Thanks @igalklebanov in that case I think a warning isn't necessary.

nikolasburk (approved):

PR #6929, opened 2025-05-27
fix: correct PrismaClient import path 

liliana-benitez:
This PR fixes an incorrect path for `PrismaClient` in "Querying the database using JavaScript and PostgreSQL" page, in "Write your first query with Prisma Client" section.

The documentation currently uses:
```js
const { PrismaClient } = require('@prisma/client')
```

But the correct import path (for this context) is:
```js
const { PrismaClient } = require('./generated/prisma')
```

The recommended import path for certain setups has been updated, and this section/page has outdated syntax, which could lead to confusion or errors for users following the docs.


content/100-getting-started/02-setup-prisma/100-start-from-scratch/110-relational-databases/250-querying-the-database-node-postgresql.mdx
```diff
@@ -40,7 +40,7 @@ main()
 
 Here's a quick overview of the different parts of the code snippet:
 
-1. Import the `PrismaClient` constructor from the `@prisma/client` node module
+1. Import PrismaClient from output folder specified in prisma init command.
```
nikolasburk:
```suggestion
1. Import `PrismaClient` from the `output` directory specified in the `prisma init` command.
```

nikolasburk (approved):
thanks for catching this @liliana-benitez 🎉

PR #6937, opened 2025-05-30
fix: prisma does not support diff in Turso yet

ThallesP:
I've been looking through the documentation for a way to compare differences between a Turso remote database, but it seems the CLI doesn't offer any `--to-turso` flag. Even the documentation itself only mentions diff support in a single line, without providing further details, maybe a silly mistake of copying the same warning from Cloudflare D1's docs?


aidankmcalister:
Hello. If you can resolve these conflicts, I would be happy to review the changes.

aidankmcalister:
Since May 30, this file has been updated to remove the previous `prisma migrate diff` mention. Thank you for taking the time to make this change. I will be closing this PR.

PR #6967, opened 2025-06-18
Fix: DC-4224 Link colors fixed

aidankmcalister:


mhessdev (approved):

Techboidesign:
@aidankmcalister @mhessdev  looks great! two things 
- **ORM** should be **indigo colored** and not _teal_ 
- There is a **small jump** for this notification box whenever you click on the left-hand navigation - is this fixable within this Fix?

PR #6986, opened 2025-06-25
fix: docs: 100-prisma-schema: missing SQL field

prabhuomkar9:
The explicit example showcases custom fields on underlying join table `assignedBy` and `assignedAt`

But the RAW SQL code missed the `assignedBy` field


nikolasburk:
Nice catch, thanks a lot for fixing this @Prabhuomkar9 🎉

PR #6991, opened 2025-06-26
fix: update Supabase guide with connection string details.

nurul3101:


nikolasburk (approved):
Loks great, thanks Nurul 🎉

Issue #6993, opened 2025-06-26
The PrismaClient import seems to be wrong


ckhatton-pvfscaffolding:
`/content/100-getting-started/02-setup-prisma/200-add-to-existing-project/110-relational-databases/250-querying-the-database-typescript-postgresql.mdx` ([link](https://github.com/prisma/docs/blob/main/content/100-getting-started/02-setup-prisma/200-add-to-existing-project/110-relational-databases/250-querying-the-database-typescript-postgresql.mdx))

Shouldn't the PrismaClient import be `import { PrismaClient } from './generated/prisma'`, and not `import { PrismaClient } from '@prisma/client'`?

Following the guide I get the error:

- `Module '"@prisma/client"' has no exported member 'PrismaClient'` in VS Code.
- `error: Cannot find module '.prisma/client/default' from ...` in the terminal.

This is after making sure to run `npx prisma generate`.

---

This would apply for the other pages that are not TypeScript.

Ace-Aryal:
Yup you are totally correct. I ran into same issue . I have created a pull request updating the docs hope it gets  merged

ckhatton-pvfscaffolding:
Hold fire @Ace-Aryal on #6997, I've come to realise that Prisma wasn't building the Prisma Client as expected - I found you have to either omit the "output" or set it (in my case) to `../node_modules/.prisma/client` in the `schema.prisma` ([source](https://stackoverflow.com/questions/67746885/prisma-client-did-not-initialize-yet-please-run-prisma-generate-and-try-to-i)). Someone also mentions it, and a shadow database, will be a required come `v7.0.0`.

`/prisma/schema.prisma`:

```prisma
generator client {
  provider = "prisma-client-js"
  output   = "../node_modules/.prisma/client"
}
```

From what I can see, nowhere in the documentation does it state to make sure to do this. In fact, the documented `output` default `node_modules/.prisma/client` is incorrect ([source](https://www.prisma.io/docs/orm/reference/prisma-cli-reference#arguments)) - I assume the actual default is `../node_modules/.prisma/client`, because none of these work:

- `node_modules/.prisma/client`
- `/node_modules/.prisma/client`
- `./node_modules/.prisma/client`

nurul3101:
From Prisma v7, it will be mandatory to provide an output path. This issue will be tackled during the Docs update for Prisma 7 release, where we will update the import path throughout the docs.

PR #6997, opened 2025-06-30
update prisma client instantiation code snippet for next js to match new version

Ace-Aryal:
The next.js primaClient instantiation code snippet is not up to date, I was getting error even after following the docs provided. Then I tried import from all the available files and it worked when I imported prismaClient from `@/generated/prisma` . So the fix I made is just changed a line of import statement. 

From 
````typescript
import { PrismaClient } from "@prisma/client";
````
to 
````typescript
import { PrismaClient } from "@/generated/prisma";
````


ckhatton-pvfscaffolding:
Hold fire on this, as mentioned in #6993.

PR #7123, opened 2025-09-13
fix: make tab items scrollable on mobile to avoid overflow

GeorgiosDrivas:
On tablet and smaller screen sizes, the tabs are going off the screen.

**Link** 
https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration/no-rust-engine

**Problem**
proof

**Solution**
Targeted both tabs with a css class in a pre-existing media query, and added an overflow-x attribute with value of scroll

## Summary by CodeRabbit

- New Features
  - Enables horizontal scrolling for the tab navigation on smaller viewports (≤1191px), making wide tab lists accessible without wrapping.

- Style
  - Introduces responsive CSS that applies overflow-x: scroll to the tabs list under narrow widths.
  - Scrollbar is consistently shown when the rule applies.
  - No visual changes on wider viewports; existing styles remain unchanged.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -678,6 +678,9 @@ hr {
   .header-github-link {
     display: none;
   }
+  .tabs-container ul.tabs{
+    overflow-x: scroll;
+  }
 }
 
 .navbar__items--right > :last-child {


aidankmcalister:
Thank you for your PR! There are some merge conflicts due to a larger change we made for Prisma 7. Would you be able to resolve those?

GeorgiosDrivas:
Of course! I will check it out and I will let you know when everything's ready for review.

GeorgiosDrivas:
After reviewing the website, it seems like the issue is fixed with the updated version, so there is no need for this issue to be open. Great job by the way, it is much better now!

PR #7151, opened 2025-10-01
feat: add deprecation warning for metrics

ankur-arch:
## Summary by CodeRabbit

- Documentation
  - Metrics docs now include a prominent deprecation warning: metrics preview deprecated in Prisma ORM v6.14.0 and planned for removal in v7.
  - Clarified compatibility: metrics supported only in Prisma ORM ≤6.13.x, unavailable in v6.14.0+; metrics blocked when engineType uses the client query compiler.
  - Installation and prerequisites updated with explicit bounded version guidance (e.g., 3.15.0–6.13.x) and concrete install commands; recommends native DB driver metrics or OpenTelemetry as alternatives.

### DIFF
--- a/content/200-orm/200-prisma-client/600-observability-and-logging/240-metrics.mdx
+++ b/content/200-orm/200-prisma-client/600-observability-and-logging/240-metrics.mdx
@@ -6,17 +6,25 @@ sidebar_class_name: preview-badge
 toc_max_heading_level: 4
 ---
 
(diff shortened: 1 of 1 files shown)


FGoessler:
@ankur-arch can we also add a note that it is not supported at all when using `engineType='client'` aka the query compiler?

mhessdev (dismissed):

PR #7204, opened 2025-10-21
docs(): update for new release

mhartington:
Adds additional documentation for the upcoming 6.18 release

## Summary by CodeRabbit

* **New Features**
  * Added a configurable engine option supporting both classic (Rust) and JS engines.
  * Introduced an env helper for type-safe environment variable access.

* **Documentation**
  * Updated configuration docs with engine usage examples, adapter guidance, and env-based patterns (replacing prior dotenv examples).
  * Added per-database metrics docs covering average response size, query duration, total egress, total operations, and cache utilization.

### DIFF
--- a/content/200-orm/500-reference/325-prisma-config-reference.mdx
+++ b/content/200-orm/500-reference/325-prisma-config-reference.mdx
@@ -348,7 +348,7 @@ Configure the schema engine your project should use.
 | -------- | ------------------ | -------- | ----------------- |
 | `engine` | `classic` or `js`  | No       | `classic`         |
 
- By default it is set to use the Rust-based schema engine, which requires that `datasource` be set
+ By default it is set to use the classic engine, which requires that `datasource` be set
  in your `prisma.config.ts`. 
 
 ```ts
--- a/content/200-orm/500-reference/325-prisma-config-reference.mdx
+++ b/content/200-orm/500-reference/325-prisma-config-reference.mdx
@@ -99,7 +99,7 @@ export declare type PrismaConfig = {
     path: string;
   };
   // Depending on the choice, you must provide either a `datasource` object or driver adapter
-  engine: 'classic': 'js'
+  engine: 'classic' | 'js'
   
 };
 ```
--- a/content/200-orm/500-reference/325-prisma-config-reference.mdx
+++ b/content/200-orm/500-reference/325-prisma-config-reference.mdx
@@ -353,7 +353,7 @@ Configure the schema engine your project should use.
 
 ```ts
 import path from "node:path";
-import { defineConfig, env } from "@prisma/config";
+import { defineConfig, env } from "prisma/config";
 export default defineConfig({
   engine: "classic",
   datasource: {
@@ -368,7 +368,7 @@ provide the adapter you wish to use
 
 ```ts
 import path from "node:path";
-import { defineConfig, env } from "@prisma/config";
+import { defineConfig, env } from "prisma/config";
 import { PrismaD1 } from "@prisma/adapter-d1";
 export default defineConfig({
   experimental: {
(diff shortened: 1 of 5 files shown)


content/200-orm/500-reference/325-prisma-config-reference.mdx
```diff
@@ -14,18 +14,22 @@ You can define your config in either of two ways:
 - Using the `defineConfig` helper:
   ```ts
   import path from "node:path";
-  import { defineConfig } from "prisma/config";
+  import { defineConfig, env } from "prisma/config";
```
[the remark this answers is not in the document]
mhessdev: Is this right?

content/200-orm/500-reference/325-prisma-config-reference.mdx
```diff
@@ -332,6 +339,55 @@ Failed to load config file "~" as a TypeScript/JavaScript module. Error: Error:
...
+export default defineConfig({
+  engine: "classic",
+  datasource: {
+      url: env('DATABASE_URL'),
+  },
+  schema: path.join("prisma", "schema.prisma"),
+});
+```
+
+Alternatively, if you are opting to use the newer Rust-free schema engine, set the engine to `js` and
```
ankur-arch, line 366:
```suggestion
Alternatively, if you are opting to use the [newer Rust-free schema engine](/orm/prisma-client/setup-and-configuration/no-rust-engine), set the engine to `js` and
```

content/250-postgres/100-introduction/250-overview.mdx
```diff
@@ -27,6 +27,15 @@ You can view the following usage metrics in your Console Dashboard:
     - Cumulative operations
     - Operations per day
 
+For details into individual databases in your workspace, each database has it's own metrics report
```
ankur-arch, line 30: Can we also add a small ## Metrics section in [the platform docs here](https://www.prisma.io/docs/platform/about#workspace)?

content/200-orm/500-reference/325-prisma-config-reference.mdx
```diff
@@ -14,18 +14,22 @@ You can define your config in either of two ways:
...
       path: path.join("db", "views"),
     },
-    typedSql: {
-      path: path.join("db", "queries"),
+    typedSql: { 
+    path: path.join("db", "queries"),
+    },
+    engine: "classic",
+    datasource: { 
+      url: env("DATABASE_URL")
```
ankur-arch, line 32: I think we need to update the following pages:

https://www.prisma.io/docs/orm/overview/databases/cloudflare-d1#3-set-up-prisma-config-file
https://www.prisma.io/docs/orm/overview/databases/turso#3-set-up-prisma-config-file

content/200-orm/500-reference/325-prisma-config-reference.mdx
aqrln: @ankur-arch your link is about the Rust-free client which is a completely different unrelated thing

aqrln: As for the Rust-free schema engine, I'm not sure if we even should document it. I guess it should be documented in some way since these properties are available in TS types, but it should be clear from the docs that it's highly experimental and isn't usable or useful yet for the general audience, it has almost nothing implemented (it only supports `db pull`/`db push` for SQLite or something like that IIRC, Alberto knows more).

mhartington: I'll just remove the "js" section all together. If it's causing this much confusion internally, then let's not have it be public for now.

ankur-arch: >I'll just remove the "js" section all together. If it's causing this much confusion internally, then let's not have it be public for now.

@mhartington I agree with this.

PR #7251, opened 2025-11-12
feat: add redirect loop checker

ankur-arch:
## Chores
* Added continuous integration checks to detect and prevent redirect loops; validation runs automatically on pull requests and main branch updates affecting redirect configurations.

### DIFF
--- a/scripts/check-redirect-loops.js
+++ b/scripts/check-redirect-loops.js
@@ -18,7 +18,7 @@ function parseRedirects(content) {
 
   for (const line of lines) {
     // Skip comments, empty lines, and the footer
-    if (line.trim().startsWith('#') || line.trim() === '' || line.includes('NO REDIRECTS BELOW')) {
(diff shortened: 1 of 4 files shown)


scripts/check-redirect-loops.js
```diff
@@ -0,0 +1,151 @@
...
+const REDIRECTS_FILE = path.join(__dirname, '../static/_redirects');
+const MAX_REDIRECT_DEPTH = 10;
+
+function parseRedirects(content) {
+  const lines = content.split('\n');
+  const redirects = new Map();
+
+  for (const line of lines) {
+    // Skip comments, empty lines, and the footer
+    if (line.trim().startsWith('#') || line.trim() === '' || line.includes('NO REDIRECTS BELOW')) {
```
ankur-arch:
```suggestion
		if (line.trim().startsWith('#') || line.trim() === '' || line.trim().includes('# NO REDIRECTS BELOW')) {
```

PR #7300, opened 2025-11-20
Fix: Add missing redirect for shadow-database docs (Issue #28586)

saikumarvasa100-hash:
The short URL https://pris.ly/d/migrate-shadow was redirecting to the old documentation path /concepts/components/prisma-migrate/shadow-database which resulted in a 404 error.

This PR adds a redirect from the old path to the new documentation location at /docs/orm/prisma-migrate/understanding-prisma-migrate/shadow-database.

Fixes #28586

## Summary by CodeRabbit

* **Chores**
  * Removed a legacy documentation redirect, so the old documentation URL no longer forwards to the updated page; users should use the current documentation path directly.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


AmanVarshney01:
I think we updated the prisly link to point to the correct location. Thanks for the PR tho 😄

PR #7348, opened 2025-12-03
DC-6336 Update deprecated prisma migrate diff flags

ArthurGamby:
## Summary by CodeRabbit

* **Documentation**
  * Updated Prisma Migrate CLI flags: `--from-schema-datamodel` is now `--from-schema`, and `--to-schema-datamodel` is now `--to-schema`.
  * Migration guides and CLI reference documentation updated with new flag names.
  * Migration examples across all database framework guides reflect the updated command syntax.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


ArthurGamby:
I've updated the deprecated `--to-schema-datamodel` / `--from-schema-datamodel` flags to `--to-schema` / `--from-schema` in the docs.

I wasn't sure if I was supposed to update only the baselining page or all occurrences, so I went ahead and updated all 10 files that used these deprecated flags (baselining, getting-started, squashing-migrations, patching-and-hotfixing, generating-down-migrations, CLI reference, and the migration guides for Drizzle/Sequelize/TypeORM/Cloudflare D1).

Let me know if you'd prefer I revert some of those and keep it scoped to just the baselining page!

aidankmcalister (approved):

PR #7480, opened 2026-02-12
DR-7287- docs: Create and use new guide about How to use Prisma with Deno

ArthurGamby:
## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for integrating Prisma with Deno, covering project setup, configuration, schema management, migrations, seeding, and server development.


aidankmcalister:
As we may have just mentioned in standup, this will need to fit the new docs structure.

Assuming it was a purposeful decision, but just in case, continue using `terminal` instead of `npm` for the Deno code blocks so they don’t allow other package managers/runtimes in the Deno-specific guide, similar to Bun.

PR #7584, opened 2026-03-03
feat: DR-7290 Charts

carlagn:
## Summary by CodeRabbit

* **New Features**
  * Added Chart component with responsive theming, legends, customizable tooltips, and support for bar/line/area/pie/stacked series plus interactive example components.

* **Documentation**
  * Comprehensive MDX docs and an examples README with live interactive demos, usage patterns, accessibility guidance, and best practices.

* **Style**
  * New global chart color tokens and per-chart theming support.


aidankmcalister:
Nice! Only note is that the value in the tooltip can sometimes be shoved up against it, maybe default to a `: `? So `Requests2,900` renders as `Requests: 2,900` instead
image

mhartington (commented):
See the comments from CodeRabbit. let's get these addressed and then merge

PR #7685, opened 2026-03-21
Native PostGIS Support for Prisma

lh0x00:
See full features at https://github.com/prisma/prisma/pull/29365

## Summary by CodeRabbit

* **Documentation**
  * Enhanced CRUD docs with a new geometry-based filtering example and updated link to geometry-field guidance.
  * Published a comprehensive "Working with geometry fields" guide covering PostGIS setup, geometry types, read/write examples, spatial predicates (near/within/intersects), distance ordering, null-handling, and migration/optimization notes.
  * Added PostGIS extension docs with Prisma schema examples, spatial query usage, and updated frontmatter metadata quoting.

### DIFF
--- a/apps/docs/content/docs/orm/prisma-client/queries/crud.mdx
+++ b/apps/docs/content/docs/orm/prisma-client/queries/crud.mdx
@@ -1,6 +1,6 @@
 ---
 title: CRUD
-description: 'Learn how to perform create, read, update, and delete operations'
+description: "Learn how to perform create, read, update, and delete operations"
 url: /orm/prisma-client/queries/crud
 metaTitle: CRUD (Reference)
 metaDescription: How to perform CRUD with Prisma Client.
@@ -115,9 +115,21 @@ const users = await prisma.user.findMany({
     posts: { some: { published: false } },
   },
 });
+
+// Filter by geometry (PostgreSQL with PostGIS)
+const nearbyLocations = await prisma.location.findMany({
+  where: {
+    position: {
+      near: {
+        point: [13.4, 52.5],
+        maxDistance: 1000, // meters
+      },
+    },
+  },
+});
 ```
 
-See [Filtering and sorting](/v6/orm/prisma-client/queries/filtering-and-sorting) for more examples.
+See [Filtering and sorting](/v6/orm/prisma-client/queries/filtering-and-sorting) for more examples, or [Working with geometry fields](/orm/prisma-client/special-fields-and-types/working-with-geometry-fields) for spatial queries.
 
 ### Select fields
 
--- a/apps/docs/content/docs/orm/prisma-client/special-fields-and-types/working-with-geometry-fields.mdx
+++ b/apps/docs/content/docs/orm/prisma-client/special-fields-and-types/working-with-geometry-fields.mdx
@@ -45,6 +45,12 @@ model Location {
 
 The field definition follows the format `Geometry(GeometryType, SRID)`. You can make the field optional by adding `?` at the end, allowing records without a position.
 
+:::info
+
+For optimal performance with spatial queries, add a GIST spatial index on your geometry columns using a [customized migration](/orm/prisma-migrate/workflows/customizing-migrations). See the [Performance considerations](#spatial-indexes) section for details.
+
+:::
+
 ### Supported geometry types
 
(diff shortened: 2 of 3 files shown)


BrenoRev:
While testing the spatial queries locally, I ran into a query planner issue on PostgreSQL 16: the planner underestimates the row count returned by the spatial CTE (estimates rows=1, actual can be 2000+), causing it to choose Nested Loop over Hash Join — which resulted in ~3.9s response time vs ~87ms.

To fix this, I wrapped all spatial queries in a Higher-Order Function (executeNearbyQuery) that enforces SET LOCAL enable_nestloop = off within the transaction scope before executing the query:

```typescript
/**
 * Executes a spatial query with SET LOCAL enable_nestloop = off.
 *
 * The PostgreSQL 16 planner underestimates the number of rows returned by the spatial CTE
 * (estimates rows=1, actual can be 2000+), choosing Nested Loop instead of Hash Join.
 * Nested Loop with 2000+ iterations causes severe degradation (3.9s vs 87ms).
 *
 * Accepts PrismaClient or TransactionClient:
 * - PrismaClient: creates a transaction internally (SET LOCAL is reverted at the end)
 * - TransactionClient: uses the existing transaction and manually reverts after the query
 *
 * PostgreSQL 17+: this function may be removed. PG17 introduced statistics propagation
 * for materialized CTEs, fixing the row estimate and allowing the planner to choose
 * Hash Join automatically without SET LOCAL.
 */
export async function executeNearbyQuery<T>(
  prisma: PrismaClient | PrismaInterfaces.TransactionClient,
  query: string,
): Promise<T[]> {
  if ('$transaction' in prisma) {
    return (prisma as PrismaClient).$transaction(async (tx) => {
      await tx.$queryRaw(PrismaInterfaces.sql`SET LOCAL enable_nestloop = off`);
      return tx.$queryRaw<T[]>(
        PrismaInterfaces.sql`${PrismaInterfaces.raw(query)}`,
      );
    });
  }

  await prisma.$queryRaw(PrismaInterfaces.sql`SET LOCAL enable_nestloop = off`);
  const results = await prisma.$queryRaw<T[]>(
    PrismaInterfaces.sql`${PrismaInterfaces.raw(query)}`,
  );
  await prisma.$queryRaw(PrismaInterfaces.sql`SET LOCAL enable_nestloop = on`);
  return results;
}
```

In the index of

```
@@index([geometry], map: "IDX_ADDRESS_GEOMETRY", type: Gist)
@@index([geometry_mercator], map: "IDX_ADDRESS_GEOMETRY_MERCATOR", type: Gist)
```

These indexes are what enable the spatial candidate scan in the CTE. The nestloop workaround is a separate concern — it corrects the join strategy after the index scan, when PG16 misjudges the CTE row count downstream.

Should the Docs reference these indexes to make that relationship explicit, or is that coupling that doesn't belong there?

lh0x00:
Thanks @BrenoRev  for mentioning it, it's extremely helpful. I've updated the docs and exported the function to support it; you can see it at https://github.com/prisma/prisma/pull/29365#issue-4109622219

aidankmcalister (approved):

lh0x00:
Since this PR hasn't merged, I've published the same PostGIS fork to npm as `@vertexa/prisma`, `@vertexa/prisma-client`, and `@vertexa/prisma-adapter-pg` (v7.8.6) — drop-in
replacements that add Geometry/Geography scalar types, near/within/intersects spatial
filters, distanceFrom ordering, and automatic GeoJSON ↔ EWKB serialization.

npm install `@vertexa/prisma` `@vertexa/prisma-client` `@vertexa/prisma-adapter-pg`
```json
{
  "pnpm": {
    "overrides": {
      "@prisma/client": "npm:@vertexa/prisma-client@7.8.6",
      "@prisma/adapter-pg": "npm:@vertexa/prisma-adapter-pg@7.8.6"
    }
  }
}
```

```prisma
model Location {
  id       Int      @id @default(autoincrement())
  name     String
  position Geometry(Point, 4326)?
}
```

```typescript
const nearby = await prisma.location.findMany({
  where: { position: { near: { point: [13.4, 52.5], maxDistance: 1000 } } },
  orderBy: { position: { distanceFrom: { point: [13.4, 52.5], direction: 'asc' } } },
  take: 10,
})
```

PR #7774, opened 2026-04-03
feat(site): performance overhaul

mhartington:
This is a big performance overhaul of the homepage

- Particle effect: swapped out the threejs for pure svg
- Testimonial: simplified implmentation to reduce excess DOM size
- static assets: minified and optimized image assets
- optimize image loading


carlagn (approved):
LGTM!

PR #7814, opened 2026-04-22
docs(docs): fix const reassignment in v7 upgrade Metrics example

### LINKED ISSUE #7810: Docs bug: const total = 0 mutated in "Metrics removed" example (Upgrade to v7)
## Summary

The `totalQueries` counter example in the ["Metrics removed"](https://www.prisma.io/docs/guides/upgrade-prisma-orm/v7#metrics-removed) section of the Upgrade to Prisma ORM v7 guide declares `total` with `const` but then attempts to reassign it, which will throw a `TypeError: Assignment to constant variable` at runtime (and fail to compile under TypeScript).

## Location

- Page: https://www.prisma.io/docs/guides/upgrade-prisma-orm/v7#metrics-removed
- Source file: `apps/docs/content/docs/guides/upgrade-prisma-orm/v7.mdx`

## Current (broken) code

```ts
const total = 0;
const prisma = new PrismaClient().$extends({
  client: {
    $log: (s: string) => console.log(s),
    async $totalQueries() {
      return total;
    },
  },
  query: {
    $allModels: {
      async $allOperations({ query, args }) {
        total += 1;          // ❌ Cannot assign to `total` because it is a constant
        return query(args);
      },
    },
  },
});
```

## Expected fix

Change `const` to `let`:

```ts
let total = 0;
```
​
## Why this matters

This is the example users are directed to when replacing the removed Metrics preview feature, so anyone copy-pasting it as a starting point will hit an immediate runtime/compile error. It also undermines confidence in the migration guide at exactly the moment users are doing a major version upgrade.

## Additional notes

- The snippet also uses `$allModels`, which scopes the counter to model operations only - raw queries (`$queryRaw`, `$executeRaw`) won't be counted. Worth a brief note if the intent is a true "total queries" counter.

nielskaspers:
## Summary

The \`totalQueries\` counter example in the [\"Metrics removed\"](https://www.prisma.io/docs/guides/upgrade-prisma-orm/v7#metrics-removed) section of the Upgrade to Prisma ORM v7 guide declares \`total\` with \`const\` but reassigns it inside \`\$allOperations\`. As written, the example throws \`TypeError: Assignment to constant variable\` at runtime (and fails to compile under TypeScript).

## Issue

Fixes #7810

## Changes

- \`apps/docs/content/docs/guides/upgrade-prisma-orm/v7.mdx\`: change \`const total = 0;\` to \`let total = 0;\` so the counter example actually runs.

## Testing

- Docs-only change. The surrounding snippet is unchanged; the only edit is the binding keyword.

## Summary by CodeRabbit

* **Documentation**
  * Fixed a code example in the Prisma ORM v7 upgrade guide to ensure proper execution.


AmanVarshney01 (approved):
Thanks for the PR @nielskaspers 🙏

PR #7853, opened 2026-04-29
fix(post-card): prevent CLS on post images

carlagn:
### What changed
`packages/ui/src/components/post-card.tsx` — the `` inside `PostCard`
was absolutely positioned inside its wrapper to fill the pre-reserved slot.

### Why the reported numbers were wrong
The original issue cited 288 of 317 images missing `width` and `height`
attributes, measured against the **rendered DOM**. That count is inflated
because Next.js `<Image fill>` intentionally emits `` tags with no
`width`/`height` attributes — it uses the parent's CSS aspect-ratio box to
reserve space instead. A DOM-level scraper treats every `fill`-mode image
as broken, even though it's correctly handled.

A source-level audit tells the real story:

| | Total | ✅ OK | ❌ Missing |
|---|---|---|---|
| Plain `` tags | 4 | 3 | **1** |
| Next.js `<Image>` (fill or w+h) | 54 | 54 | 0 |

One image was actually broken: the `` in `PostCard` had no dimensions
and no layout constraint, causing a genuine CLS hit.

### The fix
The wrapper div already had `relative aspect-video`, which reserves the
correct vertical space via CSS before the image loads. Making the ``
`absolute inset-0 w-full h-full` fills that slot — the same thing
`next/image fill` renders to in the DOM, with no framework dependency added
to the package.

## Summary by CodeRabbit

* **Style**
  * Refactored post card component styling and layout structure for improved code organization and rendering consistency.

* **Refactor**
  * Optimized image layout handling and className composition within the post card component.

### DIFF
--- a/packages/ui/src/components/post-card.tsx
+++ b/packages/ui/src/components/post-card.tsx
@@ -46,7 +46,9 @@ export function PostCard({
 
   const titleClassName = cn(
     "text-foreground-neutral font-mona-sans mt-4 mb-2",
-    isFeatured ? "text-2xl font-bold" : "text-md md:text-lg font-[650] sm:font-bold",
+    isFeatured
+      ? "text-2xl font-bold"
+      : "text-md md:text-lg font-[650] sm:font-bold",
   );
   const excerptClassName = cn(
     "text-sm text-foreground-neutral-weak line-clamp-2",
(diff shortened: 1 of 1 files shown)


mhessdev:
Superseded by `codex/seo-preview-seo-fixes`, which now contains this SEO work and is the single preview branch we are continuing from. Closing this PR to reduce review noise and keep follow-up work consolidated in one branch.

PR #7867, opened 2026-05-05
fix(docs): refer wrong raw query methods in ORM > Using Raw SQL > TypeSQL

Hephaest:
## Summary

In [Dynamic SQL Queries with Dynamic Columns](https://www.prisma.io/docs/v6/orm/prisma-client/using-raw-sql/typedsql#dynamic-sql-queries-with-dynamic-columns) section, the doc clearly mentioned:
> When you need to create a query where the columns are determined at runtime, you must use the $queryRaw and $executeRaw methods.
But in the example, it uses:
```js
const columns = "name, email, age"; // Columns determined at runtime
const result = await prisma.$queryRawUnsafe(`SELECT ${columns} FROM Users WHERE active = true`);
```
It looks counterintuitive to me, so I do check on my own. The search results confirmed my assumptions: the doc might refer to the wrong methods. 

### Proofs:
Under [`$queryRaw` considerations> ](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries#considerations)
image

Under [`$executeRaw` considerations> ](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries#considerations-1)
image

## Summary by CodeRabbit

* **Documentation**
  * Updated guidance for constructing SQL queries with dynamically determined columns.

### DIFF
--- a/apps/docs/content/docs/orm/prisma-client/using-raw-sql/typedsql.mdx
+++ b/apps/docs/content/docs/orm/prisma-client/using-raw-sql/typedsql.mdx
@@ -242,7 +242,7 @@ TypedSQL requires an active database connection to function properly. This means
 
 ### Dynamic SQL Queries with Dynamic Columns
 
-TypedSQL does not natively support constructing SQL queries with dynamically added columns. When you need to create a query where the columns are determined at runtime, you must use the `$queryRaw` and `$executeRaw` methods. These methods allow for the execution of raw SQL, which can include dynamic column selections.
+TypedSQL does not natively support constructing SQL queries with dynamically added columns. When you need to create a query where the columns are determined at runtime, you must use the `$queryRawUnsafe` and `$executeRawUnsafe` methods. These methods allow for the execution of raw SQL, which can include dynamic column selections.
 
 **Example of a query using dynamic column selection:**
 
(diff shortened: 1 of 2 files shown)


aidankmcalister (approved):
Good catch — the existing text is indeed wrong, and this fix is technically correct.

**Why `$queryRawUnsafe` / `$executeRawUnsafe` is the right choice here:**

`$queryRaw` and `$executeRaw` are implemented as [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates). This design is intentional for SQL injection safety — Prisma converts interpolated values into parameterized query arguments. However, this mechanism only works for **data values** (e.g. `WHERE email = ${email}`). Template variables **cannot** be used for identifiers like column names, table names, or SQL keywords.

From our own docs on `$queryRaw` considerations:
> "Variables cannot be used for identifiers such as column names, table names or database names, or for SQL keywords."

This means code like `prisma.$queryRaw\`SELECT ${columns} FROM Users\`` would not work as intended for dynamic column selection — the column list would be parameterized as a bound value, not interpolated as SQL identifiers, causing a query error.

`$queryRawUnsafe` (and `$executeRawUnsafe`) accept a plain string and perform direct string interpolation, which is the only way to dynamically construct column/table identifiers at runtime. The example code in the docs is correct; only the preceding text was wrong.

The PR correctly updates the text in both v6 (`content/docs/orm/v6/...`) and v7 (`content/docs/orm/...`) docs to match the example.

Approved — thanks for the thorough investigation and the screenshots as proof!

aidankmcalister (approved):

aidankmcalister:
Thank you!

PR #7924, opened 2026-05-27
Change npm install to npm ci for dependency installation

kanarian:
Because of increase in supply chain attacks in npm, better to use `npm ci` over `npm install`, moreover as per npm (https://docs.npmjs.com/cli/v11/commands/npm-ci) `npm ci` is made for the specific purpose of being used in automated environments such as a this workflow.


ankur-arch:
Thank you for contributing, @kanarian. Using `npm ci` in the CI example is the right call. The page has since moved under `orm/v6` and `orm/v7`, which is why this branch conflicts, so I re-applied your change on both current pages in #8283 with credit to you. Closing this one, and thanks again.

PR #7932, opened 2026-06-03
tem algo errado no passo 7

RAUNIFCE:
resolvi aprender somente na documentação,porém o código do passo 7 deu erro,fui pesquisar isso e descobri que O erro de tipos acontece porque process.env.DATABASE_HOST retorna string | undefined no TypeScript, mas o construtor do PrismaMariaDb espera string — sem o undefined. só coloquei os ! para ficar como not undefined

## Summary by CodeRabbit

* **Documentation**
  * Updated MySQL quickstart guide with improved environment variable handling in the Prisma adapter configuration to ensure proper validation of all required database credentials (host, user, password, and database name) for more reliable database connection setup.

### DIFF
--- a/apps/docs/content/docs/(index)/prisma-orm/quickstart/mysql.mdx
+++ b/apps/docs/content/docs/(index)/prisma-orm/quickstart/mysql.mdx
@@ -203,10 +203,10 @@ import { PrismaMariaDb } from "@prisma/adapter-mariadb";
 import { PrismaClient } from "../generated/prisma/client";
 
 const adapter = new PrismaMariaDb({
-  host: process.env.DATABASE_HOST,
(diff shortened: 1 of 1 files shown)


ankur-arch:
Obrigado por contribuir, @RAUNIFCE, e por relatar o problema com tanto detalhe. `process.env.X` is typed `string | undefined`, and the mariadb config fields are optional, so this example compiles under TypeScript's default strict settings. The error you saw appears when `exactOptionalPropertyTypes` is enabled in tsconfig, and in that case adding `!` or a fallback locally, as you did, is the right workaround. Because it depends on each project's TypeScript settings we are keeping the docs example as is, so I am closing this one. Thank you again for learning with the docs and taking the time to report back.

PR #7990, opened 2026-06-30
Fix command for scaffolding NestJS project

MarekBodingerBA:
The correct template name is `nest`, not `nestjs`.

https://github.com/prisma/create-prisma/blob/d6235e56ce34f600fa0ffb3ef0e9737b7ff2043a/src/types.ts#L16


ankur-arch:
Thank you for contributing, @MarekBodingerBA. You were right about the template name, and the NestJS guide on main now uses `--template nest`. Since the fix is already live, I am closing this one. We appreciate you taking the time to open it.

PR #7996, opened 2026-07-02
ppg-proxy series - proposal doc

pcan:


apps/blog/proposals/ppg-proxy-series.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+# 1 ppg-proxy intro: what, why and how
+
+1. What is Prisma Postgres, how/where it's deployed, what are the common access pattern to postgres in multi tenant settings (very briefly, we talked about microvms/unikernels extensively already)
+2. justifying the need for a "proxy" or "gateway": what responsibilities it takes:
+    - route postgres traffic to the target databases, preserving tenant isolation
+    - better auth scheme, api-key based, rotation without touching db credentials
+    - billing & holds based on queries, ingress/egress traffic, backup traffic, etc
+    - unified throttling, observability, access logging
```
ellotheth: The gateway will also give us a mechanism to cordon of the PPg hosts entirely from public traffic, which will be a huge security/maintenance win.

apps/blog/proposals/ppg-proxy-series.md
```diff
@@ -0,0 +1,70 @@
...
+3. Chanllenges and how we overcome them
+    - the routing data distribution: choices and tradeoffs
+    - The postgres binary protocol (message-oriented) and the open source libraries landscape
+    - buffering and the noisy neighbor problem: when horizontal scaling hides but doesn't solve the problem
+    - biting the bullet and implementing a full end-to-end low level streaming protocol capability, while still allowing selective messages inspection (e.g. error reporting or query counting)
```
ellotheth: Do you want to put our logging-to-`stdout` adventure in here too or nah 😆

pcan: YES YES YES!!!!

apps/blog/proposals/ppg-proxy-series.md
```diff
@@ -0,0 +1,70 @@
...
+
+# 2 ppg-proxy for serverless - part 1: Connection Pooling
+
+1. What are serverless workloads, how they compare to traditional stateful workloads, why they are convenient from a developer perspective, what are their limitations
+    - introducing stateless execution environment (e.g. aws Lambda) vs a traditional VM/bare metal server solution
+    - dissonance between serverless environment and "legacy" technologies such as RDBMs and their long-running persistent connection & transaction model, normally adopted in _persistent_ running environments such as ec2
+    - lack of support for native tcp/tls client connectivity (sometimes)
+    - possible approaches to bridge the gap: dedicated pooler, http/ws protocol wrapper, or pooler-as-a-service (accelerate, which does both things with its limitations)
+2. Tackling the throwaway-connection issue with pooling
+    - what's connection pooling, what's its main goal (to relieve pressure on the db for connection creation/teardown)
```
ellotheth: Another fun one here could be something about its _non_-goals, like query performance or latency, and why adding a pooler will actually slow you down sometimes.

pcan: good point!!!

apps/blog/proposals/ppg-proxy-series.md
```diff
@@ -0,0 +1,70 @@
+# ppg-proxy blogposts
+
+A proposal for a series of blogposts showing how we manage prisma postgres connectivity, discussing about the ppg-proxy, what problems it resolves, how it evolved, the challenges whe faced over time, and the revamp that we've been doing as a result.
+
+
+# 1 ppg-proxy intro: what, why and how
```
ankur-arch: Nice start!

apps/blog/proposals/ppg-proxy-series.md
```diff
@@ -0,0 +1,70 @@
...
+2. Towards a unified pipelining model for TCP & Serverless
+    - Connection lifecycle and protocol handling are separate problems, and how this helps shaping the new architecture.
+    - The BIG challenge: finding the boundary contract between the two, _while_ keeping the streaming promise.
+    - Step 1: Unifying wire protocol handling: from a mixed pgproto3/bespoke set of utilities to a well designed, cohesive raw pg protocol library with first-class streaming support.
+    - Step 2: The core pg-protocol pipeline
+    - Step 3: Serverless API & TCP terminal adapters
+    - A detour on fakepg and why we need it (testing connectivity issues, or just sequences of pre-baked messages)
+3. What's next
+    - Connection Pooling as a 2-facet system: the protocol sensing and the connection factory
+    - The plan to support advanced pooling features (first we do query detection, transaction boundary and then detecting unsupported statement as the last thing)
```
ankur-arch: @pcan here's some feedback from me and GPT:

## High-level structural feedback

* The series has strong raw material, but the structure is currently too implementation-led. It reads more like an internal architecture history than a public-facing blog series.
* The reader needs a clearer **“why should I care?”** arc before getting deep into protocol internals.
* Start with the developer/customer problem, not `ppg-proxy`.

  * “ppg-proxy intro: what, why and how” is accurate, but not very compelling as the entry point.
  * A stronger frame would be something like: **“How Prisma Postgres handles secure, scalable database connectivity.”**
  * Then introduce `ppg-proxy` as the system that makes that possible.

## Narrative arc

Right now, the progression feels like:

> proxy → pooling → HTTP/WS → revamp

A stronger arc would be:

> developer pain → platform constraints → connectivity architecture → serverless access → protocol/streaming design → unified future architecture

That gives the series a clearer reader-facing story instead of making it feel like a chronological engineering history.

## Audience clarity

The intended audience is currently unclear. Some parts speak to app developers, some to infra engineers, and some to Prisma employees.

I’d decide whether the series is primarily for:

* Prisma users who want to understand Postgres connectivity
* backend/platform engineers interested in architecture
* database/protocol nerds

My recommendation: target **senior backend/platform engineers**, but always anchor each post in user-visible product value.

## Series-level recommendations

* Each post needs a sharper promise. Before the outline, add one sentence per post answering: **“After reading this, the reader will understand X.”**
* Separate product value from engineering war stories.

  * The logging-to-stdout adventure, noisy neighbor issues, protocol rewrites, and `fakepg` are great.
  * But they should only appear where they support the main reader-facing point.
  * Otherwise, the series risks feeling like an internal postmortem.
* The proposal should define success for the series:

  * Is this meant to build trust in Prisma Postgres?
  * Explain serverless connectivity?
  * Recruit infra talent?
  * Support launch messaging?
* The structure should change depending on that goal.

## Post 1: ppg-proxy intro

* The first post is overloaded.
* It tries to explain Prisma Postgres, multi-tenancy, routing, auth, billing, throttling, observability, protocol challenges, buffering, noisy neighbors, and future plans.
* That is too much for an intro.
* Keep the intro focused on the gateway’s purpose and the core problems it solves.

Post 1 should establish the system boundary clearly:

* Is `ppg-proxy` a gateway?
* A TCP proxy?
* An auth layer?
* A routing layer?
* A metering point?
* An observability boundary?
* All of the above?

The outline lists responsibilities, but the reader needs a mental model early.

Also, the security angle should be promoted. The point about cordoning off Prisma Postgres hosts from public traffic is a big value point. That should not be a side note. It belongs in the main structure of post 1 because it makes the gateway feel necessary, not incidental.

The billing/metering point also needs careful placement. “Billing & holds based on queries, ingress/egress traffic, backup traffic” is important internally, but for an external audience it may feel vendor-centric. I’d frame it as:

* accurate usage visibility
* fair resource accounting
* abuse protection
* operational safety

## Post 2: serverless connection pooling

* Post 2 is the clearest structurally.
* The connection pooling post has a strong shape already:

  * problem
  * explanation
  * tradeoffs
  * Prisma Postgres approach

This is probably the best-structured outline in the proposal.

I’d add explicit non-goals to the pooling post. The review comment is spot on:

* pooling does not magically make queries faster
* pooling can add latency in some cases
* pooling mainly helps with connection lifecycle pressure, not query performance itself

That will make the post more trustworthy.

Also, avoid making the serverless intro too generic. “What are serverless workloads?” can easily become a 101 section readers skip. Keep it short and tie it directly to database connectivity:

* short-lived processes
* connection storms
* missing TCP support
* transaction semantics
* persistent database connections vs ephemeral compute

## Post 2 and Post 3 overlap

Post 2 and Post 3 both talk about serverless constraints and bridging TCP/TLS gaps.

I’d make the distinction explicit:

* **Post 2:** connection lifecycle pressure
* **Post 3:** transport/protocol limitations

That separation will help the series feel intentional instead of repetitive.

## Post 3: HTTP/WS API and driver

* Post 3 currently sounds too competitor-comparative.
* The Neon case study could be useful, but “poor/short-sighted model” reads unnecessarily spicy and may distract from Prisma’s own design.
* A better structure would be:

  * explain common industry approaches
  * describe their tradeoffs
  * show why Prisma chose a streaming-first design

Be careful with Data Proxy / Accelerate framing too. Lessons learned from past Prisma systems are valuable, but avoid making existing Prisma products sound broken. I’d frame it as:

> What we learned from operating query engines and managed proxies at scale.

Post 3 has the strongest technical hook: **streaming**. That should be the centerpiece.

The outline should lead with:

* why request/response breaks down
* why streaming matters
* what a streaming-first serverless Postgres API unlocks

The driver section risks becoming too broad. “A driver for every language” and JDBC examples are interesting, but they may belong in a future-looking closing section, not a major structural pillar. Keep the post centered on the API/driver contract.

## Post 4: ppg-proxy revamped

* Post 4 is too dense and too internal.
* It contains many important pieces, but it currently reads like an architecture planning document.
* For a blog post, I’d split or simplify it.
* The reader needs fewer moving parts and a more explicit payoff.

Post 4 needs a clearer before/after structure:

* **Before:** TCP proxy and serverless API evolved separately.
* **Problem:** duplicated protocol handling, inconsistent lifecycle logic, harder pooling.
* **After:** unified protocol pipeline with terminal adapters.
* **Benefit:** better pooling, testing, streaming, maintainability, and future features.

The current outline also buries one of the most interesting architectural ideas:

> Connection lifecycle and protocol handling are separate problems.

That is a strong conceptual takeaway and should be promoted earlier in Post 4.

A few specific notes:

* “AI-accelerated re-architecting” feels out of place unless it is central to the post. It risks aging badly and distracting from the actual architecture.
* `fakepg` should be a sidebar, not a main detour. It sounds fascinating, but structurally it can derail the post. Use it as a concrete testing example after the main architecture is already clear.

## Titles

The titles need work. They are descriptive but not reader-oriented.

“ppg-proxy for serverless - part 1” feels internal. I’d use titles that describe the problem/value instead:

* **Why Prisma Postgres needs a gateway**
* **Why serverless apps are hard on Postgres connections**
* **Designing a streaming API for serverless Postgres**
* **Unifying TCP and serverless connectivity in Prisma Postgres**

## Suggested repeated structure

The outline would benefit from a repeated structure across posts:

1. User/platform problem
2. Why the obvious solution breaks down
3. Prisma’s approach
4. Tradeoffs and lessons learned
5. What this unlocks next

This would make each post easier to write and easier to follow.

## Article endings

Each post should end with a concrete takeaway, not just “what’s next.”

The “what’s next” sections are useful, but each article should close with what the reader should now understand.

## Jargon

There is a lot of jargon without staged explanation.

Terms like these are fine for the audience, but they should be introduced progressively:

* ingress layer
* framing
* flow control
* terminal adapters
* clean/busy/dirty connections
* transaction boundary detection
* `pgproto3`
* sub-message streaming

## Overall recommendation

Keep the four-part series, but reframe it from:

> Here is how ppg-proxy evolved.

to:

> How Prisma Postgres is building a secure, serverless-friendly connectivity layer for Postgres.

That makes the series more valuable to external readers while still giving engineering enough room to go deep.

pcan: Thanks for the detailed pass, this reshaped the proposal. Restructured in 0fdaa605d.

Adopted:

1. A series-framing block up top: promise, audience (senior backend/platform engineers, anchored in product value), definition of success (technical trust first, recruiting signal second), and the repeated 5-step structure.
2. Your four reader-oriented titles, with a technical subtitle each.
3. A one-line "After reading, you'll understand X" promise per post, and a closing "Key takeaway" per post.
4. Post 1: the gateway mental model stated up front (it is all of those things, and co-locating them is deliberate), the security cordon promoted into the main structure, metering reframed as usage visibility, fair accounting, abuse protection, and operational safety.
5. Post 2: an explicit "Tradeoffs and non-goals" section, and a tight serverless intro tied directly to connectivity.
6. An explicit Post 2 vs Post 3 boundary: connection lifecycle pressure vs transport/protocol limitations.
7. Post 3: Neon as a neutral mechanics case study (dropped the spicy wording), Data Proxy/Accelerate framed as lessons from operating query engines and proxies at scale, and streaming leads the post.
8. Post 4: before/problem/after/benefit structure, "connection lifecycle and protocol handling are separate problems" promoted to the lead, "AI-accelerated re-architecting" dropped as a pillar, fakepg demoted to a sidebar.
9. War stories (stdout logging, noisy neighbor, fakepg) now appear only as sidebars supporting the reader-facing point.

Kept deliberately, since the series is meant to stay technically deep:

1. The Neon protocol deep-dive itself. It is the concrete anchor for "the API contract is the hard part, not the driver".
2. The full streaming chain and the client-side section (CollectableIterator, batching/pipelining), contained to the API/driver contract.
3. The JDBC / "driver for every language" material, moved to a forward-looking close as you suggested.
4. Post 4 as a single post rather than a split. The before/after structure should keep it readable. We can revisit at draft time if it runs long.

On the series goal: I set it to building technical trust in Prisma Postgres connectivity, with recruiting as a secondary effect. If it should serve launch messaging instead, say so and I will adjust the emphasis.

apps/blog/proposals/ppg-proxy-series.md
pcan: Adopted in 0fdaa605d: promoted into Post 1's main structure. The gateway mental-model section now states the cordon as a structural security and maintenance win, not a side note.

apps/blog/proposals/ppg-proxy-series.md
pcan: In 0fdaa605d it is now a Post 1 sidebar: a short war story on why observability at a gateway is harder than it looks.

apps/blog/proposals/ppg-proxy-series.md
pcan: Adopted in 0fdaa605d: Post 2 now has a dedicated "Tradeoffs and non-goals" section. Pooling does not make queries faster, can add latency, and helps connection lifecycle pressure only.

apps/blog/proposals/ppg-proxy-series.md
pcan: Thanks! Restructured in 0fdaa605d per the full review below.

PR #8010, opened 2026-07-04
docs: add datasource env() example

AlexanderPach:
## Summary
The previous example introduced env() but didn't show its usage, making it harder for users to understand how to implement it.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Prisma schema example to show `env("DATABASE_URL")` for the datasource URL, making the environment variable usage clearer.

### DIFF
--- a/apps/docs/content/docs/orm/prisma-schema/overview/index.mdx
+++ b/apps/docs/content/docs/orm/prisma-schema/overview/index.mdx
@@ -139,6 +139,7 @@ Environment variables can be accessed using the `env()` function:
 ```prisma
 datasource db {
   provider = "postgresql"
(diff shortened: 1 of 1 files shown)


AlexanderPach:
error: The datasource property url is no longer supported in schema files. Move connection URLs for Migrate to prisma.config.ts and pass either adapter for a direct database connection or accelerateUrl for Accelerate to the PrismaClient constructor. See https://pris.ly/d/config-datasource and https://pris.ly/d/prisma7-client-config

Running this on Prisma CLI Version : 7.8.0 shows that url is no longer supported. So, this PR is no longer relevant.

PR #8044, opened 2026-07-09
skills: mandatory adversarial review in content-write-blog

vanrensbird:
## What

Adds a required **adversarial review** gate to the `content-write-blog` skill (version 2026.7.9): every finished draft is reviewed by fresh-context agents with refutation missions before any PR is opened. Reviewed and approved by Martin.

## The process

- **One reviewer per applicable lens, regardless of post length**: fact refuter (every number/price/quote verified against live primary sources), code refuter (samples run fresh on stated versions, when the post has code), reader skeptic (overreach, contradictions, vendor tilt), positioning refuter (positioning-doc rules, product-status overclaims).
- **Isolation rules**: reviewers get only the draft path and their mission, never the writing conversation or intent; refutation missions with skepticism as the default.
- **Loop**: implementer fixes, reviewers re-review, repeat until a round returns zero confirmed findings; the operator receives the per-round summary, not just the verdict.
- Step 9 hand-off now points to the gate; subagents vs. an orchestrated workflow is named as an implementation choice, not a review-depth choice.

## Why

Proven on the first post written under it (serverless Postgres guide, in review separately): three rounds caught 42 findings, including two flatly wrong claims that had survived drafting and research (a product capability that does not exist, per our own docs, and a local/cloud engine conflation), a misquote of a competitor's docs, and a systematic only-vendor-with-no-drawbacks tilt invisible from inside the writing context.

## Summary by CodeRabbit

* **Documentation**
  * Updated the skill guide to a new version.
  * Clarified the hand-off step so the publishing workflow now requires an extra review step before any pull request is opened.
  * Added a new adversarial review stage with clearer review criteria and completion requirements for final publishing.


vanrensbird:
Updated per Martin's review: adversarial review is now **Step 10** of the skill's numbered process (the gate to any PR), not an appendix section, and it runs **one reviewer per applicable lens regardless of post length** (fact/reader/positioning always, code when the post has runnable code). Ready for review.

PR #8063, opened 2026-07-12
docs(docs): fix pnpm workspace client imports

### LINKED ISSUE #7944: [docs] pnpm workspaces guide: import path ./generated/client doesn't work with Prisma 7
The pnpm workspaces guide (https://www.prisma.io/docs/guides/deployment/pnpm-workspaces) shows this import in client.ts and index.ts:
`import { PrismaClient } from "./generated/client";
`

This works in Prisma 6 (prisma-client-js), where the generator creates an index.d.ts + index.js barrel in the output directory.
In Prisma 7 with the new prisma-client provider, the generator only outputs TypeScript source files (.ts) without a barrel index.ts (see issue #29525 (https://github.com/prisma/prisma/issues/29525)). The output structure is:

generated/client/
├── client.ts              ← PrismaClient lives here
├── models.ts
├── enums.ts
├── commonInputTypes.ts
├── internal/
└── models/

Since there is no index.ts or index.d.ts at generated/client/, the import "./generated/client" fails with Cannot find module './generated/client' or its corresponding type declarations.

The correct import when using Prisma 7 is:
`import { PrismaClient } from "./generated/client/client";`

Or alternatively, the guide should recommend wrapping exports through a barrel file in the database package (e.g., index.ts) to hide this implementation detail from consumers.

Suggested fix: update the guide to either:

1. Change the import paths to "./generated/client/client", or
2. Add a note explaining that Prisma 7 requires this path, and suggest using a barrel re-export for external consumers.

jaeyunjks:
Linear: DC-7944

## Summary

- Updates the Prisma Client import in the pnpm workspaces guide to point to `./generated/client/client`.
- Updates the generated client re-export to use the same Prisma 7 entry point.
- Aligns the guide with the output structure of the `prisma-client` generator, which does not generate a root `index.ts` barrel file.

## Testing

- Ran `git diff --check`.
- Confirmed no outdated `./generated/client` imports remain in the guide.
- Verified that Prisma's generated `client.ts` is the main server-side entry point and exports Prisma Client, models, enums, and generated types.

Fixes #7944

## Summary by CodeRabbit

* **Documentation**
  * Updated the pnpm workspaces deployment guide to use the current generated Prisma Client folder structure.
  * Corrected Prisma Client import and type re-export paths in the setup examples.

### DIFF
--- a/apps/docs/content/docs/guides/deployment/pnpm-workspaces.mdx
+++ b/apps/docs/content/docs/guides/deployment/pnpm-workspaces.mdx
@@ -202,7 +202,7 @@ When prompted by the CLI, enter a descriptive name for your migration.
 Once the migration is successful, create a `client.ts` file to initialize Prisma Client with a driver adapter:
 
 ```ts title="database/client.ts"
-import { PrismaClient } from "./generated/client"; // [!code ++]
+import { PrismaClient } from "./generated/client/client"; // [!code ++]
 import { PrismaPg } from "@prisma/adapter-pg"; // [!code ++]
 // [!code ++]
 const adapter = new PrismaPg({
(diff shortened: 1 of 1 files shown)


ankur-arch:
Thank you for contributing, @jaeyunjks, and for catching the wrong import path. The same fix landed through #8266 on the page's new location under `guides/v7/deployment/`, so the docs now show `./generated/client/client` as you proposed. Closing this one as superseded, with thanks for flagging it.

PR #8071, opened 2026-07-15
Replace newsletter confirmation with a welcome email

gregory-boch-prisma:
## What
- subscribe website, blog, and docs newsletter forms immediately
- send Brevo welcome template 228 only when a public contact enters list 15
- preserve original `SOURCE` attribution and record the current path in `NEWSLETTER_SOURCE`
- record consent evidence in `NEWSLETTER_CONSENT_AT` (ISO 8601) on every (re)subscription
- suppress duplicate welcomes and bound Brevo API work to a single 8-second whole-workflow budget
- add a newsletter-only unsubscribe flow that preserves transactional email delivery
- remove subscriber PII from recoverable error logs

Related Console source marker: prisma/pdp-control-plane#4560.

Two silent fixes bundled in the shared route helper, called out per review:
- blog and docs previously hardcoded `SOURCE: "website"`, so their attribution was wrong; each app now passes its own source
- docs and site previously sent a malformed comma-separated `Access-Control-Allow-Origin` header; the shared helper returns a single origin and `Vary: Origin`

`NEWSLETTER_UNSUBSCRIBE_URL` is intentionally hardcoded to production `www.prisma.io`: welcome emails triggered from previews and staging reach real inboxes, so their unsubscribe links must work in production.

## Review follow-ups in this revision
- `NewsletterSubscriptionError` → `NewsletterServiceError` with a closed union of domain codes; Brevo's own error code is preserved separately as `providerCode`
- unsubscribe tokens are now random (`crypto.getRandomValues`), no longer HMAC-derived from the Brevo API key; already-issued links keep working because lookups use the stored attribute
- `subscribeToPrismaNewsletter` validates and normalizes the email itself; route validation remains as UX
- welcome-email failures keep the subscription but now surface `code`/`status`/`providerCode` in the result and in route logs (still no subscriber PII), so a timeout is distinguishable from an inactive template during the 36 → 228 cutover
- timeouts are mapped to a dedicated `brevo_request_timed_out` code; the 8s constant is renamed `BREVO_WORKFLOW_TIMEOUT_MS` and documented as a whole-workflow budget (lookup → upsert → welcome), matching the intent above
- Brevo transport (headers, JSON handling, error conversion) is consolidated in a thin `brevoRequest` helper; business rules stay in the callers
- `listIds` documented as additive per the Brevo docs (`unlinkListIds` is the removal field, used by unsubscribe)
- welcome idempotency semantics documented: at most one welcome per address within Brevo's idempotency retention window; double-submits collapse, resubscribes inside the window get no second welcome

## Compliance — explicit rollout gate, not yet approved
- This PR moves marketing email signup from double opt-in (template 36) to single opt-in. That is a legal/product decision, not a technical one: **Legal sign-off is a pre-deploy rollout gate. It has not been obtained and this PR does not claim it.** Review flagged that DE/AT case law effectively requires DOI for marketing sends.
- Consent evidence: `NEWSLETTER_CONSENT_AT` is stored with each (re)subscription alongside `NEWSLETTER_SOURCE` and the preserved original `SOURCE`.
- Console silent opt-ins and transactional emails are unchanged. Unsubscribe removes the contact from newsletter list 15 only and never blacklists the address.

## Brevo
- template 228 is active and previewed with the new unsubscribe URL
- `NEWSLETTER_UNSUBSCRIBE_TOKEN` and `NEWSLETTER_SOURCE` contact attributes are live
- ⚠️ `NEWSLETTER_CONSENT_AT` (text) contact attribute must be created **before** this deploys — Brevo rejects payloads with unknown attributes
- template 36 stays active until this PR is deployed, then it can be deactivated

## Verification
- 27 focused validation/subscription/unsubscribe/route tests (`tsx --test` in `packages/ui`)
- site, blog, and docs type checks
- oxfmt and oxlint
- desktop (1440 px) and 390 px mobile visual check re-run 2026-07-24 on `b535238`: newsletter signup page, unsubscribe confirm state (token cookie flow), and invalid-link state

## Rollout
1. **Gate:** Legal sign-off on single opt-in for marketing email
2. **Gate:** create the `NEWSLETTER_CONSENT_AT` (text) contact attribute in Brevo
3. Deploy
4. Verify one new public subscription and one repeated subscription in Brevo
5. Verify a contact on multiple lists keeps its other memberships after a resubscribe (additive `listIds`) and after a newsletter unsubscribe (`unlinkListIds`)
6. Deactivate legacy DOI template 36

## Follow-up
- move the newsletter server modules (`newsletter-subscription.ts`, `newsletter-route.ts`) out of `packages/ui` into a dedicated package so the UI package stays client-safe (review ask)


packages/ui/src/lib/newsletter-route.ts
```diff
@@ -0,0 +1,114 @@
...
+        return NextResponse.json(
+          { error: "A valid JSON body is required" },
+          { status: 400, headers: corsHeaders },
+        );
+      }
+
+      const email =
+        typeof body === "object" && body !== null && "email" in body ? body.email : undefined;
+
+      if (typeof email !== "string" || !emailPattern.test(email.trim())) {
```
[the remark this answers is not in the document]
gregory-boch-prisma: Fixed in d6a3e87 by replacing the backtracking regex with a bounded linear-time validator and adding the reported repeated-punctuation case as a regression test.

gregory-boch-prisma:
Argos note: the single changed snapshot is unrelated to this PR's newsletter UI. Both reruns point to `apps/eclipse/tests/example.spec.ts` (`chromium/homepage.png`). The baseline captured the Docs introduction page at 1280×1316, while the branch run captured the Eclipse design-system page at 1280×9360. Argos reports this snapshot at 53% stability / 10% consistency (42 changes across 89 builds).

I did not approve the visual change. It needs an Argos reviewer to mark the cross-app capture as flaky/unrelated: https://app.argos-ci.com/developer-connections/web/builds/7932

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
+const BREVO_API_BASE_URL = "https://api.brevo.com/v3";
```
ankur-arch, line 1: **Module placement.** `@prisma-docs/ui` is documented as a shadcn-style UI component package with source-file exports; this module is server-only business logic (crypto, Brevo transport), and its sibling `newsletter-route.ts` imports `next/server`. Long-term this belongs in something like `packages/newsletter` so the UI package stays client-safe and this code can grow its own tests and dependencies. Fine to land here now — but let's file the move as a follow-up before more server code accretes in `ui`.

Within the module, the next cleanup is separating Brevo *transport* from newsletter *business rules*: a small `brevoRequest<T>(path, init)` helper (headers, fetch, `readJson`, `response.ok`, error conversion) removes the repeated request lifecycle from all five call sites and makes a future provider swap or unit test much easier. Keep it thin — consistent transport behavior, not a full SDK.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+export type NewsletterSubscriptionResult = {
+  status: "already_subscribed" | "subscribed";
+  welcomeSent: boolean;
+};
+
+export type NewsletterUnsubscribeResult = {
+  status: "already_unsubscribed" | "invalid_token" | "unsubscribed";
+};
+
+export class NewsletterSubscriptionError extends Error {
```
ankur-arch: This class is also thrown from unsubscribe and malformed-response paths, so the name is narrower than its usage — `NewsletterServiceError` fits better. While renaming, make `code` a typed union instead of a free-form string:

```ts
type NewsletterErrorCode =
  | "invalid_brevo_response"
  | "brevo_request_failed"
  | "invalid_email";
```

Right now `toBrevoError` passes through `body.code` verbatim, which mixes Brevo's vocabulary into our domain codes — log queries and caller behavior stay stable if we map provider codes to our own union at the boundary.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+export function isValidNewsletterUnsubscribeToken(token: string): boolean {
+  return unsubscribeTokenPattern.test(token);
+}
+
+function getExistingUnsubscribeToken(contact: BrevoContact | null): string | null {
+  const token = contact?.attributes[NEWSLETTER_UNSUBSCRIBE_TOKEN_ATTRIBUTE];
+  return typeof token === "string" && isValidNewsletterUnsubscribeToken(token) ? token : null;
+}
+
+async function createUnsubscribeToken(apiKey: string, email: string): Promise<string> {
```
ankur-arch: **Don't use the Brevo API key as the HMAC secret.** It couples two unrelated concerns: Brevo authentication and public unsubscribe-link identity. Because lookups go through the stored `NEWSLETTER_UNSUBSCRIBE_TOKEN` attribute, existing links survive a key rotation — but new tokens for the same email silently change after rotation, and the API key quietly gains a second job as a signing secret with different scoping needs.

Two options, either fine:

1. A dedicated `NEWSLETTER_UNSUBSCRIBE_SECRET` env var.
2. Simpler: a **random** token via `crypto.getRandomValues`. Determinism buys nothing here — you already reuse the stored token when present (`getExistingUnsubscribeToken`), and unsubscribe lookup is storage-based, so random tokens need zero extra infrastructure and are revocable by rewriting the attribute.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+  const originalSource = contact?.attributes.SOURCE;
+  const body = {
+    attributes: {
+      EMAIL: email,
+      [NEWSLETTER_UNSUBSCRIBE_TOKEN_ATTRIBUTE]: unsubscribeToken,
+      [NEWSLETTER_SOURCE_ATTRIBUTE]: source,
+      SOURCE: typeof originalSource === "string" && originalSource ? originalSource : source,
+    },
+    emailBlacklisted: false,
+    listIds: [PRISMA_NEWSLETTER_LIST_ID],
```
ankur-arch, line 253: Per Brevo's `updateContact` semantics, `listIds` is additive (`unlinkListIds` is the removal field, which the unsubscribe path correctly uses), so this shouldn't clobber other list memberships. Please verify once against a live multi-list contact during rollout anyway — silent list-loss would be hard to detect after the fact — and add a one-line comment stating "additive per Brevo docs" so nobody 'fixes' it later.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+        headers: getBrevoHeaders(apiKey),
+        body: JSON.stringify({ email, updateEnabled: true, ...body }),
+        signal,
+      });
+
+  const responseBody = await readJson(response);
+  if (!response.ok) throw toBrevoError("contact subscription", response, responseBody);
+}
+
+async function createWelcomeIdempotencyKey(email: string): Promise<string> {
```
ankur-arch, line 277: The idempotency key derives from the email alone, so it's stable across resubscriptions. Nice property: a double-submit race (two concurrent POSTs both seeing 404) collapses into one welcome email. But the flip side: depending on Brevo's idempotency retention window, a user who unsubscribes and later resubscribes may silently get **no** welcome while we still return `welcomeSent: true`.

Pick the semantic explicitly:

- *Welcome once ever per address* → current code is right; document it here.
- *Welcome once per (re)subscription* → mix a subscription event component into the digest input.

Either is defensible; the ambiguity is the problem.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+      },
+    }),
+    signal,
+  });
+
+  const body = await readJson(response);
+  if (!response.ok) throw toBrevoError("welcome email send", response, body);
+}
+
+export async function subscribeToPrismaNewsletter({
```
ankur-arch, line 310: `subscribeToPrismaNewsletter` never calls `isValidNewsletterEmail` — correctness depends on every caller remembering to validate first. The route does today; the next caller may not. The exported service should enforce its own invariant:

```ts
const normalizedEmail = email.trim().toLowerCase();
if (!isValidNewsletterEmail(normalizedEmail)) {
  throw new NewsletterSubscriptionError("Invalid newsletter email", { code: "invalid_email" });
}
```

Keep the route/UI validation too, but treat it as UX rather than enforcement.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+  apiKey,
+  email,
+  fetcher = fetch,
+  source,
+}: SubscribeToNewsletterOptions): Promise<NewsletterSubscriptionResult> {
+  const normalizedEmail = email.trim().toLowerCase();
+  const signal = AbortSignal.timeout(BREVO_REQUEST_TIMEOUT_MS);
```
ankur-arch: This single signal spans lookup + upsert + welcome, so 8s is a **whole-workflow** budget, not per request — a 6s lookup leaves 2s for everything after it. The PR description says "bound Brevo API work to 8 seconds", so maybe that's intentional; if so, rename to `BREVO_WORKFLOW_TIMEOUT_MS` and say so in a comment.

If per-request was the intent, create a fresh `AbortSignal.timeout(...)` inside each request (ideally inside a shared `brevoRequest` helper) — which also deletes the `signal` parameter from every function signature. Same question applies to `unsubscribeFromPrismaNewsletter` below.

packages/ui/src/lib/newsletter-subscription.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  const unsubscribeToken =
+    getExistingUnsubscribeToken(contact) ?? (await createUnsubscribeToken(apiKey, normalizedEmail));
+
+  await upsertContact(fetcher, apiKey, normalizedEmail, source, contact, unsubscribeToken, signal);
+
+  try {
+    await sendWelcomeEmail(fetcher, apiKey, normalizedEmail, unsubscribeToken, signal);
+    return { status: "subscribed", welcomeSent: true };
+  } catch {
```
ankur-arch: This empty catch discards the failure reason entirely — a timeout, a 401 (rotated key), a 503, and a deactivated template 228 all become indistinguishable downstream; the route can only log "completed without welcome email". The graceful `welcomeSent: false` result is right, but capture the cause:

```ts
} catch (error) {
  const details = error instanceof NewsletterSubscriptionError
    ? { code: error.code, status: error.status }
    : { code: "unexpected_error" };
  console.warn("Newsletter welcome email failed", details);
  return { status: "subscribed", welcomeSent: false };
}
```

No raw email, API key, or token in the log — the existing tests already assert PII stays out of errors, which is good.

packages/ui/src/lib/newsletter-route.ts
```diff
@@ -0,0 +1,113 @@
...
+        return NextResponse.json(
+          { error: "Newsletter service is not configured" },
+          { status: 500, headers: corsHeaders },
+        );
+      }
+
+      try {
+        const result = await subscribeToPrismaNewsletter({ apiKey, email, source });
+
+        if (!result.welcomeSent && result.status === "subscribed") {
```
ankur-arch, line 74: By the time this warns, the underlying cause was already discarded by the empty `catch` in `subscribeToPrismaNewsletter`, so `{ source }` is all we can log. Once the service surfaces the failure reason (see comment there), include `code`/`status` here — during the template-36 → 228 cutover, distinguishing "template inactive" from "timeout" is exactly the signal we'll need.

ankur-arch (commented):
## Overview

This consolidates three copy-pasted newsletter routes into a shared `createNewsletterRoute` factory plus a Brevo service module, switches from double opt-in (template 36) to immediate subscribe + welcome (template 228), and adds a token-based newsletter-only unsubscribe flow.

The decomposition into small functions is good, and the unsubscribe design is genuinely well thought out: GET stores the token in a short-lived `httpOnly` cookie and redirects (so tokens don't leak via referrer and mail scanners can't auto-unsubscribe), POST confirms, the page is `noindex`/`no-referrer`. Tests cover the main subscribe/resubscribe/unsubscribe paths.

Main asks before merge, in order:

1. **Confirm the compliance story for dropping double opt-in.** The old README explicitly said DOI "ensures compliance with GDPR"; this PR deletes that. Moving to single opt-in for marketing email is a legal/product decision, not just technical — DE/AT case law effectively requires DOI for marketing sends. Please confirm sign-off, and consider recording a consent timestamp alongside `NEWSLETTER_SOURCE` so we can evidence consent later.
2. Stop discarding the welcome-email failure reason (inline comment).
3. Clarify whether the 8s timeout is per-request or whole-workflow (inline comment).
4. Validate email inside the exported service, not only in the route (inline comment).

## Affected paths

- `POST /api/newsletter` on **site**, **blog**, and **docs** now share one implementation. Two silent fixes worth calling out in the PR description: blog and docs previously hardcoded `SOURCE: "website"` (attribution was wrong), and docs/site sent a malformed comma-separated `Access-Control-Allow-Origin` header (invalid per spec). Both are fixed by the shared helper.
- New site-only paths: `/newsletter/unsubscribe` (page) and `/api/newsletter/unsubscribe` (GET/POST).
- `NEWSLETTER_UNSUBSCRIBE_URL` is hardcoded to production `www.prisma.io` — welcome emails triggered from previews/staging will carry production unsubscribe links. Probably intentional; worth a comment saying so.
- Docs README updated to match.

## Design notes (non-blocking)

- **The lookup-before-upsert is justified here** — the GET powers `already_subscribed` welcome suppression and token reuse, which is the whole point of the PR. Don't collapse it into a blind upsert unless suppression moves provider-side.
- **Welcome send sits on the user-facing request path** (lookup → upsert → welcome, sequential). Fine at current traffic; the long-term shape is enqueue-after-subscribe so retries and latency decouple. Not blocking.
- **Tests to add next**: timeout/abort behavior, invalid-JSON Brevo responses, CORS origin fallback, unsubscribe filter returning 2 contacts, and route-level rejection of a missing/invalid body.

Inline comments follow for the file-level items.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: Brevo transport (headers, JSON handling, error conversion, timeout mapping) now lives in a thin `brevoRequest` helper used by all five call sites; business rules stay in the callers. The move out of `packages/ui` is recorded as a follow-up in the PR description — happy to file the issue once we agree on the target package name.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: renamed to `NewsletterServiceError` with a closed union (`brevo_request_failed | brevo_request_timed_out | invalid_brevo_response | invalid_email`). Brevo's own code is no longer passed through as `code` — it's preserved separately as `providerCode` at the transport boundary, so log queries stay stable while the provider detail survives.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: went with option 2 — tokens are now random via `crypto.getRandomValues` (32 bytes, same 64-hex format), so no secret is shared with the API key. Already-issued links keep working since lookups use the stored attribute; a test asserts tokens are unpredictable across subscriptions and that existing tokens are reused.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: added the "additive per Brevo docs" comment (and noted `unlinkListIds` as the removal field). Live verification against a multi-list contact is now rollout step 5 in the PR description.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: chose and documented option A — at most one welcome per address within Brevo's idempotency retention window: double-submits collapse, and unsubscribe → resubscribe inside the window intentionally gets no second welcome (a resubscribe after the window ages out gets a fresh one). A test asserts the key is stable per address and distinct across addresses.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: `subscribeToPrismaNewsletter` now normalizes (trim + lowercase) and validates itself, throwing `NewsletterServiceError` with code `invalid_email` (no PII in the message). Route/UI validation stays as UX. Test asserts the service rejects without any Brevo call.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: whole-workflow was the intent — renamed to `BREVO_WORKFLOW_TIMEOUT_MS` with a comment stating one deadline spans lookup → upsert → welcome (and the unsubscribe flow), matching the PR's 8-second bound. Timeouts/aborts now surface as a dedicated `brevo_request_timed_out` code via `brevoRequest`, with tests.

packages/ui/src/lib/newsletter-subscription.ts
gregory-boch-prisma: Addressed in b535238: the catch now captures `{ code, status, providerCode }` into the result as `welcomeFailure`, warns with exactly those fields (never the address), and still returns `welcomeSent: false`. Tests cover a 503 with provider code and a timed-out send, and assert the log contains no email.

packages/ui/src/lib/newsletter-route.ts
gregory-boch-prisma: Addressed in b535238: the route warn now spreads `welcomeFailure`, so it logs `{ source, code, status, providerCode }` — a route test asserts the `template_inactive` case logs exactly those fields with no address, which is the cutover signal you called out.

gregory-boch-prisma:
Re-review requested — all asks from the review are addressed in b535238 (inline replies on each thread). Highlights: typed `NewsletterServiceError` codes with `providerCode` at the boundary, random unsubscribe tokens, service-level validation, surfaced welcome-failure reasons, documented whole-workflow timeout, `brevoRequest` transport helper, and `NEWSLETTER_CONSENT_AT` consent evidence. The PR description now carries the compliance section (Legal sign-off on single opt-in stays an explicit pre-deploy gate — not claimed), the two silent fixes, rollout steps incl. multi-list verification and Brevo attribute creation, and the `packages/ui` move follow-up. Tests are 11 → 27; desktop + 390 px visual checks re-run on this revision. Branch is rebased on current main; all 17 checks green.

ankur-arch (approved):

PR #8082, opened 2026-07-17
docs(postgres): create-db picks closest region by default, us-east-1 is the fallback

nurul3101:
## Summary

The [npx-create-db docs page](https://www.prisma.io/docs/postgres/npx-create-db) claimed "The default region is `us-east-1`." This is stale as of create-db v1.2.1: when `--region` is omitted, the CLI geolocates the caller and provisions the database in the closest available region, falling back to `us-east-1` only when geolocation fails.

This PR updates the sentence to describe the geo-detection behavior with `us-east-1` as the fallback.

## Evidence

- `packages/create-db/src/cli/commands/create.ts` in [prisma/create-db](https://github.com/prisma/create-db/blob/main/packages/create-db/src/cli/commands/create.ts):
  ```ts
  let region: RegionId = input.region ?? "us-east-1";

  if (!input.region) {
    const userLocation = await detectUserLocation();
    region = getRegionClosestToLocation(userLocation) ?? region;
  }
  ```
- A live run from ap-southeast (2026-07-17) with no `--region` flag provisioned in `ap-southeast-1`.

## Scope check

Swept all docs pages mentioning `create-db` or `us-east-1` for the same stale claim; this was the only instance. Other `us-east-1` mentions are verbatim sample CLI output, explicit parameters (Management API, IaC, GitHub Actions), or a different product's default (Compute).

## Summary by CodeRabbit

* **Documentation**
  * Updated the CLI quick-start guide to clarify that the default database region is detected automatically, with a fallback to `us-east-1`.
  * Clarified that users can select a region manually with the `--region` option.


nurul3101:
@coderabbitai approve

PR #8108, opened 2026-07-25
Make authorEmail unique in Post model

DixonKimandu:
Add the @unique attribute to the authorEmail field in the example to show reference of a non-ID field

## Summary by CodeRabbit

* **Documentation**
  * Corrected the one-to-many relation example to mark the referenced author email field as unique, ensuring the schema example is valid.


ankur-arch:
Thank you for contributing, @DixonKimandu, and for thinking carefully about the example. In this schema `Post.authorEmail` references `User.email`, and `User.email` already carries `@unique`, which is what a relation to a non-ID field requires. Adding `@unique` to `Post.authorEmail` as well would allow only one post per author, which turns the example into a one-to-one relation rather than the one-to-many this page documents. For that reason I am closing this without merging, but we appreciate you engaging with the docs at this level of detail.

PR #8125, opened 2026-08-03
feat(docs): serve the Prisma Next error reference, with anchors matching emitted docsUrls

wmadden-electric:
Every error Prisma Next emits links to its own documentation:

```
PrismaError [CONTRACT.IDENTIFIER_INVALID]: ...
  docsUrl: https://docs.prisma.io/docs/orm/next/reference/error-reference#CONTRACT.IDENTIFIER_INVALID
```

Today that URL 404s. After this PR it renders the full error reference (244 codes) and the browser scrolls straight to the `CONTRACT.IDENTIFIER_INVALID` heading.

## The decision

We publish the error reference as a **generated page**, synced from its canonical source — `docs/reference/error-reference.md` on `prisma/prisma` `main` — and we never hand-edit it here. Upstream CI already guarantees that file documents every code the product ships, so the docs site treats it as an artifact to ingest, not content to maintain.

## How it works

**The page.** `apps/docs/scripts/generate-error-reference.mjs` converts the upstream markdown into `content/docs/orm/next/reference/error-reference.mdx`: it adds the frontmatter this site requires, rewrites repo-relative links to GitHub URLs, and refuses input that would break the MDX build (stray braces/JSX, missing or duplicate codes).

**The anchors.** This is the part most likely to bite. Fumadocs slugifies headings, so `### CONTRACT.IDENTIFIER_INVALID` would get the id `contractidentifier_invalid` — and every emitted `docsUrl` fragment (raw code: uppercase, with the dot) would silently fail to scroll. The generator therefore appends Fumadocs' explicit-anchor syntax to each code heading:

```md
### CONTRACT.IDENTIFIER_INVALID [#CONTRACT.IDENTIFIER_INVALID]
```

which makes the element id the raw code text. A new Playwright test (`apps/docs/tests/error-reference-anchors.spec.ts`) navigates to the fragment URL and asserts the heading has the raw id and lands in the viewport, so a Fumadocs upgrade can't regress this quietly.

**Staying current.** Two independent mechanisms:

1. `sync-error-reference-docs.yml` regenerates the page daily (plus manual and `repository_dispatch: error-reference-updated` triggers), verifies it, commits as Prismo, and triggers the Vercel deploy hook — the same pattern as the management-API docs sync.
2. `error-reference-check.yml` runs upstream's own verifier (`list-error-codes.mjs --verify`) against the committed page, so even if the sync breaks, drift is caught: on pushes to `main`, on PRs that touch the page or its tooling, and on a daily schedule.

The generated file is excluded from cspell so upstream wording changes can't fail spellcheck on `main`.

## Looking ahead

When Prisma 8 reaches RC, the ORM `next` channel becomes `v8` and this page moves with it to `/docs/orm/v8/reference/error-reference`; the product then flips one constant in its `docsUrl` template.

## Alternatives considered

- **Hand-maintained copy in this repo** — rejected: 244 codes and growing; it would drift the day after it lands.
- **Fetching the markdown at build time** — rejected: every docs build would depend on GitHub being reachable and would not be reproducible. Committing the generated file keeps builds hermetic; the sync workflow owns freshness.
- **Running the drift check on every PR** — tried first, walked back after review: an upstream code addition would block unrelated PRs that can't fix it. Path-restricted PR trigger + daily schedule keeps the guarantee without the blast radius.
- **Keeping slugified anchors and mapping fragments in client JS** — rejected: the product already ships raw-code fragments; explicit ids are static, testable, and need no runtime code.


sabbirbaze-lab (commented):
Md

wmadden-electric:
@coderabbitai resolve

All findings addressed in 10a1a704a: token permissions restricted, persist-credentials disabled on all checkouts, token/ref moved into env, PR trigger path-restricted with a daily schedule, and the generator now rewords the source repo's self-referential intro sentence.

PR #7479, opened 2026-02-12
chore(): upload new docs

aidankmcalister:
## Summary by CodeRabbit

* **New Features**
  * Added extensive blog content covering Prisma features, guides, and case studies.
  * Introduced automated Playwright-based testing for documentation.
  * Added new link validation workflow.

* **Documentation**
  * Updated contribution guidelines with formalized workflow and commit conventions.
  * Added developer guidance documents for AI agents and Claude integration.

* **Chores**
  * Restructured configuration files and removed obsolete workflow automation.
  * Updated linting and formatter configurations.

PR #8158, opened 2026-08-23
docs: restore nodejs_compat flag on Cloudflare deployment page

liamiepops:
The Cloudflare deployment page tells you to set `node_compat = true` in `wrangler.toml`. Cloudflare replaced that with the `nodejs_compat` compatibility flag, and the current Prisma guide at `/guides/deployment/cloudflare-workers` already uses the new form, so the two pages contradict each other.

This was already fixed. #7468 made the same change on 3 February. It was lost when #7479 migrated the docs to Fumadocs on 12 February: the fixed file was renamed into `docs.v6`, and the current v7 page came from a copy that still had the old text.

This PR restores those changes. The `compatibility_date` in the example is set to today rather than the February date.

I hit this running Prisma 7.9.1 under workerd with `nodejs_compat` set in `wrangler.jsonc`.

Three things to note:

- The commit has no Linear reference. I can't create one as an external contributor.
- The same text is still present in the v6 copy of the page. I left it alone since it's archived.
- #7479, which knocked this fix out, touched 3,630 files. Five commits touched `apps/docs/content/` between #7468 and #7479. Four added new guides, which are unaffected. The fifth, #7442, was a fix like this one, and it got lost in the merge too, so the SQL Server page reads "seconds" again instead of "milliseconds". I'm not sure what it should read though, the PostgreSQL and MySQL pages say seconds, so I'm not sure milliseconds was correct, or what the intention was there. Someone needs to untangle that one. I only checked `apps/docs/content/`, so I don't know whether anything outside the docs was affected.

## Summary by CodeRabbit

* **Documentation**
  * Updated PostgreSQL deployment instructions for Cloudflare.
  * Added a `wrangler.jsonc` configuration example using `nodejs_compat` and a compatibility date.
  * Removed outdated Cloudflare Pages workaround guidance.

### DIFF
--- a/apps/docs/content/docs/orm/prisma-client/deployment/edge/deploy-to-cloudflare.mdx
+++ b/apps/docs/content/docs/orm/prisma-client/deployment/edge/deploy-to-cloudflare.mdx
@@ -185,7 +185,7 @@ model User {
 If you are using a traditional PostgreSQL database that's accessed via TCP and the `pg` driver, you need to:
 
 - use the `@prisma/adapter-pg` database adapter (learn more [here](/orm/core-concepts/supported-databases/postgresql#using-driver-adapters))
-- set `node_compat = true` in `wrangler.toml` (see the [Cloudflare docs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/))
+- set `nodejs_compat` in `wrangler.jsonc` (see the [Cloudflare docs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/))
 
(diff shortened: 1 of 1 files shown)


ankur-arch:
Thank you for contributing, @liamiepops, and for keeping the Cloudflare page current with the `nodejs_compat` flag. The page has since been split into `orm/v6` and `orm/v7` copies, which is why this branch conflicts. I re-applied your change on both current pages in #8283 with credit to you, so your fix is now live. Closing this one.

PR #8233, opened 2026-09-09
docs: add navbar section breadcrumb and make sidebar drill-in sticky

tylerhogarth:
The docs navbar showed only `Prisma / docs`, giving no indication of which section the reader was in. This adds the section as a third crumb, and fixes two navigation faults the crumb exposed.

## Changes

- **Breadcrumb**: `src/components/layout/nav-breadcrumb.tsx` renders the current top-level section after `docs`. The link target goes through `getVersionedNavPathname`, so a v7 reader stays on v7. `src/components/layout/use-current-section.ts` lifts the section-resolution hooks out of `sidebar-nav.tsx` so the sidebar header and the navbar name the same section for a given URL. It gains a prefix-match fallback: `content/docs/orm/meta.json` does not list `v7` in its `pages`, so `/orm/v7` was absent from the ORM tab's URL set and the whole v7 tree resolved to no section, leaving both the sidebar header and the crumb blank.

- **Sidebar drill-in**: `SidebarViewProvider` now holds the resolved view rather than an override layered over a per-navigation derivation, and only re-derives when the reader changes top-level section. Previously any navigation re-derived from the URL, which ejected the reader to the all-docs list the moment they opened "Get started with Prisma" from inside the Getting Started section.

- **Getting Started target**: `src/lib/version.ts` mapped the Getting Started nav target from `/` to `/getting-started` ("Choose a Prisma 8 setup path"). Latest now resolves to `/` ("Get started with Prisma"); v7 still routes to `/v7/getting-started`, which has no page at the docs root.

- **Navbar layout**: the search trigger moves out of its centred `max-w-sm` slot into the right-hand control group and collapses to icon plus shortcut, matching the Ask AI pill beside it (`showLabel` on `LargeSearchToggle`, default unchanged for other callers). Below `lg` the Prisma lockup drops out so the section name keeps its space, and the lockup images gain `shrink-0` — as flex items with `w-auto` they were squashing horizontally around 900px instead of holding their aspect ratio.

## Why

`/` is both the docs home and the Getting Started section's index page, so the sidebar view and the crumb cannot both be derived from the URL there. The view is now state that the URL seeds but does not own, and the crumb follows the sidebar on `/` alone — every other page stays URL-derived. Giving the drilled-in state its own URL was the alternative; it would mean a second route for the same content plus the canonical, redirect and sitemap work that follows.

Verified by driving the running app over the Chrome DevTools Protocol: drill-in, cross-section moves, back, browser history, direct loads and 404 all hold, and server-rendered output matches what the client hydrates to.

## Summary by CodeRabbit

- **New Features**
  - Added contextual breadcrumbs linking to the active documentation section.
  - Added section-aware sidebar navigation across documentation versions.
  - Added responsive navigation with compact desktop search controls.

- **Improvements**
  - Updated responsive branding with full logos on larger screens and compact logo marks on smaller screens.
  - Added search controls that can display with or without labels.
  - Corrected the Latest documentation link to point to the documentation root.


ankur-arch (approved):
Rebased onto main, kept the SEO audit's aria-labels and hidden text alongside the breadcrumb and square-mark swap, applied both CodeRabbit suggestions. CI green.

PR #8293, opened 2026-09-17
Serve OpenAPI spec at standard paths

vanrensbird:
## In one line
Our API's OpenAPI spec exists, but at a URL nothing looks for. This makes it discoverable at the two paths every scanner and agent actually checks.

## For a non-engineer
An **OpenAPI spec** is a machine-readable "menu" of an API — every operation, in a format tools and AI agents read to figure out how to use it automatically. Ours is published, but at `api.prisma.io/v1/doc`. The problem: tools that go looking for an API's menu don't guess that address — by long-standing convention they check two fixed filenames, `/openapi.json` and `/swagger.json`. On our domain both return "not found." So to anything scanning us, it looks like **we have no API spec at all**.

This change puts the spec where those tools look. Concretely, it's why the ora.ai agent-readiness audit currently reports *"lacks a published OpenAPI specification"* — its single highest-impact finding for us (~+3.8 points). This is the primary fix for it.

## For an engineer
Adds two Next.js route handlers in `apps/site`:

- `GET /openapi.json` → `src/app/openapi.json/route.ts`
- `GET /swagger.json` → `src/app/swagger.json/route.ts`

Both call a shared helper (`src/lib/openapi-spec.ts`) that fetches `https://api.prisma.io/v1/doc`, injects a top-level `servers: [{ url: "https://api.prisma.io" }]` (the upstream document omits it, so a consumer of the served copy would otherwise not know the API's base URL — this mirrors `apps/docs/src/lib/openapi.ts`), and returns it as `application/json`.

Design choices:
- **Proxy, not redirect.** Some scanners don't follow cross-domain redirects, which is the whole point — so we serve the actual JSON on `www.prisma.io` rather than 302-ing to `api.prisma.io`.
- **Cached daily** via `revalidate = 86400` + `Cache-Control` (the spec only changes when the API does); `stale-while-revalidate` keeps it responsive.
- **502 fallback** if the upstream API is briefly unreachable, rather than a 500 or an empty body.
- `swagger.json` serves the identical OpenAPI 3.1 document — that legacy filename is still probed by many tools.

Tested with `node:test` (`src/lib/openapi-spec.test.ts`): `servers` injection, JSON content-type + the exact daily cache header, and both 502 paths (upstream HTTP error, and a rejecting/timed-out `fetch`). All green locally.

## Scope / follow-up
This covers the two **`www.prisma.io`** paths. The apex `prisma.io/*` 308-redirects to `www` at the platform level (same-site, standard). The matching **`api.prisma.io/openapi.json` + `/swagger.json`** are served at the source in a companion PR against `prisma/pdp-control-plane` (where the Management API lives).

## Summary by CodeRabbit

* **New Features**
  * Added publicly accessible OpenAPI specification endpoints at `/openapi.json` and `/swagger.json`.
  * OpenAPI responses now include the API server URL and caching information, with hourly refreshes.
  * Upstream specification failures return a clear gateway error response.

* **Tests**
  * Added coverage for specification retrieval, response headers, server metadata, caching behavior, and upstream HTTP, network, and timeout error handling.


apps/site/src/app/openapi.json/route.ts
```diff
@@ -0,0 +1,10 @@
+import { openApiSpecResponse } from "@/lib/openapi-spec";
+
+export const revalidate = 3600;
+
+// Serves the Prisma Postgres Management API OpenAPI spec at
```
kristof-siket: prefer JSDoc all function/constant headers. good example: https://github.com/prisma/web/pull/8293/changes#diff-1cb7e6b95968f70b389770dce485d907a08016983e93301830df25de09463600R40

apps/site/src/lib/openapi-spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  const res = await fetch(SPEC_URL, {
+    headers: { "User-Agent": "prisma-www" },
+    signal: AbortSignal.timeout(UPSTREAM_TIMEOUT_MS),
+    // Revalidate hourly: the spec changes only when the API does.
+    next: { revalidate: 3600 },
+  });
+
+  if (!res.ok) {
+    throw new Error(`Failed to fetch OpenAPI spec: HTTP ${res.status}`);
+  }
```
kristof-siket: i'd say a daily revalidate would also be fine here

kristof-siket (approved):
@vanrensbird looks good to me, added two non-blocker comments! 🙇

apps/site/src/app/openapi.json/route.ts
ankur-arch: Done in 8e21458: every function and constant across the two route handlers, the shared helper, and the test file now has a JSDoc header (including the `revalidate` exports and the test's `upstream` fixture and `stubFetch` helper). This also clears CodeRabbit's docstring-coverage warning.

apps/site/src/lib/openapi-spec.ts
ankur-arch: Agreed, done in 8e21458: revalidation is now daily (86400s). The interval lives in one exported `SPEC_REVALIDATE_SECONDS` constant that drives both the fetch's `next.revalidate` and the `Cache-Control` header. The route handlers keep a matching literal because Next.js requires `export const revalidate` to be statically analyzable, and the JSDoc on each says so.

PR #8314, opened 2026-09-23
fix(eclipse): keep accordion content in the server-rendered HTML

gregory-boch-prisma:
Closed accordion panels are unmounted, so an FAQ built with `<Accordions>` ships its **questions** as markup and its **answers** only as escaped RSC flight data inside a `<script>` tag.

### Evidence

Measured against the live site, `https://www.prisma.io/blog/agents-md-for-databases`:

| | In the rendered DOM | Inside `<script>` (RSC payload) |
| --- | --- | --- |
| FAQ questions | yes | yes |
| FAQ answer bodies | **no** | yes, escaped JSON |

Every question appears as real markup. Not one answer body does.

### Why it matters

Crawlers that do not execute JavaScript, which is most AI crawlers, see the questions and none of the answers. FAQ blocks are the most quotable part of a post: short, self-contained, already in question-and-answer shape. They are also where the `content-seo-geo` skill directs authors to put citable claims.

`<Accordions>` is used by **57 blog posts** and the docs.

### The change

`forceMount` keeps the panel in the DOM. A closed panel collapses to `h-0` and stays clipped by the existing `overflow-hidden`, so what a reader sees should be unchanged; only what is in the DOM changes.

### Not verified locally

I could not build or run this. There is no Node or pnpm on the machine it was written on, so the Eclipse and Docs Playwright suites and a visual check are down to CI and a reviewer. The specific thing to confirm by eye: that a closed panel still collapses cleanly and the open and close animations are unaffected.

### Follow-up, not in this PR

A `FAQPage` JSON-LD block emitted per post would put the same Q&A in the page as structured data and not depend on this component at all. Worth doing regardless of this fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Closed accordion panels remain mounted and collapse to zero height, preserving open and close animations.
  * Closed panels are inert, keeping their content out of keyboard navigation and the accessibility tree while closed.
  * Panel heights adjust to their natural content size, including when that content changes.
* **New Features**
  * Accordions support controlled values, allowing the open panel state to be managed externally.


gregory-boch-prisma:
@coderabbitai review

